Issue #199, opened 2026-06-06
[Cockpit] Persisted INFO logging of user actions + stream/circuit lifecycle


thefrederiksen:
## Problem / Value

The Cockpit is the **only** product component with no persisted logging. `CockpitSupervisor.cs` launches the Cockpit child with `UseShellExecute=false, CreateNoWindow=true` and no stdout/stderr capture, so every `ILogger` line goes to an invisible console; the browser side logs nothing. When the web UI misbehaves (e.g. the blank-terminal incident) there is no trace to debug from. This is the prerequisite that makes **#198/#226/#268 diagnosable instead of guesswork**, and it currently violates CLAUDE.md Rule 2 (Enterprise Logging - MANDATORY).

## Scope

**IN**
- **A. File sink:** give the Cockpit a FileLog writing to `%LOCALAPPDATA%\cc-director\logs\cockpit\cockpit-YYYY-MM-DD-<PID>.log`, same format/rotation as Director/Gateway; keep `CcDirector.Cockpit: Debug` effective for the file sink.
- **B. INFO logging of every user action** (each line carries `sid` + director endpoint): session selected (with source: rail click / needs-you nav / deep link), center + right tab switches, composer Send/Queue/Interrupt/Esc/Attach(name+size)/Speak start-stop, kebab actions, new-session open/submit (repo + options), settings open/save, fan-out, hold/wingman toggles.
- **C. Stream + circuit lifecycle:** `TerminalPane` connect attempt(wsUrl)/success/dispose, empty-DirectorBase at WARNING with the session row state; `cockpit-terminal.js` `console.log` on ws open / first binary frame (bytes + ms since connect) / close (code, reason) / each reconnect attempt, gated behind a `localStorage` `cockpit.debug=1` flag; Blazor circuit reconnect events.

**OUT**
- Gateway-side request logging (already exists in GatewayHost).
- Log-volume tuning/trimming - explicitly defer; prefer too much at INFO while the Cockpit is under active development.

## Acceptance Criteria

- [ ] After exercising the Cockpit, `%LOCALAPPDATA%\cc-director\logs\cockpit\cockpit-YYYY-MM-DD-<PID>.log` exists and contains INFO lines. (Proof: the file with content.)
- [ ] A representative sequence (select session -> switch tab -> Send -> open kebab -> new-session submit) produces one INFO line per action, each carrying `sid` + director endpoint. (Proof: annotated log excerpt mapping clicks to lines.)
[...]

PR #287, opened 2026-06-10
fix(cockpit): terminal pane recovers + shows state (#198)

thefrederiksen:
Fixes the residual half of #198 (the cross-machine half was already resolved by #268, which made the terminal stream same-origin to the Gateway). Two independent **client-resilience** defects, both verified in current code.

## Root cause

**A - blank-forever (`TerminalPane.razor`).** The pane connected only on `firstRender` and aborted if `DirectorBase` was empty. A Gateway restart mounts the pane before the session envelope is populated, so the single connect was skipped - and because the pane is `@key`ed by `SessionId`, it is never recreated to retrigger the connect. Blank forever until the user flips tabs. The `DirectorBase` precondition was also obsolete: the stream URL is same-origin (#268) and needs only `SessionId`.

**B - invisible failure (`cockpit-terminal.js`).** `ws.onclose` reconnected every 1200ms forever with no on-screen state, no `onopen`/`onerror`, and no cap. A hung/dead socket looked identical to a healthy idle stream - an empty pane with no signal.

## Fix

- **TerminalPane:** connect once `SessionId` is valid (not gated on `firstRender`), guarded by `_connectedSid` so it connects exactly once per session and retries on a later render if the attempt throws. No longer gates the stream connect on `DirectorBase` (still used for the keystroke leg in `OnInput`).
- **cockpit-terminal.js:** visible status line (`connecting...` / `reconnecting (attempt N)...` / `gave up - re-select the session to retry`), `onopen`/`onerror` + close code/reason logging, and a bounded reconnect (30 consecutive failures, ~36s). The attempt counter resets the instant real stream data arrives, so healthy long-lived sessions are unaffected.

Together: A kills the blank-forever class; B makes any future stall self-evident instead of silently blank.

## Verification

- `dotnet build src/CcDirector.Cockpit/CcDirector.Cockpit.csproj` - clean (0 warnings, 0 errors).
- **Not yet** live-verified against a real Gateway-restart in a running Cockpit - recommend watching the pane recover on screen before/after merge.

Suggested follow-up: re-sharpen #198 to this residual scope and label `flow:ready-dev` (per the maintainer note now that #268 is done). Disk-side WS logging stays with companion #199.


thefrederiksen:
## Live verification: PASS

Verified the fix end-to-end against a real Director stream across a real Gateway restart.

**Setup (no disruption to the running fleet):** an isolated Gateway on :8978 built from this branch (version-stamped `0.6.22+4a791a0`) plus this branch's Cockpit on :7471. The isolated Gateway discovered the already-running Director (:7887) via the shared FSW instances dir, so the test ran against real sessions and a real terminal stream. Only :8978 was bounced; the live fleet (7878/7470/7887) was never touched.

| # | Test | Result |
|---|------|--------|
| 1 | Terminal connects + WS proxy through the gateway | Live content rendered (not blank) |
| 2 | **Real gateway restart -> recovery (headline)** | Kill :8978 -> Blazor "Rejoining" overlay -> relaunch -> terminal repopulated live, **no blank-forever** |
| 3 | **Fix A** - fresh page reload re-mounts + connects | Pane connected on mount and streamed |
| 4 | **Fix B** asset is the one served | served `cockpit-terminal.js` (via gateway origin) contains `MAX_RECONNECT_ATTEMPTS`, `statusLine`, reconnect/gave-up text |
| 5 | **Fix B** - visible status, circuit alive | Forced a stream-only failure -> terminal showed `[stream lost, reconnecting to 127.0.0.1:8978 (attempt 3)...]` while Blazor stayed up; restored -> recovered, status cleared |

**Method notes (honest):**
- A full gateway kill also drops the Blazor circuit (both legs proxy through the gateway), so during the outage Blazor's own "Rejoining the server..." overlay shows. That is the faithful restart condition; the meaningful assertion - the terminal recovers afterward rather than staying blank - passed.
- On localhost the stream replay returns in <40ms, so under a healthy reconnect the status line flashes too fast to capture. #198's real condition (a stream that will not reconnect) was reproduced by pointing reconnects at a dead port - the status line then persisted and was captured verbatim. On a genuine slow/hung remote link it persists naturally.

Net: no blank-forever after a gateway restart (Fix A), and a hung stream now shows visible, bounded reconnect feedback instead of a silent blank pane (Fix B). Screenshots captured locally during the run.

thefrederiksen:
### Live-verification screenshots

Captured during the isolated-fleet run (gateway :8978 built from this branch, real Director :8) — see the verification comment above for the full matrix.

**1. Baseline — Terminal connected, streaming a real session through the gateway proxy**
baseline connected

**2. Gateway restart — outage (Blazor "Rejoining the server..." while :8978 is down)**
gateway restart outage

**3. Recovered — gateway back, terminal repopulated with fresh live content (no blank-forever)**
recovered after restart

**4. Fix A — full page reload, pane re-mounts and connects on its own**
fresh reload connected

**5. Fix B — stream-only failure: visible reconnect status with the Blazor circuit alive.** Buffer read verbatim: `[stream lost, reconnecting to 127.0.0.1:8978 (attempt 3)...]` (the dim ANSI status sits in the terminal pane; the surrounding Cockpit chrome confirms the circuit is up — the old silent-blank is gone).
reconnect status circuit alive

PR #295, opened 2026-06-10
fix(#293): advertise tailnet front door in handover copy + Help dialog

### LINKED ISSUE #293: Copy Handover Info (and Help dialog) leak http://127.0.0.1 Control API URL; should advertise tailnet front door when Tailscale is up
## Symptom

"Copy Handover Info" (session rail context menu) copies a Control API URL that only works on the machine that produced it:

```
Name: bpmnStudio - research
Session ID: be3ee447-ef39-462a-85ab-2b067df85c1c
Repo: D:\ReposMindzie\mindzieWeb
Director ID: 37cea6e2-da6d-418c-a5d5-7eeb417da297
Machine: SOREN_NORTH
Version: 0.6.22.0
Control API: http://127.0.0.1:7887      <-- useless from any other machine
```

The whole point of this block is to hand a session to **another** agent/machine so it can locate and talk to the session over the Control API. A `127.0.0.1` URL is unreachable from anywhere but the originating box. When this Director is on a tailnet (the default), the copied block should advertise the **tailnet front door** the session is actually reachable at, e.g. `https://soren_north.<tailnet>.ts.net:7887` -- the same address the Director already advertises to the Gateway when it registers.

This is the policy the app already commits to elsewhere ("CC Director never opens a localhost URL", `MainWindow.axaml.cs:1424-1448`), but two UI surfaces still leak loopback. User's framing: **if Tailscale is set up and not explicitly turned off, Tailscale is the default and we should never surface a localhost URL anywhere a remote consumer will read it.**

## Reproduction

1. Run a Director on a Tailscale-connected machine (Tailscale up, logged in).
2. Right-click a session in the rail -> "Copy Handover Info".
3. Paste. The `Control API:` line reads `http://127.0.0.1:<port>`.
4. From a different tailnet machine, try to use that URL -> connection refused / unreachable.

Expected: the `Control API:` line is the tailnet-reachable URL (`https://<magicdns>:<port>`). Only when Tailscale is genuinely unavailable should a loopback value appear, and it must be clearly labelled local-only.

## Root cause

The handover-copy builder hardcodes loopback instead of resolving this node's tailnet identity:

`src/CcDirector.Avalonia/MainWindow.axaml.cs:1142-1144`
```csharp
var port = app?.ControlApiHost?.Port;
if (port is > 0)
    lines.Add($"Control API: http://127.0.0.1:{port}");
```

The infrastructure to do this correctly already exists and is used by the registration path:

- `src/CcDirector.Core/Network/TailscaleIdentity.cs` -- `TryGetMagicDnsName()` returns this node's MagicDNS name (e.g. `soren_north.tailXXXX.ts.net`) or `null` when Tailscale is unavailable.
- `src/CcDirector.ControlApi/GatewayClient.cs:469-500` -- `ResolveTailnetEndpoint()` already builds exactly the right value: `https://<magicdns>:<port>` (the Tailscale Serve front door; the Director binds Kestrel to loopback only, and the Gateway's `tailscale serve --https=<port> http://localhost:<port>` maps the public HTTPS port to it). It returns `""` when there is no tailnet identity and refuses to advertise loopback. The "Control API" value in the handover block is **the same address** and should be derived the same way.

So the handover-copy and Help dialog re-implement endpoint formatting by hand and diverge from the one place that gets it right.

## Audit -- other loopback usages (the "go further" sweep)

I swept all of `src/**/*.cs`, `src/**/*.razor`, `src/**/wwwroot/*.js` for `127.0.0.1` / `localhost`. Only **two** surfaces actually leak loopback to a remote consumer. Everything else is correctly loopback (same-process / same-desktop) and must NOT be touched.

### Defects to fix (loopback handed to a remote consumer)

1. **`src/CcDirector.Avalonia/MainWindow.axaml.cs:1144`** -- "Copy Handover Info" clipboard block (the reported bug).
2. **`src/CcDirector.Avalonia/HelpDialog.axaml.cs:69`** -- the About/Help dialog's "Control endpoint" row:
   ```csharp
   controlEndpoint = host.Port > 0 ? $"http://127.0.0.1:{host.Port}" : "(no port)";
   ```
   This dialog has a "Copy" button (`_copyText`, line 45) explicitly built so the block can be pasted/shared, so the same leak applies.

### Confirmed CORRECT loopback -- do NOT change (documented so the developer doesn't "fix" them)

These are loopback by design because the URL never leaves the machine, or because the code already refuses loopback for remote callers:

- `src/CcDirector.ControlApi/GatewayClient.cs:469-500` `ResolveTailnetEndpoint()` -- already advertises MagicDNS, never loopback. **This is the reference implementation.**
- `src/CcDirector.ControlApi/ControlApiHost.cs` -- Control API binds to loopback ONLY by design; remote access is via Tailscale Serve. Correct.
- `src/CcDirector.ControlApi/InstanceRegistration.cs:51` `ControlEndpoint = http://127.0.0.1:<port>` -- written to a local `instances/<id>.json` consumed only by the **same-machine** Gateway via FileSystemWatcher; HTTP-registered Directors override with the tailnet endpoint (`DirectorRegistry`). Stays on the box. Correct.
- `src/CcDirector.Gateway/Api/GatewayEndpoints.cs` `DeriveDirectorBaseUrl()` (~1351-1374) -- has a `callerIsLoopback` guard that already refuses to hand a loopback URL to a non-loopback caller. Correct.
- `src/CcDirector.Gateway/Tailscale/TailscaleServeProvisioner.cs` -- `http://localhost:<port>` is the **backend target** of a `tailscale serve` mapping (must be loopback). Correct.
- `src/CcDirector.Gateway/Cockpit/CockpitSupervisor.cs:127` -- binds Cockpit Kestrel to loopback (`ASPNETCORE_URLS`); fronted by Serve. Correct.
- `src/CcDirector.GatewayApp/*` (GatewayTrayController, SettingsWindow, Program) -- same-process health/shutdown probes and "open settings in the browser on THIS desktop". Correct.
- `src/CcDirector.Avalonia/MainWindow.axaml.cs:1785` (`/sessions/{id}/view` "Open in Browser") and the in-process dictation roundtrip -- open on the local desktop only. Correct.
- `src/CcDirector.Core/Browser/WorkflowRunner.cs` -- in-process calls to the local browser daemon. Correct.
- `src/CcDirector.Cockpit/Program.cs:12`, `appsettings.json`, `launchSettings.json` -- dev defaults / config, overridable. Correct.
- All `*.Tests` loopback usages. Correct.

## Fix plan (executable cold)

**Goal:** the two remote-facing surfaces emit the tailnet endpoint when Tailscale is up, and only a clearly-labelled local value otherwise. Centralise the policy so it can't drift again.

1. **Add one shared resolver** for "the address this Director's Control API is reachable at", so the policy lives in ONE place (the user's "spread that across the whole application"). Put it next to the existing identity helper:
   - In `src/CcDirector.Core/Network/TailscaleIdentity.cs`, add:
     ```csharp
     /// The Control API endpoint to advertise for a given local port: the Tailscale
     /// Serve front door (https://<magicdns>:<port>) when this node has a tailnet
     /// identity, else null (no remotely-reachable URL exists).
     public static string? TryGetControlApiEndpointForPort(int port)
     {
         var dns = TryGetMagicDnsName();
         return dns is null ? null : $"https://{dns}:{port}";
     }
     ```
     (This is exactly the shape `GatewayClient.ResolveTailnetEndpoint()` produces at line 482.)
   - Optional but recommended: refactor `GatewayClient.ResolveTailnetEndpoint()` to call this helper for the MagicDNS branch, so registration and the UI share one definition. Keep its existing `_config.TailnetEndpoint` override fallback. Do not change its `""`-means-local-only contract.

2. **Fix the handover copy** -- `MainWindow.axaml.cs` `CopySessionNameAndId` (around line 1142-1144). Resolving MagicDNS shells the tailscale CLI (up to ~5s), so do it off the UI thread. The method is already `async`:
   ```csharp
   var port = app?.ControlApiHost?.Port;
   if (port is > 0)
   {
       var endpoint = await Task.Run(() =>
           TailscaleIdentity.TryGetControlApiEndpointForPort(port.Value));
       lines.Add(endpoint is not null
           ? $"Control API: {endpoint}"
           : $"Control API: http://127.0.0.1:{port} (local only - Tailscale unavailable)");
   }
   ```
   Add a `using CcDirector.Core.Network;` if not present.

3. **Fix the Help/About dialog** -- `HelpDialog.axaml.cs`. `BuildInfoRows()` is static and called synchronously from the dialog's load path; per the Responsive-UI rule it must not block on the CLI. Make the dialog populate rows asynchronously (show the dialog immediately, fill the "Control endpoint" row when resolution returns), or resolve the endpoint in the dialog's `Loaded`/`OnOpened` via `await Task.Run(...)` and update the row + `_copyText`. The value:
   ```csharp
   var endpoint = TailscaleIdentity.TryGetControlApiEndpointForPort(host.Port);
   controlEndpoint = host.Port <= 0 ? "(no port)"
       : endpoint ?? $"http://127.0.0.1:{host.Port} (local only - Tailscale unavailable)";
   ```

4. **Tests** (`src/CcDirector.Core.Tests/Network/TailscaleIdentityTests.cs` already exists -- extend it):
   - `TryGetControlApiEndpointForPort_NoTailscale_ReturnsNull` (mock/seam the MagicDNS resolver to null -- mirror how `GatewayClient` tests pin `MagicDnsResolver`; may require extracting the resolver as an injectable `Func<string?>` or testing a pure formatter that takes the dns name + port).
   - `TryGetControlApiEndpointForPort_WithMagicDns_ReturnsHttpsFrontDoor` -> asserts `https://<name>:<port>`.
   - A formatter-level test asserting the handover/help "local only" fallback string is used when the endpoint is null.

5. **Acceptance:**
   - On a tailnet machine: "Copy Handover Info" and the Help dialog "Control endpoint" both read `https://<magicdns>:<port>` (matches what the Gateway shows for this Director).
   - With Tailscale stopped: both read `http://127.0.0.1:<port> (local only - Tailscale unavailable)`.
   - Paste the tailnet value on a second tailnet machine and confirm `GET <endpoint>/healthz` answers.

## Out of scope / notes

- Do NOT change any of the "Confirmed CORRECT" list above -- those loopback uses are required.
- This does not add a network listener; the Director still binds Kestrel to loopback. We are only correcting the *advertised* address strings to match the Serve front door that already exists.

thefrederiksen:
Closes #293.

## Release Notes
"Copy Handover Info" and the About/Help dialog now show a Control API address another machine can actually reach. When this Director is on a Tailscale tailnet they advertise the Serve front door `https://<magicdns>:<port>` (the same address the Director advertises to the Gateway); only when Tailscale is genuinely unavailable do they show a clearly-labelled `http://127.0.0.1:<port> (local only - Tailscale unavailable)`.

## Changes
- `TailscaleIdentity`: add `ResolveAdvertisedControlApiEndpoint(port)` (single policy) + pure `FormatAdvertisedControlApiEndpoint(frontDoor, port)`. Reuses the existing `TryGetFrontDoorUrlForPort` instead of duplicating the front-door policy (the issue proposed a new `TryGetControlApiEndpointForPort`, but that helper already exists in effect).
- `MainWindow.CopySessionNameAndId`: resolve the endpoint off the UI thread, emit it.
- `HelpDialog`: show immediately with a `(resolving...)` placeholder, fill the Control endpoint row + Copy text in `Opened` (Responsive-UI).
- 4 new unit tests for the formatter.
- None of the "Confirmed CORRECT" loopback uses listed in the issue were touched.

## How to Test
1. On a Tailscale-connected machine, right-click a session -> "Copy Handover Info"; paste. The `Control API:` line is `https://<magicdns>:<port>`.
2. Open About CC Director; the "Control endpoint" row shows the same.
3. From another tailnet node: `GET <endpoint>/healthz` answers 200.
4. With Tailscale stopped, both read `http://127.0.0.1:<port> (local only - Tailscale unavailable)`.

## Expected Result
Both surfaces emit a tailnet-reachable URL when Tailscale is up; a labelled local-only value otherwise.

## Before / After
- Before: `Control API: http://127.0.0.1:7887` (unreachable off-box)
- After: `Control API: https://soren-north.taildb08ed.ts.net:7886` (verified reachable, healthz 200)

Proof: docs/cencon/proof/issue-293/report.html


thefrederiksen:
Developer: re-cut the branch off `origin/main` (cherry-picked only the #293 commit). PR now contains a single commit (`8cd7662`); the two issue #198 commits are gone. Re-handing to QA.

PR #804, opened 2026-06-27
Size the context gauge from the launch model id, not the transcript (#803)

thefrederiksen:
Fixes #803.

## Problem
The live context-usage gauge (#799) mis-reports Opus `[1m]` sessions. Claude's transcript records the model id **without** the `[1m]` suffix, so a 1,000,000-token session below 200k used was sized against 200k and read far too high. A live session at ~122k used showed **61%** when the truth was **~12%**. The fleet launches `opus[1m]` by default, so most Claude sessions were affected.

## Fix
- Thread the session's launch command line into `ReadContextUsage` and size the window from the **launch model id** (which preserves `[1m]`) as the authoritative signal.
- The `--model` value is parsed inside `ClaudeDriver` (it owns `ModelFlag`); handles `--model x` and `--model=x`.
- Transcript model + upward self-correction remains the **fallback** when the launch model is unknown (provider default).
- Both call sites fixed: the Control API endpoint **and** the desktop `SessionActionBar` gauge.

## Verification
- `dotnet build cc-director.sln` clean (0/0).
- **31/31** context-usage tests pass, including: `opus[1m]` launch + 121,924 used + `[1m]`-stripped transcript -> **1.0M window, ~12%** (not 61%); both `--model` forms; standard model stays 200k; no-launch-model falls back to transcript self-correction.

## Note for reviewers
A separate, orphaned Codex/pi context-usage rollout (left uncommitted by a shut-down session) was rescued to branch `rescue/agent-context-codex-pi-context-usage`. It overlaps this PR in `ClaudeContextWindow.cs`, `ClaudeDriver.cs`, `SessionContextEndpoint.cs`, and `ContextUsageTests.cs`; whoever resumes that rollout should rebase onto this fix.


thefrederiksen:
Pushed a correction after an independent Codex code review. Codex caught that the first commit wired the gauge to `Session.ClaudeArgs`, which is the raw per-session args and is **null** for the default fleet launch path (the model comes from `AgentOptions.DefaultClaudeArgs`, merged in at `BuildLaunchSpec` time). Verified empirically: a live default-launched `opus[1m]` session reports `claudeArgs=null`, so the first cut would have still read 61%.

Correction (commit 9d9395bf): persist the effective merged launch line as `Session.EffectiveLaunchArgs` (set by SessionManager at launch), and both call sites now read `EffectiveLaunchArgs ?? ClaudeArgs`. Added a production-path regression test proving the model-in-DefaultClaudeArgs path resolves to the 1M window. Build clean, 33/33 context tests pass.

PR #861, opened 2026-06-30
[Desktop] Show the Voice tab in v1 (un-gate it from alpha mode)

thefrederiksen:
## What
The desktop Voice tab no longer hides behind alpha mode, so it ships in v1 (it was alpha-gated under issue 559). Single, surgical change in `MainWindow.ApplyVoiceWingmanTabVisibility`:

```
- var voiceVisible = alpha && hasSession;
+ var voiceVisible = hasSession;     // Voice mode is a v1 feature now
```

…plus the saved-tab restore no longer forces Terminal when the persisted tab was Voice (only Wingman is still forced when alpha is off), and the related comments are updated.

## Scope (deliberately Voice-only)
Everything else stays behind `alpha_mode`, unchanged:
- Wingman tab, "Open Wingman" button
- Handovers tab, GitHub-Remote tab, FIFO, quick-launch cards, experimental menu items
- the alpha voice **settings** panel (separate from the Voice tab)

The `alpha_mode` flag and its default (OFF) are untouched. The **mobile** voice mode (`/m` PWA) was never alpha-gated.

## Behavior
- alpha OFF + active session → Voice tab **visible** (new); Wingman tab still hidden.
- No active session → Voice tab hidden (it needs a session, same as before).
- A session that had "Voice" saved as its last tab now restores to Voice instead of being bounced to Terminal.

## Verification
`dotnet build cc-director.sln` — Build succeeded, 0 warnings, 0 errors.

Note: this is a desktop (Avalonia) UI-visibility change; I verified by build + the diff (the effect follows directly from the gate). I did not capture a running-app screenshot (that needs a live desktop launch with an active session). Happy to attach one if you want it before merge.

Not merging — opened for your review per request.


thefrederiksen:
### Screenshot proof — Voice tab visible with alpha mode off

A test build of this branch (`v0.9.29 d31b705`) running with `alpha_mode` **off** and an active session. The tab row reads **Terminal | Source Control | Voice | History** — the **Voice** tab is now visible, and the **Wingman tab stays hidden** (still alpha-gated), exactly as intended.

Voice tab visible with alpha off

For contrast, the current shipped build (alpha off) shows only **Terminal | Source Control | History** — no Voice tab.

Captured from an isolated slot Director (built from this branch, launched via its own scheduled task); the image is cropped to exclude the unrelated right-hand panel and the machine name. The proof file is committed at `docs/cencon/proof/ungate-voice-tab/voice-tab-shown.png`.

PR #897, opened 2026-07-03
fix: [Dictation] compress desktop dictation to Opus before upload (#896)

thefrederiksen:
Fixes #896.

## Problem

Desktop dictation (the Dictate window) failed with a `413 Request Entity Too Large` for any recording longer than about ninety seconds. The recorder POSTed the whole clip as uncompressed 24 kHz mono WAV (about 48 kilobytes per second), so a longer recording deterministically exceeded the hosted DevThrottle transcription endpoint's fixed 4.5 MB request-body limit. The `FUNCTION_PAYLOAD_TOO_LARGE` error is generated at the platform edge before our own service code runs, so it cannot be fixed by a server change alone. A one minute fifty-one second field recording (about 5 MB) is what triggered the report.

## Fix

Encode the capture to **Ogg Opus** before the single upload, using Concentus (pure-managed, no native library). A voice-bitrate Opus stream turns two minutes of speech into a few hundred kilobytes, giving more than twenty minutes of headroom under the cap.

Both transcription modes are remote HTTP endpoints that decode Ogg Opus (OpenAI directly, and the hosted Whisper backend). The in-process **Local** mode was removed in thefrederiksen/devthrottle_internal#656, so there is no uncompressed path left to preserve and the recorder always compresses. (The original issue assumed a Local mode still existed; see the issue comment for the corrected two-mode framing.)

## Changes

- **New `CcDirector.Core/Audio/OggOpusEncoder.cs`** — encodes raw PCM16 to a complete Ogg Opus file. Encode direction only; named for the codec, not the direction, so the Gateway decode path can join it in the mobile follow-up without a rename. Documents the WebM-versus-Ogg container caveat for that follow-up.
- **`CcDirector.Avalonia/Voice/BatchDictationRecorder.cs`** — sends `dictation.ogg` instead of `dictation.wav` through the unchanged shared batch pipeline.
- **`CcDirector.Core.csproj`** — adds Concentus 2.2.2 and Concentus.OggFile 1.0.7.
- **Tests** — Ogg Opus container shape, argument guards, and a three-minute clip proven to land under the 4.5 MB cap (more than ten times smaller than the equivalent WAV, which itself exceeds the cap and reproduces the bug).

## Verification

- `CcDirector.Core` and `CcDirector.Avalonia` build clean (zero warnings, warnings-as-errors on).
- Four new `OggOpusEncoder` unit tests pass.
- All twenty-two existing desktop dictation proof tests still pass (no regression; the `WavWriter` path they exercise is unchanged).

## Remaining acceptance gate (follow-up)

Size headroom is proven by unit test. The one unverified risk is **decode correctness at the configured hosted provider** (whisper-large-v3): that it returns a correct transcript from Ogg Opus, not merely that the bytes get through. Both OpenAI and Groq accept ogg/opus, so this is expected to pass, but it needs a live round-trip with a `dt_` key to confirm. If the hosted provider were to reject ogg/opus, the safe fallbacks are FLAC (lossless) or MP3, both of which the server's container sniffer already accepts.

## Out of scope (tracked separately)

- The mobile and Gateway upload path. The phone uploads to the Gateway, which owns the mode, so any container change there is a Gateway-side follow-up.
- A presigned direct-to-storage upload flow for arbitrarily long single recordings.

### DIFF
--- a/src/CcDirector.Avalonia/Voice/BatchDictationRecorder.cs
+++ b/src/CcDirector.Avalonia/Voice/BatchDictationRecorder.cs
@@ -252,14 +252,21 @@ public sealed class BatchDictationRecorder : IAsyncDisposable
         // local cache as a side effect, #253); falls back to the local cache offline.
         var dictionary = await _dictionaryResolver.ResolveAsync(ct);
 
-        // Wrap the whole captured PCM in one WAV blob and transcribe ONCE through the
-        // shared batch pipeline. The dictionary corrector is the only text transform.
-        var wav = WavWriter.WrapPcm16(
-            pcm, MicAudioCapture.SampleRate, MicAudioCapture.Channels, MicAudioCapture.BitsPerSample);
+        // Compress the whole captured PCM to Opus (in an Ogg container) for the ONE batch upload, so a
+        // long dictation stays well under the transcription endpoint's request-body limit (issue #896:
+        // the hosted DevThrottle endpoint caps bodies at 4.5 MB, and the mic's uncompressed 24 kHz WAV
+        // at ~48 KB/s exceeded it past ~90 seconds - a hard platform 413). BOTH transcription modes are
+        // remote HTTP endpoints that decode Ogg Opus (the hosted Whisper backend and OpenAI both do),
+        // and the in-process local mode was removed in issue #887, so there is no uncompressed path to
+        // preserve. The dictionary corrector downstream is the only text transform.
+        var oggOpus = OggOpusEncoder.EncodePcm16(
+            pcm, MicAudioCapture.SampleRate, MicAudioCapture.Channels);
+        FileLog.Write($"[BatchDictationRecorder] upload packaged: file=dictation.ogg, "
+            + $"bytes={oggOpus.Length} (pcm={pcm.Length}), mode={routing.Mode.ToConfigString()}");
 
         var stopWatch = System.Diagnostics.Stopwatch.StartNew();
         using var pipeline = new BatchTranscriptionPipeline(cleanupModel: _options.DictationCleanupModel);
-        var batch = await pipeline.TranscribeAsync(wav, "dictation.wav", routing, dictionary, _profile, ct);
+        var batch = await pipeline.TranscribeAsync(oggOpus, "dictation.ogg", routing, dictionary, _profile, ct);
         stopWatch.Stop();
 
         FileLog.Write($"[BatchDictationRecorder] transcribed: rawLen={batch.RawTranscript.Length}, "
--- a/src/CcDirector.Core.Tests/Audio/OggOpusEncoderTests.cs
+++ b/src/CcDirector.Core.Tests/Audio/OggOpusEncoderTests.cs
@@ -1,0 +1,72 @@
+using System.Text;
+using CcDirector.Core.Audio;
+using Xunit;
+
+namespace CcDirector.Core.Tests.Audio;
+
+/// <summary>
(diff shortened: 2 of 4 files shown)


thefrederiksen:
## Peer review (fleet session a4a4) -- code LGTM, one open gate

Reviewed the full diff. The change is clean, minimal (4 files, +164/-5), and correctly scoped. **Approving the code; the only open risk is the live provider round-trip, which no unit test can close.**

### Verified independently (not just trusted from the report)
- **`always-compress` / no mode gate is justified.** Confirmed against `origin/main`: `TranscriptionMode` has only `Byo=0` and `DevThrottle=1` (both remote HTTP), `"local"` migrates forward, `Whisper.net` is gone from `Core.csproj`, and `Concentus` is absent on main. So thefrederiksen/devthrottle_internal#656 genuinely removed the in-process path -- there is no uncompressed branch left to preserve.
- **Server accepts the new container with zero change.** `website/api/v1/audio.js` `looksLikeContainer()` sniffs `OggS` and forwards it as-is (no PCM-wrap). The `.ogg` upload flows straight through to the configured provider.

### Strengths
- `OggOpusEncoder`: `OPUS_APPLICATION_VOIP` (right for speech), `OpusOggWriteStream` owns Ogg/Opus framing, `Finish()` flushes the tail so speech isn't clipped, `leaveOpen`+`ToArray` correct, null/odd-length guards present, little-endian `Buffer.BlockCopy` is the correct reinterpretation on win-x64.
- Encode-only, decode deliberately deferred with a clear comment; the container-aware boundary note (WebM demux for phone audio) sets up PR2 without a rename. Matches the agreed Core API shape.
- Concentus 2.2.2 + Concentus.OggFile 1.0.7 are pure-managed -- no native DLL to ship.

### Non-blocking observations
1. **Size test uses a 220 Hz sine.** Fine because Opus targets a *bitrate* not a ratio (180 s @ 24 kbps ~= 540 KB regardless of content), so the headroom claim generalizes. But a sine is the most-compressible input; white noise / multi-tone would be a stricter proxy for real-world VBR.
2. **VBR ceiling on very long clips.** 24 kbps => ~25 min under 4.5 MB. VBR can drift above target on complex passages, so a continuous ~25-min noisy dictation could theoretically approach the cap. The failure threshold moves from ~90 s to ~25 min (16x), so it's a non-issue in practice -- just worth a one-line note or a hard VBR constraint if you want a guarantee.
3. **No decode-back assertion.** Tests check magic bytes (`OggS`/`OpusHead`) but never decode the produced `.ogg` to prove it's a *valid* stream. A cheap strengthener: read it back with `OpusOggReadStream` and assert sample count ~= input. Optional -- the provider round-trip is the real proof.
4. **`EncodePcm16` hard-assumes 16-bit** (the old call passed `MicAudioCapture.BitsPerSample`). Correct given the single 16-bit mic caller; just noting the dropped parameter is intentional, not a regression.

### The one open gate (load-bearing, must close before/at merge)
Unit tests prove **size + container structure**, not that the hosted backend returns a **correct transcript** from this encoder's Ogg Opus. The risk is confined to the **DevThrottle path**: `-> Vercel proxy -> deepinfra whisper-large-v3`. (Byo `-> api.openai.com` is low-risk: OpenAI documents `ogg`.)

**Gate:** POST a sample `.ogg` from this exact encoder (24 kHz mono, 24 kbps VOIP) to `/api/v1/audio/transcriptions` with a `dt_` key and assert a correct transcript. Neither reviewer currently has a `dt_` key. Close it by (a) running that curl, or (b) a real in-app dictation smoke test in DevThrottle mode immediately post-merge.

Everything on the always-compress path and the Ogg framing checks out. Ship once the round-trip is green.

PR #899, opened 2026-07-03
fix(dictation): compress long dictation audio to Opus so it fits the upload cap (#898)

thefrederiksen:
Fixes #898.

## Problem

Whole-turn dictation captures raw PCM16 (24 kHz mono) and posts it as one WAV. PCM16 at 24 kHz is 48 KB/s, so the clip crosses Vercel's **4.5 MB serverless request-body cap at ~90 seconds** and the upload is rejected with `FUNCTION_PAYLOAD_TOO_LARGE` before the proxy code even runs (the 1:51 failure seen in the field).

## Fix

Transcode to **OGG-Opus** at the one shared remote-POST chokepoint, `BatchTranscriptionPipeline.TranscribeBatchAsync`: when the upload is a linear-PCM16 WAV at/above ~4 MB, encode it to Opus (~24 kbps, pure-managed **Concentus**, no native DLL) before the multipart POST. Opus is ~10x smaller and speech-transparent, turning the ~90 s ceiling into **~25 minutes**; both OpenAI and the DevThrottle proxy decode `audio/ogg` natively.

- **Desktop** (`BatchDictationRecorder`) and **web** (`DictationEndpoint`) both funnel through this method -> both fixed with **no call-site change**.
- **On-device Whisper** takes a different path and keeps raw WAV (it decodes PCM directly and has no network cap) -> untouched by construction.
- **Short clips / already-compressed uploads** are sent byte-for-byte as before (size-gated, zero regression for the common case).
- A genuine encode failure **throws** rather than silently shipping the oversize WAV (repo no-fallback rule).

## Changes

- `src/CcDirector.Core/Audio/OggOpusEncoder.cs` (new) - PCM16 -> OGG-Opus.
- `src/CcDirector.Core/Audio/PcmWav.cs` - `TryReadPcm16` reader (chunk-walking, PCM16-only).
- `src/CcDirector.Core/Transcription/BatchTranscriptionPipeline.cs` - size-gated transcode before the POST.
- `CcDirector.Core.csproj` - `Concentus 1.1.7` + `Concentus.OggFile 1.0.4`.
- Tests: `OggOpusEncoderTests`, `BatchTranscriptionPipelineCompressionTests`.

## Verification

- New tests (5) pass: a 100 s clip (4.8 MB WAV, over the old cap) encodes to OGG well under 4.5 MB; the actual multipart POST goes out as `dictation.ogg` with no `RIFF`; a short clip is still sent byte-identical as `.wav`.
- Full `CcDirector.Core.Tests` suite: **2634 passed, 0 failed**.
- `CcDirector.Avalonia` and `CcDirector.ControlApi` build clean (0 warnings, `TreatWarningsAsErrors`).

## Acceptance run (needs a mic + real dt_ key - reviewer to confirm)

A continuous **>2-minute** dictation returns a transcript with no 413 on BOTH the desktop Dictate dialog and the web Cockpit; local (on-device) mode still works.

### DIFF
--- a/src/CcDirector.Core.Tests/Audio/OggOpusEncoderTests.cs
+++ b/src/CcDirector.Core.Tests/Audio/OggOpusEncoderTests.cs
@@ -1,0 +1,76 @@
+using System.Text;
+using CcDirector.Core.Audio;
+using Xunit;
+
+namespace CcDirector.Core.Tests.Audio;
+
+/// <summary>
+/// Proves the OGG-Opus transcode that keeps long dictations under the hosted proxy's 4.5 MB
+/// request-body cap (issue #898): a whole-turn PCM16 WAV that would exceed the cap encodes to an OGG
+/// blob an order of magnitude smaller, and the WAV reader round-trips the format it was written with.
+/// </summary>
+public sealed class OggOpusEncoderTests
+{
+    private const int SampleRate = 24_000;   // dictation capture format
+    private const int Channels = 1;
+    private const int BitsPerSample = 16;
+
+    // Vercel serverless request-body cap. A whole-turn WAV crosses this at ~90 s; the Opus output
+    // must sit far below it. (proxy: docs/architecture/transcription-request-path.md s7)
+    private const int PlatformBodyCapBytes = 4_500_000;
+
+    /// <summary>Raw PCM16 for a low-amplitude sine tone of the given duration - non-degenerate audio.</summary>
+    private static byte[] MakePcm16(int seconds)
+    {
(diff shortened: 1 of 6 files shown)


thefrederiksen:
Superseded by #897, which independently landed the same OggOpusEncoder + Concentus dependency and compresses desktop dictation at the BatchDictationRecorder call site. This PR is a conflicting duplicate of already-merged code, so closing it. Follow-up: the web/Cockpit path (DictationEndpoint) is still uncompressed and breaks the same way past ~90s - handling that in a small separate PR that reuses the merged encoder.

PR #1147, opened 2026-07-07
Hide Voice and History tabs behind the alpha-features flag

thefrederiksen:
## What

The **Voice** and **History** session tabs are not polished enough to ship on by default, so this gates both behind the alpha-features flag, the same way the Wingman tab already is.

Previously:
- **Voice** was shown whenever a session was active (not gated).
- **History** was always visible (not gated at all).

Now both only appear when alpha mode is enabled and a session is active. With the default install (alpha off), the session tab bar shows only **Terminal** and **Source Control**.

## How

- `ApplyVoiceWingmanTabVisibility` now sets `HistoryTabButton.IsVisible` and requires alpha for the Voice tab, with a fall-back to the Terminal tab when either tab is hidden out from under the active tab.
- `HistoryTabButton` defaults to `IsVisible="False"` in the XAML so it does not flash visible before the code-behind runs.
- This method already runs on every session switch and on live alpha-flag toggles, so the tabs appear and disappear without a restart.

## Testing

Built to local slot 5 (`cc-director5.exe`), 0 warnings / 0 errors. With alpha off, only Terminal and Source Control show; flipping alpha on brings Voice and History back live.


thefrederiksen:
Closing as superseded: this pull request gated the Voice and History tabs behind the alpha-features flag, but on 2026-07-10 both tabs were deleted from the desktop app entirely (they no longer exist to hide). The repository hygiene convergence on 2026-07-11 confirmed there is nothing left for this change to apply to.

PR #1154, opened 2026-07-08
Add account nicknames to cc-outlook (#1153)

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#714.

## Problem

`cc-outlook` now routinely has two accounts (`soren.frederiksen@mindzie.com` and `email@devthrottle.com`). Sending from the non-default one required typing the full email (`-a email@devthrottle.com`), because account matching was an exact lookup on the email key. `-a devthrottle` failed with "Account 'devthrottle' not found".

## Change

Each account carries a short nickname, derived from the domain label before the top-level domain:

- `mindzie` -> `soren.frederiksen@mindzie.com`
- `devthrottle` -> `email@devthrottle.com`

The nickname is accepted anywhere `-a` / `--account` is accepted.

### Behavior

- **Resolution order** in `resolve_account()`: exact account-name (email) match first, so existing full-email usage is unchanged; then a case-insensitive nickname match. On a miss, the error lists every account name and its nickname.
- **Backfill with no re-add**: the nickname is derived on the fly when it is not stored, so the two already-configured accounts resolve by nickname immediately. Verified read-only against the real `profiles.json`: `-a mindzie` and `-a devthrottle` both resolve correctly.
- **`accounts add`**: new `--nickname` / `-n` override (defaults to the derived nickname); rejects a nickname that collides with another account (hard error, no silent fallback).
- **`accounts list`**: new Nickname column in the table and a `nickname` field in `--json` output.

## Tests

13 new tests in `tests/test_auth.py` covering derivation (subdomain, uppercase, non-email), resolve-by-stored-nickname, resolve-by-derived-nickname (backfill), exact-email-wins, case-insensitivity, unknown-lists-names, uniqueness rejection, same-account re-save, and list exposure. Full cc-outlook suite passes (97).

## Note

The changes are in the cc-outlook source. The `cc-outlook` on PATH is the installed build, so `-a devthrottle` at the terminal will work only after cc-outlook is rebuilt/reinstalled from this branch.


thefrederiksen:
For the record: this pull request closed automatically during the 2026-07-11 repository hygiene convergence because its content is ALREADY on main - the account-nickname changes to tools/cc-outlook (auth.py, cli.py, test_auth.py) landed inside commit 3067c8db (pull request thefrederiksen/devthrottle#1157, the wingman Markdown-voice fix), which carried these files along from the shared working tree. Nothing was lost; the feature is live on main. The CI failure previously blocking this pull request was an unrelated flaky HostedAgent timing test.

PR #1156, opened 2026-07-08
Fix #1155: self-heal the mobile PWA when a stale service-worker shell 404s

thefrederiksen:
Closes #1155.

## Problem
When the mobile Progressive Web App at `/m` serves a browser a previously cached shell whose router does not know a route a newer build added, the user dead-ends on React Router's raw "Unexpected Application Error! 404 Not Found". The sign-in enrollment callback (`/m/device-callback`) is exactly such a route, so this blocked sign-in entirely. Prompt service-worker activation (`skipWaiting`/`clientsClaim`, already on) does not prevent it, because the old worker renders the stale shell before the background update can reload.

## Fix
New `apps/mobile/src/components/StaleShellRecovery.tsx`:
- `RouteRecoveryBoundary` - the router's top-level `errorElement`. On a no-route match it unregisters the service worker, clears caches, and reloads once, so the browser fetches the current shell from the Gateway. A per-session guard means a genuinely missing route recovers at most once, then shows a plain message with a way back - never an infinite loop.
- `RootLayout` - clears the guard on any successful load so a later stale episode can self-heal again.

`apps/mobile/src/main.tsx` - all routes now hang off `RootLayout` with the recovery boundary. No `vite.config.ts` change (offline precache intentionally left intact).

## Verification (built + deployed to the development Gateway, driven in a real headless browser)
- Real stale-shell case: cached the fixed build, deployed a build adding a `/probe` route the cached shell lacks, navigated to it -> the fresh route rendered (not the raw 404) in one self-heal reload, no loop.
- Enrollment callback still matches and runs, never the raw 404.
- Genuine 404 recovers once then shows the friendly message - no loop.

Note: a browser already stuck on a pre-fix cached shell needs one manual "Clear site data" (or a private window) this one time, since the self-healing code only exists in builds from this fix forward.

### DIFF
--- a/apps/mobile/src/components/StaleShellRecovery.tsx
+++ b/apps/mobile/src/components/StaleShellRecovery.tsx
@@ -1,0 +1,136 @@
+import { useEffect } from "react";
+import { Outlet, isRouteErrorResponse, useRouteError } from "react-router-dom";
+
+// Self-healing recovery for a stale service-worker shell (issue #1155).
+//
+// The mobile app is an offline-capable Progressive Web App: its service worker precaches the app
+// shell and serves navigations from that cache. When a new build adds a client-side route, a
+// returning browser can still be served the PREVIOUSLY cached shell for a load (the service worker
+// serves cache-first, and a new worker only takes control on a later load). If that cached shell's
+// router does not know the route being navigated to - the device-enrollment callback
+// (/m/device-callback) is the one that bit us - React Router renders its raw "404 Not Found" and the
+// user dead-ends, unable to sign in.
+//
+// skipWaiting + clientsClaim (already on via registerType: "autoUpdate") do NOT prevent this: the old
+// worker has already rendered the stale shell before the background update+reload can win the race.
+// The only deterministic cure is to recover AT the point of failure: when a navigation matches no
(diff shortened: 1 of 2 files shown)


thefrederiksen:
The qualification soak running on this machine tonight was ended at roughly 2026-08-01 by the Mission 03 (hosted tenant-boundary) manager, and issue 1156 should know rather than find out from a missing result.

Why it was ended: it was holding the single per-user lock that serializes all Gateway test runs across every worktree on this machine, and it had stopped making progress. Three of its four partitions had finished 17 minutes earlier; partition 2's log was 0 bytes after 62 minutes, where three previous runs of that same partition each produced about 12.8 KB and completed; and its test host had burned no processor time for 30 minutes. It had no owning session in the fleet. Two escalations to the release coordinator went unanswered before it was ended.

The run should be re-run. Nothing about the qualification result is being asserted here - only that this particular run produced no result and was stopped.

One thing worth carrying into the run-book for anyone else queueing behind that lock: dotnet test abandons a lock-queued run after about five minutes and reports an empty, failed run, which reads exactly like a failed build. Setting VSTEST_CONNECTION_TIMEOUT=5400 avoids chasing a phantom failure.

PR #1203, opened 2026-07-10
fix(desktop): show session rail orange while dictations are held/queued

thefrederiksen:
## What

Makes the desktop session rail read **orange while a session has held/queued desktop dictations**, not only while a phone dictation is arriving or during the brief immediate transcription attempt.

Fixes thefrederiksen/devthrottle#1202.

## Note: re-based on the Phase 2.3 refactor

The original version of this fix layered orange onto the old `SessionStatusWingman` overlay. That mechanism was removed by the Phase 2.3 refactor (#1177 - the Director now computes only the dumb blue/red/gray standalone map; the Gateway owns overlay colors) and thefrederiksen/devthrottle#1204 added a phone-arrival orange via `Session.IsReceivingDictation`. This PR is re-implemented against that new architecture.

## Change

- New `Session.PendingDictationCount` (transient, change-notifying) - the count of held clips in the durable `PendingDictationStore` for a session.
- `SessionViewModel.StatusColorBrush` now returns orange when `PendingDictationCount > 0`, right next to the existing `IsReceivingDictation` arm, with a "Dictation queued - sending when the session is ready" reason. The view model subscribes to the new change event and repaints, exactly like the phone-arrival flag.
- `MainWindow.RefreshHeldDictationNotice` - already run on save, delivery result, and every sweeper tick - now syncs each session's count from the durable store and zeroes drained queues so the orange clears itself.

## Proof

- New Core tests (`SessionPendingDictationTests`) pin the change-notification contract the rail relies on (fires with the new value; coalesces no-op re-syncs): 2 passed.
- `CcDirector.Core` and `CcDirector.Avalonia` build clean (0 warnings / 0 errors).
- NOTE: the owner slot-5-verified the ORIGINAL (pre-refactor) implementation live; this re-based version has not yet been live-verified, only built + unit-tested. Recommend a quick re-verify on slot 5 before merge.

### DIFF
--- a/src/CcDirector.Avalonia/MainWindow.axaml.cs
+++ b/src/CcDirector.Avalonia/MainWindow.axaml.cs
@@ -3256,19 +3256,40 @@ public partial class MainWindow : Window
     {
         var store = _pendingDictationStore;
         if (store is null) return;
-        int total, parked, composerBlocked;
+        System.Collections.Generic.IReadOnlyList<global::CcDirector.Core.Dictation.PendingDictation> all;
         try
         {
-            var all = store.LoadAll();
-            total = all.Count;
-            parked = all.Count(r => r.Status == global::CcDirector.Core.Dictation.PendingDictationStatus.NeedsAttention);
-            composerBlocked = all.Count(r => r.Status == global::CcDirector.Core.Dictation.PendingDictationStatus.ComposerBlocked);
+            all = store.LoadAll();
         }
         catch (Exception ex)
         {
             FileLog.Write($"[MainWindow] RefreshHeldDictationNotice FAILED: {ex.Message}");
             return;
         }
+
+        // Keep each session's rail orange while it has held dictations (issue: only the phone-arrival
+        // lock and the brief immediate transcription attempt were orange, so a queued/held desktop clip
+        // read red). The durable store is the single source of truth; sessions with no held clips are
+        // zeroed so the overlay clears the moment their queue drains. Runs on the UI thread.
+        try
+        {
+            var perSession = all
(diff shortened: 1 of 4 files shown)


thefrederiksen:
Closing as superseded. This PR's 'orange rail while dictations are held/queued' approach (via PendingDictationCount) was replaced on main by thefrederiksen/devthrottle#1204 (desktop orange now keyed on Session.IsReceivingDictation), and thefrederiksen/devthrottle#1208 removed the desktop dictation hold-and-retry queue entirely in favour of fail-loud, no-queue delivery - so the held/queued concept this PR surfaced no longer exists. No longer needed.

thefrederiksen:
Closing as superseded. The thefrederiksen/devthrottle_internal#724 dictation workstream (Task 3b / Task 4) landed on main while this was in review and **deliberately removed the entire desktop held-dictation queue** that this fix painted orange.

`BackgroundDictationSend` on main is now explicit:

> "The behaviour is fail-loud, NO queue: ... If transcription fails, or the session is not accepting input, the words are NOT held or retried ... This deliberately replaced the old silent hold-and-retry store, whose reassurance banner everyone missed; a dictation now either goes in at once or blows up at once."

`MainWindow` no longer has `_pendingDictationStore`, the sweeper, `RefreshHeldDictationNotice`, or the held-dictation notice. So there is no longer a "held/queued desktop dictation" state to color, and the `Session.PendingDictationCount` this PR adds would never be non-zero (nothing sets it). Merging it would only add dead code.

The desktop rail already turns orange while a dictation is actively arriving via `Session.IsReceivingDictation` (#1204). The original bug thefrederiksen/devthrottle#1202 is moot under the new no-queue behaviour. Not merging.

PR #1207, opened 2026-07-10
Scheduled runs close themselves when done (#1200)

thefrederiksen:
Closes #1200.

## Problem

Scheduled cron seed runs (for example the hourly inbound-watch skill) leave their session sitting in the rail after they finish, marked "needs you" by the dumb ten-second silence timer whether or not anything actually needs a human. Running something every hour piles up leftovers to close by hand, so the automation becomes more annoying than helpful.

## What this does

An automated run now closes itself when it finishes with nothing for the human, and stays open only when it genuinely needs one.

- **AutoDismiss flag** marks scheduled seed runs (`CronJobAction.AutoDismiss`, default true), threaded `NewSessionRequest` -> `Session` -> `SessionDto`. Only flagged sessions are ever auto-closed; human-started sessions are never touched.
- **Explicit verdict**: the agent ends its run with a `CC-DISMISS` block (`verdict: done | needs-human`), mirroring the work-list `IMPL-LOOP-TERMINAL` pattern. The Director parses it from the **transcript** (robust to the alternate/full screen), not the terminal, at each turn-end and stamps `Session.DismissVerdict`, which flows up on the session snapshot.
- **Close over the stream**: the Gateway auto-dismiss sweep closes a settled auto-dismiss session whose verdict is `done` by sending the `kill` verb DOWN the Director stream (`DirectorCommandRouter` / `GatewayHost.SendCommandAsync`) - never the Director's REST Control API. The `kill` is the existing graceful kill+remove, so it leaves no phantom "interrupted" journal entry. The sweep only runs with stream mode on; there is no REST fallback by design.
- **Conservative**: no verdict, `needs-human`, a mid-turn / permission state, or no active stream all leave the session open. A run is never closed without an explicit `done`.

First consumer: the `/inbound-watch` skill emits the verdict (`done` on a quiet day, `needs-human` when anything needs Soren). That skill is an untracked local file, so it is not part of this branch.

## Tests

- `DismissVerdictSignalTests` - parsing the CC-DISMISS block (done / needs-human / none / last-wins / fenced / bad verdict).
- `AutoDismissVerdictWatcherTests` - Director-side transcript read (fake reader) + stamping, and the no-transcript case.
- `AutoDismissSweeperTests` - the selection rule (auto-dismiss + done + settled), close over the stream with the `kill` verb, no double-close, and no-stream retry.
- `PushedSessionStoreTests` - the new `SnapshotFresh` enumeration and staleness exclusion.

Full solution builds clean (0 warnings, 0 errors). Regression: Gateway 1597 passed / 0 failed, Core 2874 passed / 8 pre-existing skips / 0 failed.

## Not yet done

Live end-to-end proof on the running Gateway (stream mode on) - watching a real hourly run close itself - still to do; the tests prove the logic, not the live loop.

### DIFF
--- a/src/CcDirector.ControlApi/ControlEndpoints.cs
+++ b/src/CcDirector.ControlApi/ControlEndpoints.cs
@@ -3393,6 +3393,11 @@ internal static class ControlEndpoints
             VoiceMode = s.VoiceMode,
             OnHold = s.OnHold,
             WingmanEnabled = s.WingmanEnabled,
+            // Scheduled-run auto-dismiss (issue #1200): the flag + the agent's parsed verdict ride the
+            // snapshot/delta path up to the Gateway's auto-dismiss sweep, which closes the session over the
+            // stream on "done". Reported straight from the Session; the Gateway is the actor.
+            AutoDismiss = s.AutoDismiss,
+            DismissVerdict = s.DismissVerdict,
             // Raw local facts for the Gateway color fold (issue #1177, Phase 2). Reported straight from
             // the Session; the Director does NOT fold them into a color here (StatusColor is unchanged).
             IsBrandNew = s.IsBrandNew,
--- a/src/CcDirector.ControlApi/SessionCommandExecutor.cs
+++ b/src/CcDirector.ControlApi/SessionCommandExecutor.cs
@@ -396,6 +396,18 @@ internal static class SessionCommandExecutor
         session.WingmanEnabled = req.WingmanEnabled;
         FileLog.Write($"[SessionCommandExecutor] create: sid={session.Id} wingmanEnabled={session.WingmanEnabled}");
 
+        // Scheduled-run auto-dismiss (issue #1200): a cron seed marks the session auto-dismiss so it closes
+        // itself when it finishes with nothing needing a human. Only then attach the verdict watcher, which
+        // parses the agent's CC-DISMISS sentinel off the transcript at each turn-end and stamps the verdict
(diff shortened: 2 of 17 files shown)


thefrederiksen:
### Live end-to-end proof added (commit 7148279b)

The "not yet done" item is now done as an integration test over the **real** Director↔Gateway SignalR stream, not mocks. Two tests in `StreamCommandTests` (a real `GatewayHost` with stream mode on + a real Director-side `GatewayStreamClient` + real `SessionManager`):

- `AutoDismiss_DoneVerdict_ClosesSessionOverTheStream_EndToEnd` — an auto-dismiss session with a `done` verdict: the verdict rides **up** on the pushed session snapshot, the auto-dismiss sweep sends the `kill` verb **down** the live stream, and the Director actually kills + removes the session (asserted gone from the roster). The whole loop.
- `AutoDismiss_NeedsHumanVerdict_LeavesSessionOpen_EndToEnd` — the mirror: a `needs-human` verdict is not closed.

Also made the harness push production-faithful DTOs (the real `ControlEndpoints.Map`, the same mapper `ControlApiHost` pushes with) so the pushed session carries `Status` + the `AutoDismiss`/`DismissVerdict` fields.

All 30 `StreamCommandTests` pass; full Gateway suite (1597) and Core suite (2874) green.

Note: this proves the loop in-process over the genuine stream. I deliberately did **not** stand up a test Gateway on the tailnet — a leftover test Gateway once stole the tailnet front door (443) and broke production, so an in-process real-stream test is the safe faithful proof.

PR #1218, opened 2026-07-10
[Cockpit] Fleet sweep stability (Gateway backend) - grace window + stale envelope (#1215)

thefrederiksen:
Part of #1215 (Cockpit improvement plan, phase 6). Gateway BACKEND half only. The Cockpit rendering of the three states is a follow-up handled by the Cockpit Manager on this same issue.

## Diagnosis

The blink is a presentation defect in the roster aggregation, not the registry. In `src/CcDirector.Gateway/Api/GatewayEndpoints.cs`, the `GET /sessions` handler dropped a Director's entire session group on any single failed poll: on a non-null poll `error` it added a `MachineErrorDto` and `continue`d, removing that Director's sessions from the aggregated `sessions` for that refresh. The next successful poll put them back, so the Cockpit saw the sessions vanish and reappear. `DirectorRegistry` keeps the Director registered across a one-cycle miss (60 s heartbeat timeout, 3 min unreachable-evict, circuit opens only after 3 probe failures), so only the per-refresh session list was being lost. Full write-up posted as a comment on #1215 and in `docs/proof/issue-1215/diagnosis.md`.

## Fix

New `src/CcDirector.Gateway/Discovery/FleetRosterCache.cs`: a per-Director last-known-good snapshot with a defined grace window.

- Grace-window constant: `FleetRosterCache.GraceWindowPollCycles = 3` (named constant with an explanatory comment). Counted in poll cycles, matching how the registry's reachability circuit counts consecutive failures. Three cycles is far shorter than the registry's outer eviction bounds, so a genuinely down machine still reaches Offline promptly and within those bounds - the fix never extends them. No registry constant is changed.
- On a successful poll: store the snapshot, reset the failure streak -> Online.
- On a failed poll, within the grace window and with a stored snapshot: serve the stored snapshot marked Wobbly (with last-seen timestamp + age). The `GET /sessions` handler now runs these stale sessions through the SAME enrichment pipeline, so entries change appearance in place and the list never reflows. The owner-prune is skipped for a stale serve (a non-answering Director is not authoritative).
- On a failed poll past the grace window: Offline - drop the sessions exactly as before (historical behaviour preserved for a real outage).

This is a defined presentation state, not a silent retry that hides an outage (project law: no fallback programming).

## Envelope shape (what the Cockpit must render)

`GET /sessions?envelope=true` keeps `sessions` and `machineErrors` unchanged and adds a `directors` array (one reachability record per Director). camelCase on the wire:

- `directors[].directorId` : string (matches `session.directorId`)
- `directors[].machineName` : string
- `directors[].state` : string, one of `"online"` | `"wobbly"` | `"offline"`
- `directors[].lastSeenUtc` : string | null (ISO 8601 UTC)
- `directors[].lastSeenAgeSeconds` : number | null (0 when online, null when never seen)
- `directors[].error` : string | null (failure reason for wobbly/offline; null when online)

A Wobbly Director's sessions STAY in `sessions` (served stale). The Cockpit joins `session.directorId` -> `directors[].state` to dim in place. Example of the stale marker:

```json
{
  "sessions": [
    { "sessionId": "s-200", "directorId": "dir-B", "name": "tests", "activityState": "Idle" }
  ],
  "machineErrors": [],
  "directors": [
    {
      "directorId": "dir-B",
      "machineName": "desktop-2",
      "state": "wobbly",
      "lastSeenUtc": "2026-07-10T12:00:05Z",
      "lastSeenAgeSeconds": 5,
      "error": "timeout"
    }
  ]
}
```

Full before/after in `docs/proof/issue-1215/roster-envelope-shape.md`.

## Client-core

`packages/client-core/src/fleet/fleetClient.ts`: added the `DirectorReachability` type and the `ReachabilityState` union (`online`/`wobbly`/`offline` consts), and extended `SessionsEnvelope` with `directors`. Envelope TYPE only - no files under `apps/cockpit/` or `apps/mobile/` were touched.

## Tests

`src/CcDirector.Gateway.Tests/FleetRosterCacheTests.cs` covers the three required transitions: reachable -> wobbly -> reachable, reachable -> wobbly -> offline, offline -> reachable; plus never-reachable -> offline, idle-clock recompute on a stale serve, and forget-on-evict. Full Gateway test project: 1606 passed, 0 failed. client-core `fleetClient.ts` typechecks clean.

## Notes

- Live roster JSON capture returned HTTP 401 (auth gate ON by default), so the envelope proof is documented from the contract types and the unit-tested behaviour, per the issue's "captured JSON or documented shape" option.
- No test Gateway process was spawned; verification is via unit tests only.

### DIFF
--- a/docs/proof/issue-1215/diagnosis.md
+++ b/docs/proof/issue-1215/diagnosis.md
@@ -1,0 +1,84 @@
+# Issue #1215 diagnosis - where a single failed poll drops a Director's sessions
+
+## Summary
+
+The blink is a presentation defect in the roster aggregation, exactly as the issue
+predicted. The `DirectorRegistry` mechanisms (heartbeat timeout, reachability circuit,
+unreachable cooldown, stale sweeper) are sound and are NOT the cause. The defect is in
+the aggregated `GET /sessions` handler: when a single poll to one Director fails, that
+Director's sessions are removed from the aggregated response for that refresh, then
+reappear on the next successful poll.
+
+## Exact code path
+
+File: `src/CcDirector.Gateway/Api/GatewayEndpoints.cs`, the `GET /sessions` handler
+(the roster fan-out that aggregates every Director's sessions).
+
+1. Fan-out per Director (before this change, lines 424-473). For each Director the
+   handler either serves from the pushed cache, short-circuits with an error for a
+   flagged / cooling-down / no-endpoint Director, or pulls over HTTP via
+   `client.ListSessionsWithStatusAsync(ep, includeExitedActual)`. A failed pull returns
+   `(Director: d, Sessions: null, Error: <reason>)` and calls
+   `registry.RecordUnreachable(d.DirectorId, error)`.
+
+2. Aggregation loop (before this change, the original lines 480-492):
+
+   ```csharp
+   foreach (var (d, sessions, error) in results)
+   {
+       if (error is not null)
+       {
+           machineErrors.Add(new MachineErrorDto
+           {
+               DirectorId = d.DirectorId,
+               MachineName = d.MachineName,
+               Error = error,
+           });
+           continue;   // <-- THE DEFECT: the Director's sessions are dropped for this refresh
+       }
+       if (sessions is null) continue;
+       ...
+   }
+   ```
+
+   The `continue` on any non-null `error` means a SINGLE failed poll cycle removes that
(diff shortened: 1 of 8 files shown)


thefrederiksen:
Folded into #1226 (the combined issue #1215 pull request), which includes these exact backend commits plus the Cockpit three-state rendering. Closing to keep one pull request per issue.

PR #1222, opened 2026-07-10
[Cockpit] Handover-info Gateway endpoint + client-core call (#1214)

thefrederiksen:
## What

Backend and shared-client half of #1214: expose the desktop app's "Handover info" for a session to the browser through the Gateway. The browser talks only to the Gateway and never learns a Director address. This does NOT touch any Cockpit or mobile UI (the session menu view is built separately by the Manager).

## Endpoint path

- Director Control API: `GET /sessions/{sessionId}/handover` (new, read-only)
- Gateway proxy: `GET /sessions/{sessionId}/handover` (new, read-only)

## JSON shape (`HandoverInfoDto`)

```json
{
  "sessionId": "string",
  "displayName": "string",
  "repoPath": "string",
  "directorId": "string",
  "machineName": "string",
  "version": "string"
}
```

This mirrors the desktop "Copy Handover Info" clipboard block (name, session id, repo, director id, machine, version) with ONE deliberate exclusion: the Director's Control API endpoint. That endpoint is a Director address, and #1214 requires the browser never to learn one, so it is omitted from the DTO entirely. A test asserts the wire body carries no `http://` / control-endpoint field.

## Auth behavior

The Gateway route is a normal (non-public) session route, so the existing global `AuthMiddleware` gates it exactly like `/sessions/{sid}/summary`: a request with no valid credential gets **401** before any Director lookup. `HasValidToken` accepts the shared machine token OR an enrolled per-device key (Bearer or cookie), so the phone/desktop per-device keys work unchanged. Behavior:

- **401** without a valid token or device key.
- **404** when the session is unknown to every Director.
- **502** when the owning Director is unreachable (never a silent empty body).

The owning Director is resolved server-side via the same `LocateSessionAsync` the other proxied session endpoints use; the resolved Control API endpoint stays server-side.

## client-core

`getHandover(sessionId, signal?)` in `packages/client-core/src/api/client.ts` fetches the Gateway-relative route with `authHeaders()` and throws `GatewayError` on any non-2xx. Returns the typed `SessionHandover` interface (same fields as the DTO, camelCase).

## Tests

- ControlApi (real `ControlApiHost`): happy-path identity block, unknown-session 404, invalid-guid 400, and a no-Director-address-leak assertion.
- Gateway (real `GatewayHost`, auth on): 401 without a credential, 404 with a valid token but no owning Director.
- client-core (vitest): well-formed 200 shape, URL-encoded relative route + Bearer, partial-body degrade, 404 and 502 both throw `GatewayError`.

`dotnet build` + the handover `dotnet test` filter are green (6 passed); client-core vitest + typecheck green (5 passed).

Part of #1214.

### DIFF
--- a/docs/proof/issue-1214/handover-fields.md
+++ b/docs/proof/issue-1214/handover-fields.md
@@ -1,0 +1,48 @@
+# Issue 1214 - what the desktop "Handover info" shows today
+
+## Where it lives in the desktop app
+
+The desktop Avalonia app exposes handover info as the session context-menu item
+"Copy Handover Info" (see `src/CcDirector.Avalonia/MainWindow.axaml.cs`, the menu
+item built near line 2199 and the text builder `CopySessionNameAndId` near line
+2375). Its stated purpose (from the method's own summary comment) is:
+
+> "everything another agent needs to locate the session and talk to it."
+
+It copies a plain-text block to the clipboard. It is NOT the rich transcript
+"summary" (files touched / recent commands / open todos) served by
+`GET /sessions/{sid}/summary` - it is the small identity/locate block.
+
+## The exact lines the desktop copies
+
+The block is built from these fields (in order):
+
+- `Name`         - the session display name (`SessionViewModel.DisplayName`)
+- `Session ID`   - the stable session id (`Session.Id`)
+- `Repo`         - the repository / working directory (`RepoPath`)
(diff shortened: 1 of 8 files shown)


thefrederiksen:
Folded into #1225 (the combined issue #1214 pull request), which includes these exact backend commits plus the Cockpit session-menu UI. Closing to keep one pull request per issue.

PR #1280, opened 2026-07-11
[Cockpit] De-duplicate the copy-pasted formatting helpers (#1261)

thefrederiksen:
Closes #1261.

## What changed

The shared `fleet/format` module owns `repoBasename`, `humanizeState`, `relativeTime`, `uptime`, `shortId`, and `portOf`, yet three pages re-implemented their own copies. Every local copy is deleted; each page imports the shared helper.

- **ExesView** - dropped local `portOf`, `repoBasename`, and `relativeTime`. The unique `fmtSize` (build-slot size) stays local.
- **WingmanQueueView** and **FeedbackView** - dropped their local `shortId`, now import the shared one.
- **format.ts** - exported `portOf` (it was module-private) so the Exes page reuses the same parser.

## No behavior change - the one subtlety

ExesView's local `relativeTime` rendered hours as **"3h 20m"** - that is the shared **`uptime()`** format, not the shared `relativeTime()` which renders **"3h"**. Both of its call sites are durations ("up X", "built X ago"), so they now use shared **`uptime()`**, which preserves the exact rendered text (verified in-browser: "port 7879 - v1.1.0 - up 3h 20m", "built 2h 10m ago"). `portOf` is used with a `?? "?"` fallback so the empty-endpoint cell still shows "?".

## Session names instead of identifier prefixes

The issue asks to show the session name **where it is available from the data in hand**. Checked both payloads:
- The **Wingman queue snapshot** (`WingmanInFlight` / `WingmanQueueEntry` / `WingmanRecentBrief`) carries only `sessionId`.
- The **Feedback corpus item** (`BriefFeedbackItem`) carries only `sessionId`.

Neither has a session-name field, so the short identifier stays (per the manager's direction). Proper `sessionId` to name resolution should ride on the shared roster store (**issue 1239, Wave 2**) rather than an ad-hoc roster fetch here that 1239 would supersede. ExesView already shows session names. **Net result: no visible text changes** - this is a pure consolidation.

## Acceptance criteria

- [x] One implementation of each helper remains in `apps/cockpit`; the three files import it. (Local `portOf`/`repoBasename`/`relativeTime` in ExesView and `shortId` in Wingman + Feedback are deleted.)
- [x] No behavior change anywhere except identifiers replaced by names where the name is available. (No name is available in the two payloads, so nothing visible changed; the Exes duration output is preserved by mapping to `uptime()`.)

## Tests

`npm run test --workspace @devthrottle/cockpit` - 67 tests pass (11 new in `format.test.ts` locking `portOf`, `repoBasename`, `shortId`, `uptime` - the helpers all three pages now depend on). Build (`tsc` + `vite`) passes; ESLint clean.

## Proof

Rendered the real `ExesView` against a stubbed Gateway: the port cell, "up 3h 20m", "built ... ago", the repo basenames, and the slot sizes all render identically to before, confirming the consolidation is behavior-preserving. Screenshot in the mission QA report.

Built from a fresh isolated worktree off origin/main; only the five listed files are touched.

### DIFF
--- a/apps/cockpit/src/exes/ExesView.tsx
+++ b/apps/cockpit/src/exes/ExesView.tsx
@@ -11,6 +11,7 @@ import {
 import { gatewayErrorMessage } from "@devthrottle/client-core/api/client";
 import { dotColor } from "@devthrottle/client-core/sessions/ordering";
 import { ConfirmDialog, EmptyState, ErrorBanner, LoadingState, StatusMessage, useFlash } from "../components";
+import { portOf, repoBasename, uptime } from "../fleet/format";
 
 // The Exes management page (issue #977, epic #967) - the React port of the Blazor Cockpit
 // Exes.razor(.css) (#183). It lists the Directors running on THIS computer + their sessions and the
@@ -150,8 +151,8 @@ export function ExesView() {
                       {dir.slot == null ? "no slot" : `slot ${dir.slot}`}
                     </span>
                     <span className="ex-meta">
-                      PID <b>{dir.pid}</b> &middot; port <b>{portOf(dir.controlEndpoint)}</b> &middot; v
-                      {dir.version ?? "?"} &middot; up {relativeTime(dir.startedAt)}
+                      PID <b>{dir.pid}</b> &middot; port <b>{portOf(dir.controlEndpoint) ?? "?"}</b> &middot; v
+                      {dir.version ?? "?"} &middot; up {uptime(dir.startedAt)}
                     </span>
                     <span className="ex-spacer" />
                     {dir.directorUrl && dir.directorUrl.trim().length > 0 && (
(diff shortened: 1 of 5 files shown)


thefrederiksen:
Merging with an administrator override. The only failing check is the pre-existing flaky pair CcDirector.Gateway.Tests.StreamIntegrationTests.StreamModeOff_DoesNotMapTheHub and LauncherStreamIntegrationTests.StreamModeOff_DoesNotMapTheLauncherHub, which fail intermittently on main itself (recent main runs show roughly half failing on exactly these two) and are SignalR hub-mapping integration tests with no relationship to this change. This pull request is frontend-only (ExesView, FeedbackView, WingmanQueueView, fleet/format.ts and its new test) and contains zero C# changes, so it cannot affect these .NET tests; the .NET job does not run the frontend vitest suite at all. The actually-affected tests - the cockpit vitest suite, 67 passing - are green, and the .NET check was re-run three times with the same unrelated flaky failure each time. Recommend a separate issue to stabilize those two flaky tests (noted in the mission QA report).

PR #1311, opened 2026-07-11
fix(dictation): stop wrapped desktop dictations from being typed in twice (#1308)

thefrederiksen:
## The problem

A long desktop dictation that wraps across the composer's rows is injected into the session **twice, back to back**, then submitted once - the whole spoken message duplicated. Reproduced live in two sessions (one was even named "devthrottle - double text").

## Root cause

Desktop dictation delivers through the shared echo-verified submit (`TerminalSubmit.SharedSubmitAsync`), which watches the terminal's raw byte stream to confirm the composer echoed the typed text before pressing Enter. When the text **wraps** across composer rows, the terminal repaints it interleaved with box borders, the "esc again to clear" footer hint and cursor moves, so the typed text never appears in the byte stream as one contiguous run - even though it is sitting in the composer.

The echo check gets a **false negative**, so the code presses Escape to clear the composer and retypes the whole message. But a **single Escape does not clear Claude's composer** (it needs Escape twice - hence the "esc again to clear" hint), so the second type lands on top of the first, and the doubled text is submitted.

This code is already on `main`, so running latest code does **not** fix it - this pull request does.

## The fix

Consult the **rendered screen** as a second opinion. When the byte-stream echo check misses, concatenate and normalize the current screen rows (wrap points, borders and padding all fall outside the echo alphabet, so wrapped text reconstructs into the original) and, if the typed text is present, press Enter instead of disturbing the composer with an Escape-and-retype.

- `TerminalSubmit.cs` - threads an optional `screenSnapshot` through the submit path and adds `ScreenShowsText`.
- `Session.cs` - one line, passes `screenSnapshot: SnapshotScreenRows` (already exists on `main`).
- The dictation lock on `main` is left fully untouched - this change only adds the screen-snapshot second opinion.

## Tests

Two regression tests (all 14 `TerminalSubmit` tests pass):
- A torn-byte-stream wrapped dictation is rescued by the screen snapshot and submits **once**, with no stray Escape.
- A composer that genuinely never received the text still fails loudly.

## Not yet done

Unit-verified only. Recommend a live check in the running app (dictate a long message that wraps, confirm it lands once) before merge, since the bug is fundamentally a rendering-timing issue.

### DIFF
--- a/src/CcDirector.Core.Tests/Drivers/TerminalSubmitTests.cs
+++ b/src/CcDirector.Core.Tests/Drivers/TerminalSubmitTests.cs
@@ -224,6 +224,65 @@ public sealed class TerminalSubmitTests
         Assert.Equal(new byte[] { 0x0D }, backend.WrittenBytes[1]);
     }
 
+    // ===== screen-grid fallback (issue #1308) ======================================================
+    // A long dictation that WRAPS across composer rows is repainted interleaved with box borders and
+    // footer hints, so the byte stream may never carry the typed text as one contiguous run even
+    // though it is sitting in the composer. The rendered screen is consulted as a second opinion
+    // before Escape/retype/throw.
+
+    [Fact]
+    public async Task EchoVerifiedSubmit_EchoTornInByteStream_ScreenSnapshotRescuesAndPressesEnter()
+    {
+        var backend = new RecordingSessionBackend { Buffer = new CircularTerminalBuffer() };
+        var text = "Do you understand what it is I am talking about? And can you please also rename this session";
+        // The byte stream only ever shows torn fragments interleaved with the footer hint.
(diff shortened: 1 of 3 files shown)


thefrederiksen:
Closing as superseded. Verified that the double-type fix this pull request delivers (the 'consult the rendered screen as a second opinion' logic in TerminalSubmit.cs - ScreenShowsText / screenSnapshot) is already on main, byte-for-byte identical, merged via #1310 (commit 2370cf78, 'verify against the rendered screen, never lose the words'). This branch's Session.cs is additionally stale: it still carries the old dictation lock that #1310 deliberately removed, so merging would risk a regression. No unmerged work here. The fix reaches installed machines via a stable release cut (held intentionally until another implementation lands).

PR #1512, opened 2026-07-14
fix(sessions): make an explicit "hold my session" during a live turn stick (deferred hold)

thefrederiksen:
## The bug
Asking to put a session **on hold while it is still working** didn't stick. `OnHold` was set immediately, but the very turn that was in flight then ended, and the turn-END auto-lift (which surfaces an ordinary snooze at the next "needs you") cleared the hold right back off. Net: "put my session on hold" appeared to do nothing.

## The fix — a deferred, durable hold
`Session.RequestHold(onHold)`:
- **un-hold** clears the hold and any pending deferral at once;
- a hold requested **while a real turn is in flight** is **deferred** (a new `_pendingHold` flag) instead of set now;
- a hold requested while the session is already settled applies **immediately** (unchanged).

When the session next settles, `SetActivityState` applies the deferred hold **durably** — `OnHold` goes true and, unlike the automatic snooze, is **not** lifted by that same turn boundary. It parks the session held (red or idle) until the user comes back to it. A fresh submission before it applies **supersedes** the deferral (the user is driving again).

The hold command executor routes through `RequestHold`, and `HoldResponse` gains a `Pending` flag so a caller can tell the user "it'll hold when it finishes what it's doing" rather than implying it's already held.

## Safety
The existing turn-END auto-lift (for a snooze already in place *before* the turn) is **unchanged** — the new path is purely additive, so the prior hold tests pass untouched.

## Verification
- Core + ControlApi + Gateway build clean.
- Full `CcDirector.Core.Tests`: **3016 passed, 8 skipped, 0 failed** (incl. 5 new deferred-hold tests: defer-then-apply at red and idle settles, immediate hold when settled, release-clears-pending, superseded-by-submit).

### DIFF
--- a/src/CcDirector.ControlApi/SessionCommandExecutor.cs
+++ b/src/CcDirector.ControlApi/SessionCommandExecutor.cs
@@ -293,9 +293,16 @@ internal static class SessionCommandExecutor
         if (session is null)
             return DirectorCommandResult.Fail(DirectorCommandStatus.NotFound, "session not found");
 
-        session.OnHold = onHold;
-        FileLog.Write($"[SessionCommandExecutor] hold: session={guid} onHold={onHold}");
-        return DirectorCommandResult.Success(Serialize(new HoldResponse { OnHold = session.OnHold }));
+        // RequestHold defers an EXPLICIT hold that arrives mid-turn so it applies durably when the turn
+        // ends, instead of being bounced by the turn-END auto-lift (the "put my session on hold while it is
+        // still working" case). An immediate hold / un-hold behaves as before.
+        var outcome = session.RequestHold(onHold);
+        FileLog.Write($"[SessionCommandExecutor] hold: session={guid} onHold={onHold} outcome={outcome}");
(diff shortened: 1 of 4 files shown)


thefrederiksen:
PAUSED - do not merge yet. Coordination: session 3a2239de is redesigning hold/snooze as a unified `HoldState` state machine (collapsing `OnHold` + `_turnInFlight` + this PR's `_pendingHold` into one enum, plus the Gateway/cockpit/mobile/desktop labels), which fully supersedes this point-fix. This PR's deferred-hold semantics + 5 tests are being folded into that design; Soren decides who implements. Holding here with a written reason per trunk rules rather than merging.

thefrederiksen:
Closing in favor of the unified HoldState state machine (owner decision: session 3a2239de designs and builds it). The deferred-hold semantics and the 5 tests from this PR are being folded into that work and credited. The branch `fix-hold-current-session` is intentionally NOT deleted so that work can lift the code verbatim.

PR #1558, opened 2026-07-14
Move the prompt log to the Gateway (#1551)

thefrederiksen:
Follow-up to #1552, which put the log in the wrong place. The log belongs on the **Gateway**.

## Why

Two reasons, both the owner's, and both decisive:

- **The Gateway is what the whole fleet reports to.** Anyone asking for history asks it and the answer
  is already there. Nothing has to go hunting across machines.
- **The Gateway is what moves to the server.** The log moves with it, because it was never somewhere
  else to begin with.

## The split

**The Director captures and pushes. The Gateway stores. The Director keeps no copy.**

That split is forced, not chosen:
- Only the Director ever sees a prompt - desktop-local input never reaches the Gateway - and only the
  Director knows whether it was typed or spoken and from which surface.
- Only the Gateway sees the whole fleet.

Same shape as the stats spine that already exists, where the Director observes and the Gateway
aggregates.

## Gateway side

| | |
|---|---|
| `GatewayPromptLog` | THE log. Daily jsonl, unbounded retention. |
| `POST /prompts` | a Director pushes what it captured |
| `GET /prompts` | anyone wanting history reads it here |

`PromptRecord` moves to `Gateway.Contracts` so both sides share it, and gains **`machine`** — the
Gateway holds the fleet, so without it two developers' records are indistinguishable.

A batch spanning days (a backfill, or one crossing midnight) is split into the right daily files rather
than dumped into today's — otherwise a week's reading is wrong.

## Director side

- **`ConversationLog` and `InputOriginLog` are deleted.** No Director-side log remains.
- `InputOriginBuffer` replaces the origin log: in memory, bounded per session. An origin only has to
  survive the seconds between submitting and the turn ending. If the Director dies in that window, the
  prompt is honestly `unknown` rather than guessed.
- **A message is marked recorded only once the Gateway confirms the write.** Nothing else holds it, so
  marking before a failed push would lose the prompt permanently. A refused push is retried at the next
  turn, where the transcript still has it. This is why the push is acknowledged with a count instead of
  being fire-and-forget.
- The sink resolves the **current** `GatewayClient` per push rather than capturing one — the Director
  rebuilds its client on reconfigure, and a captured instance would go stale and silently stop
  recording.
- No Gateway configured means no log, and it says so rather than pretending.

The ingest state left on the Director is a **watermark, not a log**: hashes only, never text.

## Proof

- **17 Core tests** — including a refused push being retried rather than lost, and every record
  carrying its machine.
- **14 Gateway tests** — 9 over the store, 5 driving `POST` then `GET` against a **real HTTP server** on
  an ephemeral port: a Director pushes, the Gateway serves it back, two machines stay distinguishable.
- Core **3058 green**, Gateway **2170 green**, whole solution builds.

Everything from #1552 survives unchanged: capture at the choke point, the origin join, the context id,
the Gemini logs.json fix, source-scoped dedupe. Only the destination changed.


thefrederiksen:
Superseded: this branch was cut from the pre-squash `feat/prompt-log` rather than from main, so it carried three commits already merged as f14b2f22 and could not compute mergeability. Reopened as a clean cherry-pick onto current main - same change, no duplicate history.

PR #1774, opened 2026-07-18
feat(gateway): persist the workflow catalog and extract the mission conduct (Workflows phase 1)

thefrederiksen:
Phase 1 of the approved Workflows plan (mission 9a955739: Mission becomes the first built-in, user-defined, cross-agent Workflow).

## What this does

- The workflow catalog moves off the three hard-coded C# literals onto the EF data layer: a `workflows` head table (identity + lifecycle), immutable `workflow_versions` rows (draft/published/superseded, canonical bundle content hash, authorship), and `workflow_files` for helper files. All tenant-scoped, one new migration (`AddWorkflows`).
- The three built-ins (mission, standalone, standalone-with-review) are seeded at startup from embedded markdown resources. The mission body is a FAITHFUL extraction of `.claude/skills/mission/SKILL.md` - same conduct, new home. Two mechanical self-reference edits were made, each visible in this diff: the preservation warning now says "THIS WORKFLOW" instead of "THIS FILE", and the brief checklist points at `cc-devthrottle workflow instructions mission` instead of linking the file. The skill file itself is untouched (it is stubbed only in phase 6, after the live cross-agent proof).
- Built-ins follow the ours/yours trade (owner ruling): editable and versioned like any workflow, never deletable, and newer shipped content auto-publishes ONLY while the published content still matches what the previous binary shipped. A customized workflow is never touched; the newly shipped hash is still recorded so a later reset-to-shipped knows what this binary carries.
- `GET /gateway/workflows[/{id}]` keeps the exact legacy read shape and gains additive fields only (`version`, `isBuiltIn`, `updatedUtc`, `hasDraft`, `contentHash`). The cockpit workflows page is untouched.

## Proof

- Full solution suite: all seven test projects, 6,272 passed, 0 failed.
- Fidelity test compares the embedded mission body against the skill file modulo exactly the two listed edits - and was proven able to fail: a deliberate one-character drift in the embedded body turned it red; restoring turned it green.
- Existing endpoint tests pin the legacy wire shape (ids, order, seats, 404, case-insensitive lookup) and now run against the persisted store; a new test covers the additive fields; store tests cover fresh seeding, idempotent re-seeding across a restart, the uncustomized auto-upgrade path, and the customized-left-alone path.

## Not in this phase

Authoring routes (drafts, publish, If-Match), the `cc-devthrottle workflow` CLI, the run spine, seated sessions, the skill-file stub, and the cockpit UI replacement - phases 2 through 8 of the plan.

### DIFF
--- a/src/CcDirector.Gateway.Contracts/WorkflowDtos.cs
+++ b/src/CcDirector.Gateway.Contracts/WorkflowDtos.cs
@@ -1,0 +1,66 @@
+namespace CcDirector.Gateway.Contracts;
+
+/// <summary>
+/// One step of a workflow on the wire: a named piece of work, who does it, who reviews it, and what
+/// finishing it means. Reviewer is null when the step has no separate review seat - which is itself a
+/// statement the workflow is making, not an omission. The field names and shape are FROZEN as the
+/// legacy catalog contract (issue #1617); the Cockpit's workflowsClient.ts types mirror them.
+/// </summary>
+public sealed class WorkflowStepDto
+{
+    public string Name { get; set; } = "";
+    public string Description { get; set; } = "";
+    public string Doer { get; set; } = "";
+    public string? Reviewer { get; set; }
+    public string Done { get; set; } = "";
+}
+
+/// <summary>
+/// One declared outcome criterion of a workflow: what "done and accepted" means for a run of it. The
+/// workflow AUTHOR declares these; a run seeds its per-criterion results from them (the governance
+/// outcome spine, issue #1771). Structured on purpose - governance needs per-criterion evaluation -
+/// while the conduct itself stays in the instruction markdown.
+/// </summary>
+public sealed class WorkflowOutcomeCriterionDto
+{
+    /// <summary>Stable slug identifying the criterion within its workflow (e.g. "merged-pr").</summary>
+    public string CriterionId { get; set; } = "";
+
+    /// <summary>What must be true for a run to count as delivered on this criterion.</summary>
+    public string Description { get; set; } = "";
+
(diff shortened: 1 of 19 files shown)


thefrederiksen:
Independent inspection (different agent family) was called on this diff before merge and returned six findings. Disposition:

1. Migration not discoverable - REFUTED: the inspection was fed a diff that deliberately excluded the generated Designer and model-snapshot files to keep it focused; both are in the commit (the Designer carries the migration attribute, the snapshot is updated), and the reseeding test exercises a real migrate-open cycle over the same file.
2. Tenant scoping without tenant-composite keys - VALID observation, pre-existing layer-wide convention: every table in this data layer (cron, worklists) keys the same way; the local install is single-tenant. Composite keys are a layer-wide rework for the multi-tenant hosted Gateway phase, tracked there rather than patched piecemeal here.
3. Rollback treated as an upgrade - INTENDED semantics, comment was imprecise: an uncustomized built-in tracks the RUNNING binary's shipped content in both directions (the injected-text ours channel behaves the same way); a rollback republishes the older conduct on purpose and the minted version row is the honest record. Comments now say so.
4. Case-insensitive lookup over stored ids - holds by invariant: the only write path (the seeder) writes lowercase slugs, and the authoring API in phase 2 validates the slug pattern at the boundary.
5. Content-hash invariant unguarded by tests - FIXED: a new test pins determinism and sensitivity of the bundle hash to every content dimension (metadata, steps, instructions, files); a constant-hash substitution now fails the suite.
6. Fidelity test's unchecked replacements - FIXED: each listed edit now asserts its source phrase exists exactly once in the skill file before replacing, so phrase drift can no longer make an edit a silent no-op.

Workflow tests re-run green after the fixes (17 passed).

PR #1775, opened 2026-07-18
fix(director): push a session delta when mobile view mode changes

thefrederiksen:
**Mission: Dumb Clients - slice A** (first of the sequence).

## What
`WireDoorbellPush` subscribed to activity, hold, the three colour inputs and the wingman gate, but **not** to `OnViewModeChanged`. `ControlEndpoints.Map` stamps `VoiceMode` (derived from `ViewMode`) onto the pushed `SessionDto`, and the mobile roster reads it to decide a row's link target (voice sessions open on their Voice tab) and its voice affordance. So switching a session to voice mode told the Gateway nothing until the next ten-second re-push - the phone kept the stale `voiceMode`. Same one-line shape and rule as the existing colour-fact pushes: **if a client reads it, it pushes.**

## Scope (deliberately minimal)
- One `OnViewModeChanged` subscription, added identically to the existing anonymous handlers.
- A `rePushInterval` **test seam** on `ControlApiHost` (production passes `null`, keeping the 10s default) used only to isolate the regression test.
- The test suite `DoorbellColourFactPushTests` is generalized to `DoorbellFactPushTests` (event-to-stream fact propagation); the four colour tests keep their history.

## Proof it can fail
`VoiceMode_ReachesTheGateway_OnItsOwnPush` runs the real Gateway + real Director + real `WireDoorbellPush`. It observes an explicit `false` baseline, then asserts **both** transitions (`false→true`, `true→false`) inside a 4s window, with the periodic re-push disabled for the run - so a constant stamp or a coincidental snapshot cannot satisfy it. Verified: with the subscription removed the test fails by timing out with the exact symptom; the four colour tests stay green; restored, all five pass. Full Gateway.Tests suite green.

## Explicitly out of scope (follow-up slice)
The doorbell handlers are anonymous and are **not** detached when a session is removed - a **pre-existing** property of all of these subscriptions (a late change on a removed session can map a disposed session). This slice does not change that behaviour; it is fixed for **all** the handlers together in a dedicated follow-up (doorbell subscription lifetime + concurrency + idempotent attach), which also adds a received delta-vs-snapshot seam to make the reconnect-reseed isolation a hard assertion.

### DIFF
--- a/src/CcDirector.ControlApi/ControlApiHost.cs
+++ b/src/CcDirector.ControlApi/ControlApiHost.cs
@@ -203,6 +203,7 @@ public sealed class ControlApiHost : IAsyncDisposable
     // (StartControlApi runs before StartScheduler), so we capture an accessor, not the instance.
     private readonly Func<Core.Scheduler.SchedulerService?>? _schedulerAccessor;
     private readonly string? _instancesDirectory;
+    private readonly TimeSpan? _rePushInterval;
     private bool _stopped;
     private bool _stateServicesStarted;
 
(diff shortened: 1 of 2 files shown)


thefrederiksen:
Deferred. The one-line view-mode push is correct but the reviewer showed it must sit on top of a doorbell-subscription lifetime/concurrency fix (all handlers), which is its own slice. Reopening/redoing as part of that follow-up. Closing to keep the board clean.

PR #1776, opened 2026-07-18
feat(gateway): workflow authoring API - drafts, If-Match, transactional publish, reset, archive (Workflows phase 2)

thefrederiksen:
Phase 2 of the approved Workflows plan (mission 9a955739). Phase 1 (#1774) persisted the catalog; this makes it agent-authorable.

## What this does

- **Drafts are the safety boundary.** POST creates a workflow as a draft (invisible to the catalog); PUT {id}/draft replaces the draft's content wholesale, minting the next version number when drafting over a published workflow. An optional If-Match content hash refuses stale edits with 409 instead of clobbering a concurrent author.
- **Publish is the fleet-visible act**, and it is strict and atomic: it demands instructions and at least one step with a doer and a done, then flips draft to published and the previous published version to superseded in one save.
- **Pinned history resolves forever.** GET {id}/instructions?version=N and GET {id}/files/{name}?version=N serve any version row - superseded or archived - because a workflow run that pinned a version must always be able to answer what conduct governed it. The default (no version) read follows catalog semantics.
- **Built-ins are editable with reset-to-shipped** (owner ruling): customizing mission works like any edit; POST {id}/reset republishes the running binary's shipped content as a new version via the same builder the startup seeder uses, so shipped-content-as-a-version has exactly one definition.
- **Archive is soft** and never touches a built-in.
- Validation at the store boundary: slug ids, 200 KB instruction cap, bare validated helper-file names with an allow-listed extension set (.py .md .txt .json), 256 KB per file, 20 files per version, unique outcome-criterion slugs.

## Proof

- Full solution suite: all seven test projects, 6,286 passed, 0 failed.
- 14 new tests: store-level (create/conflict/slug rules, full-replacement draft writes with hash recompute, If-Match stale vs current, draft-over-published minting, the publish gate, supersede atomicity, pinned reads after supersede AND archive, built-in customize-then-reset round-trip, reset refused on user workflows, archive semantics, file rules) plus an HTTP round-trip covering create - skeletal publish refused - fill draft - publish - raw markdown read - stale If-Match 409 - duplicate id 409 - archive - built-in delete refused.

## Next

Phase 3: the cc-devthrottle workflow command group (list/show/instructions/pull/push/publish/versions/delete/reset/materialize).

### DIFF
--- a/src/CcDirector.Gateway.Contracts/WorkflowDtos.cs
+++ b/src/CcDirector.Gateway.Contracts/WorkflowDtos.cs
@@ -64,3 +64,79 @@ public sealed class WorkflowDto
     /// <summary>The canonical content hash of the published version (the exact bundle a run pins).</summary>
     public string ContentHash { get; set; } = "";
 }
+
+/// <summary>A helper file carried by a workflow version, with full content (authoring payloads).</summary>
+public sealed class WorkflowFileDto
+{
+    public string FileName { get; set; } = "";
+    public string Content { get; set; } = "";
+}
+
+/// <summary>A helper file reference without its content (detail responses; content is served raw by
+/// <c>GET /gateway/workflows/{id}/files/{fileName}</c>).</summary>
+public sealed class WorkflowFileInfoDto
+{
+    public string FileName { get; set; } = "";
+    public string ContentHash { get; set; } = "";
+}
+
+/// <summary>One row of a workflow's version history (no content bodies).</summary>
+public sealed class WorkflowVersionInfoDto
+{
+    public int Version { get; set; }
+    public string Status { get; set; } = "";
+    public string ContentHash { get; set; } = "";
+    public string AuthoredBy { get; set; } = "";
+    public string? ChangeNote { get; set; }
+    public DateTime CreatedUtc { get; set; }
+    public DateTime? PublishedUtc { get; set; }
+}
+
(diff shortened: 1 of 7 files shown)


thefrederiksen:
Independent inspection (different agent family) returned five findings. Disposition:

1. Mutable drafts served by pinned-read routes - FIXED: explicit-version instruction and file reads now exclude the draft row (a draft is mutable and cannot be pinned history); drafts remain readable through the versions/{n} detail route, which states the status. Test added.
2. Concurrent publish/reset across two store instances - NOT A PHASE 2 DEFECT: the Gateway constructs exactly one store instance per process and the data layer's documented model is single-process, single-writer with an instance lock (every store in the layer - cron, work lists - has the same shape). Two Gateways over one database is not a supported topology. A filtered unique index on published status is a sensible layer-wide hardening if that topology ever arrives.
3. If-Match check/write race across store instances - same disposition as 2: within the one process that exists, the instance lock serializes the check and the write.
4. Unbounded authoring fields - FIXED: names capped at 200 characters, prose fields at 4000, 50 steps and 50 criteria per version, on top of the existing instruction/file caps. Test added.
5. Null collection entries producing 500s - FIXED: steps, criteria, and files lists reject null entries with a clear 400. Test added.

Workflow tests re-run green after the fixes (31 passed).

PR #1778, opened 2026-07-18
feat(cli): the workflow command group - agents author fleet workflows end-to-end (Workflows phase 3)

thefrederiksen:
Phase 3 of the approved Workflows plan (mission 9a955739). Phase 1 (#1774) persisted the catalog, phase 2 (#1776) made it authorable over HTTP; this gives agents the CLI surface, which is the primary authoring path by design.

## What this does

- New cc-devthrottle workflow group: list / show / instructions / versions / pull / push / publish / materialize / reset / delete, all registered in the agent-discoverable actions list.
- instructions prints the RAW markdown, byte-identical - this output goes straight into an agent's context, so no rich rendering and no wrapping.
- Authoring round-trips a skill-like directory (workflow.json + instructions.md + helpers/). Pull writes a .workflow-hash sidecar; push sends it as If-Match so a stale copy gets a clean conflict instead of clobbering a concurrent author, and writes the new hash back on success.
- Push creates the workflow when new (existence checked via the versions route - the catalog read rightly 404s a draft-only workflow) and updates the draft otherwise. Authorship defaults from CC_SESSION_ID.
- materialize writes a published version's bundle to %LOCALAPPDATA%/cc-director/workflows/<id>/<version>/, hash-keyed and idempotent, printing absolute paths so any agent can run helpers with its own shell. Drafts are refused.
- Small Gateway addition: the version-detail response now carries helper-file content (the raw files route rightly refuses drafts, and pull must round-trip drafts for the authoring loop).

## Proof

- Full solution suite: all seven projects, 6,287 passed, 0 failed. CLI suite: 81 passed (9 new: actions discoverability, raw-markdown output pinning, pull directory shape + sidecar, push round-trips exactly what pull wrote with If-Match riding along, create-when-new, refusing a directory declaring a different id, authoring-version pick preferring the draft).
- LIVE end-to-end proof, run locally against a real isolated GatewayHost on an ephemeral port with the real Python CLI from source: list showed the built-ins; instructions mission returned the conduct with its sentinel strings (THE FOUR LAWS, merged-to-origin/main); pull -> edit -> push produced draft v2; publish made it the fleet-read version; reset restored the shipped conduct as v3 while the customized v2 stayed readable as pinned history; materialize wrote the machine cache and printed the paths; versions showed published/superseded/gateway:reset history. The throwaway proof harness was removed after the run (it shells out to python, which the committed suite must not depend on).

## Next

Phase 4: the workflow-run spine (#1771) - runs pinned to a version, lifecycle separate from acceptance, participants, criteria results.

### DIFF
--- a/src/CcDirector.Gateway.Contracts/WorkflowDtos.cs
+++ b/src/CcDirector.Gateway.Contracts/WorkflowDtos.cs
@@ -72,12 +72,14 @@ public sealed class WorkflowFileDto
     public string Content { get; set; } = "";
 }
 
-/// <summary>A helper file reference without its content (detail responses; content is served raw by
-/// <c>GET /gateway/workflows/{id}/files/{fileName}</c>).</summary>
+/// <summary>A helper file in a version-detail response. Carries the full content: the detail route is
+/// the authoring read (the CLI pulls a complete directory from it, drafts included - the raw
+/// <c>files/{fileName}</c> route serves only pinned, non-draft versions).</summary>
 public sealed class WorkflowFileInfoDto
 {
     public string FileName { get; set; } = "";
     public string ContentHash { get; set; } = "";
+    public string Content { get; set; } = "";
 }
 
 /// <summary>One row of a workflow's version history (no content bodies).</summary>
--- a/src/CcDirector.Gateway.Tests/WorkflowAuthoringTests.cs
+++ b/src/CcDirector.Gateway.Tests/WorkflowAuthoringTests.cs
@@ -97,9 +97,13 @@ public sealed class WorkflowAuthoringTests : IDisposable
         Assert.Equal("verify.py", updated.Files[0].FileName);
 
         // A DRAFT is never served by the pinned-read routes (it is mutable, so it cannot be pinned
-        // history); the file becomes readable the moment the version publishes.
+        // history); the file becomes readable the moment the version publishes. The AUTHORING read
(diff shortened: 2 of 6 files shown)


thefrederiksen:
Independent inspection (different agent family) returned eight findings. Disposition:

1. Server-supplied file names could escape the pull/cache directories - FIXED: a bare-name safety check refuses separators, traversal, and drive colons before any write, on both pull and materialize. Test added (a hostile fileName fails cleanly and writes nothing).
2. A missing sidecar silently disabled the stale-copy protection - FIXED, made LOUD: push now refuses to update without the sidecar and says why; --force is the explicit override. Test covers both refusal and override.
3. Re-pull could resurrect server-deleted helpers - FIXED: pull mirrors the server by clearing the helpers directory first. Test added.
4. Materialize trusted the hash sidecar without checking the files exist - FIXED: every listed file must exist or the bundle rewrites.
5. Instructions output appended a newline - FIXED: prints exactly what the Gateway returned.
6. Universal-newline translation could alter content on the round-trip - FIXED: exact newline handling on every content read and write; a test pins carriage-return content surviving pull-then-push byte-faithfully.
7. A sidecar write failure after a successful server update left a confusing state - FIXED: a clean message states the draft WAS updated and how to resynchronize.
8. Wrong-shaped workflow.json produced a traceback - FIXED: clean validation message.

CLI suite re-run green after the fixes (86 passed).

PR #1779, opened 2026-07-18
feat(gateway): the workflow-run spine - every mission now opens a governance run (Workflows phase 4, #1771)

thefrederiksen:
Phase 4 of the approved Workflows plan (mission 9a955739), and the foundation issue thefrederiksen/devthrottle_internal#856 asked for: the outcome spine attaches to workflow RUNS, with Mission as the built-in workflow's run.

## What this does

- New workflow_runs table (tenant-scoped, one migration). A run pins the exact published version that governed it at creation - version row id, number, canonical content hash - and published versions referenced by runs are never deleted, so what conduct governed this run is answerable forever.
- Lifecycle and acceptance are SEPARATE: lifecycle is created / active / awaiting-human / succeeded / failed / abandoned with legal transitions enforced in one place (terminal is final; started/completed timestamps stamp automatically), acceptance is pending / accepted / rejected / waived with accepter identity and timestamp. Verified yield reads acceptance, never mere completion.
- Criteria results seed pending from the pinned version's declared outcome criteria and CANNOT be invented on a run. Proof links carry evidence. Participants are persisted run-to-session membership (session, agent kind, role, machine, join/leave history) recorded on the run directly - standalone runs have no Mission to derive membership from.
- Creating a mission through the Gateway now opens a run of the built-in mission workflow and returns the additive workflowRunId on the mission DTO. Mission stays the grouping record; the run is the governance object (reference, not merge).
- REST: POST/GET/GET{id}/PATCH /gateway/workflow-runs with workflow/status/mission filters. CLI: workflow runs (list) and workflow run <id> (detail), in the agent-discoverable actions.

## Proof

- Full solution suite: all seven projects, 6,296 passed, 0 failed. CLI suite: 86 passed.
- Eight new run-store tests: version pinning survives the catalog moving on; unknown/draft-only/archived workflows refuse runs; legal-moves-only lifecycle with terminal final; acceptance independent of lifecycle (accepted while active; back-to-pending clears the stamp); criteria update the seeded set only; duplicate participant join is a no-op and leave stamps history; list filters.
- One new HTTP test: POST /missions returns workflowRunId; the run is pinned to the published mission conduct's hash, anchored to the mission, and lists by missionId.

## Next

Phase 5: discoverability index + version-pinned seated sessions + the cross-agent live proof.

### DIFF
--- a/src/CcDirector.Gateway.Contracts/MissionDto.cs
+++ b/src/CcDirector.Gateway.Contracts/MissionDto.cs
@@ -20,6 +20,12 @@ public sealed class MissionDto
 
     /// <summary>The parent Mission this one nests under (a tree of Missions), or null for a root Mission.</summary>
     public Guid? ParentMissionId { get; set; }
+
+    /// <summary>ADDITIVE (Workflows mission, phase 4, issue #1771): the workflow run opened beside
+    /// this Mission when it was created through the Gateway - a mission is a run of the built-in
+    /// "mission" workflow. Null on reads that do not resolve it and on missions predating the spine.
+    /// Existing clients ignore it.</summary>
+    public Guid? WorkflowRunId { get; set; }
 }
 
 /// <summary>
--- a/src/CcDirector.Gateway.Contracts/WorkflowRunDtos.cs
+++ b/src/CcDirector.Gateway.Contracts/WorkflowRunDtos.cs
@@ -46,6 +46,10 @@ public sealed class WorkflowRunProofLinkDto
 /// </summary>
 public sealed class WorkflowRunParticipantDto
 {
+    /// <summary>The CANONICAL fleet session id - the Director-minted session GUID every other
+    /// per-session record (token/spend statistics, transcripts, the roster) also keys on. This is
+    /// governance's join from a run to the effort it consumed (issue #1771, spine item 3); never put
+    /// any other identifier here.</summary>
     public string SessionId { get; set; } = "";
     public string AgentKind { get; set; } = "";
     public string Role { get; set; } = "";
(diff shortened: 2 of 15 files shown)


thefrederiksen:
Independent inspection (different agent family) returned five findings; the governance session (#1771) also reviewed the run contract as joint owner and is green. Disposition:

1. Mission creation not atomic across the two stores - FIXED for every checkable failure: the mission workflow's runnability is validated BEFORE the Mission record is written, so the expected failure (unrunnable workflow) can no longer orphan a mission. The residual window - process death exactly between the JSON mission write and the EF run write - is documented in place and closes when the legacy JSON mission store retires onto the EF layer.
2. Non-pending acceptance could lack or inherit its actor - FIXED (and extended per governance's ask): accepted, rejected, AND waived all require an EXPLICIT acceptedBy on every decision; the previous accepter is never inherited, so a later rejection cannot be misattributed. Timestamp always server-stamped. Tests assert all three refusals and the waiver stamping.
3. Partial criterion patches silently reset status - FIXED: criteria patches now use a dedicated update type whose null fields mean unchanged (no status default), and the evaluation timestamp is server-stamped, never caller-supplied. Test pins a note-only patch preserving met.
4. Duplicate criterion ids blending inconsistently - FIXED: one patch may name a criterion once; duplicates refuse. Test added.
5. Unbounded run listing - FIXED: ordered and bounded in the database, default page 200, ceiling 1000, limit exposed on the endpoint.

Full suite re-run green after the fixes (all seven projects, 6,296 passed). Governance confirmation on record: both pre-merge asks satisfied, zero further concerns, clear to merge.

PR #1780, opened 2026-07-18
feat(director): the workflow index rides the fleet preamble (Workflows phase 5a)

thefrederiksen:
Phase 5a of the approved Workflows plan (mission 9a955739): discoverability. Sessions learn the workflow catalog exists at launch; the conduct itself stays pull-based (phase 3's CLI), so the per-session context cost is a few lines, paid once.

## What this does

- New [WORKFLOW_INDEX] placeholder in the fleet preamble: one line per published workflow (id + summary, truncated at 160 characters) plus the fetch command. Rendered like every other placeholder - a user's own injected text gets the index only where they wrote the token.
- New Director-side WorkflowIndexStore, a structural sibling of InjectedTextStore: RefreshAsync fetches GET /gateway/workflows off the launch path (on Gateway-connect and the existing refresh timer), renders the block once, disk-caches it; ActiveIndex() reads the cache synchronously at launch. Never-reached Gateway = empty index = exactly what sessions got before the feature.
- Delivery paths opt in explicitly (both hook endpoints + the Pi preamble writer); a null store renders no index, keeping every test and preview hermetic by default.
- The migration-era byte-for-byte oracle retired on its own charter's terms (first deliberate rewording) and its promised replacement ships in the same change: a golden snapshot of FleetPreambleTemplate.Default with a received-file approval flow, so every future preamble change is a reviewable diff.

## Proof

- Full solution suite: all seven projects, 6,294 passed, 0 failed (13 retired oracle cases out, 11 new tests in).
- New tests: index format pinned (one line per workflow, fetch command present, clean block end, runaway summaries truncated), cache round-trip through a fresh store, refresh against a stub Gateway with the fleet token asserted, empty catalog renders nothing (no floating header), the default template carries the placeholder, a seeded index lands in the rendered preamble, and a null index store leaks neither the token nor a header.

## Next

Phase 5b: version-pinned seated sessions (spawn --workflow/--role, participants recorded on the run) and the cross-agent live proof.


thefrederiksen:
Independent inspection (different agent family) returned four findings. Disposition:

1. Authored summaries injected verbatim - FIXED (half of it had already landed as a hardening push that crossed the inspection): whitespace runs collapse to one space, control characters including ANSI escapes strip outright, ids cap at 64 characters, and at most 50 entries render with an honest 'and N more' line. Tests cover newline forgery, escape stripping, id runaway, and the entry cap.
2. Cache corruption or a refresh/read race could break session launch - FIXED: the cache write is atomic (temp + move), and a corrupt/unreadable cache degrades loudly to no index (the documented empty state) instead of escaping as an exception into every launch; the next successful refresh heals it. Test writes garbage JSON and asserts a clean empty read.
3. Revoked content riding a dead Gateway forever - FIXED: a cache older than 24 hours injects nothing until a refresh succeeds. The cost is only discoverability (the CLI still lists everything); the refresh timer keeps a healthy Director far inside the window. Test covers both sides of the ceiling.
4. Golden test trimmed all trailing newlines - FIXED: exact comparison with exactly one concession (the single final newline editors append on save), so added or removed blank lines at the template's end are real diffs again.

Preamble and index tests re-run green (62 passed).

PR #1787, opened 2026-07-18
feat(director,gateway): version-pinned seated sessions - spawns seat on workflow runs (Workflows phase 5b)

thefrederiksen:
Phase 5b of the approved Workflows plan (mission 9a955739). 5a (#1780) gave every session the catalog index; this seats sessions on runs.

## What this does

- NewSessionRequest gains workflowRunId plus Gateway-stamped workflowId/workflowVersion. BOTH spawn legs seat: the Gateway relay (remote) validates and stamps like it already stamps MissionName; the local /fleet/spawn leg resolves through new GatewayClient run lookups with the thefrederiksen/devthrottle#1548 fail-loud posture (unreachable Gateway = 502, never 'unknown run').
- A mission spawn AUTO-SEATS onto the mission's governance run - no new flags; an explicit seat is available via session spawn --workflow-run (unknown ids refused in plain English).
- The spawned session is recorded as a run PARTICIPANT with the canonical fleet session GUID (the thefrederiksen/devthrottle_internal#856 run-to-session membership governance joins effort on).
- The seat persists on the session, surfaces on the SessionDto (additive fields), and the fleet preamble appends a [Workflow seat] paragraph: role, run, the conduct fetch command PINNED to the run's version (never a moving head), and the fail-closed rule. It rides the same preamble every agent family already receives.
- Backward compatible with the PRODUCTION Gateway: a Gateway predating the run spine (404 on the run routes) means spawn-unseated, exactly the pre-feature behavior. The existing mission-attach test - whose Gateway stub faithfully models that older Gateway - went red and caught this before it shipped; without the fix, a Director updated ahead of the Gateway would have 502'd every mission spawn.

## Proof

- Full solution suite: all seven projects, 6,306 passed, 0 failed - with the honest full Gateway count (2,620), after an earlier partial run was caught by the count check and re-run.
- New WorkflowSeatTests over the real local-spawn path (real Director Control API, real RawCli spawns, Gateway stub over loopback HTTP): the mission spawn stamps run id + workflow id + pinned version on the session; the participant PATCH reaches the Gateway with the session GUID, agent kind, role, and machine; the seated session's preamble carries the seat paragraph with the pinned --version fetch and the STOP rule; an unknown explicit run id 400s and records nothing; an unseated spawn is unchanged and carries no paragraph.

## Sequencing note

The fleet-LIVE cross-agent proof (a real Claude and a real Codex session fetching the conduct through the CLI) requires the production Gateway to run a binary with the workflow routes - it runs after the next Gateway deployment, and phase 6 (stubbing the old mission skill file) stays gated on it. Phase 7 (cockpit list + add) is independent and next.

### DIFF
--- a/src/CcDirector.ControlApi/ControlEndpoints.cs
+++ b/src/CcDirector.ControlApi/ControlEndpoints.cs
@@ -79,6 +79,7 @@ internal static class ControlEndpoints
                 var text = FleetPreamble.BuildForSession(
                     session.Id.ToString(), name, Environment.MachineName, session.RepoPath, user,
                     workflowIndex: new WorkflowIndexStore());
+                text = AppendSeatParagraph(text, session);
                 return Results.Text(text, "text/plain");
             }
             catch (Exception ex) when (ex is InjectedTextUnavailableException or FleetPreambleTemplateException)
@@ -127,6 +128,7 @@ internal static class ControlEndpoints
                 text = FleetPreamble.BuildForSession(
                     session.Id.ToString(), name, Environment.MachineName, session.RepoPath, user,
                     workflowIndex: new WorkflowIndexStore());
+                text = AppendSeatParagraph(text, session);
             }
             catch (Exception ex) when (ex is InjectedTextUnavailableException or FleetPreambleTemplateException)
             {
@@ -243,7 +245,7 @@ internal static class ControlEndpoints
         // Create a session LOCALLY through the shared SessionCommandExecutor (issue #1177 Phase 1), then build
         // the identity-stamped 201 response. Used by the local branch of POST /fleet/spawn; the Machine routing
         // field is advisory here (the routing decision is made before the request reaches this Director).
-        async Task<IResult> CreateLocalSessionAsync(NewSessionRequest? req)
+        async Task<IResult> CreateLocalSessionAsync(NewSessionRequest? req, Func<Session, Task>? onCreated = null)
         {
             var command = new DirectorCommand
             {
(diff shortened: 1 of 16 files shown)


thefrederiksen:
Independent inspection (different agent family) returned four findings. Disposition - all four FIXED:

1. Seat stamped after launch (Pi missed it entirely; hook agents could race it) - FIXED: a new beforeLaunch hook on session creation stamps the role and the seat before Pi's preamble file is written and before the process starts. Pi's file now carries the seat paragraph (test added), and the paragraph moved to one shared builder so no agent family can receive a differently-worded seat.
2. Relay participant write could turn a successful spawn into a retryable 500 - FIXED: exception boundary; the failure is reported loudly in the log beside the already-running session.
3. Forged/injected seats via the trusted down-channel - FIXED at two layers: the shared paragraph builder validates the workflow id against the catalog slug shape (a non-slug seat renders nothing and logs), and the create path refuses a malformed seat as a clean bad request before any process starts. Tests cover newline and case forgeries. The residual fact that a fleet-token-bearing caller can stamp a syntactically valid seat without Gateway validation is the same coarse-bearer trust model every stamped field (mission name included) already lives under - documented, and the honest paths always resolve through the Gateway.
4. New Gateway recording participants for sessions on old Directors - FIXED: membership records only when the Director's reply carries the seat back; otherwise the session is logged as started unseated, which is the truth.

Suite: all seven projects green, with the Gateway suite re-run STANDALONE for the honest full count (2,620 passed) after the solution-level runner twice reported a silently truncated Gateway total - that runner quirk is worth its own issue and is noted in the report.

PR #1790, opened 2026-07-18
feat(cockpit): the workflows page becomes a list + detail + agent-handoff add dialog (Workflows phase 7)

thefrederiksen:
Phase 7 of the approved Workflows plan (mission 9a955739): the UI replacement. TypeScript and CSS only - no Gateway or Director changes.

## What was wrong

The old page rendered every workflow fully expanded: a wall at three workflows, unusable at ten; fed by what used to be C# literals, it could not show a user workflow, a version, or a draft; no add; and the actual conduct - the thing the whole feature distributes - was invisible.

## What this does

- LIST: one compact row per workflow (name, Built-in/Custom badge, Draft-waiting badge, version, step count, one-line summary), linking to the detail page.
- DETAIL (/workflows/:id): metadata + the step summary rendered small (the machine-readable shape, demoted), then the instruction markdown rendered read-only via the existing sanitized markdownToHtml from client-core - the exact text an agent fetches with cc-devthrottle workflow instructions, and labelled as such.
- ADD: a dialog asking ONLY name + one-line summary. Submit creates a DRAFT on the Gateway (invisible to the fleet until an agent publishes) and shows a copyable handoff prompt covering the pull, edit, push, publish loop. No markdown box, no step builder - authoring is agent-driven by design.
- Client: getWorkflow / getWorkflowInstructions / createWorkflow added beside the existing read; additive catalog fields typed optional so an older Gateway still reads; suggestWorkflowId pinned by tests to always produce the Gateway's slug shape.

## Proof

- Strict workspace typecheck green (the build gate every configuration runs).
- client-core: 540 tests passed including the new suggestWorkflowId suite; cockpit bundles cleanly.
- The wire endpoints this page consumes are covered by the Gateway suite from phases 1-2 (legacy shape pinned; create-draft and instructions routes proven over HTTP).

## Remaining phases

Phase 6 (stub the old mission skill file) stays gated on the fleet-live cross-agent proof, which needs the production Gateway's next deployment. Phase 8 (distribution extras) is owner-gated.

### DIFF
--- a/apps/cockpit/src/main.tsx
+++ b/apps/cockpit/src/main.tsx
@@ -18,6 +18,7 @@ import { DirectorsView } from "./fleet/DirectorsView";
 import { DirectorDetailView } from "./fleet/DirectorDetailView";
 import { ScheduleView } from "./schedule/ScheduleView";
 import { WorkflowsView } from "./workflows/WorkflowsView";
+import { WorkflowDetail } from "./workflows/WorkflowDetail";
 import { DictionaryView } from "./dictionary/DictionaryView";
 import { TranscriptsView } from "./transcripts/TranscriptsView";
 import { YourThrottleView } from "./throttle/YourThrottleView";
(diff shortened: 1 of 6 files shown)


thefrederiksen:
Independent inspection (different agent family) returned five findings; it also independently verified the sanitized-renderer claim from the renderer's code (raw HTML escaped, link schemes allow-listed) and found no XSS. Disposition - all five FIXED:

1. Torn reads on the detail page - the conduct is now fetched pinned to the version the metadata names, sequentially.
2. Backdrop dismissal during a slow create (half-born draft, lost handoff, retry id-conflict) - the backdrop only dismisses while idle.
3. Older Gateways mislabeling built-ins as Custom - the kind badge renders only when the field is reported.
4. Unhandled clipboard rejection - honest visible degrade pointing at the selectable text.
5. Nonexistent theme tokens - replaced with the shell's real --warn / --attention / --accent.

Re-proof after fixes: strict workspace typecheck green, client-core 540 passed, cockpit 145 passed, bundle clean.

PR #1817, opened 2026-07-18
feat(gateway,director,cli): the owner's switch - any workflow can be turned off, built-ins included (register redesign, slice 1)

thefrederiksen:
Slice 1 of the register redesign (approved mockup direction A). The switch semantics are the owner ruling verbatim.

## What this does

- Every workflow head gains Enabled (default true; the migration backfills existing rows at the database level, so upgrading can never silently switch workflows off).
- OFF means: dropped from every agent's launch briefing (index builder filters; absent field from an older Gateway = enabled); the default conduct read refuses with a clear 400 naming the re-enable command (pinned explicit-version reads keep resolving - a seated run's conduct never vanishes under it); no new runs (store + EnsureRunnable) and no new seats on either spawn leg (run projections carry workflowEnabled, defaulting true for older Gateways so Director-ahead compatibility holds).
- The owner ruling on missions: a mission whose workflow is off still gets created - UNGOVERNED, loudly logged, no run record - and the next mission after re-enable is governed again. Never a 400.
- Nothing deleted, ever: catalog lists off workflows with honest state; history, versions, and past runs stay; the flip is instant both ways fleet-wide.
- Surface: POST /gateway/workflows/{id}/enable + /disable; cc-devthrottle workflow enable/disable (agent-discoverable); State column on workflow list.

## Proof

- Six new tests pin every semantic: catalog still lists with state; default read refused with the exact message while pinned reads resolve; new runs refused and restored by the flip; past runs report state; the briefing index drops off workflows and counts honestly (absent field = enabled); and over real HTTP: disable -> 400 conduct read -> UNGOVERNED mission created (no run id) -> enable -> governed mission again; unknown id 404s.
- Suites: Gateway standalone 2,797 passed 0 failed (full honest count - see the note), Core 3,272, CLI 86, all other projects green.
- Test-runner note: the intermittent silent-truncation defect (#1791) fired twice during this slice's gates, including on a standalone run (190 reported as everything while 2,773 were discoverable). Evidence added to the issue; every count above is from verified-complete runs.

## Next

Slice 2: the register layout itself (approved mockup direction A) - state spine, switch, provenance and activity columns, lifecycle strip, explainer/Help.

### DIFF
--- a/src/CcDirector.ControlApi/ControlEndpoints.cs
+++ b/src/CcDirector.ControlApi/ControlEndpoints.cs
@@ -685,6 +685,13 @@ internal static class ControlEndpoints
                             statusCode: StatusCodes.Status502BadGateway);
                     }
 
+                    if (seatRun is not null && !seatRun.WorkflowEnabled)
+                    {
+                        // The owner turned this workflow OFF: no new seats; spawn unseated, loudly.
+                        FileLog.Write($"[ControlEndpoints] POST /fleet/spawn: workflow " +
+                                      $"'{seatRun.WorkflowId}' is OFF - spawning UNSEATED");
+                        seatRun = null;
+                    }
                     if (seatRun is not null)
                     {
                         req.WorkflowRunId = seatRun.Id;
--- a/src/CcDirector.Core/Sessions/WorkflowIndexStore.cs
+++ b/src/CcDirector.Core/Sessions/WorkflowIndexStore.cs
@@ -153,18 +153,19 @@ public sealed class WorkflowIndexStore
     /// </summary>
     public static string BuildIndexText(IReadOnlyList<CatalogWorkflow> workflows)
     {
-        if (workflows.Count == 0)
+        var inForce = workflows.Where(w => w.Enabled != false).ToList();
+        if (inForce.Count == 0)
             return "";
 
         var text = new StringBuilder();
         text.Append("[Workflows] Named ways of working this fleet defines - usable by ANY agent. Before taking\n");
         text.Append("on work that matches one, fetch its conduct and FOLLOW it:  cc-devthrottle workflow instructions <id>\n");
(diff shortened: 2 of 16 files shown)


thefrederiksen:
Independent inspection (different agent family) returned four findings. Disposition:

1. Missing/archived workflow silently treated as the owner's OFF - FIXED on this branch BEFORE the verdict landed (self-caught while briefing the inspection): the check is three-valued; only an explicit FALSE is a choice, null (missing/unseeded store) keeps the loud pre-check refusal. Silently ungoverned missions are exactly the gap the outcome spine closes.
2. A Director one version behind the switch can seat onto a just-disabled workflow via its own resolution path until it updates - ACCEPTED as a bounded transitional window, documented here: the alternative (redacting disabled-workflow runs from the run-read surface) would corrupt governance reads, the harm is limited to a briefing paragraph pointing at conduct that remains readable as pinned history, and the window closes when the Director updates - which on this fleet happens with the Gateway.
3. No actor or audit on the switch - FIXED: enable/disable REQUIRE ?by= (anonymous flips refused 400 with a clear message - the run-acceptance posture applied to governance changes), the actor is recorded in the log, and the CLI names its session automatically. The durable audit trail is governance's append-only event ledger (#1771), which hangs off these same ids.
4. Mission-create racing a flip - FIXED for both windows: the pre-check refusal cannot fire for an off workflow (three-valued check), and a disable landing between the pre-check and the run create is honored - the mission is created UNGOVERNED and logged, never an error for a mission that was in fact created.

Gates: Gateway standalone verified-complete 2,798 passed 0 failed; workflow filter 50 passed; CLI 86 passed. (The thefrederiksen/devthrottle_internal#858 runner truncation fired twice more during this slice; every count above is from verified-complete runs.)

PR #1819, opened 2026-07-18
feat(cockpit): the workflows register - ledger layout, switch on every row, lifecycle strip, on-page explainer (register redesign, slice 2)

thefrederiksen:
Slice 2 of the register redesign (the layout the owner approved from the mockups). Slice 1 (#1817) shipped the switch end to end on the Gateway, Director, and CLI; this slice gives the cockpit the register that presents it.

## What changes

**The list becomes a register** (`WorkflowsView.tsx` rewritten):
- One row per workflow with a 5px state spine on the left edge: green = in force, amber = draft waiting, gray = off. The row reads as a ledger entry, not a card.
- The owner's switch on every row, built-ins included. Turning OFF asks for confirmation (it removes the workflow from every agent's briefing and refuses new runs and seats); turning ON is direct. Nothing is deleted either way.
- Provenance column: published version + last update date. Activity column: run counts rolled up client-side from the run spine; an off workflow states "agents will not see or run this"; if the runs read fails (an older Gateway), the column says "activity unavailable" instead of blanking the register.
- The lifecycle strip above the register states the delivery model as page furniture: any agent authors -> draft -> publish -> in force everywhere, instantly; nothing to deploy, no restart, running missions keep their pinned version.
- A collapsible explainer under the register answers the standing questions (out of the box, no restart, pinned runs, off) on the page itself.
- The add dialog is unchanged in behavior (draft + copyable agent handoff prompt).

**The detail page** (`WorkflowDetail.tsx`):
- The same switch and state label in the facts row, with the same confirm-on-off semantics.
- An off banner when the workflow is off. The conduct still renders, because the detail read is pinned to the workflow's published version and pinned reads bypass the off refusal by design.
- Built-ins gain a "Reset to shipped" action (confirmed): the shipped content becomes a new published version; customized history stays readable by version number.

**Client** (`workflowsClient.ts`): `WorkflowRunSummary` + `getWorkflowRuns`, `setWorkflowEnabled` (actor required, per the Gateway contract), `resetWorkflow`, and the additive `enabled` flag on `WorkflowDefinition` (absent on an older Gateway means enabled).

**Styles** (`workflows.css`): rewritten for the register on the shell theme tokens; the dead card and compact-row styles from the two earlier layouts are removed. Scrollable panes use always-visible scrollbars per the house rule.

## Verification

- `npm run typecheck --workspaces --if-present`: client-core, cockpit, mobile all clean.
- `npm run test --workspace @devthrottle/client-core`: 544 passed (50 files).
- `npm run test --workspace @devthrottle/cockpit`: 145 passed (14 files).
- `npm run build --workspace @devthrottle/cockpit`: production build clean.
- No C#/Python changes in this slice; the seven-project suite ran green on slice 1 (#1817), which this branch is cut on top of.

### DIFF
--- a/apps/cockpit/src/workflows/WorkflowDetail.tsx
+++ b/apps/cockpit/src/workflows/WorkflowDetail.tsx
@@ -3,11 +3,13 @@ import { Link, useParams } from "react-router-dom";
 import {
   getWorkflow,
   getWorkflowInstructions,
+  resetWorkflow,
+  setWorkflowEnabled,
   type WorkflowDefinition,
 } from "@devthrottle/client-core/workflows/workflowsClient";
 import { gatewayErrorMessage } from "@devthrottle/client-core/api/client";
 import { markdownToHtml } from "@devthrottle/client-core/history/historyMarkdown";
-import { ErrorBanner, LoadingState } from "../components";
+import { Button, ConfirmDialog, ErrorBanner, LoadingState } from "../components";
 
 // One workflow, in full (Workflows mission, phase 7). The list row answered "what exists"; this page
 // answers "what does it actually say": the metadata and step summary up top (the machine-readable
@@ -20,6 +22,8 @@ export function WorkflowDetail() {
   const [workflow, setWorkflow] = useState<WorkflowDefinition | null>(null);
   const [instructions, setInstructions] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
+  const [pendingOff, setPendingOff] = useState(false);
+  const [pendingReset, setPendingReset] = useState(false);
 
   const load = useCallback(
     async (signal?: AbortSignal) => {
@@ -47,6 +51,12 @@ export function WorkflowDetail() {
     return () => ctrl.abort();
   }, [load]);
 
+  const flip = async (enabled: boolean) => {
+    if (id === undefined) return;
+    await setWorkflowEnabled(id, enabled, "cockpit");
+    await load();
+  };
+
   return (
     <div className="page wf">
       <header className="ui-page-header">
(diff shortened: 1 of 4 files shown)


thefrederiksen:
Adversarial inspection (gpt-5.6-sol) reported seven findings. Disposition:

1. REFUTED (reported as blocker): "the detail loader reads instructions through the unversioned endpoint, so a disabled workflow's detail page 400s." The loader fetches the definition first and then fetches the conduct PINNED to `wf.version` (`getWorkflowInstructions(id, wf.version, signal)`), and pinned reads bypass the off refusal by design. Every workflow in the register has a published version (never-published drafts are not listed), so the pin is always present.

2. FIXED: run counts from a bounded newest-runs window were presented as all-time totals. The column is now labeled "Recent activity" and the empty state says "no recent runs".

3. FIXED: a failed enable/disable/reset died as an unhandled promise rejection with the switch silently unchanged. Flip and reset now catch, land the message in the page error state (with Retry), and re-render from the Gateway's truth.

4. FIXED: the register rendered an actionable switch even when an older Gateway omitted the `enabled` field (and therefore lacks the switch routes). The switch now renders only when the Gateway reported the flag, matching the detail page's gating.

5. FIXED: the narrow-pane collapse keyed on viewport width, so a wide window with the session rail open could leave the pane too narrow and clip the provenance and activity columns under `overflow: hidden`. The register pane is now a size container and the collapse is a container query on its actual width.

6. FIXED: a mutation-triggered detail refresh carried no abort signal, so a late response after routing to another workflow could paint workflow A's state under workflow B's URL. Loads now carry a generation and a superseded load drops its result.

7. FIXED: the explainer disclosure button carries `aria-expanded` and `aria-controls`.

Gates after the fixes: typecheck clean across client-core, cockpit, and mobile; client-core 544 passed (50 files); cockpit 145 passed (14 files); cockpit production build clean.

PR #1825, opened 2026-07-18
feat(director): the workflow index invites agents to improve workflows, not just follow them

thefrederiksen:
The [Workflows] block every session is briefed with gains one fixed closing
line naming the authoring commands: pull, push, publish, with the two facts an
agent needs to use them safely (drafts are private; publish is fleet-wide and
instant). An agent that finds a workflow wrong or outdated now knows the
sanctioned move is to fix the workflow, not to silently work around it. The
line renders only under a non-empty index, so a Director that has never
reached a Gateway still injects nothing.

### DIFF
--- a/src/CcDirector.Core.Tests/Sessions/WorkflowIndexStoreTests.cs
+++ b/src/CcDirector.Core.Tests/Sessions/WorkflowIndexStoreTests.cs
@@ -63,15 +63,22 @@ public sealed class WorkflowIndexStoreTests : IDisposable
     {
         // Agents are invited to IMPROVE workflows, not just obey them (owner ruling 2026-07-18):
         // the block's last line names the authoring commands. An empty catalog still injects
-        // nothing - the invitation never floats without an index above it.
+        // nothing - the invitation never floats without an index above it. The sneaky entry pins
+        // the boundary: authored data is collapsed to one physical line by the sanitizer, so no
+        // summary can push extra lines below itself, displace the fixed footer, or forge one.
         var text = WorkflowIndexStore.BuildIndexText(new[]
         {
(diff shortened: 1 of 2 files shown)


thefrederiksen:
Adversarial inspection (gpt-5.6-sol) reported two blockers and one should-fix. Disposition:

1. REFUTED ("breaks the line-oriented format contract"): the index block was never a records-only format - it already opens with two header lines and can end with an "...and N more" overflow line, none of which match the entry shape. No consumer parses the block; it is rendered prose injected into agent context, and the "one line per workflow" promise is about preamble cost and injection containment, not machine parsing. Rendering the invitation inside BuildIndexText is deliberate: it must appear only under a non-empty index, which the template placeholder cannot express.

2. REFUTED ("authored-content prompt-injection path adjacent to a mutation authorization"): the trust boundary is upstream - catalog summaries can only be authored through the authenticated Gateway API by the owner's own fleet agents, and the sanitizer collapses each to a single capped printable line. The footer grants no capability an agent does not already have: the CLI is on every PATH, its action listing names the same commands, and the block's HEADER already carries a fetch-and-follow instruction, which is strictly more powerful than the footer. Agents-may-publish with drafts as the safety boundary is the owner's explicit ruling (2026-07-17); approval gates are a later governance feature by design.

3. FIXED (missing adversarial boundary test): the new pinning test now includes a hostile summary carrying newlines and a forged copy of the footer text, and asserts the exact physical line count (two header lines + one line per workflow + the footer) plus the collapsed single-line rendering of the hostile entry - proving authored data cannot add lines, displace the fixed footer, or forge one.

Gates: full seven-project suite green standalone (Core 3,265 passed / Gateway 2,808 passed on the rebased branch; the other five projects passed pre-rebase and are untouched); the strengthened index tests pass (15/15).

PR #1841, opened 2026-07-19
fix(gateway): keep a voice-mode session yellow until its voice is ready

thefrederiksen:
## What

On the phone, a voice-mode session was flashing **red \"needs you\"** in the gaps between voice-generation attempts, then flipping back to yellow when the next attempt began - until the voice was finally ready. It should stay **yellow \"preparing voice\"** the whole time.

## Why it happened

The rail color and the Voice screen are two separate Gateway folds. The Voice screen (`VoiceDisplayFold`) already stays yellow while retrying. The rail (`SessionOrdering.IsVoicePreparing`) held yellow **only while a synthesis attempt was literally in flight** (`VoiceGenerating`), so it fell back to the base red state in every gap. Voice was modeled as a transient overlay on a red session rather than as a state of its own.

## The change

Voice mode is now a first-class state. `IsVoicePreparing` holds yellow while the session is in voice mode, waiting, and has **no playable audio yet** (`VoiceGenerating || !VoiceAudioReady`) - held across the gaps and retries - and turns red only once `VoiceAudioReady` is true (there is finally something to play and act on).

This restores the `|| !VoiceAudioReady` hold that was removed on 2026-07-08. That removal traded the flicker for a wedge-proof rule, but the wedge it feared (a permanently failing voice sitting yellow forever, because \"no audio yet\" and \"gave up\" were the same value) is a **voice-reliability** concern to fix at the source - not a reason to flash red. The target is a sub-minute average; anything over three minutes is an exception to be flagged and fixed, and voice will get a terminal \"gave up\" state separately.

## Agreement check

`VoiceAudioReady` is a Gateway-only fact (the desktop never sets it), and it is now a fold input for the first time. The `StateAgreementCheck` treats it accordingly: the desktop reconstruction strips it, and a voice-ready row (Gateway red, desktop yellow) is reported as a real divergence. The old \"voice being prepared diverges\" case now **agrees** (both yellow).

The phone renders `effectiveColorHex`/`stateLabel` verbatim, so this one Gateway fold change fixes every surface.

## Tests

- `SessionOrderingTests`: no-audio-not-generating now asserts **yellow** / Active; added audio-ready = NeedsYou.
- `GatewayAddsAFactTests`: the red control now pins `VoiceAudioReady = true` (the real out-of-window case).
- `AgreementCheckFaultInjectionTests`: new audio-ready divergence test; companion proving voice-being-prepared now agrees.

Targeted suites green: Gateway.Tests 119 passed, Avalonia.Tests fault-injection 18 passed.

### DIFF
--- a/src/CcDirector.Avalonia.Tests/AgreementCheckFaultInjectionTests.cs
+++ b/src/CcDirector.Avalonia.Tests/AgreementCheckFaultInjectionTests.cs
@@ -237,21 +237,44 @@ public sealed class AgreementCheckFaultInjectionTests
     }
 
     [Fact]
-    public void VoiceBeingPrepared_IsYellowOnTheGatewayAndRedOnTheDesktop_AndIsReported()
+    public void VoiceAudioReady_IsRedOnTheGatewayAndYellowOnTheDesktop_AndIsReported()
     {
-        // THIS IS NOW THE ROW THE LIVE GATEWAY ACTUALLY SERVES, which it was not when this was written.
-        // A companion test used to sit below this one arguing that the real Gateway stamps BriefingState
-        // ="Briefing" in the same breath as VoiceGenerating, so this row was a shape production never
-        // emitted. That stamp is deleted (gap 5) - the Gateway adds VoiceGenerating and nothing else - so
-        // this row is exactly what the fleet serves, and the companion went with the stamp it described.
-        var row = Waiting("voice");
+        // THE VOICE DIVERGENCE, as it stands since the 2026-07-19 widening of IsVoicePreparing (yellow while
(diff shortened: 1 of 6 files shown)


thefrederiksen:
Merging past the red .NET check: it is the known DirectorRegistry/host FileSystemWatcher dispose-race (System.ObjectDisposedException in DI ServiceProvider disposal, aborting the run at the Gateway suite teardown), reproduced on two CI attempts but not a test failure. The exact Gateway.Tests suite passes 2848/0/10 locally on this branch, and this change only touches the display-color fold, a comment, and the agreement-check reconstruction - nothing in the host/DI lifecycle.

PR #1873, opened 2026-07-19
fix(core): authenticate the director-startup telemetry report

thefrederiksen:
Closes thefrederiksen/devthrottle#1855.

## The root cause is broader than the issue title says

The issue reads as "hosted rejects the very device key it just issued". It is not that.

`DevThrottleDirectorStartupTelemetryReporter` built its request as:

```csharp
using var request = new HttpRequestMessage(HttpMethod.Post, endpoint)
{
    Content = new StringContent(body.ToJsonString(), Encoding.UTF8, "application/json"),
};
```

**No `Authorization` header, on any path.** Every other Director-to-Gateway call sets `Bearer {_config.Token}` from `GatewayConfig` (see `GatewayClient`), and the Gateway's gate is host-wide. So this report was refused by *any* Gateway with auth on - **self-host included** - and it has been since the gate went on by default. Hosted surfaced it first only because hosted is authenticated by construction.

The credential was never missing. `GatewayCredentialStore.SaveEnrolledKey` writes the enrolled per-device key straight into `gateway.token` in `config.json`, which is the same key the tunnel and the cockpit reads use in the same boot. The reporter loaded `gateway.url` from that config and simply never read `.token`.

The second half of the issue - that it **fails silently** - is what made this survive. The caller swallows every failure by design, so the only symptom was an absence, and an absence looks exactly like nobody having started a Director.

**Suggested issue re-scope:** this is "the startup telemetry reporter never authenticates", not "hosted rejects the key". Worth correcting, because the current framing blames hosted for a general auth omission.

**Scope, ruled.** thefrederiksen/devthrottle#1855 originally expected the route to "record the event against that device's tenant". That expectation has been **dropped and the issue rescoped**: director-startup telemetry is **global operational data with no per-account attribution**.

The decisive fact, verified by reading the endpoint rather than inferred: it writes a process-global record line and, only when a forward URL is set, enqueues the body with `bearer: null` and no tenant field - and **there is no tenant-facing read route for this telemetry anywhere**. Nothing serves one tenant's startup record to another, so there is no cross-tenant leak here to close. This is the operator seeing operational data - the attribution side of the isolation-versus-attribution distinction, not a tenancy boundary.

Attribution is deferred to **#1875**, which records the hard part: the unsolved problem is how a hosted Gateway authenticates a per-account forward to the cloud when it holds no account token. That will recur for any hosted-to-cloud attributed forward, so it needs its own design rather than a tenant column bolted onto the queue - a tenant field on a queue whose forward cannot authenticate is a half-partition.

**The safeguard that makes this rescope honest rather than a quiet drop**, also recorded on thefrederiksen/devthrottle_internal#886: it is safe *only* because there is no tenant-facing read of this telemetry. If anyone ever adds one, attribution must come first - at that point it stops being operational data and becomes tenant data.

**Related, verified while here:** `DevThrottleLoginTelemetryReporter` has the structurally identical omission - it POSTs `/telemetry/login` with no `Authorization` header, and that path is not on the auth middleware's exact-match public list, so the same default gate refuses it. Booked as **#1876**, to get the same fix shape *and* the same two security guards (host-match, no-redirect), each with its own revert-proof - rather than being widened into this pull request.

## The fix

1. **Send the credential.** The report attaches `Authorization: Bearer <gateway.token>`, exactly as every other Director-to-Gateway call does, so the report ARRIVES and is recorded instead of being silently refused.

   **Scope, corrected.** The first version of this pull request claimed the Gateway "resolves the tenant from that key, so the event lands against the right account with nothing further to plumb". **That was false and I had not verified it.** The receiving endpoint never reads the request's tenant: it writes a process-global log line and, when forwarding is configured, enqueues the raw body globally with `bearer: null` and no tenant field. So this fixes the 401 and the local record; it does **not** attribute the event to an account. Caught in review.

2. **Only to its own Gateway.** `DEVTHROTTLE_STARTUP_TELEMETRY_URL` can point the report at an arbitrary address for a test, proof or staging run. Attaching the key unconditionally would hand this machine's per-device Gateway key - which authenticates *every* Director-to-Gateway call - to whatever host that variable names, by setting one environment variable. The Bearer therefore rides only on a target whose scheme, host and port match the configured Gateway. A different scheme does not qualify: sending the key over plain `http` to a host we know as `https` is exactly the downgrade this refuses.

3. **A refused credential no longer reads like a transient.** 401 and 403 get their own loud log line saying what was refused, that it is a configuration fault that will not recover on its own, that telemetry from this machine will be missing until it is fixed, and whether a credential was sent at all. The report stays best-effort - it must never delay or fail startup.

## A prior decision this reverses, deliberately

`ReportStartupAsync_SendsNoAuthorizationHeader` asserted the POST carries **no** `Authorization` header, ever, on the #642 reasoning that "the Director holds no credential, and the Gateway attaches its own token on the cloud forward (#639)".

The second half is still true and is enforced Gateway-side. The first half conflated two different hops: the Director does hold a credential **for the Gateway**, and the inbound hop is gated like every other Gateway route. The test is kept, narrowed to what survives - with no gateway token configured there is nothing to send and the reporter must not invent one - and the reasoning is recorded in place.

Also fixed while here: the constructor now reads `config.json` **only** when the target was not supplied, so the test-isolation guarantee the existing tests rely on ("every test passes an explicit `gatewayUrl` so the reporter never reads the test machine's config.json") extends to the credential. A test can never pick up the developer's own `gateway.token`.

## Proof

**Revert-prove, run verbatim against the final file.** Delete the two lines:

```csharp
if (authenticated)
    request.Headers.Authorization = new ...AuthenticationHeaderValue("Bearer", _gatewayToken);
```

The build succeeds, and **exactly two** tests redden on a null `Authorization` header: `SendsTheConfiguredGatewayTokenAsBearer` and `SendsTheCredentialWhenTheOverridePointsAtItsOwnGateway`. The remaining nine stay green - **including the cross-host test, which asserts the header is absent and therefore cannot detect this revert at all.**

That last point is why the own-gateway control exists. Without it, deleting the credential entirely would leave the cross-host guard green and the deletion would look harmless - a guard that only ever asserts absence cannot tell "correctly withheld" from "never sent".

**Redirect guard.** `AllowAutoRedirect = false` on the reporter's own handler, because a redirect moves the destination *after* the host-match check has passed. Proven behaviourally over real sockets against a server that 302s to a different origin, driving the real shipped handler - not by asserting a configuration flag. Revert-prove: set `AllowAutoRedirect = true`; the build succeeds and exactly this one test reddens on `Assert.Equal(0, destination.Requests)` with `Expected 0 / Actual 1` - the redirect target really was contacted - eleven others green. The exception is *recorded* rather than required so the security fact is the assertion that reddens first, not "the redirect was followed at all".

This deliberately does not rest on .NET clearing `Authorization` across origins. That is true today and it is framework internals; the guard should be visible in our own file.

**Gateway side.** Four new wire tests pin the half the client fix depends on and that nothing covered: a per-device key is accepted (202), the shared machine token is also accepted (so self-host is not broken by a device-key-only route), no credential is refused (401 - the reported failure reproduced exactly at the wire), and a bogus key is refused (so the two acceptance tests pass because the credential was valid, not because the route waves everything through).

These prove **admission, not attribution**. Any valid key gets in, including one bound to no tenant - which is the correct behaviour for telemetry that is global by construction, and the tests now say so explicitly rather than implying a tenancy property they do not test.

## Tests

All seven test projects run, green:

| Project | Passed | Skipped | Total |
|---|---|---|---|
| Gateway | 2872 | 10 | 2882 |
| Core | 3306 | 8 | 3314 |
| Avalonia | 247 | 0 | 247 |
| HostedAgent | 86 | 0 | 86 |
| Engine | 62 | 0 | 62 |
| Launcher | 27 | 0 | 27 |
| Terminal.Avalonia | 21 | 0 | 21 |

Gateway 2882 total against the ~2861 baseline (four added here); Core 3314 total (four added here).

## Note on CI coverage

The `#1864` repair of folding the setup test projects into `cc-director.sln` is **not** load-bearing for this change. The reporter is `src/CcDirector.Core/Account/`, the endpoint is `src/CcDirector.Gateway/Api/`, and both projects' tests are already in the solution and already gate a merge. Checked rather than assumed.

### DIFF
--- a/src/CcDirector.Core.Tests/Account/DevThrottleDirectorStartupTelemetryReporterTests.cs
+++ b/src/CcDirector.Core.Tests/Account/DevThrottleDirectorStartupTelemetryReporterTests.cs
@@ -42,7 +42,7 @@ public sealed class DevThrottleDirectorStartupTelemetryReporterTests
     }
 
     [Fact]
-    public async Task ReportStartupAsync_SendsNoAuthorizationHeader()
+    public async Task ReportStartupAsync_SendsNoAuthorizationHeader_WhenNoGatewayTokenIsConfigured()
     {
         var handler = new CapturingHandler(HttpStatusCode.OK);
         var reporter = new DevThrottleDirectorStartupTelemetryReporter(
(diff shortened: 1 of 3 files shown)


thefrederiksen:
Architect ruling on the credential-scoping decision in this change, recorded here for durability.

**The host-match gate is correct and stays.** Attaching the per-device Gateway key only when the target's scheme, host and port match the configured Gateway is the right principle - a credential is bound to its audience and must never travel to a destination that is not its owner. Attaching it unconditionally would turn a single environment variable into a credential-exfiltration primitive, since that key authenticates every Director-to-Gateway call.

**The redirect path must be closed, not just noted.** A fire-and-forget telemetry POST has no legitimate reason to follow a redirect, so `AllowAutoRedirect` should be **false** on this reporter's client. That makes an off-host header leak structurally impossible. It must not rest on the fact that .NET strips `Authorization` on cross-origin redirects by default - that is framework internal behaviour, and a security guard should be explicit and independently verifiable rather than dependent on it.

Review should adversarially attack **both**: attempt to defeat the host/scheme/port match, and confirm that no redirect can carry the Bearer off-host.

The re-scope of the underlying issue (this is a general 'never authenticated' bug that also affects self-host, not a hosted-only key rejection) is correct - the issue title has been updated to match.

PR #1888, opened 2026-07-19
fix(gateway): deny the stats dashboard on hosted (#1848, stats half)

thefrederiksen:
Closes the **stats half** of thefrederiksen/devthrottle#1848. The prompts half landed separately as thefrederiksen/devthrottle#1860.

## The defect

`GET /stats/data` served, **fleet-globally and to any tenant's device key**: every repository name the fleet has driven, the per-agent and per-model tallies, the **token spend**, and the hourly activity series. `GET /stats` serves the dashboard that reads it.

The handler was a **parameterless lambda** - it took no `HttpContext` at all, so it could not resolve a tenant even in principle. Same signature-level shape as the prompt log.

## Denied, not partitioned - and why

There is **nothing to partition by**. These are pre-aggregated fleet totals with **no tenant anywhere in the schema**, so a per-tenant answer would have to be recomputed from an attribution that was never recorded. Inventing one is a **half-partition**, which is worse than an honest refusal because it looks like isolation.

The concept does not survive the move either: this is the **owner's private view of his own gateway**, and on shared hosted infrastructure "the owner" is not a thing. There is no correct per-tenant answer to serve here - only a disclosure to close.

Contrast the prompts half, which *was* partitioned: prompt text is per-session, already tenant-attributable when it arrives, and its storage unit can physically carry the tenant. That is the test - **partition when the data is already attributable and its storage unit can carry the tenant; deny when it is an aggregate whose tenant was never recorded.**

## What changed in the rework

Rejected in review on **four** points. Each is answered below, in the review's own order.

| Review point | Answered by | State |
|---|---|---|
| 1. One `MapGroup` filter, not `DenyOnHosted` per handler; prove a newly mapped probe route **refused on hosted AND served with hosted mode explicitly off** | group filter + probe test in **both** directions | done |
| 2. An explicit hosted-off control driving both routes and their **real** page/feed payloads | `HostedStatsSelfHostControlTests` | done |
| 3. A **named** un-deny row per denied route with its real restore condition | two rows on thefrederiksen/devthrottle_internal#891 | done |
| 4. Each real production-line revert preceded by a successful build and run through the **full** Gateway suite; then restore, rebuild, rerun full | revert leg done; **restored leg outstanding** | 1 of 2 |

### 1. ONE group filter, not a guard per route

The deny is now a single `AddEndpointFilter` on the route **group** both stats routes are mapped into, replacing the `DenyOnHosted()` call that was repeated in each handler. It follows the owner-settings deny in `SettingsEndpoints.cs`, which is our own better pattern and which this pull request originally did not carry over.

A per-route guard **rots**: the moment someone adds a route to that file it is undefended, and nothing fails. A group filter runs before every route in the group **including routes that do not exist yet**.

That difference is invisible to any test that only drives `/stats` and `/stats/data` - a per-route guard and a group filter behave identically on the routes that exist today, which is exactly what makes the per-route shape dangerous. So the proof is a **future-route probe**:

`HostedStatsGroupFilterTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own`

It maps a **brand-new route** onto the group and asserts it is refused, with **no deny written for it anywhere**. `StatsPageEndpoint.Map` now returns its `RouteGroupBuilder` so that property is statable from outside the file - that return value exists solely to make this test possible.

**Proven in both directions, as the review asked.** The same probe path is also driven with hosted mode **explicitly off** by `HostedStatsSelfHostControlTests.A_route_added_to_the_group_still_serves_on_self_host`, over **both** non-hosted forms (absent, and present-but-`"0"`), asserting it is **served**. One direction alone cannot tell a working gate apart from a brick: without the served half, a filter that refused everything unconditionally would pass every hosted assertion in the file while having silently killed the route for self-host too.

A second control, `A_route_outside_the_group_still_serves_on_hosted`, maps a route **outside** the group and shows it still serves on hosted — so the passing hosted tests are the filter doing its job rather than the host refusing everything.

### 2. The self-host control is now EXPLICIT

`HostedStatsSelfHostControlTests` sets `CC_GATEWAY_HOSTED` **itself**, to **both** non-hosted forms that occur in practice - **absent**, and **present-but-`"0"`** - asserts `GatewayHostedMode.IsHosted` is actually false so no test can silently run in the mode it thinks it is not in, and then drives **both** routes.

It asserts the **real payloads**, not the absence of the refusal string: the seeded bucket counts (7 turns / 700 characters, voice / phone), the ranked repository (`devthrottle`, 7 turns), the honesty caveats, the feed's full property set, and the dashboard markup (`Your Throttle`, `/stats/data`, `What you have spent`). An empty-but-successful response would satisfy "the refusal is absent" while still being a broken self-host, so absence is not the assertion.

Leaning on `StatsPageEndpointTests` - as this pull request previously did - proves nothing about self-host: those tests never mention `CC_GATEWAY_HOSTED` and pass only because the runner happens to leave it unset. If that ambient default ever flipped they would keep passing while self-host was completely broken.

### 3. Un-deny debt, booked SEPARATELY per route

Two named rows under thefrederiksen/devthrottle_internal#891, one for `/stats` and one for `/stats/data` **individually**, each stating what self-host still does with it, what hosted does now, and the real condition before the guard comes off: https://github.com/thefrederiksen/devthrottle_internal/issues/891#issuecomment-5017821923

They are separate because their conditions are different. `/stats/data` is blocked on the stats database being partitioned or remodelled to record a tenant at aggregation time - today it is one SQLite file with **no tenant column** - plus thefrederiksen/devthrottle_internal#880, plus an unasked product question. `/stats` is blocked on **all of that AND** a rewrite of the page, which is written as the single owner's view. One shared row would let the schema work read as satisfying both, and the dashboard would come back on hosted still rendering an owner's-eye view.

## The deny shape, unchanged

- Gated on `GatewayHostedMode.IsHosted` **directly**, never an optional or nullable argument - a security branch that depends on an optional argument **fails open** when a caller forgets it.
- **404** with a body of **exactly one `error` property**, asserted by parsing the JSON and enumerating the property set as an **allow-list** - not a substring match, which cannot see an extra leaked field. 404 rather than 403 because on hosted this route does not exist as a concept; 403 would imply the right credential could reach it, and none can.
- It **REFUSES**, it does not serve empty. A zeroed series is a **false statement** where an absent one is merely absent - the mistake `/healthz` made when it zeroed its fleet counts and anything monitoring them read a permanently dead fleet.

## Revert-proof

Both legs on the **same head** (`90cf460f`), each preceded by a build confirmed at **0 errors, 0 warnings**, both run through the **full** Gateway suite under the machine-wide test lock (#1918). A run after a failed build executes a stale binary and reports a false pass; a filtered slice cannot see whether an existing test already covered the behaviour, nor collateral damage.

The revert **deletes the `AddEndpointFilter` block outright**, leaving `MapGroup("")` so the file still compiles, with **no per-route guard put back** - the hosted deny is absent entirely. Never `if (false)`: unreachable code is a build error here.

| Leg | Failed | Passed | Skipped | Total |
|---|---|---|---|---|
| Restored (filter in place) | **0** | 2965 | 10 | 2975 |
| Reverted (filter deleted) | **8** | 2957 | 10 | 2975 |

**The counts reconcile arithmetically:** 2957 + 8 = 2965, the restored passed count, with total and skipped unchanged. Nothing else moved.

**All 8 reds arrived as ASSERTIONS. Zero crashes** - seven `Assert.Equal() Failure: Values differ`, one `Assert.DoesNotContain() Failure: Sub-string found`:

| Reddened with the filter deleted | Arrived as |
|---|---|
| `HostedStatsGroupFilterTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own` | assertion |
| `HostedStatsGroupFilterTests.Both_stats_routes_are_refused_on_hosted_through_the_group_filter` (`/stats`) | assertion |
| `HostedStatsGroupFilterTests.Both_stats_routes_are_refused_on_hosted_through_the_group_filter` (`/stats/data`) | assertion |
| `HostedStatsDenyTests.The_stats_feed_is_refused_to_an_enrolled_tenant` | assertion |
| `HostedStatsDenyTests.The_stats_page_is_refused_too` | assertion |
| `HostedStatsDenyTests.The_refusal_is_not_a_zeroed_dashboard` | assertion |
| `HostedStatsDenyTests.The_refusal_carries_nothing_but_the_refusal` (`stats`) | assertion |
| `HostedStatsDenyTests.The_refusal_carries_nothing_but_the_refusal` (`stats/data`) | assertion |

**That last row is why this pull request has an extra commit.** An earlier revert leg reconciled perfectly to the same eight - and one of them was a **crash**, not an assertion: `JsonDocument.Parse` threw `'<' is an invalid start of a value`, because with the guard deleted `/stats` serves the HTML dashboard and the test parsed the body before establishing it was JSON. It still went red, so it looked correct in every count-based check. But a crash proves only that the mutation broke something upstream; it cannot say **what** was served in place of the refusal, which is the entire claim a deny revert-proof makes. Parsing is itself an unstated assertion about format. Both parse sites now assert status and media type first, so the same revert reddens as *"expected NotFound, got OK"* and *"expected application/json, got text/html"*.

**Two things only the full suite could show.** Nothing **anywhere else** in the suite reddened, so this behaviour was **not already covered**. And there was **no collateral red**.

**Green in BOTH directions, deliberately** - each is a control, and a control that moves with the change under test is not a control:

| Green through the revert | Why it must be |
|---|---|
| `An_unauthenticated_caller_is_still_rejected` | The host-wide auth gate runs before the group filter, so the deny must not have opened the route up as a side effect. |
| `A_route_outside_the_group_still_serves_on_hosted` | Proves the filter is scoped to the group, not a blanket refusal on the whole application. |
| all of `HostedStatsSelfHostControlTests` | Self-host is the control for this mission and must not move when the hosted guard does. Includes the served half of the probe proof. |

**The tests were confirmed to have EXECUTED, by name** - all 16 stats cases listed individually, 16/16 passed - rather than inferred from a green total. A green suite does not tell you your tests ran; it tells you nothing failed, and those differ exactly when a harness silently drops a class.

## Tests - all seven projects

| Project | Failed | Passed | Skipped | Total |
|---|---|---|---|---|
| Gateway | 0 | 2965 | 10 | 2975 |
| Core | 0 | 3306 | 8 | 3314 |
| Avalonia | 0 | 247 | 0 | 247 |
| HostedAgent | 0 | 86 | 0 | 86 |
| Engine | 0 | 62 | 0 | 62 |
| Launcher | 0 | 27 | 0 | 27 |
| Terminal.Avalonia | 0 | 21 | 0 | 21 |

Gateway is at head `90cf460f`. The six non-Gateway projects were run at the pre-rebase head and are unaffected by this diff, which touches only `CcDirector.Gateway` and its test project.

Rebased onto `e1b8825c`, which carries the test lock (#1918), thefrederiksen/devthrottle#1916 and v1.6.0. The branch continuous-integration red previously noted was the known unrelated account-status case; thefrederiksen/devthrottle#1916 fixes it, and it did **not** appear in either leg here - the allowance is retired.

## The pattern, for the other hosted deny pull requests to normalize onto

Four things make a hosted deny reviewable. Copy all four; each exists because leaving it out produced a defect that got through review at least once on this mission.

### 1. One group filter, never a guard per route

```csharp
var app = outer.MapGroup("");
app.AddEndpointFilter(async (ctx, next) =>
{
    if (DenyOnHosted() is { } denied) return denied;
    return await next(ctx);
});
```

The empty prefix keeps every route path written out in full, so the self-host surface is unchanged and the diff stays readable.

**Why it is not a guard per handler.** A repeated guard passes *exactly the same tests* as a group filter for the routes that exist today. That is what makes it dangerous rather than merely untidy: the difference only appears on the route somebody adds next, when it is open on hosted by default and **nothing fails**. The deny rots silently as the file grows.

Gate on `GatewayHostedMode.IsHosted` **directly** — never an optional or nullable argument, which **fails open** the moment a caller omits it.

### 2. The probe test — and it must be proven in BOTH directions

This is the assertion the whole change exists for, and it is the one most likely to be skipped, because everything looks green without it.

- **Refused on hosted:** map a **brand-new** route onto the group and assert it is refused, with **no deny written for it anywhere**. This is the only test that distinguishes a group filter from a per-route guard.
- **Served with hosted mode explicitly off:** drive the **same** probe path with hosted mode explicitly disabled and assert it is **served**.

One direction alone cannot tell a working gate apart from a brick. Without the served half, a filter that refused everything unconditionally would pass every hosted assertion in the file while having silently killed the route for self-host too.

To make this statable at all, the map method must **return its `RouteGroupBuilder`** — otherwise nothing outside the file can say anything about routes added to the group.

Add the scoping control too: a route mapped **outside** the group still serves on hosted, proving the filter is scoped rather than a blanket refusal.

### 3. Self-host proven explicitly, never inherited

Set `CC_GATEWAY_HOSTED` **yourself**, to **both** non-hosted forms — **absent**, and **present-but-`"0"`** — and assert `GatewayHostedMode.IsHosted` is actually false so no test can silently run in the mode it thinks it is not in.

Then assert the **real payloads**: the seeded counts, the ranked rows, the actual page markup. **Not** the absence of the refusal string — an empty-but-successful response satisfies "the refusal is absent" while still being a completely broken self-host.

A control that rests on the test runner's ambient default proves nothing: if that default ever flips, it keeps passing while self-host is broken.

### 4. Assert the refusal shape as an allow-list, and make every red an ASSERTION

- **404**, not 403 — on hosted the route does not exist as a concept; 403 would imply the right credential could reach it, and none can.
- Body is **exactly one `error` property**, asserted by parsing the JSON and **enumerating the property set** as an allow-list. Not a substring match, which cannot see an extra leaked field, and not a deny-list of today's keys, which rots as the payload grows.
- **Refuse, never serve empty.** A zeroed series is a *false statement* where an absent one is merely absent.
- **Assert the status and media type BEFORE parsing the body.** This one is easy to miss and this pull request got it wrong first time round: with the guard deleted, `/stats` serves the HTML dashboard, and `JsonDocument.Parse` died with `'<' is an invalid start of a value`. The test still went red, so it looked fine in a count — but a **crash** proves only that the mutation broke something upstream, and cannot say **what** was served in place of the refusal. That is the whole claim a revert-proof makes. Assert status and media type first and the same revert reddens as *"expected NotFound, got OK"* and *"expected application/json, got text/html"*.

### The proof standard

Both legs on the **same head**, each preceded by a build confirmed at **0 errors** — a run after a failed build executes a stale binary and reports a false pass. Revert by **deleting** the filter outright; never `if (false)`, which is unreachable code and a build error here.

Run the **full** Gateway suite both ways, not a filtered slice. A filtered revert-proof answers only "do my new tests fire?" — it cannot see whether an existing test already covered the behaviour, and it cannot see collateral damage.

Then reconcile the counts arithmetically rather than eyeballing them: **revert passed + revert failed must equal restored passed**, with total and skipped unchanged. If it does not reconcile, something else moved and that needs finding before either number is reported.

Confirm your tests **executed**, by name — a green suite does not tell you your tests ran, it tells you nothing failed, and those differ exactly when a harness silently drops a class.

### DIFF
--- a/src/CcDirector.Gateway.Tests/HostedStatsDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedStatsDenyTests.cs
@@ -496,14 +496,21 @@ public sealed class HostedStatsSelfHostControlTests : IDisposable
     }
 
     /// <summary>
-    /// The mirror of the future-route probe: on self-host the group filter lets a route added to the group
-    /// through untouched. Without this, "the filter refuses everything, always" would pass every hosted test
-    /// in this file.
+    /// THE SECOND HALF OF THE FUTURE-ROUTE PROOF, and the review asked for both halves by name: the probe
+    /// route must be REFUSED on hosted and SERVED with hosted mode EXPLICITLY off. The hosted half is
+    /// <see cref="HostedStatsGroupFilterTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own"/>;
+    /// this is its mirror, on the SAME probe path, with hosted mode stated rather than assumed - and stated in
+    /// both non-hosted forms, absent and present-but-"0".
+    ///
+    /// Without this half, "the filter refuses everything, always" would pass every hosted test in this file
+    /// while having silently killed the route for self-host too. One direction alone cannot tell a working
+    /// gate apart from a brick.
     /// </summary>
-    [Fact]
-    public async Task A_route_added_to_the_group_still_serves_on_self_host()
+    [Theory]
+    [MemberData(nameof(NonHostedValues))]
+    public async Task A_route_added_to_the_group_still_serves_on_self_host(string? hostedValue)
     {
-        DeclareSelfHost(null);
+        DeclareSelfHost(hostedValue);
 
         var (app, http) = await StatsGroupProbeHost.StartAsync(
             SeededAggregator(),
--- a/src/CcDirector.Gateway.Tests/HostedStatsDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedStatsDenyTests.cs
@@ -3,7 +3,9 @@ using System.IO;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Linq;
 using System.Net.Sockets;
+using System.Text.Json;
 using System.Threading.Tasks;
 using CcDirector.Gateway;
 using Xunit;
@@ -27,8 +29,9 @@ namespace CcDirector.Gateway.Tests;
 ///
 /// IT IS A REFUSAL, NOT AN EMPTY DASHBOARD. Serving zeroed series would be a FALSE statement rather than an
 /// absent one - the same mistake /healthz made when it zeroed its fleet counts and anything monitoring them
-/// read a permanently dead fleet. The tests below assert the refusal AND the absence of the payload keys, so
-/// a future "helpful" empty-shape response fails here rather than passing as a deny.
+/// read a permanently dead fleet. The refusal body is asserted as an EXACT PROPERTY SET - one error field and
+/// nothing else - rather than as an absence of known payload keys. A deny-list of today's keys was tried and
+/// review broke it in one move; see the note on that test for why an allow-list is the only shape that holds.
 ///
 /// Revert-prove: delete the <c>DenyOnHosted()</c> guard from either route and that route's tests go RED -
 /// the data route with a 200 carrying the fleet-global keys, the page route with the HTML dashboard.
@@ -84,16 +87,33 @@ public sealed class HostedStatsDenyTests : IAsyncLifetime
         Assert.Equal(HttpStatusCode.NotFound, resp.StatusCode);
     }
 
-    [Fact]
-    public async Task The_stats_feed_discloses_none_of_its_payload()
+    [Theory]
+    [InlineData("stats/data")]
+    [InlineData("stats")]
+    public async Task The_refusal_carries_nothing_but_the_refusal(string path)
     {
-        // Asserting the status alone would let a future "helpful" change serve an empty-but-shaped payload
-        // under some other code and still pass. These are the keys that carry the disclosure - repository
-        // names, agent and model tallies, token spend - and none of them may appear in any form.
-        var body = await (await Get("stats/data", _key)).Content.ReadAsStringAsync();
+        // AN ALLOW-LIST, NOT A DENY-LIST, and the difference is the whole test.
+        //
+        // This first asserted that a handful of known payload keys were absent. Review broke it in one move:
+        // a denial envelope carrying generatedAtUtc, timeZone, concurrency and notCaptured passed all five
+        // tests. Worse, the real feed has keys a substring deny-list silently misses - "tokenSpendByHour"
+        // does not contain the string "tokenSpend" once the closing quote is included - so a data-bearing
+        // partial dashboard could pass while every listed key was "absent".
+        //
+        // A deny-list also rots by construction: it protects against the payload as it is TODAY, and every
+        // field added to the feed later is unprotected until someone remembers to add it here. Asserting the
+        // property set is EXACTLY one error field inverts that - anything empty-shaped, anything new, and
+        // anything metadata-looking reddens automatically without this test being touched.
+        var resp = await Get(path, _key);
+        var body = await resp.Content.ReadAsStringAsync();
 
-        foreach (var key in new[] { "repos", "agents", "models", "tokenSpend", "hourlyTurns", "buckets", "wingman" })
-            Assert.DoesNotContain($"\"{key}\"", body, StringComparison.Ordinal);
+        using var doc = JsonDocument.Parse(body);
+        Assert.Equal(JsonValueKind.Object, doc.RootElement.ValueKind);
+
+        var properties = doc.RootElement.EnumerateObject().Select(p => p.Name).ToArray();
+        Assert.Equal(new[] { "error" }, properties);
+        Assert.Equal("the stats dashboard is not available on the hosted gateway",
+            doc.RootElement.GetProperty("error").GetString());
     }
 
     [Fact]
@@ -111,6 +131,8 @@ public sealed class HostedStatsDenyTests : IAsyncLifetime
     {
         // The /healthz lesson, applied here on purpose: a zeroed count is a FALSE statement where an absent
         // one is merely absent. A caller must not be handed a dashboard that reads "no work has been done".
+        // The exact-property-set test above is what actually enforces this; this one states the intent in the
+        // terms the mistake was originally made in, so the reason survives even if the shape of the check changes.
         var body = await (await Get("stats/data", _key)).Content.ReadAsStringAsync();
 
         Assert.DoesNotContain("\"turns\":0", body, StringComparison.Ordinal);
--- a/src/CcDirector.Gateway.Tests/HostedStatsDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedStatsDenyTests.cs
@@ -154,6 +154,15 @@ public sealed class HostedStatsDenyTests : IAsyncLifetime
         var resp = await Get(path, _key);
         var body = await resp.Content.ReadAsStringAsync();
 
+        // ASSERT THE STATUS AND THE MEDIA TYPE BEFORE PARSING, so this test reddens as a STATEMENT rather
+        // than as a crash. Without these two lines, deleting the guard makes /stats serve the HTML dashboard
+        // and JsonDocument.Parse dies with "'<' is an invalid start of a value" - the test still goes red, but
+        // it goes red as a JsonReaderException, which proves only that the mutation broke something upstream.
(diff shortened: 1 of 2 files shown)


thefrederiksen (commented):
REJECT

1. `/stats` and `/stats/data` each call `DenyOnHosted` inside their own handlers instead of sitting behind one `MapGroup` filter. A later stats route is open on hosted by default. Settle this by putting the family behind one group filter and proving a newly mapped probe route is refused on hosted and served with hosted mode explicitly off.

2. The cited `StatsPageEndpointTests` never explicitly clear `CC_GATEWAY_HOSTED`. They prove the runner's ambient default, not that the same gate leaves self-host working. Add an explicit hosted-off control that drives both routes and their real page/feed payloads.

3. #1886 is mentioned generically, but there is no named un-deny row for each denied route with its real restore condition. List `/stats` and `/stats/data` separately so the deny is a tracked promise rather than a silent permanent removal.

4. The guard reverts were exercised only in a focused slice. The mission requires each real production-line revert to be preceded by a successful build and run through the full 2900+ Gateway suite, with the intended assertions red, controls green, and a counted summary; then restore, rebuild, and rerun full.

The direct `GatewayHostedMode.IsHosted` gate, 404 response, exact one-`error`-property allow-list, and current target assertions are otherwise sound. The branch CI red is the known unrelated account-status case, not this diff.

PR #1895, opened 2026-07-19
docs(testing): a plain-English guide to trying the hosted gateway

thefrederiksen:
The Architect has ruled the owner can dogfood hosted as the only tenant. This is what he needs to actually do it.

## Why it is a separate Director

His everyday Director on this machine **is** the one running the whole agent fleet. Enrolling it at hosted would move every running session with it. So dogfooding has to be a second, isolated Director — not a reconfiguration of his.

The guide opens with that, in those terms, because it is the one thing that must not be misunderstood.

## What it gives him

Three double-clickable scripts, and a page written for him rather than for an engineer:

| Script | What it does |
|---|---|
| `Start hosted DevThrottle.cmd` | starts it; refuses politely if already running |
| `Stop hosted DevThrottle.cmd` | asks it to shut down cleanly, so it leaves no stray "interrupted" entry |
| `Show what the cloud sees.cmd` | asks the **cloud** over the internet and prints the answer |

The third is the important one. It does not ask the program on his PC whether it thinks it is connected — it queries the hosted gateway directly and prints the Directors and sessions it reports. "Am I really on hosted?" becomes something he can **see** rather than something he is told.

## What does not work yet, stated plainly

Listed so none of it reads as his mistake:

- **No cockpit and no phone view on the cloud image at all** — the assets are deliberately excluded from the container, so the desktop window is the only way to see hosted sessions today (#1892, now being treated as a go-live gate, since the stated reason for excluding them is out of date)
- **Dictation is off by design** (#1884)
- **Cross-account separation is not observable** from a single account
- **`/login` rejects a device key** that works fine as a cookie — a usability landmine that would otherwise convince him his key is wrong (#1892)

## Note on the migration guard

This guide is only safe to hand over **after thefrederiksen/devthrottle#1891 merges** and the slot build is rebuilt from it. Until then, starting a Director against a cleaned isolated root re-copies the owner's real session history straight back into it — `CopyFileIfNewer` copies whenever the destination is missing, and the migration runs on every start.

Script copies live beside the guide because their absolute paths make them machine-specific, and one-copy knowledge has cost this project real time today.

### DIFF
--- a/docs/testing/hosted-dogfood/Show what the cloud sees.cmd
+++ b/docs/testing/hosted-dogfood/Show what the cloud sees.cmd
@@ -1,0 +1,27 @@
+@echo off
+REM =====================================================================
+REM  Asks the CLOUD gateway what it can see, and prints the answer.
+REM
+REM  This is the honest check that you are really on hosted. It does not
+REM  ask the program on this PC - it asks the cloud directly, over the
+REM  internet, and prints whatever the cloud says.
+REM
+REM  If your sessions appear here, they genuinely reached the cloud.
+REM =====================================================================
+
+powershell -NoProfile -ExecutionPolicy Bypass -Command ^
+  "$key = (Get-Content 'D:\ReposFred\_wt\hosted-clientleg\testroot\config\director\gateway-token.txt' -Raw).Trim();" ^
+  "$gw  = 'https://devthrottle-gw.azurewebsites.net';" ^
+  "Write-Host '';" ^
+  "Write-Host ('Asking the cloud gateway: ' + $gw);" ^
+  "Write-Host '';" ^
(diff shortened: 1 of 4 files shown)


thefrederiksen:
Merging with the .NET check **red**, knowingly and on the business owner's explicit call.

The failure is thefrederiksen/devthrottle_internal#894 — `HostedAccountStatusTests.Enrolled_without_a_recorded_email_is_signed_in_with_the_identity_absent` — and it is **categorically unrelated to this diff**, which is one markdown file and three `.cmd` scripts. The same assertion has now failed on three separate documentation-only branches plus a storage guard in a different assembly. A markdown file cannot change what a tenant registry returns.

It was **investigated rather than re-run**: three hypotheses ruled out by evidence (parallel-class race, subject collision, root baked into a static), with the mechanism deliberately left unnamed because it has not been observed. See the investigation on thefrederiksen/devthrottle_internal#894.

Overriding an investigated, documented, categorically-unrelated red check is a different act from overriding an uninvestigated one. Recording that distinction here so the override is auditable rather than a shrug.

PR #1898, opened 2026-07-19
fix(gateway): deny the whole owner-settings group on the hosted gateway

thefrederiksen:
Closes thefrederiksen/devthrottle#1863. Un-deny ledger: thefrederiksen/devthrottle_internal#908.

**This branch is a rebuild, not an edit.** The head this pull request carried until now (`c0ab3540`) was the abandoned first attempt: two files, twenty-two routes, based at `4fe2c3bb`. It was rejected twice and its base predated almost everything since merged. What is here now was written from scratch against current `origin/main` (`f5bb926f`), covers **thirty-one** routes across **three** endpoint classes, and answers all three reviews point by point. Nothing was salvaged from the old head.

## The defect

Thirty-one routes across `SettingsEndpoints`, `AiModelsEndpoint` and `TelemetryConsentEndpoint` read and write **process-global configuration with no tenant dimension anywhere in them**. `config.json` is one file for the whole process. On the hosted Gateway that means every write was a **fleet-wide mutation performed by whichever authenticated caller happened to send it** - and public signup on the backing account project is open, so "any authenticated caller" is the public.

Two of them are not disclosures but actions on shared infrastructure:

- `POST /gateway/brain/restart` starts a real coding-agent process, with permission prompts skipped, in a directory under the shared storage root.
- `PUT /gateway/injected-text` rewrites the preamble every agent launched on that machine is given at startup.

And `GET /gateway/injected-text` hands back the owner's own custom agent-launch instruction text - his words - to any caller.

## Denied, not partitioned - and why

There is **nothing to partition by**. These are not per-tenant records whose storage unit could carry a tenant; they are single global values, one per key. A per-tenant answer would have to be invented from an attribution that was never recorded, which is a **half-partition** - worse than an honest refusal, because it looks like isolation.

The concept does not survive the move either: this is the **owner's control panel for his own Gateway**, and on shared hosted infrastructure "the owner" is not a thing. Same reasoning already applied to the stats dashboard on thefrederiksen/devthrottle#1888.

**404 rather than 403**: on hosted these routes do not exist as a concept, so "not here" is the truthful answer. A 403 would imply some credential could reach them, and none can.

**It refuses, it does not serve empty.** An empty settings snapshot is a *false* statement where an absent one is merely absent.

## What the reviews asked for, and where each is answered

| Review point | Answered by | State |
| --- | --- | --- |
| 1. Five write routes had no deny proof - training capture, addressing mode, snooze default, snooze presets, time zone | All thirty-one routes and verbs are now one table, `OwnerSettingsRoutes`, driven by both sides | done |
| 2. Self-host controls never cleared the hosted variable and covered only fragments | `HostedOwnerSettingsSelfHostControlTests` sets it itself, to **both** non-hosted forms, asserts it took, and drives the whole family | done |
| 3. No named un-deny row per denied route | thefrederiksen/devthrottle_internal#908 - thirty-one named rows, each with its own condition | done |
| 4. No future-route probe | `HostedOwnerSettingsGroupFilterTests` plus `HostedOwnerSettingsSelfHostProbeTests`, **both directions**, all three families | done |
| 5. Media type not asserted before parsing, so a masked route crashes instead of distinguishing | Status and media type asserted before every parse, on both sides | done |
| 6. Revert-proof run only against a focused slice | Re-registered below against the new structure; **outstanding - needs the test lane** | **pending** |
| 7. Both `outer` and the guarded group in scope at every route mapping - 31 uncounted bypasses | Each family's routes moved into a private `MapRoutes` receiving only the guarded group; the bypass no longer compiles | done |
| 8. Brain restart had no served-side handler proof, and the mutation plan would spawn a process | Injected restart seam - exact path and verb, handler receipt, destructibility control, no process on any arm | done |
| 9. Transcription mode had no observable write effect | Proved by a seeded transition off the opposite stored value | done |
| 10. The uniform no-purge ledger claim is not established | Withdrawn and corrected - every row now requires quarantine or migration first | done |

## The deny shape

Each of the three endpoint classes puts **one endpoint filter on the route group it maps**, gated on `GatewayHostedMode.IsHosted` **directly** - never on an optional or nullable argument, which fails **open** the moment a caller omits it. The group uses an empty prefix, so every route path stays written out in full and the self-host surface is byte-identical.

A guard repeated per handler **rots**: the route added to one of these files next year would be open on hosted by default and nothing would fail. A group filter runs before every route mapped below it, **including routes that do not exist yet**.

That difference is invisible to any test that drives only the thirty-one routes that exist today - a per-route guard and a group filter behave identically on all of them. **That is precisely why the property the design rests on is the one nothing otherwise tests.** So all three `Map` methods now return their `RouteGroupBuilder`, and `HostedOwnerSettingsGroupFilterTests` maps a **brand-new route** onto each group and finds it refused with no deny written for it anywhere. The served half - the same probe paths, hosted mode explicitly off, in both non-hosted forms - is `HostedOwnerSettingsSelfHostProbeTests`. One direction alone cannot tell a working gate from a brick: a filter that refused everything unconditionally would pass every hosted assertion while having silently killed the routes for self-host too.

`TelemetryConsentEndpoint` is mapped onto the **outer** builder, not into the settings group, deliberately: it carries its own filter, and two filters over the same routes would mean deleting either one alone left them still refused - so neither could be attributed by a revert. One family, one filter, one thing to revert.

## How thirty-one routes are covered - the rule, stated rather than sampled

**Every route and verb is driven on both sides. There is no sampling.** The route table is written once and both sides read it, which is what makes "every route, both directions" checkable rather than a sentence in a pull request.

Hosted side: all thirty-one, each refused with a body whose property set is **exactly** `["error"]`, asserted as an allow-list over the whole set - not a substring, which cannot see an extra leaked field, and not a deny-list of today's keys, which rots as the payload grows.

Self-host side, the harder half, because "the refusal is absent" is satisfied by an empty 200, by a single-page-application shell, and by a route that was deleted. Every route gets a **positive** fact instead, by one of four rules:

| Rule | Routes | What it asserts |
| --- | --- | --- |
| Real payload, field by field | 11 reads | The actual fields, and where independently knowable, the actual value read back through the Core config class |
| Independently re-read effect | 13 writes | The value goes over the wire, then comes back out of the **configuration store**, not the response - a handler that accepted and wrote nothing returns the same 200 |
| Served, and reaches a handler - the deliberately NARROWED claim | `PUT /gateway/transcription-mode` | See below. This route can carry no stronger claim, and that is proved rather than asserted |
| Effect that is a **reset**, proved against a sentinel | `PUT /gateway/ai-provider` | A value the reset cannot produce is written first, so "unchanged" and "reset" cannot be confused |
| Handler-unique receipt | 4 routes | A status and sentence only that one handler can produce - the brain-config rejection, the two credential-resolution 503s, and autostart's `{ supported: false }` |

That is thirty. The thirty-first is `POST /gateway/brain/restart`, and it is the **one deliberate exception**: its whole job is to start a process, so invoking it on the served side would spawn one. It is proved instead by a **routing-table existence fact** - a GET to that path answers 405 with `Allow: POST`, a positive statement from the routing table that a POST endpoint is registered there, on a probe host with no single-page-application fallback where an unmapped path answers a bare 404 with an empty `Allow`. Put beside the POST being refused 404, the two say: the route exists on hosted, and the thing turning it away is the deny. **It proves the route is registered; it does not exercise the handler and makes no claim to.** That is written at the test rather than left to be discovered.

### The one narrowed claim, and why narrowing was the honest answer

Review asked for `PUT /gateway/transcription-mode` to either **seed a persistent transition** or have its **claim explicitly narrowed** if the route has no distinguishable mutation.

**The transition was attempted first, and continuous integration reddened it - correctly.** Seeding the store to the other enum value and reading it back returns `DevThrottle`. The setting is **single-valued by construction**: `TranscriptionMode.Byo` serialises to `"devthrottle"`, and `Parse` maps every legacy value (`"byo"`, `"openai"`, `"local"`, null, blank) forward to `DevThrottle`. No reachable stored state differs from what the route writes, so **no seeding strategy could distinguish the handler from a no-op**. That failure is what establishes the second option is the right one; it was not chosen to avoid work.

So the claim is now stated as exactly what it proves - the route is served on self-host and reaches a handler that reports the setting - and it is **no longer counted as an independently re-read effect, because it is not one**.

The narrowing is itself proved: `The_transcription_mode_setting_is_single_valued_so_no_write_is_observable` pins that every value the type can serialise writes `"devthrottle"` and every value it parses resolves to `DevThrottle`. **That test also reddens the day a second transcription mode is added**, so the caveat carries its own expiry date rather than outliving its cause and being read years later as a still-true fact.

One further caveat, called out rather than left looking like the others:

- `GET /gateway/ai/models` is proved by receipt rather than payload, because a real catalog would require a provider credential and an outbound call.

## Format facts before parsing, on both sides

Status and media type are asserted **before** any body is parsed. Parsing is itself an unstated assertion about format: on this Gateway a 404 is not necessarily JSON - the single-page-application fallback answers unmatched paths and can reply with plain text, or on a release host a 200 with HTML - so a test that parses first turns a real finding into a **crash**. A crash proves something upstream broke; it cannot say **what was served in place of the refusal**, which is the entire claim a deny makes. This was review point 5 and it applies to the served side just as much.

## Test isolation - and why it ships here rather than separately

Six existing test classes drive routes that are now mode-sensitive. Before this change those routes answered the same way in both modes, so none of them cared what `CC_GATEWAY_HOSTED` said; now a hosted-mode class running in parallel would turn their 200 into a 404, intermittently, for reasons entirely outside their own files. **The deny is what creates that need, so the fix belongs in the change that creates it** - shipping the deny without it would be shipping a defect into somebody else's tests.

They move into a new `GatewayHostedMode` collection with parallelization disabled. That is **strictly stronger** than the `DirectorRoot` collection five of them carried - running alone subsumes running serially with a subset - so nothing was weakened by the move. A search of `origin/main` for every one of the thirty-one route paths across all test projects returns exactly these six files, so the move is complete rather than sampled.

## The un-deny sentence - CORRECTED, and it is not what I first claimed

The first version of this pull request said the sentence was uniformly *"the write is stopped"* across all thirty-one routes and that **no purge would be needed** before a future un-deny. **That was an overclaim. It is withdrawn.**

The error was in the frame. The sweep behind it asked *"which routes write these configuration keys"*, found every one inside the three files being denied, and concluded nothing else could write. But a deny closes **one door**. The question that supports a no-purge claim is not whether the denied routes write - it is whether **anything** writes.

The widened sweep:

- `CcDirectorConfigService.MergePatch` is the single write authority for `config.json`, and it is called from **many production surfaces outside these three files** - the desktop application, the Director's control interface, onboarding, the browser-default store, the agent registry, telemetry settings. Those are other processes, which makes the claim a **deployment-isolation** claim, and deployment isolation is not proved anywhere.
- A hosted `config.json` **can already hold values written before this deny existed**, or by an operator. The deny stops future writes through these routes; it does nothing about what is already there.

Re-reading the store after nineteen requests - all the original evidence did - refutes neither.

### The sweep, stated precisely enough to be audited

The question asked was **"what writes this state"**, not "what writes it through these routes". Three legs:

1. **Is `MergePatch` the only production write authority for `config.json`?** Yes. `CcDirectorConfigService` has exactly one write path - `MergePatch`, which serialises to a temporary file and atomically moves it over `config.json`. A search for any direct write to that file across the repository returns **only test fixtures**, no production code.
2. **Does anything inside the hosted Gateway process write these keys outside the three denied files - in particular at host startup?** **No.** Every `MergePatch` call and every `*Config.Set` call reachable in `CcDirector.Gateway` and `CcDirector.HostedAgent` is inside `SettingsEndpoints.cs`, `AiModelsEndpoint.cs`, or `TelemetryConsentEndpoint.cs`. There is no startup writer and no background writer. This was checked specifically, because the sibling key-vault unit found exactly that failure - a writer running at host startup, before the hosted check everyone assumed guarded it. **That failure mode does not exist here**, and it is recorded as checked rather than assumed.
3. **Does anything OUTSIDE that process write the same file?** **Yes** - the desktop application, the Director's control interface, onboarding, the browser-default store, the agent registry and the telemetry settings all call `MergePatch`. They are other processes, so whether they can reach a hosted Gateway's `config.json` is a **deployment-isolation** question, and deployment isolation is not proved anywhere.

Leg 3, plus the fact that a hosted `config.json` **can already hold values written before this deny existed**, is why the no-purge claim is withdrawn. Leg 2 is why the narrower claim - the write through these handlers is stopped - survives.

**So every row is now the second kind of sentence:**

> Removing the deny requires ALSO purging or partitioning what accumulated behind it - specifically, the pre-existing process-global settings must be quarantined or migrated before any tenant-scoped un-deny, unless deployment isolation is proved at that time.

The narrower statement that **is** proved, and is kept as such: **the write through these handlers is stopped** - every one of the nineteen non-read routes is refused before its handler runs, proved by an independent re-read of the configuration store, so no tenant can add to that state through this surface while the deny holds. That is real, and it is simply not the same claim.

thefrederiksen/devthrottle_internal#908 carries this correction in full, at the top, above the thirty-one rows.

## The bypass the earlier revision shipped, and why it now does not compile

The previous head advertised three bypassable primitives. **That count was wrong, and the review was right to reject it.** Both `outer` (ungrouped) and the guarded group were in scope where every production route was mapped, so **any one of the thirty-one routes could be moved onto `outer` by a one-word edit** - opening that route alone while every other stayed correctly denied. The future-route probes could not see this: they prove a route *placed on the group* inherits the filter, never that the production routes are *on* the group. The true count was thirty-four, and thirty-one of them had no proof at all.

Each family's routes now live in a private `MapRoutes` that receives **only** its guarded `RouteGroupBuilder`. Inside it there is nothing to map onto except the guarded group, so the bypass is **inexpressible rather than merely unlikely**. This is the shape the key-vault rework uses in `VaultEndpoints`, which is already reviewed; it is worth comparing the two directly.

Verified rather than asserted: injecting `outer.MapGet(...)` inside `MapRoutes` fails to build with

```
error CS0103: The name 'outer' does not exist in the current context
```

`TelemetryConsentEndpoint.Map(outer)` is called from `SettingsEndpoints.Map`, the one method that still holds `outer`, precisely so `outer` need not stay in scope where routes are mapped.

**Thirty-four primitives down to three, by design.** Three is the floor for three families: a shared guard helper would still leave three deletable call sites, so it would reduce the number by argument rather than by construction. Collapsing to a single primitive is possible - one guarded group shared by all three families - but it would force one refusal message across three families and merge three revert attributions into one. That is a design call worth making deliberately rather than as a side effect of chasing a smaller number.

## Revert-proof - re-registered against the new structure

**Outstanding; needs the test lane.** Registered before the runs so a predicted red that fails to appear is a finding.

**Three mutations, three full Gateway suite runs, plus a green baseline and a restored run.** Each mutation **deletes the `AddEndpointFilter` block outright**, leaving `MapGroup("")` so the file compiles, with no per-route guard put back. Never `if (false)` - unreachable code is a build error here. Every run is preceded by a build at 0 errors and 0 warnings, because a run after a failed build executes a stale binary and reports a false pass.

The thirty-one mapping bypasses are **not** mutation-proved, deliberately: they are compile errors, and a mutation arm that cannot build is not an arm. That is the reduction the scope split buys.

| Mutation | Predicted red | Must stay green |
| --- | --- | --- |
| A. Delete the filter in `SettingsEndpoints.Map` | The 22 settings rows of `Every_owner_settings_route_is_refused_on_hosted`; `The_refusal_is_not_an_empty_settings_snapshot`; `A_refused_write_does_not_reach_the_configuration_store`; `The_refused_brain_restart_never_invokes_the_restart_action`; the `settings` row of the future-route probe | The 9 models and telemetry rows; `An_unauthenticated_caller_is_still_rejected`; all 3 rows of `A_route_outside_the_group_still_serves_on_hosted`; the brain-restart registration fact; every self-host control and probe row |
| B. Delete the filter in `AiModelsEndpoint.Map` | The 7 models rows; `A_refused_write_does_not_reach_the_configuration_store`; the `models` probe row | The 24 settings and telemetry rows; `The_refusal_is_not_an_empty_settings_snapshot`; `The_refused_brain_restart_never_invokes_the_restart_action`; all the same controls |
| C. Delete the filter in `TelemetryConsentEndpoint.Map` | The 2 telemetry rows; `A_refused_write_does_not_reach_the_configuration_store`; the `telemetry` probe row | The 29 settings and models rows; `The_refusal_is_not_an_empty_settings_snapshot`; all the same controls |

**Every predicted red must arrive as an assertion, not a crash.** Reds are classified by arrival; a crash red counts as a failure of the proof, because it cannot say what was served.

**The spawn hazard is gone.** The earlier plan knowingly drove the live restart path on mutation A and could have left a coding-agent process behind - the reviewer was right that this is not an acceptable proof harness. The hosted deny class now installs a non-spawning restart seam, so **no arm of this sweep can start a process**, and `The_refused_brain_restart_never_invokes_the_restart_action` turns that seam into a stronger deny assertion than the status code alone.

## Continuous integration - green on attempt 2, and the red is UNATTRIBUTED rather than fixed

`Build & Test (web)`: pass. `Build & Test (.NET)`: **failed on attempt 1, passed on attempt 2, on the identical commit `eebedebb`** with nothing changed between them (run 29733899417, `run_attempt` 1 = failure, 2 = success).

**This is stated as an unattributed red, not a fixed one.** A re-run is a diagnostic, never a remedy, and one flip on a timing-sensitive test does not establish a cause. It establishes only that the result is not deterministic on this commit.

The red, named specifically rather than waved at:

- `CcDirector.Gateway.Tests.TunnelMechanismProofTests.Producer_blocksOnAStalledSink_overTheRealSignalRStreamBuffer`, at line 304 - `Assert.Equal(pinned, Volatile.Read(ref yielded))`, expected 35, actual 57.
- That assertion requires an up-stream producer, already pinned by full socket and pipe buffers, to advance by **exactly zero** over a further one-second window. On a loaded shared runner it advanced by 22.

**Why this change is not the cause, stated precisely rather than glibly.** The test boots a real `GatewayHost`, and this branch does modify `GatewayHost.cs` - so "my diff does not touch that file" would be false and is not the claim. The claim is narrower and checkable: the entire production change to that file is **two added code lines and no deletions or modifications** -

```
+    internal Func<CancellationToken, Task> BrainRestartAction { get; set; }
+        BrainRestartAction = ct => Brain.RestartAsync(ct);
```

- a property and one constructor assignment, on the brain-restart path, which this test never invokes. Nothing in the hub, the streaming path, the MessagePack protocol, or the backpressure mechanism is altered by this branch. The other three production files are the owner-settings endpoint classes.

**The prior head `403f904e` is the control**: that run reddened on two of my own transcription-mode tests and this tunnel test **passed**. So the tunnel test has passed and failed on both heads of this branch, which is the shape of a timing flake rather than a regression.

**Booked as its own finding: thefrederiksen/devthrottle_internal#910**, carrying the two-attempt evidence, the run and job identifiers, and the mechanism - the test declares the producer "pinned" after roughly 800 milliseconds without advance, then asserts it advances by **exactly zero** over a further second. That promotes a temporal sample to a permanent property, which no widening of either window can make deterministic; the file's last commit, thefrederiksen/devthrottle#1471, was exactly such a widening. That history is recorded there as evidence the problem is **known**, explicitly not that it is acceptable.

thefrederiksen/devthrottle_internal#910 also notes that the flaky assertion is **stricter than the property the test exists to prove**: backpressure is already asserted soundly by the two `< total` checks, neither of which is flaky.

**No standing flaky-test exemption is claimed here, and thefrederiksen/devthrottle_internal#910 says so in its own text.** This red is named, bounded, and re-checked on its own head, and any future occurrence would have to be re-checked the same way rather than waved through by citing the issue.

## Build

`CcDirector.Gateway` and `CcDirector.Gateway.Tests` build at **0 errors, 0 warnings** on this head.

### DIFF
--- a/src/CcDirector.Gateway.Tests/AiModelsEndpointTests.cs
+++ b/src/CcDirector.Gateway.Tests/AiModelsEndpointTests.cs
@@ -16,7 +16,7 @@ namespace CcDirector.Gateway.Tests;
 /// /gateway/ai-provider snapshot, and the list route reports "not signed in" (503) with no key - never a
 /// crash or a silent empty list.
 /// </summary>
-[Collection("DirectorRoot")]
+[Collection("GatewayHostedMode")]
 public sealed class AiModelsEndpointTests : IAsyncLifetime
 {
     private readonly string _root;
--- a/src/CcDirector.Gateway.Tests/AiProviderEndpointTests.cs
+++ b/src/CcDirector.Gateway.Tests/AiProviderEndpointTests.cs
@@ -15,7 +15,7 @@ namespace CcDirector.Gateway.Tests;
 /// ephemeral port with CC_DIRECTOR_ROOT redirected to a temp dir, so writes round-trip an isolated
 /// config.json rather than the user's real one.
 /// </summary>
-[Collection("DirectorRoot")]
+[Collection("GatewayHostedMode")]
 public sealed class AiProviderEndpointTests : IAsyncLifetime
 {
     private readonly string _root;
--- a/src/CcDirector.Gateway.Tests/BrainConfigEndpointTests.cs
+++ b/src/CcDirector.Gateway.Tests/BrainConfigEndpointTests.cs
@@ -24,7 +24,7 @@ namespace CcDirector.Gateway.Tests;
 /// round-trip); the default is claude + opus when unset; and an unknown agent id is rejected
 /// (no-fallback).
 /// </summary>
-[Collection("DirectorRoot")]
+[Collection("GatewayHostedMode")]
 public sealed class BrainConfigEndpointTests : IAsyncLifetime
 {
     private readonly string _root;
--- a/src/CcDirector.Gateway.Tests/CockpitUrlEndpointTests.cs
+++ b/src/CcDirector.Gateway.Tests/CockpitUrlEndpointTests.cs
@@ -13,11 +13,18 @@ namespace CcDirector.Gateway.Tests;
 
 /// <summary>
 /// Endpoint-level proof that the P1 fix is WIRED IN, not just present as a helper: the live Gateway
-/// <c>GET /cockpit</c>, the <c>CockpitUrl</c> on <c>GET /gateway/about</c>, and the <c>cockpit.url</c> on
-/// <c>GET /gateway/settings</c> all hand back <c>{base}/cockpit</c> in hosted mode, where base comes from
-/// <c>CC_GATEWAY_PUBLIC_URL</c>. These are the tests that go red if any of the THREE call sites is reverted
-/// to <c>TailscaleIdentity.TryGetFrontDoorBaseUrl()</c> (which, with no tailnet in the test host, yields
-/// null where the public URL should be) - a green pure-resolver test cannot catch a mis-wired call site.
+/// <c>GET /cockpit</c> and the <c>CockpitUrl</c> on <c>GET /gateway/about</c> both hand back
+/// <c>{base}/cockpit</c> in hosted mode, where base comes from <c>CC_GATEWAY_PUBLIC_URL</c>. These are the
+/// tests that go red if either call site is reverted to <c>TailscaleIdentity.TryGetFrontDoorBaseUrl()</c>
+/// (which, with no tailnet in the test host, yields null where the public URL should be) - a green
+/// pure-resolver test cannot catch a mis-wired call site. These two are the surfaces a hosted client actually
+/// reads for the public URL.
+///
+/// <c>GET /gateway/settings</c> once carried a third <c>cockpit.url</c> copy, but that route is part of the
+/// owner-settings group DENIED on the hosted Gateway (issue #1863), so on hosted it serves the refusal and no
+/// cockpit URL at all - proved by <see cref="Hosted_settings_route_is_denied_and_carries_no_cockpit_url"/>.
+/// That copy is unused anyway: boot, navigation, and URL discovery do not consume it; its only reader is the
+/// settings-page load, which renders a load-error state on the deny.
 ///
 /// Only the HOSTED direction is asserted at the endpoint: it is deterministic (no tailscale involved). The
 /// self-host direction depends on whether a tailnet exists on the build host, so its byte-identical proof
@@ -66,24 +73,23 @@ public sealed class CockpitUrlEndpointTests
     }
 
     [Fact]
-    public async Task Hosted_settings_cockpit_url_returns_configured_public_cockpit_url()
+    public async Task Hosted_settings_route_is_denied_and_carries_no_cockpit_url()
     {
         await WithHostedGateway(PublicBase, async (http, gateway) =>
         {
-            // The third call site (first cut missed it): /gateway/settings.cockpit.url must hand back the
-            // SAME {base}/cockpit, not the raw front-door root it emitted before. Reverting this call site
-            // to TryGetFrontDoorBaseUrl() turns this red (null in the test host).
+            // /gateway/settings is part of the owner-settings group DENIED on hosted (issue #1863), so it no
+            // longer serves the cockpit.url it once did - it answers the refusal. The live public-URL surfaces
+            // are GET /cockpit and GET /gateway/about, asserted above; cockpit.url is unused by boot,
+            // navigation, or URL discovery.
             //
-            // MH-2: authenticate with a per-device key, not the shared token (rejected on hosted).
+            // MH-2: authenticate with a per-device key, not the shared token (rejected on hosted) - so this
+            // reaches the deny, not the auth gate.
             var deviceKey = gateway.Devices.Register("cockpiturl-settings", "PHONE").DeviceKey;
             using var req = new HttpRequestMessage(HttpMethod.Get, "gateway/settings");
             req.Headers.Authorization = new AuthenticationHeaderValue("Bearer", deviceKey);
             using var resp = await http.SendAsync(req);
-            resp.EnsureSuccessStatusCode();
 
-            using var doc = JsonDocument.Parse(await resp.Content.ReadAsStringAsync());
-            var url = doc.RootElement.GetProperty("cockpit").GetProperty("url").GetString();
-            Assert.Equal(ExpectedCockpit, url);
+            await OwnerSettingsRoutes.AssertIsNothingButTheRefusal(resp, OwnerSettingsRoutes.SettingsRefusal);
         });
     }
 
--- a/src/CcDirector.Gateway.Tests/GatewayHostTests.cs
+++ b/src/CcDirector.Gateway.Tests/GatewayHostTests.cs
@@ -18,6 +18,7 @@ namespace CcDirector.Gateway.Tests;
 /// the gateway's REST API over loopback. This covers the discovery flow, proxy
 /// routing, auth middleware, and error paths.
 /// </summary>
+[Collection("GatewayHostedMode")]
 public sealed class GatewayHostTests : IAsyncLifetime
 {
     private ControlApiHost _director = null!;
--- a/src/CcDirector.Gateway.Tests/GatewayHostedModeCollection.cs
+++ b/src/CcDirector.Gateway.Tests/GatewayHostedModeCollection.cs
@@ -1,0 +1,25 @@
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// The collection for every test whose result depends on the PROCESS-WIDE <c>CC_GATEWAY_HOSTED</c>
+/// variable - both the tests that SET it and the tests that would be broken by somebody else setting it.
+///
+/// Two rules put a class in here:
+///   1. It sets <c>CC_GATEWAY_HOSTED</c> itself. The owner-settings deny classes set it to OPPOSITE values -
+///      hosted in the deny classes, explicitly not-hosted in the self-host controls - so two of them running
+///      concurrently would each be reading the other's mode.
+///   2. It drives a route that is REFUSED on hosted (issue #1863 denied the whole owner-settings group).
+///      Before that deny those routes answered the same way in both modes, so a class driving them did not
+///      care what the variable said. Now it does: a hosted-mode class running in parallel would turn its
+///      200 into a 404, intermittently and for reasons entirely outside that test file.
+///
+/// <c>DisableParallelization</c> is what actually delivers rule 2 - the collection runs on its own, so no
+/// other collection can flip the mode underneath it. That also subsumes what <see cref="DirectorRootCollection"/>
+/// does for its members, which is why the classes moved in here no longer carry that attribute.
+/// </summary>
+[CollectionDefinition("GatewayHostedMode", DisableParallelization = true)]
+public sealed class GatewayHostedModeCollection
+{
+}
--- a/src/CcDirector.Gateway.Tests/GatewayHostedModeCollection.cs
+++ b/src/CcDirector.Gateway.Tests/GatewayHostedModeCollection.cs
@@ -16,8 +16,16 @@ namespace CcDirector.Gateway.Tests;
 ///      200 into a 404, intermittently and for reasons entirely outside that test file.
 ///
 /// <c>DisableParallelization</c> is what actually delivers rule 2 - the collection runs on its own, so no
-/// other collection can flip the mode underneath it. That also subsumes what <see cref="DirectorRootCollection"/>
-/// does for its members, which is why the classes moved in here no longer carry that attribute.
+/// other collection can flip the mode underneath it. That is STRICTLY STRONGER than what
+/// <see cref="DirectorRootCollection"/> gives its members - running alone subsumes running serially with a
+/// subset - which is why the five classes moved in here no longer carry that attribute. Nothing was
+/// weakened by the move.
+///
+/// WHY THIS SHIPS WITH THE DENY RATHER THAN SEPARATELY. Before issue #1863 these routes answered the same
+/// way in both modes, so none of this was needed. The deny is what creates the need, so the fix belongs in
+/// the change that creates it: shipping the deny without this collection would leave six test classes
+/// nobody here wrote failing intermittently, for reasons entirely outside their own files. A change that
+/// makes someone else's tests flaky is a change that shipped a defect.
 /// </summary>
 [CollectionDefinition("GatewayHostedMode", DisableParallelization = true)]
 public sealed class GatewayHostedModeCollection
--- a/src/CcDirector.Gateway.Tests/HostedOwnerSettingsDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedOwnerSettingsDenyTests.cs
@@ -1,0 +1,526 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Net.Sockets;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using CcDirector.Core;
+using CcDirector.Core.Configuration;
+using CcDirector.Gateway.Api;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// One place naming EVERY route in the owner-settings group, with the verb each one actually answers on.
+/// Both sides of the proof read this table: the hosted side asserts every row is REFUSED, the self-host
+/// side asserts every row is SERVED. Writing the table once is what makes "every route, both directions"
+/// a checkable claim rather than a sentence in a pull request.
+/// </summary>
+public static class OwnerSettingsRoutes
+{
+    /// <summary>The refusal a route in <c>SettingsEndpoints</c> answers with on hosted.</summary>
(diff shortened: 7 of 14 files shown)


thefrederiksen (commented):
REJECT

1. The deny suite covers every read but omits five writes: `PUT /gateway/wingman/training-capture`, `/gateway/addressing-mode`, `/gateway/snooze-default`, `/gateway/snooze-presets`, and `/gateway/time-zone`. Those routes have no deny-closed or exact-one-`error`-property proof. Drive each with an enrolled tenant key.

2. The cited self-host suites never explicitly clear `CC_GATEWAY_HOSTED` and cover only fragments of this 24-route family. Add explicit hosted-off controls over the whole route family, or a justified full read/write matrix that proves the shipped self-host behavior rather than ambient runner state.

3. There is no named un-deny row per denied route with the real remodel/partition or intentional non-restoration condition. Without that ledger, 24 routes become a silent permanent removal.

4. The group filter is structurally correct but has no future-route probe. Map a probe through the filter seam and prove hosted refuses it while explicit hosted-off serves it.

5. The filter revert was run only against 43 focused tests. Repeat the real production-line revert after a successful build in the full 2900+ Gateway suite, record the counted summary with intended reds and controls green, then restore/rebuild/rerun full.

The direct hosted-mode gate, one group-level filter, 404 response, and allow-list helper are otherwise sound.

thefrederiksen (commented):
Re-audit against the measured 404/SPA masking bar confirms the served-side proof is insufficient at `c0ab3540`.

The hosted refusal helper parses JSON without first asserting `application/json`, so an absent/masked route can fail as a parser crash rather than a distinguishing assertion. More importantly, the claimed self-host controls are pre-existing `AiProviderEndpointTests` / `BrainConfigEndpointTests` / `AiModelsEndpointTests`; they never set `CC_GATEWAY_HOSTED` explicitly and do not drive every denied path+verb from the route inventory. They therefore cannot prove the exact routes still resolve to their own handlers on self-host. Add a table-driven dual-sided row for every route: hosted media-type-first plus exact one-error refusal, and explicit-hosted-off handler-positive payload/receipt/effect on the same path+verb. A future-route probe needs both hosted refusal and self-host served halves. The proof driver must abort unless its baseline is green and complete, then mutate one bypassable primitive per full Gateway run and restore exact head green. This is in addition to the previously reported missing write routes. Verdict remains **REJECT**.

thefrederiksen:
REJECT on exact head `7e054482` (static review; no lane consumed).

1. The advertised three-primitive design is not what the code implements. In all three `Map` methods, both `outer` and filtered `app` remain in scope while each production route is mapped. Any one route can therefore bypass the gate by changing `app.Map*` to `outer.Map*`; that compiles and independently opens only that route. The future-route probes prove a route placed on the group inherits the filter, but cannot prove production routes remain on it. Move each family’s route mappings into a private method that receives only its filtered `RouteGroupBuilder` (the shape now used by the vault and recording reworks), or mutation-prove every independent mapping bypass with full-suite arms.

2. `POST /gateway/brain/restart` has no served-side handler proof. A GET that produces 405/Allow: POST proves only route registration, not that the exact POST reaches the intended handler. The required positive control is the exact path+verb with a handler-specific receipt. Add a safe injected brain/process seam. The current mutation-A plan knowingly invokes the live restart path and may leave a coding-agent process behind; that is not an acceptable proof harness.

3. `PUT /gateway/transcription-mode` still has no independently observable write effect: it writes the sole/default value and then echoes it. That cannot distinguish a real write from a no-op. Seed/prove a persistent transition, or explicitly narrow the claim if the route has no distinguishable mutation.

4. The uniform “nothing accumulates / no purge” ledger is not established by re-reading after these nineteen HTTP requests. The underlying process-global `config.json` can already contain shared values from before the deny, and `CcDirectorConfigService.MergePatch` is also used by other production surfaces/processes. At minimum, scope the claim to these handlers and require quarantine/migration of the pre-existing global settings before a future tenant-scoped un-deny unless deployment isolation is proved.

Because the security primitive count and two positive controls are wrong, do not spend the Gateway lane on the registered five-run sweep yet. Rework, rebase onto current main, and register the complete baseline/mutation/reconciliation proof against the new exact head.

thefrederiksen:
STANDING: HELD
eebedebb62feb8a542bb9fda6e7315b0d46ebec5

The rework answers the prior static findings: production mappings can no longer select the unguarded builder, the restart path has a non-spawning handler receipt, the transcription claim is honestly narrowed, and the un-deny condition now requires quarantine or migration.

Clearance is held on the proof the PR itself still marks outstanding. On this exact head, execute and report the registered green baseline, each of the three filter-deletion arms, and the exact-head restored run, with complete totals reconciled before scoring, every predicted red classified at its stated assertion, controls unchanged, and no unexpected reds. Green CI is necessary but does not substitute for that revert attribution.

thefrederiksen:
STANDING: REJECTED
eebedebb62feb8a542bb9fda6e7315b0d46ebec5

The rebuild closes the prior local proof defects: all 31 mappings are structurally confined to their guarded groups; brain restart now has a non-spawning exact-handler seam plus a destructive failure control; the transcription-mode claim is honestly narrowed and carries an expiry test; and the global-config ledger now requires quarantine/migration rather than claiming no purge.

The shared boundary remains three endpoint-group filters, so it is post-binding. All three future-route probes are parameterless GETs and cannot exercise the pre-binding property. The outside-group controls use unrelated paths rather than same-namespace neighbours, leaving over-refusal at each family boundary unproved.

The registered mutation plan covers only the three filter removals. Each deny is still a conjunction of attachment and an independently wrong-able hosted gate direction, so three separate gate-inversion arms are also required. Those arms must predict both hosted and self-host reds while keeping neighbours green.

The registered full-suite proof is outstanding, the CI red remains explicitly unattributed, and the head predates current `main`.

Rework must adopt a cleared shared pre-binding primitive, use body-bound future POSTs with observable no-binding evidence, add meaningful neighbours, register the three independent gate-direction arms, and then run the complete serialized proof on the rebased final head.

PR #1899, opened 2026-07-19
fix(gateway): store per-device keys hashed at rest, never in plaintext

thefrederiksen:
Closes #1878.

## The defect

`DeviceRegistry` wrote every enrolled device's per-device key into
`config/director/devices.json` in the clear, and `IsValidDeviceKey` /
`DeviceTypeForKey` / `TenantForKey` each compared a presented key against that
stored plaintext.

That key is a bearer secret: it authenticates every Director-to-Gateway call. On
the hosted Gateway the single file also holds the device-to-account-to-tenant
map for every tenant, on a mounted Azure Files share. So anything that could read
one file could impersonate every tenant's Director - which is also what turned a
process spawned with a working directory on that same share (#1863) from a
code-execution problem into a credential-exfiltration one.

## The fix

Each key is stored only as a one-way SHA-256 hash. The Gateway never needs the
key itself: it returns the plaintext to its owner at enrollment and thereafter
only VERIFIES a presented key, the same way a password store does. A read of
`devices.json` now yields no usable credential.

Two design choices, both deliberate and both written down in the code:

- **Plain SHA-256, not a slow password-stretching function.** The secret is a
  256-bit value the registry generated with a cryptographic random number
  generator, not a human-chosen password, so there is no guessable search space
  for stretching to defend - and this lookup is the hot path, running on every
  authenticated request. No per-record salt either: salts defend low-entropy
  secrets against precomputation, which does not apply here.
- **Key lookup is unchanged in shape.** The three public lookups now share one
  private resolver that hashes the presented key once and compares that digest
  against each active record's stored digest with
  `CryptographicOperations.FixedTimeEquals`.

### What the constant-time property does and does not cover

Stated precisely, because the first draft of this pull request overstated it.

The **digest comparison** is constant-time: it always reads both digests in
full, so it does not leak how many leading bytes of a wrong digest happened to
match. That partial-match signal is the one that would otherwise let an attacker
walk a guess toward a stored value, and it is the part that has to be
constant-time.

The **lookup around it is not** constant-time and no longer claims to be. It
walks the records, skips inactive and unhashed ones, and returns as soon as it
matches, so its running time varies with how many devices are enrolled and where
a match falls. What that can leak is coarse registry shape, not anything about
the secret: the compared value is the output of a hash of the presented key, so a
wrong guess yields an unrelated digest whose position carries no information.
Making the walk constant-time would defend nothing here. The code says all of
this at `FindActiveByKey`, and the three public lookups point at it rather than
repeating a shorter, looser version.

## Enrollment stays safe to retry

Hashing the stored key removed the registry's ability to hand a re-enrolling
device its existing key back, which turned every duplicate `RegisterIfAbsent`
call into a **rotation**. That was a real regression, and a serious one: a caller
that had not yet durably saved the first response - a lost or delayed response, a
double submit, two calls in flight - would be left holding a key that a later
call had already retired, locked out until a human re-enrolled it. An outage that
only happens on a retry is the kind that does not surface until real devices are
enrolling over a flaky mobile connection. The method's name and its prior
contract both promised idempotence, so the property is restored rather than
documented as a hazard.

An issued key now stays replayable for a short window held **in this process's
memory only**. A duplicate enrollment inside that window returns the same key and
rotates nothing. It is never serialized, never written to `devices.json`, and does
not survive a restart, so the at-rest property this change exists for is
untouched. The whole method is serialized, so two calls in flight converge on one
issue instead of racing to rotate.

Past the window a further enrollment rotates in place: same single registry entry,
same account and tenant binding, same cloud mapping, and still exactly one key
that validates, with the previous one retired. That one-valid-key-per-device cap
is what the #1136 auto-mint key leak guardrail actually defends (the leak was
enrollment ACCUMULATING valid credentials), and it holds on both branches.

Two details that exist to keep the mechanism provable rather than merely correct:

- The window is enforced in **exactly one place**, the expiry check. An earlier
  draft also short-circuited the recording path on a zero window; that second
  guard was removed because it would have let a broken expiry check still look
  correct under the zero-window setting the tests use - the guard would have
  hidden the defect it sat next to.
- A replay is checked against what the record **currently** verifies, so a key
  superseded by an explicit re-pairing is never handed back.

## The migration story for existing keys

Real machines are enrolled against the live hosted box right now and there is no
re-enrollment path that does not involve a human, so invalidating a stored key
would take the box down.

`Load` handles it: a record that carries a plaintext `DeviceKey` and no
`DeviceKeyHash` is hashed on the spot, the plaintext is nulled, and the file is
rewritten. The device keeps working with the key it already holds; its account
and tenant binding, cloud mirror id, platform, and device type all survive. From
then on the record has no `deviceKey` field at all.

Proven by `ADeviceEnrolledBeforeTheChange_StillAuthenticatesAfterIt` and four
sibling tests, which start from a hand-written pre-change `devices.json` in
exactly the shape the old code emitted - plaintext key, hosted account and tenant
binding, cloud id - and assert the old key still authenticates, still resolves to
its bound tenant, still works after a further restart, and that the plaintext is
gone from the file.

### What the migration does NOT do - the claim, narrowed

The migration rewrites the **live** registry file. That is the whole of what a
file write can prove, and the earlier wording here ("the plaintext leaves the
disk") overclaimed. Rewriting a file cannot reach a copy of the old contents that
something else already took: a storage-share snapshot, a backup, a soft-delete
retention window, or file-system history all retain whatever the file held
before, and no write from inside the Gateway touches any of them.

On hosted, the registry sits on a mounted Azure Files share. Snapshots and soft
delete there are account-level and share-level settings held **outside this
repository** - a search of the repository finds no configuration that enables,
disables, or references them - so whether pre-change copies are being retained is
a property of the deployed storage account and has to be checked there rather
than inferred from code. This pull request does not assert either way.

Where such retention is found, two things follow, both separate operational work
tracked separately from this change:

1. Purge the retained copies of the pre-change `devices.json`.
2. Treat every key that appeared in plaintext in those copies as exposed and
   **rotate it**. Rotation, not file cleanup, is what ends the exposure - a copy
   may already have been read.

`docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md` carries the same
narrowed statement, so the document and the code agree.

## Documentation

`docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md` gains a section
classifying the device registry explicitly as a high-value store - what it holds,
that it used to hold plaintext, the at-rest handling now, the retry-safety
window, the note that hashing removes the bearer secret from the file but not the
metadata, and the "what the migration does not do" limits above.

### Retention of that window is enforced, not asserted

The window holds an issued key in memory, and what makes that acceptable is that
it is short-lived - so short-livedness is enforced by a clock this type owns.

A first version got this wrong in a way that looked right, and it is recorded
here because the shape recurs: pruning ran only inside the replay lookup, which
is reached only when an ALREADY-ENROLLED device asks to replay. A device that
enrolls once under its own identifier and never comes back - the ordinary
fresh-enrollment path - never reached it, so its plaintext key stayed strongly
referenced for the life of the process, and the dictionary grew with every fresh
identifier. **The property that justified the design was the one nothing tested.**

Refusing to hand back an expired key and bounding how long the key is kept are
two different jobs, and only the second one needs a clock:

- The registry starts a recurring eviction pass **the moment it first retains a
  key**, so retention and the clock that bounds it begin together.
- The pass drops every expired issue whether or not anything ever enrolls again.
- It runs at **half** the window, so the real bound is one and a half windows
  rather than two - by default, at most seven and a half minutes. The documented
  bound has to be the true one.
- The registry is now `IDisposable`; disposal stops the pass and drops everything
  retained. `GatewayHost.StopAsync` disposes it with its other timers.
- The lazy prune is gone. Eviction has exactly one owner, so a broken pass cannot
  be masked by another path tidying up on its way past.

Coverage is deterministic - a hand-driven scheduler fires the pass, so nothing
waits on a clock - and asserts retention **directly** through an internal count
rather than inferring it from behaviour that would still look correct while the
plaintext sat in memory. The control is a key still inside its window, which must
SURVIVE a pass; without it every retention test is satisfied by a pass that
clears the dictionary unconditionally and silently destroys the retry safety the
window exists to provide.

### Why the exposure is bounded rather than removed

The alternative is to retain nothing and have the caller present the key it
already holds. It was considered and not chosen, because it does not cover the
concurrent case: two enrollment calls in flight for a device that has no key yet
cannot present anything, so the second must rotate and the caller can be left
holding the first. That is the lock-out this window exists to prevent, so
removing the retention would trade a bounded, now-enforced memory exposure for an
uncovered lock-out path. It would also change the request contract across both
enrollment endpoints and every client that calls them.

It is nonetheless the better end state - it would let the window shrink to near
zero and cover the common double submit without retaining anything - and is
raised as follow-up work on the enrollment contract rather than folded into this
change.

## The previous head went continuous-integration green while carrying this defect

Stated plainly, because it is the honest reason the new tests exist and the
answer to anyone who later asks why a retention window needed eleven mutation
runs.

Head `db531999` passed every check - the full .NET suite across all seven test
projects and the web suite - while the replay cache was retaining plaintext
device keys for the life of the process. The suite was not careless and it was
not lucky. **Nothing exercised the property, so nothing could go red.** No test
ever asked "after a device enrolls once and never comes back, is its key still in
memory?", and a suite cannot answer a question nobody asks.

The general form is worth keeping: a green suite tells you the head is not
BROKEN in the ways it already knows how to check. It never tells you the head is
PROVEN. The gap between those two is exactly the size of the properties nothing
tests - and the property most likely to go untested is the one that JUSTIFIED the
design, because it feels like a premise rather than a behaviour. Short-livedness
was the premise that made an in-memory plaintext window acceptable here, and it
was the one thing with no coverage.

That is why the new tests assert retention directly, through a count of retained
keys, rather than through behaviour. Behaviour was already green.

## Proof status

The revert-proof recipes were run and passed on an earlier head. The branch has
since been rebased onto current `origin/main` and reworked twice, so they are
**pending a re-run on the current head** together with the round-two mutation
set - eleven runs, pre-registered with predicted reds in a comment below. The
machine's single Gateway test lane is held by other work; this queues behind it
rather than running two suites at once.

The baseline will be taken fresh at this head and reconciled against itself -
never against a count quoted from an earlier head or from another unit. A run
with no summary line is not a result in either direction. M4, the enrollment
lock, remains a DECLARED GAP: its red depends on thread interleaving, so it is
reported as unproven rather than promoted on a single green.

### DIFF
--- a/docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md
+++ b/docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md
@@ -25,6 +25,35 @@ token injected into the page**, so reaching the URL (or viewing page source) was
 unrevocable fleet control. That token is gone from the shell. The phone now holds only (C) - a local,
 per-device, individually revocable key that is not the master token and not a cloud credential.
 
+### 1a. The device registry is a high-value store (issue #1878)
+
+`DeviceRegistry` persists to `config/director/devices.json` under `CC_DIRECTOR_ROOT`. That one file is
+both the Gateway's credential store for credential (C) and, on the hosted Gateway, the authoritative
+device-to-account-to-tenant map for every tenant on the box - and on hosted it sits on a mounted Azure
+Files share. It must be treated as a high-value store wherever it is discussed, on self-host and on
+hosted alike.
+
+It used to hold each device key in **plaintext**, so anything that could read the file could impersonate
+every enrolled device - which is also what made a process spawned with a working directory on that same
+share (issue #1863) a credential-exfiltration problem rather than only a code-execution one.
+
+Each key is now stored **only as a one-way SHA-256 hash**. The Gateway never needs the key itself: it
+issues the plaintext to its owner once at enrollment and thereafter only VERIFIES a presented key, the
+same way a password store does. A read of `devices.json` therefore yields no usable credential. Two
+consequences follow and are deliberate:
+
+- The hash is a plain SHA-256, not a slow password-stretching function, because the secret is a 256-bit
+  value the Gateway generated with a cryptographic random number generator, not a human-chosen password.
+  There is no guessable search space to stretch against, and this lookup runs on the hot path of every
+  authenticated request.
+- A device that re-enrolls is issued a **fresh** key in place - the registry has no plaintext to hand
+  back. Its entry, its account and tenant binding, and its cloud mapping are all preserved, and it still
+  has exactly one valid key. A device that simply keeps using the key it already holds is unaffected: a
+  registry file written before the change is migrated on load, so keys issued before this keep working.
+
+The file remains a real target for its non-secret contents (which accounts and tenants exist, and which
+machines are enrolled). Hashing removes the bearer secret from it, not the metadata.
+
 ---
 
 ## 2. Trust boundaries (what crosses which line)
@@ -130,6 +159,7 @@ forever unseen.
 | A leaked phone key compromises the account/fleet | (C) is one revocable device key; it is not (A), (B), (D), or the master token |
 | A stolen/lost phone keeps access | Revoke the one device from the website; the reconcile sweep drops (C) |
 | Cloud outage locks the phone out | Steady-state validation of (C) is fully local; no cloud dependency after enrollment |
+| Reading the Gateway's registry file yields every device's key | (C) is stored only as a one-way hash, so `devices.json` holds no usable credential (issue #1878, section 1a) |
 
 ---
 
--- a/docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md
+++ b/docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md
@@ -51,8 +51,32 @@ consequences follow and are deliberate:
   has exactly one valid key. A device that simply keeps using the key it already holds is unaffected: a
   registry file written before the change is migrated on load, so keys issued before this keep working.
 
+Enrollment stays safe to RETRY. Because the registry can no longer hand back a stored key, a repeated
+enrollment would otherwise rotate the key on every call - and a caller that lost or had not yet saved the
+first response would then be holding a retired key and locked out. So an issued key stays replayable for a
+short window held **in this process's memory only**: a duplicate or retried enrollment inside that window
+gets the same key back and nothing rotates. It is never serialized and does not survive a restart, so the
+at-rest property is unaffected. Past the window a further enrollment rotates in place as described above.
+
 The file remains a real target for its non-secret contents (which accounts and tenants exist, and which
 machines are enrolled). Hashing removes the bearer secret from it, not the metadata.
+
+### What the migration does NOT do
+
+The migration rewrites the **live** registry file. That is the limit of the claim, and it should not be
+read as secure erasure. Rewriting a file cannot reach a copy of the old contents that something else
+already took - a storage-share snapshot, a backup, a soft-delete retention window, or file-system history
+all retain whatever the file held before, and no amount of writing from inside the Gateway touches them.
+
+On the hosted Gateway the registry sits on a mounted Azure Files share, where snapshots and soft delete are
+account-level and share-level settings held outside this repository; nothing in the repository configures or
+disables them, so whether pre-change copies are being retained is a property of the deployed storage account
+and has to be checked there rather than inferred from code. Where such retention is found, two things follow
+and both are separate operational work, tracked separately from this change:
+
+1. Purge the retained copies of the pre-change `devices.json`.
+2. Treat every key that appeared in plaintext in those copies as exposed and **rotate it**. Rotation, not
+   file cleanup, is what actually ends the exposure - a copy may already have been read.
 
 ---
 
--- a/docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md
+++ b/docs/architecture/mobile/DEVICE_AUTH_SECURITY_MODEL.md
@@ -58,6 +58,19 @@ short window held **in this process's memory only**: a duplicate or retried enro
 gets the same key back and nothing rotates. It is never serialized and does not survive a restart, so the
 at-rest property is unaffected. Past the window a further enrollment rotates in place as described above.
 
+That in-memory window is a real, if small, exposure, and the thing that makes it acceptable is that it is
+short-lived - so short-livedness is **enforced, not assumed**. The registry starts a recurring eviction pass
+the moment it first retains a key, and that pass drops every issue whose window has closed regardless of
+whether any device ever enrolls again. Plaintext is held for at most one and a half windows, no more than
+seven and a half minutes by default, and shutting the registry down ends it at once.
+
+The subtlety is worth recording, because the first implementation got it wrong in a way that looked right:
+it checked an entry's age only when an already-enrolled device asked to replay. That correctly refuses to
+hand back an expired key, but it does nothing about retention, because the ordinary case - a device that
+enrolls once under its own identifier and never comes back - never reaches the check. Every such key stayed
+resident for the life of the process. Refusing to serve an expired secret and bounding how long the secret
+is kept are two different jobs, and only the second one needs a clock of its own.
+
 The file remains a real target for its non-secret contents (which accounts and tenants exist, and which
 machines are enrolled). Hashing removes the bearer secret from it, not the metadata.
 
--- a/src/CcDirector.Gateway.Contracts/DeviceEnrollment.cs
+++ b/src/CcDirector.Gateway.Contracts/DeviceEnrollment.cs
@@ -51,7 +51,9 @@ public sealed class DeviceRegistrationResponse
 /// <summary>
 /// One device's public-facing entry in the Gateway device registry (issue #469): the
 /// host-readable record used to list registered devices. The per-device key itself is NEVER
-/// included - the registry serves identity and status, not the secret.
+/// included - the registry serves identity and status, not the secret. It DOES carry a NON-SECRET
+/// masked key identity (<see cref="KeyPrefix"/> / <see cref="KeyLast4"/>, issue #1899) so a listing can
+/// tell devices apart by key without ever substituting or exposing the raw key.
 /// </summary>
 public sealed class RegisteredDeviceDto
 {
@@ -66,4 +68,15 @@ public sealed class RegisteredDeviceDto
 
     /// <summary>The device's status (e.g. <c>active</c>, <c>revoked</c>).</summary>
     public string Status { get; set; } = "";
+
+    /// <summary>
+    /// The NON-SECRET first few characters of the device's key (issue #1899) - masked display metadata so a
+    /// host can recognise which key a device holds without the registry ever returning the raw key. Reveals a
+    /// handful of a 256-bit key and is not a credential. Empty for a record with no recorded key identity.
+    /// </summary>
+    public string KeyPrefix { get; set; } = "";
+
+    /// <summary>The NON-SECRET last few characters of the device's key (issue #1899), the trailing half of the
+    /// masked key identity. Not a credential.</summary>
+    public string KeyLast4 { get; set; } = "";
 }
--- a/src/CcDirector.Gateway.Contracts/DeviceEnrollment.cs
+++ b/src/CcDirector.Gateway.Contracts/DeviceEnrollment.cs
@@ -3,8 +3,9 @@ namespace CcDirector.Gateway.Contracts;
 /// <summary>
 /// A co-located Director's request to enroll with its own Gateway using the DevThrottle account
 /// sign-in instead of a pairing code (issue #1069). The Director POSTs this to
-/// <c>/devices/enroll-signed-in</c>; the Gateway mints (or, if this device already has one, returns)
-/// the Director's own per-device key - gated on the Gateway being signed in to DevThrottle AND the
+/// <c>/devices/enroll-signed-in</c>; the Gateway issues the Director's own per-device key - a fresh one
+/// if this device is already enrolled, since the registry keeps only a hash of the key it issued and so
+/// has no plaintext to hand back (issue #1878) - gated on the Gateway being signed in to DevThrottle AND the
 /// caller being a loopback same-machine connection. There is no pairing code: signing in is the
 /// authorization. Carries no credential of its own; the loopback origin plus the Gateway's signed-in
 /// account are the proof.
--- a/src/CcDirector.Gateway.Tests/DeviceKeyAtRestTests.cs
+++ b/src/CcDirector.Gateway.Tests/DeviceKeyAtRestTests.cs
@@ -6,18 +6,21 @@ using Xunit;
 namespace CcDirector.Gateway.Tests;
 
 /// <summary>
-/// Issue #1878: the device registry file used to hold every enrolled device's per-device key in
+/// Issue #1878 / #1899: the device registry file used to hold every enrolled device's per-device key in
 /// PLAINTEXT. That key is a bearer secret - it authenticates every Director-to-Gateway call - and on the
 /// hosted Gateway the one file holds every tenant's device on shared infrastructure, so anything that
 /// could read the file could impersonate every tenant's Director.
 ///
-/// These tests pin the fix from both ends:
+/// These tests pin the fix from several ends:
 ///   1. WHAT IS WRITTEN: the file must contain no usable credential - the plaintext key must not appear
 ///      in it, and what does appear must be the one-way hash.
 ///   2. WHAT KEEPS WORKING: a device enrolled BEFORE the change - whose key exists only in the plaintext
 ///      file and in that machine's credential file - must still authenticate afterwards, and must still
 ///      resolve to its bound tenant. There is no re-enrollment path that does not involve a human, so a
 ///      change that invalidated those keys would take the live hosted box down.
+///   3. NO PLAINTEXT IS EVER RETAINED (issue #1899): plaintext is disclosed at exactly one point - the
+///      enrollment response - and kept nowhere. A REPEAT enrollment cannot re-reveal the current key; it
+///      rotates to a fresh one instead. The registry holds no plaintext cache to leak or to replay.
 /// </summary>
 public sealed class DeviceKeyAtRestTests : IDisposable
 {
@@ -210,6 +213,23 @@ public sealed class DeviceKeyAtRestTests : IDisposable
     }
 
     [Fact]
+    public void Migration_RecordsTheMaskedKeyIdentity_FromThePlaintextBeforeDroppingIt()
+    {
+        // Issue #1899: a migrated device must gain the same non-secret masked key identity a freshly-enrolled
+        // one has, computed from the plaintext while it is still in hand and BEFORE it is dropped - otherwise a
+        // pre-change device would list with no key identity at all.
+        const string alreadyIssuedKey = "pre-change-key-Ab3xQ9zK-do-not-invalidate";
+        WritePreChangeStore("device-already-enrolled", alreadyIssuedKey, "tenant-already-enrolled");
+
+        var entry = Assert.Single(new DeviceRegistry(StorePath).List());
+
+        Assert.Equal(alreadyIssuedKey.Substring(0, 8), entry.KeyPrefix);
+        Assert.Equal(alreadyIssuedKey.Substring(alreadyIssuedKey.Length - 4), entry.KeyLast4);
+        // And the masked identity is not the whole key - it is a hint, never a credential.
+        Assert.False(alreadyIssuedKey.Contains(entry.KeyPrefix + "..." + entry.KeyLast4, StringComparison.Ordinal));
+    }
+
+    [Fact]
     public void Migration_DoesNotMakeEveryKeyValid()
     {
         // The control on the migration: hashing an existing record must not turn the registry into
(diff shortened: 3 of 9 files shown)


thefrederiksen:
## Pre-registered mutation set and predicted reds

Registered BEFORE the runs, so that a predicted red which does not appear is a
finding rather than something to be explained away afterwards. The machine's
single Gateway test lane is held by another change; this set is queued behind it.
One primitive per full-suite run, never combined.

A red counts only when it fails on an assertion the test's own logic raised. A
red that arrives as a crash, a build failure, or a framework error is laundered
and proves nothing about the protection, so every red is classified by how it
ARRIVED before it is counted.

Each line below earns its own run under the bypassability rule: it can be
individually wrong while everything else stays correct, and the protection still
breaks. Overlap does not exempt any of them. Where two guards overlapped, the
count was reduced by removing the redundant guard from the DESIGN - see M2 - not
by arguing the run away.

### M1 - the replay branch

Mutation: delete the `if (TryReplayIssuedKey(...))` block from
`RegisterIfAbsent`, so every repeat call rotates. This is the pre-fix behaviour.

Predicted red, and nothing else:
- `EnrollingTwice_ReturnsTheSameKey_SoARetryCannotLockTheCallerOut`
- `ARetriedEnrollment_DoesNotRetireTheKeyTheCallerAlreadyHolds`
- `EnrollingTwice_StillLeavesExactlyOneValidKey`
- `ConcurrentEnrollmentsOfOneDevice_AllResolveToTheSameKey`
- `RegisterIfAbsent_SameDeviceTwice_WithinTheReplayWindow_ReturnsTheSameKey`
- `DestructibilityControl_AfterARePairing_TheSupersededKeyIsNotReplayedBack`
  (its first assertion is satisfied, but it asserts a rotation happened through
  the replay path; expected to stay GREEN - flagged here so that if it reds it is
  read as an over-broad mutation, not as extra confirmation)

Predicted green controls: all three remaining destructibility controls, both
`ReEnrollingADevice_PastTheReplayWindow*` tests, every migration test, and both
file-content canaries.

### M2 - the replay expiry check

Mutation: delete `if (DateTime.UtcNow - issue.IssuedAtUtc >= _replayWindow)
return false;` from `TryReplayIssuedKey`, so a replay never expires.

Predicted red:
- `DestructibilityControl_OnceTheReplayWindowCloses_ARepeatEnrollmentRetiresThePreviousKey`
- `ReEnrollingADevice_PastTheReplayWindow_LeavesExactlyOneValidKeyAndOneEntry`
- `RegisterIfAbsent_SameDeviceTwice_LeavesOneEntryAndOneValidKey`
- `RegisterIfAbsent_ManyCalls_NeverAccumulateDevicesOrKeys`

Note on design, not on argument. This mutation was originally NOT detectable.
`RecordReplayableIssue` also short-circuited on a zero window, so with the expiry
check deleted the zero-window tests still saw no replay entry and stayed green -
a second guard concealing the failure of the first. The redundant short-circuit
was removed from the code so the window is enforced in exactly one place and this
run can fail. The run count went down because the design changed, not because the
overlap was argued to be sufficient.

### M3 - the supersession check

Mutation: delete `if (!string.Equals(issue.KeyHash, existing.DeviceKeyHash, ...))
return false;` from `TryReplayIssuedKey`, so a stale replay entry is handed out
even after the record's key has moved on.

Predicted red:
- `DestructibilityControl_AfterARePairing_TheSupersededKeyIsNotReplayedBack`

That test was added specifically for this line. Without it M3 was undetectable:
the pre-existing re-pairing control never called `RegisterIfAbsent` after the
`Register`, so nothing ever asked the replay path for a superseded key.

### M4 - the enrollment lock

Mutation: remove `lock (_enrollLock)` from `RegisterIfAbsent`.

Predicted red:
- `ConcurrentEnrollmentsOfOneDevice_AllResolveToTheSameKey`

**Declared weakness, stated rather than glossed.** This red is PROBABILISTIC. The
test races sixteen parallel enrollments and asserts they converge on one key;
without the lock the interleaving that produces divergence is very likely but not
guaranteed, so a single green run of M4 does not establish that the lock is
unnecessary, and a single red does not establish a deterministic proof. This is
the one line in the set that does not have a deterministic single-mutation proof,
and it is reported as a gap rather than counted as proven. The sequential
retry-safety property - which is the reviewer's named scenario for a lost or
delayed response and a double submit - is proven deterministically by M1 and does
not rest on this lock.

### M5 and M6 - the two original reverts, re-run

The branch has been rebased onto current `origin/main` and has gained the
retry-safety work, so the revert-proof from the pre-rebase head does not carry
over: a revert proof is only a proof on the head it ran against, and both prior
runs must be repeated here together with the full suite rather than a filtered
subset.

- **M5** - add `DeviceKey = key,` back into the record built by `Register`.
  Predicted red: `EnrolledKey_IsNeverWrittenToTheRegistryFileInPlaintext`,
  `EveryEnrolledKey_IsAbsentFromTheFile_EvenWithManyDevices`. Also now expected:
  `DestructibilityControl_ReplayIsNotOfferedAcrossARestart`, whose final
  assertion reads the file for the plaintext.
- **M6** - delete the migration block from `Load`.
  Predicted red: `ADeviceEnrolledBeforeTheChange_StillAuthenticatesAfterIt`,
  `ADeviceEnrolledBeforeTheChange_StillResolvesToItsBoundTenant`,
  `LoadingAPreChangeStore_ScrubsThePlaintextKeyFromDisk`,
  `AMigratedDevice_KeepsWorkingAcrossTheNextRestartToo`.

### The baseline

Six mutation runs, each preceded and followed by a restored run. The baseline
must be green AND explicitly complete on this rebased head - a run with no
summary line is not a result in either direction, and the reconciliation
`passed + failed under the mutation == passed under the restored run` must
compute. Every test named above must be confirmed to have EXECUTED by name; a
mutation whose predicted red is absent is recorded as a finding, not smoothed
over.

thefrederiksen:
## Response to the first review

A note on the record first: **the first review's text is not on this pull request.**
`gh pr view 1899 --json reviews`, `--json comments`, and all three REST surfaces
(`/pulls/1899/reviews`, `/pulls/1899/comments`, `/issues/1899/comments`) each
return zero. The ruling reached this work unit only as a relayed summary. I have
acted on all of it, but the primary artifact could not be read, so anyone
checking my responses against the reviewer's own words cannot do so from here.
Worth fixing for the next round - a relayed summary is not the artifact.

### 1. The idempotence defect - FIXED, not accepted

Taken as the serious one, and agreed on the reasoning: an outage that only
happens on a retry does not surface until real devices are enrolling over a flaky
connection. Rather than prove that every caller serializes, the property was
restored, because the method's NAME and its prior contract both promise it.

That the prior contract was relied upon is not a matter of inference. The hosted
enrollment path reasons about it by name in its own tenant-namespacing comment,
which argues that namespacing keeps enrollment "idempotent for one account (same
tenant -> same hash)". Rotating on every call had quietly falsified the premise
that comment rests on.

An issued key now stays replayable for a short window held in this process's
MEMORY ONLY - never serialized, gone on restart - so a duplicate enrollment
returns the same key and rotates nothing. `RegisterIfAbsent` is serialized end to
end, so two calls in flight converge on one issue instead of racing. Past the
window a further call rotates in place, so the #1136 one-valid-key-per-device cap
holds on both branches. Details in the pull request body.

Two things were changed specifically so the fix is provable rather than merely
present:

- The window is enforced in **exactly one place**. A first draft also
  short-circuited the recording path on a zero window; with two guards, deleting
  the expiry check still left the zero-window tests green - the redundant guard
  concealed the failure of the one that mattered. It was removed from the design.
  The mutation count went down because the code changed, not because the overlap
  was argued to be sufficient.
- A replay is checked against what the record **currently** verifies, so a key
  superseded by an explicit re-pairing is never handed back. The pre-existing
  re-pairing control never called `RegisterIfAbsent` afterwards, so nothing had
  ever asked the replay path for a superseded key; a test was added that does.

Every idempotence assertion is paired with a **destructibility control**, so that
"the key survived a second call" reads as protection rather than as the registry
having quietly lost the ability to retire keys at all. Four controls: an explicit
re-pairing retires the previous key, a repeat past the window retires it, a
restart offers no replay (which is also the direct evidence nothing was
persisted), and a superseded key is not resurrected.

### 2. The overclaim - narrowed, in all three places

Agreed without reservation: writing a file cannot prove secure erasure. "The
plaintext leaves the disk" is now "the plaintext is gone from the **live**
registry file", in the pull request body, the `DeviceRegistry` type
documentation, the migration comment, the migration log line, and
`DEVICE_AUTH_SECURITY_MODEL.md`.

On whether such retention exists - checked rather than assumed, and the honest
answer is that **it cannot be determined from this repository**. A search finds no
configuration that enables, disables, or references Azure Files share snapshots
or soft delete; both are account-level and share-level settings on the deployed
storage account, held outside version control. So this pull request asserts
neither way, and says so. What it does say is what follows if retention IS found,
and that this is separate operational work: purge the retained copies, and
**rotate** the affected keys - because rotation, not file cleanup, is what ends
an exposure that may already have been read.

That leaves a real open item for whoever can inspect the storage account. It is
deliberately not claimed as closed here.

### 3. The wording - corrected

"Constant-time lookup" was wrong and is gone. The claim is now a constant-time
**digest comparison**, with the scope spelled out at `FindActiveByKey` and the
three public lookups pointing at it instead of each repeating a looser version:
the comparison always reads both digests in full, so it does not leak how many
leading bytes of a wrong digest matched; the surrounding walk is variable-time
and does not claim otherwise, and what it can leak is coarse registry shape
rather than anything about the secret.

### The continuous integration red

It cleared on a clean rebase, so the explanation held - but the cause is better
than "flake". The red was a single genuine `Assert.False()` failure in
`HostedAccountStatusTests.Enrolled_without_a_recorded_email_is_signed_in_with_the_identity_absent`,
which arrived as an assertion the test itself raised, not as a crash. `origin/main`
carries **#1916**, which stopped `HostedAccountStatusTests` sharing an account
subject with other suites - a real cross-suite state bleed, fixed. This branch
was based before it.

So this was not an intermittent condition to be waited out: it was a defect that
had already been fixed on main and that a stale base kept reproducing. The
previously-red test is confirmed `Passed` **by name** on the rebased head. There
is no known-failure allowance left to carry here, and carrying one would now
conceal whatever arrives next.

thefrederiksen:
## Review record: original ruling and current-head verdict

This comment records my original ruling in my own words; the earlier fleet message was a notification, not a durable review artifact, and relying on a relayed summary was a process failure.

### Original design ruling

- **Plain SHA-256 with no per-record salt is the correct verifier here.** These device keys are 32-byte values generated by a cryptographic RNG, not human passwords. An attacker who obtains a digest still faces the infeasible 256-bit key space; password stretching adds cost to every authenticated Gateway request without buying a meaningful defense, and a salt does not help against precomputation for independent high-entropy random secrets.
- **The constant-time claim applies to the digest comparison, not to the lookup.** `CryptographicOperations.FixedTimeEquals` is appropriate for comparing the supplied digest with one stored digest. The surrounding registry scan skips records and returns on a match, so it is variable-time and must not be described as a constant-time lookup.
- **Concern 1 — enrollment idempotence:** the pre-rework implementation made every repeated `RegisterIfAbsent` call rotate the key. A lost or delayed response, a retry, a double submit, or parallel calls could therefore leave the client persisting a key that a later call had already retired. That was a serious regression in the method's prior retry-safe contract and needed to be fixed or disproved at every caller.
- **Concern 2 — secure-erasure overclaim:** rewriting `devices.json` removes plaintext from the live file, but it cannot establish erasure from Azure Files snapshots, backups, soft-delete retention, or other historical copies. The claim needed to be narrowed to the live registry file, with retained-copy purge and affected-key rotation tracked as separate operational work.

### Assessment of `db5319990f1b473a1a8156af58e91d0d0614f0c2`

The rework resolves most of that ruling. The comparison wording is now precise. The migration preserves valid legacy records' key and tenant/account/cloud/type bindings while rewriting the live JSON without `DeviceKey`. The retention language is properly narrowed and explicitly leaves purge plus key rotation as operational work. The in-process retry behavior is also materially corrected: `RegisterIfAbsent` is serialized, repeats inside the window replay one key, expiry rotates in place, and the stored-hash supersession check prevents the sequential re-pairing path from resurrecting an old key.

There is still one blocking correctness/security finding in the new replay mechanism:

- **The plaintext replay cache does not actually expire on the five-minute deadline.** [`PruneExpiredIssues`](https://github.com/thefrederiksen/devthrottle/blob/db5319990f1b473a1a8156af58e91d0d0614f0c2/src/CcDirector.Gateway/Pairing/DeviceRegistry.cs#L272-L279) is called only from [`TryReplayIssuedKey`](https://github.com/thefrederiksen/devthrottle/blob/db5319990f1b473a1a8156af58e91d0d0614f0c2/src/CcDirector.Gateway/Pairing/DeviceRegistry.cs#L246-L257). A first enrollment for a new device bypasses that method, records its plaintext key in `_replayableIssues`, and performs no pruning. Therefore, if devices enroll once under distinct IDs—the ordinary fresh-enrollment path—every plaintext key remains strongly referenced after five minutes, potentially for the process lifetime. An explicit `Register` likewise leaves an older replay entry resident. That contradicts the class and field documentation that keys are held only for the short replay window, broadens what a later process-memory disclosure can recover, and lets the additional dictionary grow with fresh device IDs. Expired entries need real deadline-driven eviction (or an equivalent design that removes their strong references without requiring a repeat enrollment), plus deterministic coverage of the distinct-device/no-repeat path. Merely checking age when someone asks to replay prevents an expired key from being returned; it does not bound plaintext retention.

**Standing: CHANGES REQUESTED on this head.** The original hash, migration, comparison-scope, idempotence, and erasure-wording decisions are otherwise acceptable. At the time of this verdict the web check is green and the .NET check is still running; my local targeted run timed out behind the machine's saturated shared .NET test lane, so I am not substituting it for the pending CI result.

thefrederiksen:
## Response to the second review - head `0844c5ab`

The finding is correct and I am not arguing with any part of it. Quoting the
decisive sentence back, because it is the one that matters:

> Merely checking age when someone asks to replay prevents an expired key from
> being returned; it does not bound plaintext retention.

That is exactly the error. `PruneExpiredIssues` ran only inside
`TryReplayIssuedKey`, which is reached only when an already-enrolled device asks
to replay. A device enrolling once under a distinct identifier - the ordinary
fresh-enrollment path - never reached it, so its plaintext key stayed strongly
referenced for the life of the process, and the dictionary grew with every fresh
identifier. An explicit `Register` left its older entry resident the same way. I
had also written into the class documentation that keys are held only for a short
window, so the code and the claim disagreed, and the claim was the false one.

I want to name the shape rather than just patch it, because it is the same shape
twice in one unit. **The property that justified the design was the one nothing
tested.** The in-memory plaintext window was defensible *only* because it was
short-lived; short-livedness was therefore the load-bearing property, and it had
no test. The case that would have caught it - a device that enrolls and never
comes back - is the ordinary case, not an exotic one. I tested the retry
behaviour I had just built and never tested the assumption that made building it
acceptable.

### What changed

Retention now has a clock of its own.

- The registry starts a recurring eviction pass **the moment it first retains a
  key**, so retention and the clock that bounds it begin together. A registry
  that never issues a key never starts a clock.
- The pass drops every issue whose window has closed, whether or not anything
  ever enrolls again.
- It runs at **half** the window, so the true bound is one and a half windows,
  not two. At the full window length an entry recorded just after a pass could
  live nearly twice the window - which is not what the documentation claims, and
  the documented bound has to be the real one. Default: at most seven and a half
  minutes.
- `DeviceRegistry` is now `IDisposable`; disposing stops the pass and drops
  everything retained. `GatewayHost.StopAsync` disposes it alongside its other
  timers.
- The lazy prune inside the replay lookup is **removed**. Eviction has exactly
  one owner, so a broken pass cannot be masked by another path tidying up on its
  way past - the same reasoning that removed the zero-window short-circuit from
  the recording path in the previous round. The age check inside the replay
  lookup stays, because it does a different job: it stops an expired key being
  *returned*. It never bounded retention and no longer pretends to.

### Deterministic coverage, not probabilistic

A hand-driven scheduler captures the eviction pass instead of running it on a
timer, so both halves are proven without waiting on a clock - waiting on a real
timer would make the result depend on machine load, which is not a proof, and I
am not adding a second probabilistic red to this unit.

- `TheRegistry_SchedulesARecurringEvictionPass_AsSoonAsItRetainsAKey` - proves a
  pass exists at all; without it the policy below is dead code.
- `ADeviceThatEnrollsOnceAndNeverReturns_DoesNotRetainItsPlaintextKey` - the
  exact broken path, one device, one enrollment, nobody ever knocks again.
- `ManyDevicesEnrollingOnceUnderDistinctIds_DoNotAccumulatePlaintextKeys` - the
  growth half, twenty-five distinct identifiers.
- `AnExplicitRePairing_DoesNotLeaveTheOlderReplayableKeyResident`.
- `Disposing_DropsEveryRetainedPlaintextKey`.
- `TheEvictionInterval_IsHalfTheWindow_AndNeverZero` - asserts the arithmetic
  behind the documented bound, including that a zero window cannot become a busy
  loop.

Retention is asserted **directly**, through an internal
`RetainedReplayableKeyCount`, rather than inferred from behaviour that would
still look correct while the plaintext sat in memory. That is the specific
mistake being corrected, so the assertion had to name the quantity.

The control is `EvictionControl_AKeyStillInsideItsWindow_IsNotEvicted`. Without
it every test above is satisfied by a pass that unconditionally clears the
dictionary - which would silently destroy the retry safety the window exists to
provide. Eviction has to be driven by the deadline, not by being called.

### Why I bounded the exposure rather than removing it

Asked directly, and it deserves a direct answer rather than a default.

The alternative is to stop retaining the secret and have the caller present the
key it already holds: if the presented key verifies, do not rotate, and echo back
what the caller supplied. The server would then never keep plaintext at all.

I did not choose it, for two reasons, and the first is the one that decides it:

1. **It does not cover the concurrent case.** Two enrollment calls in flight for a
   device that has no key yet - one of the three scenarios in the original ruling
   - cannot present anything, because neither has anything to present. The first
   call issues a key, the second finds an active record, has no presented key to
   verify, and must rotate. The client saves whichever response it processes last
   while the registry holds the other, and that is the lock-out this unit exists
   to prevent. Removing the retention would trade a bounded, now-enforced memory
   exposure for an uncovered lock-out path. That is not a strict improvement, and
   I do not think "no plaintext in memory" is worth reintroducing the defect the
   window was built to close.
2. It is a request-contract change across the hosted and signed-in enrollment
   endpoints and every client that calls them, which is a wider blast radius than
   this security unit should take on while it is the thing standing between the
   registry and plaintext at rest.

So the honest summary is: retention is required to cover the concurrent case
under the current contract, and the right response was to make the bound real
instead of asserted. **I still think the presented-key idea is the better end
state** - it would let the window shrink to near zero and cover the common double
submit without retaining anything - and it is worth raising as follow-up work on
the enrollment contract rather than smuggling into this pull request. I am
flagging it rather than claiming it.

### On the checks

The .NET check has since gone green on the previous head `db531999` - so
continuous integration was green on a head that carried this defect. That is not
a complaint about the check; it is the point. Nothing exercised the property, so
nothing could go red. A green run is only as good as what was asked of it.

I am still not admitted to the machine's shared test lane and have not run the
suite locally. The mutation set for this round is registered separately.

thefrederiksen:
## Pre-registered mutation set, round two - head `0844c5ab`

Registered before any run. A predicted red that does not appear is a finding, not
something to be explained afterwards. One primitive per full-suite run, never
combined. Every red is classified by how it ARRIVED - a failure on an assertion
the test itself raised counts; a crash, a build failure, or a framework error is
laundered and proves nothing.

This supersedes the earlier registration only where it overlaps; M1 and M3 from
round one are unchanged and still required, M2 is restated because the code it
targeted moved, and M4/M5/M6 are unchanged.

### New this round

**M7 - the eviction pass is scheduled at all.**
Mutation: in `StartEvictionPassOnce`, return before scheduling.
Predicted red: `TheRegistry_SchedulesARecurringEvictionPass_AsSoonAsItRetainsAKey`,
`ADeviceThatEnrollsOnceAndNeverReturns_DoesNotRetainItsPlaintextKey`,
`ManyDevicesEnrollingOnceUnderDistinctIds_DoNotAccumulatePlaintextKeys`,
`AnExplicitRePairing_DoesNotLeaveTheOlderReplayableKeyResident`.
The first fails on the scheduling assertion; the rest fail inside `RunOnce`,
which throws with "the registry never scheduled an eviction pass". That throw is
raised by the test harness's own logic about the condition under test, so it
counts - but it arrives as an exception rather than an `Assert`, and I am
flagging that here in advance so it is classified deliberately and not waved
through.

**M8 - the eviction policy honours the deadline.**
Mutation: in `EvictExpiredReplayableIssues`, remove the age comparison so it
removes every entry unconditionally.
Predicted red: `EvictionControl_AKeyStillInsideItsWindow_IsNotEvicted`.
Predicted GREEN: all four retention tests above - they use a zero window, where
evicting everything is indistinguishable from evicting the expired. That is
precisely why the control exists, and it is the only test that can catch this
mutation. If the control does not go red here, the control is not doing its job
and the retention tests prove nothing about the deadline.

**M9 - retention actually starts.**
Mutation: in `RecordReplayableIssue`, drop the `StartEvictionPassOnce()` call.
Same predicted reds as M7. Registered as its own run because it is a distinct
line that can be individually wrong while `StartEvictionPassOnce` itself stays
correct - the pass exists but nothing ever arms it.

**M10 - the interval that produces the documented bound.**
Mutation: in `EvictionIntervalFor`, return the window itself rather than half.
Predicted red: `TheEvictionInterval_IsHalfTheWindow_AndNeverZero`,
`TheRegistry_SchedulesARecurringEvictionPass_AsSoonAsItRetainsAKey` (the interval
assertion).
This one is deliberately included even though the behaviour still "works": the
documented retention bound is one and a half windows, and at the full window
length the true bound is two. A documented bound that nothing checks is how this
unit got into trouble in the first place.

**M11 - disposal releases the plaintext.**
Mutation: in `Dispose`, remove `_replayableIssues.Clear()`.
Predicted red: `Disposing_DropsEveryRetainedPlaintextKey`.
Its own run: it can be individually wrong while the timer teardown stays correct,
and the protection - no plaintext outlives the registry - still breaks.

### Carried forward, unchanged and still required

- **M1** delete the replay branch from `RegisterIfAbsent`.
- **M2** delete the age check from `TryReplayIssuedKey` (the line is unchanged;
  the lazy prune that used to sit beside it is gone, which makes this mutation
  cleaner than it was).
- **M3** delete the supersession check from `TryReplayIssuedKey`.
- **M4** remove `lock (_enrollLock)`. **Still declared PROBABILISTIC** - its red
  depends on interleaving. It remains the one line in this unit without a
  deterministic single-mutation proof, and I continue to report it as a gap
  rather than count it proven.
- **M5** add `DeviceKey = key,` back into the record built by `Register`.
- **M6** delete the migration block from `Load`.

### Count and baseline

**Eleven mutation runs**, each preceded and followed by a restored run, on a
baseline that must be green AND explicitly complete on head `0844c5ab`. A run
with no summary line is not a result in either direction. The reconciliation
`passed + failed under the mutation == passed under the restored run` must
compute for every one, and every test named above must be confirmed to have
EXECUTED by name.

I am not admitted to the shared test lane and will not queue into it. Eleven runs
is the number; scheduling is the Manager's call.

thefrederiksen:
## Continuous integration on `0844c5ab` - green, and what that does not mean

Run [29726883238](https://github.com/thefrederiksen/devthrottle/actions/runs/29726883238)
completed **success**, both jobs: `Build & Test (.NET)` and `Build & Test (web)`.

All 29 `DeviceKeyAtRestTests` are confirmed `Passed` **by name**, including the
seven added this round:

- `TheRegistry_SchedulesARecurringEvictionPass_AsSoonAsItRetainsAKey`
- `ADeviceThatEnrollsOnceAndNeverReturns_DoesNotRetainItsPlaintextKey`
- `ManyDevicesEnrollingOnceUnderDistinctIds_DoNotAccumulatePlaintextKeys`
- `AnExplicitRePairing_DoesNotLeaveTheOlderReplayableKeyResident`
- `EvictionControl_AKeyStillInsideItsWindow_IsNotEvicted`
- `Disposing_DropsEveryRetainedPlaintextKey`
- `TheEvictionInterval_IsHalfTheWindow_AndNeverZero`

**The limit of that result, stated deliberately.** This green proves the new
tests PASS. It does not prove they can FAIL, and that is the claim that matters.
The previous head `db531999` was equally green while the replay cache retained
plaintext for the life of the process - the suite could not see it, because
nothing asked. A test that cannot fail is indistinguishable from one that passes.

So this run is a precondition for the proof, not the proof. The eleven
pre-registered mutation runs are what establish that each protection is
load-bearing, and they have not been run - the machine's shared test lane is held
by other work and I have not queued into it. The baseline for those runs will be
taken fresh at this head and reconciled against itself, never against a count
quoted from an earlier head.

M4, the enrollment lock, remains a declared gap: its red depends on thread
interleaving, so it will not be promoted on a single green.

thefrederiksen:
## Mutation-run protocol, registered before admission

Recorded here rather than kept in conversation, because the whole value of
pre-registration is that it cannot be adjusted after seeing a result. This
supplements the eleven-run set above; it does not change any mutation or any
predicted red.

### The failure this defends against

A mutated file can fail to reach the assembly the tests actually execute. The
build then **succeeds while compiling nothing**, and the run exercises the
previous binary. The existing "abort on a failed build" rule does not catch it,
because a skipped build is not a failed build - it succeeds silently, having done
nothing.

The dangerous direction is a mutation that never lands. The arm comes back green,
which reads as **a predicted red that did not appear** - and under this unit's own
rules that is escalated as a FINDING. So the result is not a lost arm, it is a
FABRICATED one, and the more rigorous the pre-registration the more convincing the
fabrication looks. A protocol that produces false findings is worse than no
protocol.

### Rules

1. **Mutate by editing the file in place. Restore with `git checkout -- <path>`.**
   Never move or copy a backup file over a mutated one. A backup carries an older
   timestamp, so the build judges the assembly up to date and skips the rebuild.
2. **Verify the compile actually happened on every arm AND every restore** - not
   only on the mutation. A restore that does not land leaves the mutation in the
   binary and poisons the run that is supposed to re-establish the baseline.

### How the verification is done

Two independent checks, because the first is mechanical and the second is the one
that explains a failure.

**Primary - hash the assembly that is actually executed.** The test host loads
`src/CcDirector.Gateway.Tests/bin/Debug/net10.0/CcDirector.Gateway.dll` - the
COPY in the test project's output directory, not the one in the Gateway project's
own output. Those are two distinct ways a change can fail to arrive: a skipped
compile, and a skipped copy. The copy is what runs, so the copy is what gets
hashed. .NET builds are deterministic, so identical source produces an identical
assembly and the comparison is exact:

- `H0` = hash at the restored baseline.
- After mutating and building, `H1`. **Require `H1 != H0`.** If they are equal the
  mutation never reached the executed binary: the arm is **VOID and is re-run**,
  and it is never reported as a predicted red that did not appear.
- After restoring and building, `H2`. **Require `H2 == H0`.** Byte-identical is
  the proof the restore truly restored, which a passing test suite alone does not
  establish.

**Secondary - read the build line.** Confirm the output shows a real compile
rather than an up-to-date skip, on the mutation and on the restore. This is what
tells me WHY a hash comparison failed when it does.

### What the report will say

For every one of the eleven arms: the three hashes, that `H1 != H0` and
`H2 == H0`, and that the build line showed an actual compile on both the mutation
and the restore. Any arm that cannot show that is re-run rather than interpreted.

The reconciliation is unchanged - `passed + failed under the mutation == passed
under the restored run` - and every named test is confirmed to have EXECUTED by
name. A run with no summary line is not a result in either direction.

M4 remains a declared gap. A single green does not settle it, for exactly the
reason the body of this pull request now gives about green runs.

thefrederiksen:
STANDING: REJECTED
Exact head: `0844c5ab0023694e3a07bea5beb5544869d2c441`

The persisted-hash migration and unified verification path are sound on static inspection, but the bounded-retention guarantee is not yet proved in the production shape.

All deterministic eviction tests construct the registry with the hand-driven scheduler, so they exercise the `_scheduleEviction` branch. Production takes the separate `_evictionTimer = new Timer(...)` branch. Deleting only that production branch leaves every current retention test green while removing the production clock. That is an independently bypassable primitive not present in the eleven-arm plan. Refactor to one scheduler call-site whose default implementation owns the real timer and whose test implementation captures the callback, or add a deterministic canary and full-suite arm for the production branch.

Two registered proof defects also remain:

- M4 is explicitly probabilistic. The enrollment serialization primitive needs a deterministic interleaving seam/barrier and an assertion red; a declared probabilistic gap cannot clear a security unit.
- M7 predicts several failures through `HandDrivenEviction.RunOnce` throwing `InvalidOperationException`. Under the proof rule those are crash/exception reds, not assertion reds. Make the harness assert scheduling as a format/precondition fact before invocation so every counted red is raised by an explicit test assertion.

After those changes, correct the contradictory M1 prediction list, rebase onto current main, and register/run the complete exact-head set with clean baseline, one full-Gateway arm per remaining primitive, restored green after each arm, assembly/source/build identity checks, complete summaries, and reconciliation. Current exact-head CI is green but the mutation proof is still pending.

PR #1900, opened 2026-07-19
fix(gateway): tenant-key the dictation upload and transcript store

thefrederiksen:
Closes the store half of thefrederiksen/devthrottle_internal#889: tenant-key the dictation upload and transcript store so a caller cannot claim another account's upload id.

This does NOT re-enable dictation on hosted. That is thefrederiksen/devthrottle_internal#889 part (a) and is held: the session locate inside the completion deliberately still resolves in the local partition, so a hosted session is still not found and the route stays as dead as it is today. Partitioning first is the whole point - converting the route first would arm this store rather than close it.

## The defect

`VoiceUploadStore` had ONE GLOBAL ROOT and keyed every directory, chunk and record solely by the caller-supplied upload id. `DictationDeliveryRecord` carried no tenant. The upload, chunk, ack and abandon legs authenticated a DEVICE and never asked whose upload it was. The in-flight completion cache was static and keyed by upload id alone.

So after account A completed upload id X, account B posting `Idempotency-Key: X` read A's terminal record and was handed A's TRANSCRIPT; and before X reached a terminal state, B could overwrite A's chunks, ack or abandon A's record, or race its completion. A GUID is not a tenant boundary - secrecy of an identifier is not authorization, and upload ids travel in client logs, retries and store-and-forward queues.

## What changed

1. **The store is partitioned by tenant** - the on-disk root and the record. `ForTenant(tenant)` returns a store whose every path, gate, chunk and record lives inside that tenant's partition, so the isolation is the DIRECTORY, not a predicate a later edit could forget. The same upload id is now legally two different uploads in two accounts.
2. **The record carries its tenant**, stamped in the one place a record is written and required to match on every read - an independent second check behind the directory, so a mis-computed partition root cannot silently hand over a transcript.
3. **The static single-flight completion cache and the upload-to-session map are keyed by tenant and upload id.** Those are in-memory and no on-disk partition can guard them.
4. **All five legs - upload, chunk, complete, ack, abandon - resolve and authorize the tenant** from the authenticated device key, using the same seam the prompt log and the cockpit read path use. Deny-by-default: on hosted, an authenticated key with no bound tenant gets 403, never the local partition.
5. **An unscoped leg is no longer something a person can write.** `GatewayDictationEndpoint.Map` does not accept or capture a `VoiceUploadStore` at all - it takes a `DictationTenantGate`, which privately owns the only raw store and whose `TryOpen` cannot return an unscoped one, with the refusal folded into the same call so a caller who skips the check gets no store rather than an unchecked one. The two static caches got the same treatment: the tenant is a required parameter and the key is composed inside the cache type. The realistic defect was never an edit to one of these five legs - it was the sixth leg added later by someone who never read the reasoning around them.

The tenant id becomes a PATH COMPONENT here, so only the one canonical lowercase form the registry mints is accepted. A traversal spelling (`..` resolves to the parent partition) and an uppercase alias (same directory on Windows and Azure Files, different identity to the case-sensitive tenants table) are REFUSED rather than normalised, and the combined path is asserted to resolve inside the partition root. Worked example followed: `GatewayPromptLog`.

**Self-host is unchanged and is the control**: the local tenant keeps the exact root it has always used, nothing migrates, and no upload moves.

## Why this is being re-proved on current main

> thefrederiksen/devthrottle#1915 partitions voice state per tenant and this partitions dictation state per tenant - two partitions of the same storage root, landed independently, **never once executed in the same suite.**

That is the case for re-proving this change against current main rather than merging evidence gathered on an older base, and it is checkable by anyone in a minute.

Main has moved seven commits since the base the first full proof ran on, four of them merges touching adjacent surface: the stats deny (#1888), the voice partition (#1915), the entitlement gate (#1901), and pinned-root storage behaviour (#1891). They bring three brand-new Gateway test classes - `WingmanVoiceTenantPartitionTests`, `HostedStatsDenyTests`, `HostedEntitlementGateTests` - plus changes to six existing ones. A full-suite revert exists to answer "did this break something else"; asked against a suite that does not contain the code this change now sits beside, that question was unanswerable.

**Pre-registered suspect, written down before the run rather than assembled after it:** thefrederiksen/devthrottle#1891 changed legacy-data migration behaviour for an explicitly pinned root, and the tenant tests here pin `CC_DIRECTOR_ROOT` to a temporary directory - the same mechanism from opposite sides. If the baseline surprises us, that is the first place to look. If it turns out to be something else entirely, that will be said plainly; a wrong pre-registered suspect is still worth more than a right retrospective one.

## Proof

The evidence for this change is the full-suite mutation proof in the pull request comments, not this section. **Eight one-primitive mutations, each a full Gateway suite, plus a restored run at the exact pushed head**, with every red named, its kind (assertion or crash) reported, its actual assertion text quoted, and each arm reconciled arithmetically against the restored count.

Two results worth stating here because they are easy to overstate elsewhere:

- **The record tenant stamp is not a disclosure guard.** Dropping it while the ownership check stays strict makes account records unattributed and therefore **unreadable** - a correctness and availability failure. Its reds are all "the record was absent", including account A failing to read account A's own transcript. It earns its own proof because the design requires the persisted ownership marker; it does not demonstrate cross-tenant disclosure.
- **`BelongsHere` is the disclosure guard**, and its red says so concretely: account B's partition returning account A's record with A's transcript in it.

Two defences are deliberately left without canaries, documented at the guards themselves: the path-containment belt in `PartitionRootFor` (unreachable while `IsMintedAccountTenant` admits only canonical lowercase minted identifiers - the comment names predicate loosening as the point it becomes live and must gain one) and the partition-container skip in the pending-record projection (a clarity and cost skip; the real boundary on that line is `BelongsHere`, which is separately canaried). Neither got a test invented for it, because the only test either could have is one that cannot fail.

## Test counts

Reported with the full-suite proof at the head being merged, across all seven test projects. The earlier counts in this body's history referred to an older base and an older suite composition and are superseded - including a `HostedAccountStatusTests` failure that was genuinely pre-existing at the time and has since been fixed on main by thefrederiksen/devthrottle#1916.

No web workspace is touched (no TypeScript or contract change).

### DIFF
--- a/src/CcDirector.Gateway.Tests/DictationOmittedTenantBoundaryTests.cs
+++ b/src/CcDirector.Gateway.Tests/DictationOmittedTenantBoundaryTests.cs
@@ -1,0 +1,395 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Net.Http.Json;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using CcDirector.Core;
+using CcDirector.Core.Tenancy;
+using CcDirector.Gateway.Api;
+using CcDirector.Gateway.Discovery;
+using CcDirector.Gateway.Pairing;
+using CcDirector.Gateway.Tenancy;
+using CcDirector.Gateway.Transcription;
+using CcDirector.Gateway.Voice;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// Issue #1884, the FAIL-OPEN case: a HOSTED Gateway whose dictation endpoint was wired with NO tenant
+/// boundary must REFUSE on all five legs, never quietly serve the shared self-host partition.
+///
+/// Why this is its own test class. The tenant boundary is a SECURITY argument, and it used to be declared
+/// optional and nullable (<c>HostedTenantBoundary? tenantBoundary = null</c>) with <c>ResolveTenant</c>
+/// answering <see cref="TenantId.Local"/> whenever it was absent - WITHOUT ever asking
+/// <see cref="GatewayHostedMode.IsHosted"/>. Correctness therefore rested on every present and future call
+/// site remembering a one-word argument. Any hosted call site, test, or rewire that omitted it sent
+/// register, chunk, complete, ack and abandon to the base shared root - reopening transcript disclosure,
+/// chunk overwrite, ack, abandon and completion-cache joining - and NOTHING failed loud to say so. An
+/// optional security argument is indistinguishable from a resolved one at the call site, which is exactly
+/// why that mistake survives review.
+///
+/// TWO DEFENCES, and this class proves the runtime one. The compile-time one is that
+/// <c>GatewayDictationEndpoint.Map</c> now takes the boundary as a REQUIRED parameter, so omitting it will
+/// not build; the tests below have to force a null through explicitly (<c>null!</c>) to reproduce the
+/// miswire at all, which is the point - the accident is now unrepresentable, and even the deliberate
+/// version is refused.
+///
+/// EVERY LEG, not one. A fix proven on one leg says nothing about the other four: each leg calls
+/// <c>ResolveTenant</c> itself, so each is its own opportunity to fall back to Local.
+///
+/// EVERY ABSENCE CLAIM HAS A POSITIVE COMPANION. Each leg asserts the concrete refusal STATUS (403), the
+/// exact refusal the deny path - and only the deny path - produces, and the response's property set as an
+/// ALLOW-LIST (a substring check cannot see an extra leaked field). On top of that, a secret is seeded in
+/// the LOCAL partition under the very same upload id, and each leg is checked to have left it alone: not
+/// disclosed, not overwritten, not retired, not abandoned. Before the fix, the register leg alone handed
+/// that secret transcript straight back in its terminal-register body.
+///
+/// AND THE ROUTES ARE PROVEN TO EXIST. A test that drives a verb or path production does not map gets a
+/// framework 405/404 BEFORE endpoint selection, so no handler runs and the test passes for the wrong
+/// reason. Every request here is replayed against an IDENTICAL harness differing ONLY in the boundary
+/// argument (<see cref="Wired"/>), which answers each of them successfully - so a 403 above is the guard
+/// refusing, never a route that does not exist.
+///
+/// REVERT-PROVE: restore <c>ResolveTenant</c> to <c>boundary is null ? TenantId.Local : ...</c> and all five
+/// tests below go RED - the unwired legs answer 200 and the register leg hands back the seeded secret.
+///
+/// The assembly runs sequentially (TestParallelization), so toggling CC_GATEWAY_HOSTED and the storage root
+/// here is safe; both are restored in DisposeAsync.
+/// </summary>
+public sealed class DictationOmittedTenantBoundaryTests : IAsyncLifetime
+{
+    private const string GatewayToken = "test-token";
+
+    /// <summary>The thing an unwired hosted leg must never read, overwrite, retire, or abandon.</summary>
+    private const string LocalSecretTranscript = "self-host-partition-secret-transcript";
+    private const string LocalSecretAudio = "self-host-partition-secret-audio";
+
+    /// <summary>The one message the deny path produces. Asserted exactly, so a 403 from anywhere else fails.</summary>
+    private const string DenyMessage = "no tenant is bound to this request";
+
+    private readonly string _storageRoot =
+        Path.Combine(Path.GetTempPath(), "cc-dictation-nobound-" + Guid.NewGuid().ToString("N"));
+
+    private string? _priorHosted;
+    private string? _priorRoot;
+
+    /// <summary>The miswired host: hosted mode ON, no tenant boundary. The subject of every test here.</summary>
+    private Harness _unwired = null!;
+
+    /// <summary>The identical host WITH the boundary. The control that proves the routes and harness work.</summary>
+    private Harness _wired = null!;
+
+    private TenantId _tenant;
+
+    public async Task InitializeAsync()
+    {
+        _priorHosted = Environment.GetEnvironmentVariable("CC_GATEWAY_HOSTED");
+        _priorRoot = Environment.GetEnvironmentVariable("CC_DIRECTOR_ROOT");
+        Environment.SetEnvironmentVariable("CC_GATEWAY_HOSTED", "1");
+        Environment.SetEnvironmentVariable("CC_DIRECTOR_ROOT", _storageRoot);
+
+        // A canonical lowercase minted GUID - the ONE tenant spelling the registry produces and the upload
+        // store accepts as a partition name.
+        _tenant = new TenantId(Guid.NewGuid().ToString("D"));
+
+        _unwired = await Harness.StartAsync(_storageRoot, "unwired", _tenant, wireBoundary: false);
+        _wired = await Harness.StartAsync(_storageRoot, "wired", _tenant, wireBoundary: true);
+    }
+
+    public async Task DisposeAsync()
+    {
+        await _unwired.DisposeAsync();
+        await _wired.DisposeAsync();
+        Environment.SetEnvironmentVariable("CC_GATEWAY_HOSTED", _priorHosted);
+        Environment.SetEnvironmentVariable("CC_DIRECTOR_ROOT", _priorRoot);
+        try { if (Directory.Exists(_storageRoot)) Directory.Delete(_storageRoot, true); } catch { /* cleanup */ }
+    }
+
+    // ===== leg 1: register =========================================================================
+
+    [Fact]
+    public async Task Register_refuses_on_hosted_when_the_boundary_was_omitted()
+    {
+        // THE CONCRETE DISCLOSURE this class exists for. Seed a delivered upload in the LOCAL partition -
+        // the base shared root, where self-host's dictations live - and register that same id against the
+        // unwired hosted host. Before the fix the missing boundary resolved to Local, the terminal-register
+        // short-circuit fired, and the response body carried the seeded transcript verbatim.
+        var id = Guid.NewGuid().ToString();
+        LocalStore().MarkDelivered(id, submitted: true, movedOn: false, transcript: LocalSecretTranscript);
+
+        var (status, body) = await Read(await _unwired.RegisterAsync(id));
+        AssertRefusal(status, body);
+
+        // The local record is untouched: still delivered, still its own words.
+        var record = LocalStore().ReadRecord(id);
+        Assert.NotNull(record);
+        Assert.Equal(DictationDeliveryState.Delivered, record!.State);
+        Assert.Equal(LocalSecretTranscript, record.Transcript);
+
+        // Positive companion: the SAME request on the SAME harness WITH a boundary is served, so the 403
+        // above is the guard refusing and not a route that does not exist or a host that answers nothing.
+        var wired = await Read(await _wired.RegisterAsync(id));
+        Assert.Equal(HttpStatusCode.OK, wired.status);
+        Assert.DoesNotContain(LocalSecretTranscript, wired.body, StringComparison.Ordinal);
+        var wiredRegister = JsonDocument.Parse(wired.body).RootElement;
+        // The fresh-register shape, which ONLY the genuine handler produces: an upload id, and no terminal
+        // short-circuit - the local partition's delivered record was not this host's to see.
+        Assert.False(string.IsNullOrWhiteSpace(wiredRegister.GetProperty("upload_id").GetString()));
+        Assert.False(wiredRegister.TryGetProperty("terminal", out _));
+    }
+
+    // ===== leg 2: chunk ============================================================================
+
+    [Fact]
+    public async Task Chunk_refuses_on_hosted_when_the_boundary_was_omitted()
+    {
+        // Staged audio in the local partition. An unwired hosted chunk must not be able to overwrite it,
+        // extend it, or even confirm that the upload id exists.
+        var id = Guid.NewGuid().ToString();
+        var local = LocalStore();
+        local.Register(id);
+        await local.StoreChunkAsync(id, 0, Encoding.UTF8.GetBytes(LocalSecretAudio), null, default);
+        Assert.Equal(LocalSecretAudio, await StagedTextAsync(local, id));
+
+        var (status, body) = await Read(await _unwired.PutChunkAsync(id, 0, "overwritten-by-the-unwired-host"));
+        AssertRefusal(status, body);
+        Assert.Equal(LocalSecretAudio, await StagedTextAsync(local, id));
+
+        // Positive companion: PUT on this exact path is a real production route and the wired host stores
+        // the chunk - into its OWN tenant partition, leaving the local bytes alone.
+        await _wired.RegisterAsync(id);
+        var wired = await Read(await _wired.PutChunkAsync(id, 0, "wired-host-audio"));
+        Assert.Equal(HttpStatusCode.OK, wired.status);
+        Assert.True(JsonDocument.Parse(wired.body).RootElement.GetProperty("ok").GetBoolean());
+        Assert.Equal(LocalSecretAudio, await StagedTextAsync(local, id));
+    }
+
+    // ===== leg 3: complete =========================================================================
+
+    [Fact]
+    public async Task Complete_refuses_on_hosted_when_the_boundary_was_omitted()
+    {
+        // The other door onto the same disclosure: complete's cached-terminal-outcome short-circuit. It also
+        // reaches the STATIC single-flight cache, which an unwired host would key as "Local|<id>" - joining
+        // the self-host partition's in-flight run.
+        var id = Guid.NewGuid().ToString();
+        LocalStore().MarkDelivered(id, submitted: true, movedOn: false, transcript: LocalSecretTranscript);
+
+        var (status, body) = await Read(await _unwired.CompleteAsync(id));
+        AssertRefusal(status, body);
+
+        // Nothing of the local record was disclosed and nothing of it moved.
+        var record = LocalStore().ReadRecord(id);
+        Assert.NotNull(record);
+        Assert.Equal(DictationDeliveryState.Delivered, record!.State);
+        Assert.Equal(LocalSecretTranscript, record.Transcript);
+
+        // Positive companion: the wired host reaches its own handler on this exact verb and path. It is NOT
+        // refused, and it does not see the local partition's terminal outcome either.
+        var wired = await Read(await _wired.CompleteAsync(id));
+        Assert.NotEqual(HttpStatusCode.Forbidden, wired.status);
+        Assert.NotEqual(HttpStatusCode.MethodNotAllowed, wired.status);
+        Assert.DoesNotContain(LocalSecretTranscript, wired.body, StringComparison.Ordinal);
+    }
+
+    // ===== leg 4: ack ==============================================================================
+
+    [Fact]
+    public async Task Ack_refuses_on_hosted_when_the_boundary_was_omitted()
+    {
+        // An ack DELETES state - it retires the durable tombstone that stops a delivered turn being injected
+        // twice. An unwired hosted host must not be able to retire the self-host partition's tombstone.
+        var id = Guid.NewGuid().ToString();
+        LocalStore().MarkDelivered(id, submitted: true, movedOn: false, transcript: LocalSecretTranscript);
+
+        var (status, body) = await Read(await _unwired.AckAsync(id));
+        AssertRefusal(status, body);
+        Assert.NotNull(LocalStore().ReadRecord(id));
+        Assert.Equal(LocalSecretTranscript, LocalStore().ReadRecord(id)!.Transcript);
+
+        // Positive companion: the wired host's ack really does run - it answers the handler's own shape and
+        // reports retired=false for an id that is not in ITS partition, which is a refusal by partition, not
+        // a broken route. The local tombstone is still standing.
+        var wired = await Read(await _wired.AckAsync(id));
+        Assert.Equal(HttpStatusCode.OK, wired.status);
+        var wiredRoot = JsonDocument.Parse(wired.body).RootElement;
+        Assert.True(wiredRoot.GetProperty("ok").GetBoolean());
+        Assert.False(wiredRoot.GetProperty("retired").GetBoolean());
+        Assert.NotNull(LocalStore().ReadRecord(id));
+    }
+
+    // ===== leg 5: abandon ==========================================================================
+
+    [Fact]
+    public async Task Abandon_refuses_on_hosted_when_the_boundary_was_omitted()
+    {
+        // Abandon DISCARDS staged audio and clears the session lock. An unwired hosted host must not be able
+        // to resolve or discard the self-host partition's live dictation.
+        var sessionId = Guid.NewGuid().ToString();
+        var id = Guid.NewGuid().ToString();
+        var local = LocalStore();
+        local.Register(id);
+        local.MarkPending(id, sessionId);
+        await local.StoreChunkAsync(id, 0, Encoding.UTF8.GetBytes(LocalSecretAudio), null, default);
+        Assert.True(local.IsSessionLocked(sessionId));
+
+        var (status, body) = await Read(await _unwired.AbandonAsync(id));
+        AssertRefusal(status, body);
+        Assert.Equal(DictationDeliveryState.Pending, local.ReadRecord(id)!.State);
+        Assert.Equal(LocalSecretAudio, await StagedTextAsync(local, id));
+        Assert.True(local.IsSessionLocked(sessionId));
+
+        // Positive companion: the wired host's abandon really does run on this exact verb and path - it
+        // answers the handler's own shape, inside its own partition, and the local dictation is still live.
+        var wired = await Read(await _wired.AbandonAsync(id));
+        Assert.Equal(HttpStatusCode.OK, wired.status);
+        Assert.True(JsonDocument.Parse(wired.body).RootElement.GetProperty("abandoned").GetBoolean());
+        Assert.Equal(DictationDeliveryState.Pending, local.ReadRecord(id)!.State);
+        Assert.True(local.IsSessionLocked(sessionId));
+    }
+
+    // ===== the refusal itself ======================================================================
+
+    /// <summary>
+    /// The refusal, asserted positively: the concrete status, the exact message only the deny path emits,
+    /// and the response's property set as an ALLOW-LIST. The allow-list matters - a substring check for what
+    /// must be absent cannot see an EXTRA field that leaked, and the whole hazard here is a body that
+    /// carries another partition's state.
+    /// </summary>
+    private static void AssertRefusal(HttpStatusCode status, string body)
+    {
+        Assert.Equal(HttpStatusCode.Forbidden, status);
+        var root = JsonDocument.Parse(body).RootElement;
+        Assert.Equal(JsonValueKind.Object, root.ValueKind);
+        Assert.Equal(new[] { "error" }, root.EnumerateObject().Select(p => p.Name).ToArray());
+        Assert.Equal(DenyMessage, root.GetProperty("error").GetString());
+        Assert.DoesNotContain(LocalSecretTranscript, body, StringComparison.Ordinal);
+        Assert.DoesNotContain(LocalSecretAudio, body, StringComparison.Ordinal);
+    }
+
+    // ===== helpers =================================================================================
+
+    /// <summary>The LOCAL (base, shared, self-host) partition - the one an unwired hosted leg would take.</summary>
+    private VoiceUploadStore LocalStore() => new VoiceUploadStore(Harness.UploadRoot(_storageRoot));
+
+    private static async Task<string> StagedTextAsync(VoiceUploadStore store, string uploadId)
+    {
+        var assembled = await store.AssembleAsync(uploadId, 1);
+        return assembled.Audio is null ? "" : Encoding.UTF8.GetString(assembled.Audio);
+    }
+
+    private static async Task<(HttpStatusCode status, string body)> Read(HttpResponseMessage resp)
+    {
+        using (resp) return (resp.StatusCode, await resp.Content.ReadAsStringAsync());
+    }
+
+    /// <summary>
+    /// A minimal host that maps ONLY the dictation endpoint. The two instances are built by the same code
+    /// with the same dependencies and differ in EXACTLY ONE argument - whether the tenant boundary is
+    /// supplied - so a difference in behaviour between them can only be that argument.
+    /// </summary>
+    private sealed class Harness : IAsyncDisposable
+    {
+        private readonly WebApplication _app;
+        private readonly HttpClient _http;
+
+        private Harness(WebApplication app, HttpClient http)
+        {
+            _app = app;
+            _http = http;
+        }
+
+        internal static string UploadRoot(string storageRoot) => Path.Combine(storageRoot, "dictation-uploads");
+
+        internal static async Task<Harness> StartAsync(string storageRoot, string name, TenantId tenant, bool wireBoundary)
+        {
+            var devices = new DeviceRegistry(Path.Combine(storageRoot, name, "devices.json"));
+            var deviceKey = devices.Register("dev-" + name, "M-" + name).DeviceKey;
+            devices.SetAccountBinding("dev-" + name, "sub-" + name, tenant.Value);
+
+            // BOTH harnesses share ONE upload root, so "the unwired host did not touch the local partition"
+            // is a claim about the very bytes the wired host would have to reach through its own partition.
+            var uploads = new VoiceUploadStore(UploadRoot(storageRoot));
+
+            var builder = WebApplication.CreateBuilder();
+            builder.Logging.ClearProviders();
+            var app = builder.Build();
+            app.Urls.Add("http://127.0.0.1:0");
+
+            var transcription = new GatewayTranscriptionService(
+                new KeyVault(Path.Combine(storageRoot, name, "keyvault.json")));
+
+            // The ONE argument under test. The unwired case has to be forced through with null! because the
+            // parameter is REQUIRED - the compile-time half of the fix - so the accidental version of this
+            // miswire cannot be written at all.
+            HostedTenantBoundary? boundary = wireBoundary
+                ? new HostedTenantBoundary(new AsyncLocalTenantContext(), devices)
+                : null;
+
+            GatewayDictationEndpoint.Map(app,
+                new DirectorRegistry(Path.Combine(storageRoot, name, "instances")),
+                owners: null,
+                token: GatewayToken,
+                transcription,
+                new TranscribingSessions(),
+                uploads,
+                devices,
+                boundary!);
+
+            await app.StartAsync();
+            var http = new HttpClient
+            {
+                BaseAddress = new Uri(app.Urls.First()),
+                Timeout = TimeSpan.FromSeconds(30),
+            };
+            http.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", deviceKey);
+            return new Harness(app, http);
+        }
+
+        // The verbs and paths below are the production ones, and the wired harness answering each of them
+        // is what proves it: a verb or path production does not map is answered 405/404 by the framework
+        // BEFORE endpoint selection, so no handler would run on either harness.
+        internal async Task<HttpResponseMessage> RegisterAsync(string uploadId, string? sessionId = null)
+        {
+            using var req = new HttpRequestMessage(HttpMethod.Post, "/dictation/upload")
+            {
+                Content = JsonContent.Create(new { sessionId = sessionId ?? Guid.NewGuid().ToString(), baselineBufferBytes = 0 }),
+            };
+            req.Headers.Add("Idempotency-Key", uploadId);
+            return await _http.SendAsync(req);
+        }
+
+        internal async Task<HttpResponseMessage> PutChunkAsync(string uploadId, int index, string text)
+        {
+            using var content = new ByteArrayContent(Encoding.UTF8.GetBytes(text));
+            content.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");
+            return await _http.PutAsync($"/dictation/{uploadId}/chunk/{index}", content);
+        }
+
+        internal Task<HttpResponseMessage> CompleteAsync(string uploadId)
+            => _http.PostAsJsonAsync($"/dictation/{uploadId}/complete",
+                new { sessionId = Guid.NewGuid().ToString(), totalChunks = 1 });
+
+        internal Task<HttpResponseMessage> AckAsync(string uploadId)
+            => _http.PostAsync($"/dictation/{uploadId}/ack", content: null);
+
+        internal Task<HttpResponseMessage> AbandonAsync(string uploadId)
+            => _http.PostAsync($"/dictation/{uploadId}/abandon", content: null);
+
+        public async ValueTask DisposeAsync()
+        {
+            _http.Dispose();
+            await _app.StopAsync();
+            await _app.DisposeAsync();
+        }
+    }
+}
--- a/src/CcDirector.Gateway.Tests/DictationOmittedTenantBoundaryTests.cs
+++ b/src/CcDirector.Gateway.Tests/DictationOmittedTenantBoundaryTests.cs
@@ -111,6 +111,26 @@ namespace CcDirector.Gateway.Tests;
 /// belt in PartitionRootFor (unreachable while IsMintedAccountTenant stays strict). Neither got a test
 /// invented for it, because the only test either could have is one that cannot fail.
 ///
+/// A SURVIVAL CLAIM NEEDS ITS PRECONDITION ESTABLISHED, NOT ASSUMED - the trap these tests fell into first.
+///
+/// Every leg here asserts that something SURVIVED a refused operation: the tombstone is still there, the
+/// record is still Pending, the audio is unchanged. That is only evidence of a refusal if the operation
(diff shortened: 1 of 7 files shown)


thefrederiksen (commented):
REJECT

1. `GatewayDictationEndpoint.Map` accepts optional nullable `HostedTenantBoundary? tenantBoundary = null`, and `ResolveTenant` returns `TenantId.Local` whenever that argument is omitted. It never asks `GatewayHostedMode.IsHosted` before taking that fallback. Any hosted call site, test, or future rewire that forgets the optional argument sends all five legs to the shared Local root, reopening transcript reads, chunk overwrite, ack, abandon, and completion-cache joining. Gate the fallback directly on hosted mode so hosted plus missing/unresolved boundary denies, make the production boundary non-optional as a second defence, and test hosted plus omitted boundary across all five legs while explicit hosted-off still uses Local.

2. The nine production-line reverts were run only against 7 HTTP tests and 14 store tests. Each real revert needs a successful build and full 2900+ Gateway run with counted intended reds and controls green, followed by a restored full run.

The physical per-tenant directory, canonical lowercase minted-GUID validation before path combination, refusal of `..`, `../..`, uppercase and SYSTEM aliases, bidirectional same-id audio/transcript canaries, record tenant stamp, tenant-keyed static caches, all-five-leg coverage, and deliberately held Local session lookup/order are otherwise sound.

thefrederiksen:
## Revert attribution: what the combined run does and does not prove

This change has **thirteen** production lines, not nine. The count kept drifting because two guards
share the same text (`if (IsPartitionContainer(dir)) continue;` appears in both `SweepAbandoned` and
the pending-record projection) and were being counted as one.

**Attribution comes from single-line filtered probes, not from the combined revert.** Each line was
reverted alone, built (build line checked every time), and run filtered. The combined all-lines
revert answers two different questions - did this break something else in the suite, and was any of
it already covered elsewhere - and it answers those regardless of how many lines were pulled.

**The combined run does NOT separately prove L1, L2, L4, L5, L8 and L10.** Reverting
`VoiceUploadStore.PartitionRootFor` (L7) collapses every tenant onto one root, which makes the
per-leg `ForTenant` calls irrelevant and reddens those six lines' tests by itself. Three of L7's
failures are `NullReferenceException` or `IOException` - they never reach the distinguishing
assertion at all. Those six lines are attributed by their filtered single-line probes and by nothing
else. Stating this plainly rather than letting a reader infer that one combined run proved thirteen.

| Line | Filtered single-line revert result |
|---|---|
| L1 register leg `ForTenant` | register leaks the seeded transcript |
| L2 chunk leg `ForTenant` | local staged audio overwritten |
| L3 complete leg `ForTenant` | complete leaks the seeded transcript |
| L4 ack leg `ForTenant` | a foreign ack retires the record |
| L5 abandon leg `ForTenant` | a foreign abandon resolves the record |
| L6 `CacheKey` at `GetOrAdd` | both tenants observe one identical cache key |
| L7 `PartitionRootFor` | 17 red; the subsumer described above |
| L8 `IsMintedAccountTenant` | the uppercase casing alias is no longer refused |
| L8b path containment | **no canary** - unreachable by construction |
| L9 record stamp / `BelongsHere` | a foreign record is returned; the session lock crosses |
| L10 `SweepAbandoned` skip | the sweep eats another tenant's staging |
| L10b pending-projection skip | **no canary** - redundant with `BelongsHere` |
| L11 `ResolveTenant` hosted gate | all five omitted-boundary legs, Forbidden to OK |

**Two lines have no canary, deliberately, and neither got a test invented for it.** L8b cannot be
reached because `IsMintedAccountTenant` already refuses every value that could escape the root; L10b
is redundant because the container holds no record file and `BelongsHere` on the same line already
refuses a foreign record. The only test either could have is one that cannot fail, which reads as
coverage while proving nothing. Both are documented at the guard itself, including the condition
under which L8b becomes reachable and must then be given a canary - the moment someone loosens the
minted-tenant rule.

The full-suite combined revert run and its restored companion are still owed and are queued for the
test lane.

thefrederiksen (commented):
Re-review of corrected head b06eac69: the original fail-open defect is fixed. `Map` now requires the boundary, `ResolveTenant` gates directly on `GatewayHostedMode.IsHosted`, and hosted null/non-hosted/unresolved boundaries return the exact 403 refusal before any store or completion-cache access. All five legs have omission canaries that assert the one-property refusal body and preservation of Local state; the canonical physical tenant partition, record stamp, tenant-keyed caches, and traversal/uppercase/system refusals remain sound. Current verdict remains REJECT for evidence only: the PR body still documents focused revert runs (7/7 HTTP and 14/14 store), not the required successful-build plus full ~2,900-test Gateway production-line revert with every red named and the restored full suite tied to this exact pushed head. Add that two-leg proof; no further static code defect is presently identified.

thefrederiksen (commented):
Head 35c510bc adds an honest thirteen-line attribution map and explicitly declares two defenses without canaries. I accept those exceptions: L8b containment is unreachable while `IsMintedAccountTenant` admits only canonical lowercase GUIDs, and the comment names predicate loosening as the point where it becomes live and must gain a canary; L10b is only a clarity/cost skip because a partition container has no record and the actual `BelongsHere` security check is separately canaried. The map also correctly refuses to infer individual attribution for six lines subsumed by the partition-root collapse or masked by exceptions. This does not replace the remaining full-suite proof: verdict stays REJECT until a clean serialized production-line revert yields a complete named-red inventory and the restored full suite matches this exact pushed head.

thefrederiksen:
## Mutation map for validation — thefrederiksen/devthrottle#1900

**Rule applied:** mutate every line whose *individual* wrongness could break isolation while everything
else stays correct. Exempt only when bypass is **structurally impossible** — never because a test also
reddens under another primitive's global revert.

**I took the wrapper option. It was clean.** The five per-leg scopings and the two cache call-site families
are now structurally incapable of bypass, so they collapse legitimately rather than by exemption.

### What makes bypass impossible (asserted with the mechanism, not by claim)

- `GatewayDictationEndpoint.Map` **no longer takes a `VoiceUploadStore`.** It takes `DictationTenantGate`.
  `grep` for an unscoped store identifier in that file returns **zero**. A sixth leg has nothing to call
  except `TryOpen`, which cannot return an unscoped store, and the deny is folded into the same call — a
  caller that skips the check has no store at all.
- The two static caches are now `TenantKeyedCache<T>`. The tenant is a **required parameter** of every
  operation and the key is composed inside the type. All four `CacheKey(...)` compositions are inside the
  cache; **none at a call site.**

### The eight primitives

| # | Primitive | Mutate it to | While these stay correct | Guarantee that breaks |
|---|---|---|---|---|
| P1 | `DictationTenantGate.TryOpen` | `_uploads` instead of `_uploads.ForTenant(resolved)` | ResolveTenant, PartitionRootFor | all five legs operate on the shared root |
| P2 | `ResolveTenant` hosted gate | restore `boundary is null ? Local : …` | gate, partition | hosted Gateway wired without a boundary serves the self-host root |
| P3 | `PartitionRootFor` | return `_partitionBase` for all | gate, ResolveTenant | every tenant collapses onto one directory |
| P4 | `IsMintedAccountTenant` | loosen to a character allow-list | partition computation | two identities differing only in case name one folder |
| P5 | `WriteRecordMarker` tenant stamp | drop `with { Tenant = … }` | `BelongsHere` | records unattributed, accepted in the local partition |
| P6 | `BelongsHere` | `=> true` | the stamp | a record physically in the wrong partition is served |
| P7 | `TenantKeyedCache` key (`CacheKey`) | drop the tenant | everything on disk | two accounts share one in-memory completion entry and transcript |
| P8 | `SweepAbandoned` container skip | delete the skip | everything else | one local age-sweep deletes every other tenant's staged audio |

**P5 and P6 are two, not one.** The stamp can be dropped while the check stays right, and the check can be
neutered while the stamp stays right. Each leaks alone. I previously reported these as a single line; that
was wrong under this rule.

### Exempt

- **E1** five per-leg scopings → absorbed into P1, structurally (mechanism above).
- **E2** two cache call-site families → absorbed into P7, structurally (mechanism above).
- **E3** pending-projection container skip — settled; redundant, `BelongsHere` on the same line is the real
  boundary and it has canaries.
- **E4** `PartitionRootFor` containment belt — settled; unreachable while `IsMintedAccountTenant` is strict.

### Count

**8 mutation runs + 1 restored run = 9 full Gateway runs.** Without the refactor it would have been 13.

### Two things I will not overstate

1. **The refactor is new code that has never been executed.** `DictationTenantGate` and `TenantKeyedCache`
   build clean, and the structural claims are verified by `grep`, but no test has run against them. The map
   is contingent on the refactor passing its first run.
2. **Crash-hardening is done but unverified.** Five reds previously arrived as crashes (four
   `NullReferenceException`, one file-in-use `IOException`). Bare `!` dereferences are replaced with an
   asserting `MustStillExist`, and the two `File.Copy` setups now assert the partitions differ *before*
   copying, so a collapse reddens by assertion instead of crashing during setup. Not yet observed.

thefrederiksen:
## Two corrections to the mutation map above — both narrowing my own claims

Head is now `50b1c7a9`. Both corrections are recorded in the code, not just here.

### 1. P5 (record tenant stamp) does not demonstrate disclosure

My map's consequence column for P5 was wrong. P5 and P6 remain **two** primitives — each is
independently bypassable — but their failure modes differ, and I had blurred them:

- **P6 `BelongsHere` is the disclosure guard.** Neuter it and a record physically present in the wrong
  partition is handed to the wrong account.
- **P5 the stamp is not.** With the primary partition intact, dropping the stamp makes account records
  **unattributed and therefore refused** — a correctness and availability failure, not a disclosure one.

P5 still earns its own run, because the accepted design requires the persisted ownership marker. But I
will report its **actual** assertion and **actual** failure, and describe the consequence as what it is.

The general form, which is the reason this matters: *a mutation showing a line is load-bearing does not
show it is load-bearing for the security property.* Two claims; only the observed one is mine to make.

### 2. What actually carries the structural impossibility

My claim was that a grep for an unscoped store identifier in the endpoint file returns zero. **That is
true and it is all I proved.** It is not the same as "an unscoped store cannot be passed."

Two private helpers on the completion path — `RunCompleteCoreAsync` and `MapNonOkTranscription` — still
take a bare `VoiceUploadStore` in their signatures and would accept an unscoped one if such a store were
ever in reach. **None can be today: `DictationTenantGate` privately owns the only raw store and hands out
nothing but partitioned views.** So the property holds by that type's **encapsulation**, not by the type
system at those two signatures.

The exemption from per-leg proof was granted on structural impossibility, and it *is* structurally
impossible today — but a reviewer who reads "cannot be written" and then finds a permissive signature
should doubt the rest of the map, so the precise version is now recorded at the gate itself.

### Why I am not doing the stronger form now

Giving `TryOpen` a distinct scoped-handle type, constructible only by the gate, and taking that in both
helpers would move the guarantee into the type system — where it survives someone later making a raw
store reachable for an unrelated reason. It is the right end state and is recorded as a follow-up.

**I am declining it in this change, and not because of the effort.** This structure has been
independently inspected and is what the eight mutation runs will execute against. Changing it now would
mean the thing proven is not the thing validated, and re-validation costs more than the hardening buys.
It goes in next, when it is not racing a proof.

Count unchanged: **8 mutation runs plus 1 exact-head restored run.**

thefrederiksen:
## Pre-run correction: all five deny legs had an unestablished precondition

Head `5179d473`. **Test-only** — no production file touched, so the inspected gate and keyed-cache
structure is unchanged.

### The defect

Each of the five legs asserts that something **survived** a refused operation: the tombstone is still
there, the record is still `Pending`, the audio is unchanged. That is evidence of a refusal **only if the
operation would otherwise have destroyed it** — and none of the five established that.

The starting state a survival claim depends on is not "a record exists". It is **"a record exists *that
this operation would destroy*"**, and the second half was assumed in all five.

As written, a no-op ack leg, a no-op abandon leg, or a chunk leg that stored nothing at all would have
left everything standing *exactly as a refusal does*, and every assertion would still have passed. Five
tests that could not fail, wearing the shape of strong ones.

**Why it hid:** a refusal test asserts that *nothing happened*, so "nothing happened" can never by itself
separate **the refusal** from **the operation being inert**. The evidence is identical under both.

### The fix

Each leg now closes with a **destructibility control** — the same operation, permitted, really does retire
the tombstone / resolve the record / overwrite the bytes. Ordered last, because it consumes the fixture.
Register and complete additionally assert the terminal state that arms the short-circuit before attempting
it, rather than trusting the seeding.

Paired with the deny fingerprint already present (exact 403, exact refusal message, response property set
as an allow-list — which proves the request reached the guard rather than a missing route), each leg now
states the whole sentence: **the thing was destructible, something tried to destroy it, it was refused,
and it survived.** Nothing shorter than that sentence is a refusal proof.

### Scope: local, not systemic — checked rather than assumed

I read the pre-existing `DictationTenantIsolationTests` to find out whether this was my mistake or the
file's convention. **They do not have the defect.** `An_ack_never_retires_another_accounts_record` ends
with account A retiring the very upload id account B failed to retire, and the abandon test opens with a
control abandon that succeeds — both establish destructibility already.

So this was a local defect in the tests I added, not a systemic one in the tenant test suite. That
distinction decides whether the fix is five tests or a sweep of the file, and it is only knowable by
reading the other tests rather than assuming the author had been careful.

### General form, recorded for reuse

> **"X did not happen" is worthless unless you have shown X *can* happen in that fixture.**

It is the positive-control rule in its destructive costume.

### What this would have cost

Eight mutation runs against five canaries that could not fail would have produced eight green-looking
runs certifying nothing — and **reconciling perfectly**. Arithmetic would not have caught it.
Assertion-versus-crash would not have caught it. Only asking *what would make this test fail* caught it.

Count unchanged: **8 mutation runs plus 1 exact-head restored run**, pinning `5179d473`.

thefrederiksen:
# Full-suite mutation proof — head `5179d473`

Eight mutations, **one primitive per run**, full Gateway suite each, then a restore run at exact head.
Every run's build line was checked; a failed build aborts the run, so no result comes from a stale binary.

## Artifact

- `local == remote == 5179d473` (confirmed by fetch, not assumed).
- Base `e1b8825c`. **Note: `origin/main` has since moved to `712ef1ff`.** I deliberately did not rebase
  mid-proof — the structure reviewer 2 inspected is the structure the runs executed against, and changing
  it between inspection and evidence breaks the chain. Happy to rebase and re-run if required.
- Continuous integration **read, not inferred**: `Build & Test (.NET)` **pass** (17m6s),
  `Build & Test (web)` **pass** (1m2s), run `29712158339`, at this head.
- Attribution scan: branch commits clean, pull request title and body clean.

## Per-primitive matrix

| # | Primitive | Failed | Passed | Sum | Reds | Crash-reds |
|---|---|---|---|---|---|---|
| P1 | `DictationTenantGate.TryOpen` | 10 | 2965 | **2975** | 10 | **0** |
| P2 | `ResolveTenant` hosted gate | 5 | 2970 | **2975** | 5 | **0** |
| P3 | `PartitionRootFor` | 17 | 2958 | **2975** | 17 | **0** |
| P4 | `IsMintedAccountTenant` | 3 | 2972 | **2975** | 3 | **0** |
| P5 | record tenant stamp | 8 | 2967 | **2975** | 8 | **0** |
| P6 | `BelongsHere` | 1 | 2974 | **2975** | 1 | **0** |
| P7 | `TenantKeyedCache` key | 1 | 2974 | **2975** | 1 | **0** |
| P8 | `SweepAbandoned` container skip | 1 | 2974 | **2975** | 1 | **0** |
| — | **RESTORE at exact head** | **0** | **2975** | **2975** | — | — |

**Reconciliation computed, not eyeballed:** every mutation's `passed + failed = 2975`, which equals the
restored run's passed count. `Skipped = 10` and `Total = 2985` are constant across all nine runs.

**Baseline complete, not merely green.** The 10 skips are all environment-gated live-Postgres tests
(`PostgresProviderProofTests`, `GatewayDatabaseLivePostgresProofTests`, `GatewayHostBootSmokeTests`) — no
tenant test is skipped. The three tenant classes plus the thefrederiksen/devthrottle#1916 canary were confirmed **executed by name**
at exact head (26 tenant tests + `Enrolled_without_a_recorded_email_is_signed_in_with_the_identity_absent`,
all Passed). The default console logger prints only skipped and failed tests, so a green suite genuinely
cannot show you what ran — that check had to be made explicitly.

## What each red actually asserts

**P2** — all five deny legs fail on the refusal status itself: `Expected: Forbidden, Actual: OK`. The
hosted gate is the only thing standing between an unwired hosted host and the shared root.

**P1** — the deny legs redden through their **wired positive companion**, not their deny claim, and that
is correct: the unwired host still refuses (the deny happens in `ResolveTenant`, upstream of the gate),
while the *wired* host now escapes its partition and destroys the local fixture —
`Expected: "self-host-partition-secret-audio", Actual: "wired-host-audio"`, and the register/complete
bodies disclose `self-host-partition-secret-transcript`. Attributable to the gate's scoping precisely.

**P6 is the disclosure guard**, concretely:
`Assert.Null() Failure … Actual: DictationDeliveryRecord { Transcript = alpha-secret-transcript, Tenant = d1f91ea3-… }`
— account B's partition handing over account A's record, with A's transcript in it.

**P5 is not disclosure, exactly as predicted.** Every one of its eight reds is a record becoming
*unreadable*: `Assert.NotNull() Failure: Value is null`, `the record was absent`,
`A's own abandon must have written A's tombstone (the record was absent)`, and at register
`Not found: "alpha-account-secret-transcript"` — that is A's **own** positive control failing to read A's
**own** transcript. With the primary partition intact, dropping the stamp leaves account records
unattributed and therefore refused: **a correctness and availability failure, not a disclosure one.** P5
earns its run because the design requires the persisted ownership marker; it does not demonstrate
cross-tenant disclosure and is not reported as if it does.

**P7** — `Collection: ["ab18b6ce…", "ab18b6ce…"]`: two tenants observing one identical cache key.
**P8** — `Expected: 1, Actual: 2`: the sweep eating another tenant's staging.
**P4** — `Assert.Throws() Failure: No exception was thrown`: the uppercase casing alias no longer refused.

## Three claims I had been holding open — now closed

1. **The refactor had never executed.** It has now, across nine full runs. Closed.
2. **The crash-hardening was believed, not observed.** Nine runs, **zero crash-reds**. The four
   `NullReferenceException`s and the `IOException` that P3 previously produced are gone; P3 now yields 17
   assertion-reds. Observed. Closed.
3. **The five destructibility controls were unverified**, written by the same person who wrote five
   canaries that could not fail. **Closed, with a caveat I am stating rather than hiding:** no
   destructibility control was ever itself the failing assertion, so none is *directly* proven by a red of
   its own. What proves the shape is non-vacuous is that the same `MustStillExist` mechanism **did** fail,
   twice, when its premise was false — under P5 and P1, with the exact text `(the record was absent)`. A
   control of this shape demonstrably fails when the fixture is not in the required state. That is the
   honest strength of the claim: the mechanism is proven, my five specific instances are proven only
   insofar as they use it.

## Seven project counts, all at this head

| Project | Result |
|---|---|
| Gateway | 2975 passed, 0 failed, 10 skipped |
| Core | 3306 passed, 0 failed, 8 skipped |
| Engine | 62 passed, 0 failed |
| HostedAgent | 86 passed, 0 failed |
| Launcher | 27 passed, 0 failed |
| Avalonia | 247 passed, 0 failed |
| Terminal.Avalonia | 21 passed, 0 failed |

## Lock behaviour

The lock **never made me wait**: acquired immediately on all nine runs and named no holder. The box was
verified clear before I started, so this run exercised the uncontended path only — it is not evidence
about the waiting path.

## Scope — read this before treating the merge as permission to configure anything

This advances the voice chain on the **store** side only: dictation uploads, chunks, delivery records and
transcripts are now tenant-keyed, and the in-memory completion caches with them. **It does not enable
voice.** The transcription key stays out, and the routes and sweeps still pass a single hard-coded tenant —
the session locate inside the completion is deliberately still `TenantId.Local`, so a hosted session is
not found and dictation stays dead on hosted exactly as it is today.

So this is **a partition that holds, not an arming step.** What makes the callers actually use it is the
routes-and-sweeps conversion that follows. Partitioning first is the point: converting the routes first
would have armed this store rather than closed it.

thefrederiksen (commented):
The `5179d473` eight-arm proof is internally sound: every mutation is one primitive, all eight independently reconcile to the green and explicitly complete 2975-pass baseline, all reds are assertions, P5 is correctly limited to availability/correctness while P6 carries disclosure, the restored exact head is green, and CI plus the other six projects pass.

It is not merge-final evidence, however. The branch is still based on `e1b8825c`; current `origin/main` is `7a6d514b` and has since added #1888, #1915, #1901, pinned-root storage behavior, and new Gateway tests. Those changes overlap `GatewayHost` and storage semantics, and the old full-suite arms necessarily could not observe tests that were not in that assembly. Please rebase onto current main, inspect the range-diff and any conflict resolution, then rerun on the rebased final head: green+complete baseline first (abort otherwise), eight separate one-primitive full Gateway arms, exact-head restored green, exact-head CI, and the six non-Gateway projects. Preserve the precise P5 availability-only claim. Verdict remains **REJECT** pending that merge-final evidence.

thefrederiksen:
# Merge-final mutation proof, rerun on the rebased head `7e4208ef`

This is the evidence reviewer 2 asked for: rebase onto current main, inspect the rebase, then rerun
everything on the rebased final head. Nothing here is carried over from the `5179d473` run - every
number below was produced tonight against this head.

## 1. The rebase

- `origin/main` at the moment I started was **`f5bb926f`**, not the `7a6d514b` named in the verdict -
  main moved again (`#1929`) while the verdict was being written. I rebased onto `f5bb926f`.
- The branch turned out to already sit on `7a6d514b`, so the rebase carried one commit of main.
- **Verified with `git range-diff 7a6d514b..eb0142a1 f5bb926f..7e4208ef`, not with a rebuild.** All
  seven commits came back as `=` - byte-identical patches, nothing crept in:

  ```
  1:  9701f98f = 1:  44c4f2f7 fix(gateway): tenant-key the dictation upload and transcript store
  2:  130cd8fd = 2:  27c12107 fix(gateway): close the fail-open when the dictation boundary is omitted
  3:  3f46c9c4 = 3:  39fd10fc docs(gateway): record the two guards that have no canary, and why
  4:  e78b7cc4 = 4:  cd654468 refactor(gateway): make an unscoped dictation leg impossible to write
  5:  8621f6a7 = 5:  c699aa8f test(gateway): record why a partition test asserts before it copies
  6:  c91ba50c = 6:  3abada4a docs(gateway): state the tenant guarantees at the precision they were proved
  7:  eb0142a1 = 7:  7e4208ef test(gateway): establish the precondition every survival claim rested on
  ```

- **No conflict arose, so none was resolved.** Stating that plainly because "how did you resolve the
  conflict" has no answer here, and silence would read as an omission.
- `local == remote == 7e4208ef`, confirmed by fetching and comparing `git rev-parse HEAD` against
  `git ls-remote`, not assumed. Every run below is pinned to that head with a clean working tree.

The assembly did change under the rebase, exactly as reviewer 2 warned: the Gateway suite grew from
**2985 to 3035** tests (`#1888` stats deny, `#1915` voice-state partition, `#1901` entitlement gate,
`#1929`). So the old arms genuinely could not have observed 50 of the tests these arms ran.

## 2. How the mutations were applied and undone - read this before the matrix

**Mutated by editing the source file in place; restored with `git checkout -- <file>`.** No backup file
was ever moved or copied over a source file. That matters because a restore that carries an older
timestamp lets the build judge the assembly up to date and skip recompiling, and a **skipped** build is
not a **failed** build - it succeeds silently having done nothing, so the "abort on a failed build" rule
does not catch it.

Three independent pieces of evidence that every mutation and every restore reached the executed binary:

1. **Positive control, measured.** Before the P8 restore the built `CcDirector.Gateway.dll` was stamped
   `04:22:47`. `git checkout --` stamped the source `04:31:29`, and the rebuild advanced the assembly to
   `04:31:32`. The restore reaches the binary.
2. **Build timings.** Every arm's build recompiled in 3.5-6.2 seconds against 1.78 seconds for the one
   build where no source had changed.
3. **The decisive one: all eight arms produced pairwise-distinct red sets** - computed, 8 distinct of 8,
   not eyeballed. A stale binary can only ever reproduce the *previous* arm's reds. Eight different red
   sets, each matching the primitive it names, is positive evidence rather than an absence of evidence.

## 3. The baseline - and the aborted run I am not hiding

**The first baseline run ABORTED and is discarded.** The test host crashed and `dotnet test` still
printed a cheerful `Passed! - Failed: 0, Passed: 2698, Skipped: 5, Total: 2703` followed by
`Test Run Aborted.` That count is **partial and is not a result in either direction**. The crash:

```
Unhandled exception. System.InvalidOperationException: The collection has been marked as complete with regards to additions.
   at CcDirector.Core.Utilities.FileLogWriter.Enqueue(String line) ... FileLogWriter.cs:line 66
   at CcDirector.Gateway.Streaming.GatewayStreamRegistry.TimeoutUnclaimed(String streamId) ... line 76
```

A `GatewayStreamRegistry` open-timeout timer fires after `FileLogWriter.Stop()` has called
`CompleteAdding()`. Both files are untouched by this branch and untouched by the new main commits
(`GatewayStreamRegistry` last changed in `#1414`, `FileLogWriter` in `#277`), so this is a pre-existing
latent shutdown race, timing-dependent. **I re-ran once as a diagnostic to classify it as deterministic
or flaky - not as a remedy - and it did not recur in the ten runs that followed.**

The lasting consequence is that I hardened the runner: **every run below is checked for
`Test Run Aborted` / `Test host process crashed` and marked INVALID if present**, because otherwise a
truncated run's partial count is indistinguishable from a real one. All ten counted runs reported
`RUN COMPLETED (no abort, no host crash)` and totalled 3035.

**The baseline that counts is green AND explicitly complete:**

- `Failed: 0, Passed: 3025, Skipped: 10, Total: 3035`
- **Every skip accounted for.** All 10 are environment-gated live-Postgres tests -
  `PostgresProviderProofTests` (5), `GatewayDatabaseLivePostgresProofTests` (4),
  `GatewayHostBootSmokeTests` (1). **No tenant test is among them.**
- **The tenant classes executed BY NAME**, confirmed from the result file rather than inferred from a
  green total: `DictationTenantIsolationTests` 7, `DictationOmittedTenantBoundaryTests` 5,
  `VoiceUploadStoreTenantPartitionTests` 14 - **26 tenant tests, and the same 26 executed in every one
  of the ten runs**, arms included. A green total says nothing failed, not that anything ran.

## 4. The eight arms - one primitive per full-suite run

| # | Primitive | Failed | Passed | Sum | Reds | Crash-reds | Predicted |
|---|---|---|---|---|---|---|---|
| P1 | `DictationTenantGate.TryOpen` | 10 | 3015 | **3025** | 10 | **0** | 10 |
| P2 | `ResolveTenant` hosted gate | 5 | 3020 | **3025** | 5 | **0** | 5 |
| P3 | `PartitionRootFor` | 11 | 3014 | **3025** | 11 | **0** | 17 - see below |
| P4 | `IsMintedAccountTenant` | 3 | 3022 | **3025** | 3 | **0** | 3 |
| P5 | record tenant stamp | 8 | 3017 | **3025** | 8 | **0** | 8 |
| P6 | `BelongsHere` | 1 | 3024 | **3025** | 1 | **0** | 1 |
| P7 | `TenantKeyedCache` key | 1 | 3024 | **3025** | 1 | **0** | 1 |
| P8 | `SweepAbandoned` container skip | 1 | 3024 | **3025** | 1 | **0** | 1 |
| - | **RESTORE at exact head `7e4208ef`** | **0** | **3025** | **3025** | - | - | 0 |

**Reconciliation, computed and not eyeballed:** for every arm, `passed + failed = 3025`, which equals
the restored run's passed count exactly. `Skipped = 10` and `Total = 3035` are constant across all ten
runs. Predictions were **pre-registered before any arm was run**, including the prediction that the new
tests would move nothing (the primitives live in three files, and the only production reference to
`VoiceUploadStore` outside them is the wingman utterance path, which never touches records, `BelongsHere`
or `ForTenant`).

**Zero crash-reds across eight arms.** Every red was classified by HOW IT ARRIVED, from the failure
message and stack in the result file. A red that arrives as a runtime fault is laundered and proves
nothing; none did. Two reds arrive through `Assert.True(record is not null, claim)`, whose message
carries only the claim text and names no exception type - those are the test's own logic raising, and
the stack confirms the top frame is the assertion helper, not a dereference.

## 5. The P3 deviation - I predicted 17 and got 11, and the cause is mine

**The difference is my mutation, not the new tests, and I am not reaching for the more comfortable
explanation.** The previous proof mutated `PartitionRootFor` to "return `_partitionBase` for all",
which also bypasses the `IsMintedAccountTenant` validation guard earlier in the same method. That makes
two things wrong at once and pulls P4's reds into P3's arm. I mutated only the final line:

```diff
-        return combined;
+        return _partitionBase;
```

which leaves the validation intact and collapses only the *directory decision* - the one thing P3 names.
That is the stricter reading of one-primitive-per-run: P3 is now "the root is wrong while the legality
rule is right", and P4 remains "the legality rule is wrong while the root computation is right". Under
this form P3 and P4 are disjoint, and P4 still reddens its own 3 independently.

Old count 17, new count 11, reason: a deliberately narrower mutation. Noticed against the
pre-registration and diagnosed, not quietly updated.

One related note in the same spirit: **P4's loosened allow-list reddens 3 of the 7 non-minted theory
cases**, because `""`, `"   "`, `".."` and `"a/b"` are still refused by a letter-digit-dash allow-list.
That is correct for a narrow mutation and is why the count is 3 rather than 8.

## 6. What each red actually asserts

**P2** - all five omitted-boundary deny legs fail on the refusal itself:
`Assert.Equal() Failure ... Expected: Forbidden, Actual: OK`. The hosted gate is the only thing between
an unwired hosted host and the shared self-host root.

**P1** - the deny legs redden through their **wired positive companion**, not their deny claim, which is
correct: the unwired host still refuses upstream in `ResolveTenant`, while the *wired* host escapes its
partition - `Expected: "self-host-partition-secret-audio", Actual: "wired-host-audio"`, and the register
and complete bodies disclose `self-host-partition-secret-transcript`.

**P3** - the partition collapses: `The_same_upload_id_is_two_different_uploads_in_two_tenants` fails
`Assert.False ... Actual: True`; the sweep eats another tenant's staging (`Expected: 1, Actual: 2`); a
chunk that should be `NotFound` returns `OK`. `A_record_carrying_another_tenant_is_refused_even_inside_this_partition`
reddens on its **setup precondition** - `Assert.NotEqual` on the two partition paths - which is exactly
the crash-hardening working: the fixture asserts the partitions differ *before* copying, so a collapse
reddens by assertion instead of crashing during setup.

**P4** - `Assert.Throws() Failure: No exception was thrown`, three times: the uppercase casing alias,
`"system"`, and `"not-a-guid"` are no longer refused.

**P6 is the disclosure guard**, concretely - account B's partition handing over account A's record:
`Assert.Null() Failure: Value is not null ... DictationDeliveryRecord { State = Delivered, Transcript = alpha-secret-transcript, Tenant = 16195c4e-... }`

**P5 is NOT disclosure, and the precise claim is preserved.** Every one of its eight reds is a record
becoming *unreadable* or a positive control failing: `Assert.NotNull() Failure: Value is null`,
`A's own abandon must have written A's tombstone (the record was absent)`,
`the sweep must not have destroyed another tenant's pending record (the record was absent)`, and at
register `Not found: "alpha-account-secret-transcript"` - A's **own** control failing to read A's **own**
transcript. With the primary partition intact, dropping the stamp leaves account records unattributed
and therefore **refused**: an availability and correctness failure, **not a disclosure one**. P5 earns
its run because the design requires the persisted ownership marker; it does not demonstrate cross-tenant
disclosure and is not reported as if it does.

**P7** - `Assert.Contains() Failure ... Collection: ["7c569dda5e1e465e83ea98657e2a24a2", "7c569dda5e1e465e83ea98657e2a24a2"]`:
two tenants observing one identical cache key.

**P8** - `Assert.Equal() Failure: Expected: 1, Actual: 2`: one local age-sweep eating another tenant's staging.

## 7. Seven project counts, all at `7e4208ef`

| Project | Result |
|---|---|
| Gateway | 3025 passed, 0 failed, 10 skipped |
| Core | 3312 passed, 0 failed, 8 skipped |
| Engine | 62 passed, 0 failed |
| HostedAgent | 86 passed, 0 failed |
| Launcher | 27 passed, 0 failed |
| Avalonia | 247 passed, 0 failed |
| Terminal.Avalonia | 21 passed, 0 failed |

**Continuous integration read, not inferred**, at this exact head: run `29723544314`,
`Build & Test (.NET)` **success** (17 minutes 8 seconds), `Build & Test (web)` **success** (58 seconds).

## 8. Rules left FAIL-OUTSTANDING

Stated rather than claimed.

1. **The five destructibility controls are still not directly proven.** No control was itself the
   failing assertion in any arm. What *is* proven is that the mechanism they use fails when its premise
   is false - `MustStillExist` reddened with the exact text `(the record was absent)` under P1, P3 and
   P5. The mechanism is proven; my five specific instances are proven only insofar as they use it.
2. **The test lane's waiting path is untested.** The lock was acquired immediately on all eleven runs
   and named no holder. This exercised the uncontended path only.
3. **The abort race is detected, not fixed.** `FileLogWriter.Stop()` racing a `GatewayStreamRegistry`
   open-timeout timer can truncate any Gateway suite run and still print a `Passed!` line. My runner now
   flags it, and no counted run here was truncated, but the underlying race is pre-existing, is not this
   branch's to fix, and will do this to somebody else.
4. **Scope, unchanged and worth repeating:** this is a partition that holds, **not** an arming step. It
   does not enable voice. The routes and sweeps still pass a single hard-coded tenant, so a hosted
   session is not found and dictation stays dead on hosted exactly as it is today.

## 9. Wall time, measured

Full Gateway suite at this head: **7 minutes 11 seconds to 9 minutes 58 seconds**, mean about
**7 minutes 58 seconds**; about **9 minutes** end to end per arm including the build. The whole
sequence - one aborted baseline, one good baseline, eight arms, one restore - ran
**07:09:40 to 08:47:11, 1 hour 37 minutes 31 seconds**, plus about 7 minutes for the six non-Gateway
projects. Budget roughly **1 hour 45 minutes** for a full ten-run set, and carry an allowance for one
abort re-run.

thefrederiksen:
RE-REVIEW of proof artifact `5020434614` and head `7e4208ef`: the merge-final mutation artifact is accepted on its own terms. The PR is held only for final-head convergence described below.

The P3 deviation is correctly diagnosed. The earlier ?return `_partitionBase` for all? form bypassed the validation branch as well as the directory decision. Six otherwise-refused values reach `PartitionRootFor` after the outer validity check and were therefore folded into that old arm; the new mutation changes only the final `return combined` and leaves `IsMintedAccountTenant` intact. That makes P3 ?valid tenants collapse to one root? while P4 independently loosens admissibility. Eleven reds instead of seventeen is evidence of a narrower, one-primitive mutation, not missing coverage. P4?s three reds are also consistent with the declared letter/digit/dash loosening: uppercase GUID, `system`, and `not-a-guid` pass it, while traversal/blank cases remain refused.

The ten counted runs satisfy the proof protocol: green and explicitly complete 3025-pass baseline; all ten skips accounted for and none tenant-related; 26 named tenant tests present in every run; eight separate non-crash assertion-red arms; exact reconciliation on every arm; exact-head restored green; exact-head CI and six other projects. The discarded 2703-total run is correctly classified as aborted despite its misleading `Passed!` line. Completeness must now mean expected total plus absence of abort/host-crash markers, not merely the presence of a summary.

The 8/8 pairwise-distinct red sets are strong anti-stale evidence here, but not logically decisive by themselves: they discriminate stale carry-over because these arms have disjoint, non-empty signatures. I accept freshness from the combination of those signatures, matching primitive-specific reds, observed source/assembly timestamp advance, and recompilation timings?not from distinctness as a universal rule.

The declared fail-outstanding items do not block this unit:

- The five destructibility controls assert concrete allowed effects in the final tests. They were not each mutation-proven by making its operation inert, so the artifact correctly avoids claiming that stronger sensitivity result; that is not another independently bypassable production security primitive in this patch.
- The suite-lock waiting path belongs to the already-merged lane mechanism, not dictation partitioning.
- The abort race is pre-existing and did not occur in any counted run. It must be recorded/routed as a separate finding because it can invalidate future evidence, but it does not invalidate the runs whose totals and non-abort markers were checked.
- The ?partition, not arming? scope is precise and preserved.

Final-head hold: while this 97-minute proof was running, main advanced from `f5bb926f` to `f2028deb` via thefrederiksen/devthrottle#1928. I inspected that range: it touches Avalonia connection choice and setup-engine files/tests only, not Gateway production or the Gateway test assembly, so it is orthogonal and the evidence may carry under the patch-identical rebase rule. thefrederiksen/devthrottle#1937 is also required to merge before the queued units and changes proof/test infrastructure, not these seven PR files. Rebase thefrederiksen/devthrottle#1900 after the required main convergence, post a range-diff showing all seven patches remain `=`, and require current exact-head CI. If the rebase is conflict-free and patch-identical, no further ten-run mutation sequence is required. Any conflict, changed patch, or Gateway-overlapping main commit reopens the exact-head proof requirement.

thefrederiksen:
# Booked, not fixed: the upload store's public constructors bind the shared root and the local tenant

Found by a sibling unit working in a neighbouring file, deliberately not edited by it, and deliberately
**not** added to this branch. Recorded here so it lands with the routes-and-sweeps conversion instead of
reopening a proven unit. Verified against `VoiceUploadStore.cs` at `7e4208ef`, not taken on description.

## The finding

```csharp
public VoiceUploadStore() : this(CcStorage.VoiceTurnUploads()) { }

/// <summary>Test seam: stage under an explicit root instead of the shared storage dir.</summary>
public VoiceUploadStore(string root) : this(root, TenantId.Local, root) { }

private VoiceUploadStore(string root, TenantId tenant, string partitionBase)
```

**The zero-argument path is the unsafe one.** Writing `new VoiceUploadStore()` in a new endpoint file
tomorrow binds a shared cross-tenant staging root with no argument and nothing that looks wrong in
review. It is the inverse of every collapse this mission has rewarded: we have been making the wrong
state inexpressible, and here the wrong state is the **default**, reached by writing *less* code.

## One sharpening, because booking it as stated would under-fix it

The finding is real and it is **broader than "the zero-argument constructor"**. Both public constructors
bind `TenantId.Local`:

- `new VoiceUploadStore()` binds `CcStorage.VoiceTurnUploads()` - the **voice-turn** root.
- `new VoiceUploadStore(root)` binds **whatever root you name**, and it is also `TenantId.Local`.

That second one is the path to the **dictation** root - `new VoiceUploadStore(CcStorage.DictationUploads())`
is exactly how `GatewayHost.cs:457` reaches the audio this unit partitions. So a fix that only removes or
guards the zero-argument constructor would leave the one-argument public constructor as an unguarded door
onto the very staging this pull request tenant-keys. **Both constructors are the unit of the fix.**

## What this narrows in MY OWN claim on this pull request - the part worth the booking

I wrote, of `DictationTenantGate`, that the property holds because "this gate **privately owns the only
raw store** and hands out nothing but partitioned views." Scoped to the dictation endpoint file and its
two private helpers, that sentence is **true**. Read as a statement about the process, it is **false**:
any new file can mint a fresh raw store with one public constructor call, and the gate owns nothing about
that.

That is the same shape the sibling reported about its own deny, turned on my unit: **every sentence was
true and none was exhaustive.** My structural-impossibility claim is a claim about *one file*, and I
should not be read as having claimed more. The mutation proof is unaffected - P1 measures the gate's
scoping and reddens 10 tests, and that measurement stands - but "a leg cannot be written unscoped" means
*in that file*, not *in this process*.

## The distinction this unit does and does not earn

The tidy version of this - "a partition survives a new reader, a deny does not" - is **too strong, and
this finding is the counter-example.** Corrected, the honest form is narrower and less satisfying:

- A **deny** protects a **route family**.
- A **partition** protects an **access path**.
- **Neither protects a store that anyone can still instantiate.**

Closing the state requires closing **construction** as well as access. A new file that mints a raw store
with one public constructor call bypasses this partition exactly as it would bypass a deny.

This unit is still the stronger of the two moves, and the partition is the right shape - **but it earns
its claim only for callers who obtain their store through `ForTenant` or the gate.** The constructor gap
is precisely the hole: a new reader can construct *around* the partition rather than through it. Its
construction route is not yet closed to match. Saying otherwise would be the victory-lap version, and it
is not true.

## Why nothing is reachable today - checked, not assumed

The only production zero-argument site is `GatewayWingmanVoiceEndpoint.cs:277`
(`var uploads = uploadStore ?? new VoiceUploadStore();`), the wingman utterance path, which uses
`Register`/`Exists`/chunk/complete and never touches the record marker, `BelongsHere` or `ForTenant`.
That surface is separately tracked in **#1896**. Every other construction site is a test.

The gap is a future one, and the reason it is worth booking rather than shrugging at: **no test in this
repository would go red the day someone writes that new reader.** A group filter defends the family it is
attached to and cannot defend a family nobody has written yet.

## Where the fix belongs

With the routes-and-sweeps conversion under **#1884**, alongside **#1896**, not here. The shape that
matches the rest of the mission: make the tenant a **required** argument of construction, so the raw
shared root is reachable only through a private route and an unscoped store cannot be minted at all -
the same move that made `GatewayDictationEndpoint.Map` take the gate instead of the store. Doing it now
would change the structure the eight-arm proof executed against and cost the whole arm set for something
that is not reachable today.

Not fixed on this branch. Booked.

thefrederiksen:
DEPENDENCY HOLD at exact head `7e4208efc316da41fc95ff7dd1b7ccbefbd4447d`: the booked constructor closure in thefrederiksen/devthrottle#1945 does not yet establish the requiredness property on which the process-wide structural claim depends.

Its standing reflection canary accepts any public constructor containing a `TenantId` parameter. Mutation to `TenantId tenant = default` left all four required-tenant tests green while tenant omission compiled. It also surveys public instance constructors only, so a public factory could mint a store without entering the canary.

This does not invalidate thefrederiksen/devthrottle#1900's eight-arm partition evidence for callers that enter through the gate. It does suspend the broader construction closure: before that dependency can discharge “a store cannot be minted unscoped,” its canary must redden when the tenant gains a default and it must cover public factories (or narrow the claim and pay the corresponding bypass proof).

thefrederiksen:
Superseded by thefrederiksen/devthrottle#1945. thefrederiksen/devthrottle#1945 is a strict superset stacked on thefrederiksen/devthrottle#1900 (contains all seven thefrederiksen/devthrottle#1900 commits verbatim plus the required-tenant construction hardening). Rather than merge both, thefrederiksen/devthrottle#1945 will be reconstructed on current main - preserving this PR's tenant-partition mechanism (partitioned VoiceUploadStore paths/records, tenant-keyed dictation caches, DictationTenantGate, fail-closed tenant resolution) plus the newer main work (#1952 sweep safeguards, thefrederiksen/devthrottle#1973 voice routing, thefrederiksen/devthrottle#1903/#1907 hosted denies) - and merged as the single vehicle. Analysis: docs/reviews/production-readiness/mission/analyze-uploadstore-mtr10.md.

PR #1901, opened 2026-07-19
feat(gateway): gate hosted enrollment on a paid entitlement

thefrederiksen:
The paid-entitlement gate at hosted enrollment. Without it, enrolling against the hosted Gateway is **free**, and the billing side sells what anyone can take for nothing.

Implements the gateway half of the cross-team contract in `hosted-entitlement-gate-design.md`: the `gateway.entitlements` table already exists live and the scoped Gateway role already holds SELECT on it.

## Where the gate sits, and why exactly there

Between **subject-verification** and **tenant-mint**, per the design:

```
validate token (existing) -> subject verified (existing)
  -> NEW: entitlement check      -> not entitled: 402, no tenant, no key
  -> mint/lookup tenant (existing)
  -> bind device (existing)
```

It is before the mint deliberately: **minting a tenant is itself giving something away**. An unpaid account must leave no trace and hold no key - and with no device key there is no tunnel, no cockpit and no mobile, which is the literal meaning of "approval before you can use a hosted tenant".

## Three outcomes, not two - this is the whole correctness of the gate

| Outcome | Meaning | Answer |
|---|---|---|
| `Entitled` | read succeeded, valid entitlement | enroll |
| `NotEntitled` | read **succeeded**, no valid entitlement | **402**, mint nothing |
| `Unknown` | read **FAILED** - we do not know | **503 retry**, mint nothing, deny nothing |

A boolean would collapse *"I looked and there is nothing"* with *"I could not look"*, and those demand **opposite** answers. The failure modes are not symmetric:

- A false **402** locks out a customer who has paid.
- A false **mint** gives the product away for nothing - **and it is silent**, because a successful enrollment looks exactly like a correct one.

So: **the mint happens only on a confirmed entitlement.** Ignorance mints nothing and denies nothing; it says try again. The registry never throws - a caller forced to catch will eventually catch in the wrong place and turn ignorance into a denial.

## Policy, stated rather than implied

Entitled when the record says `active`, **or** `past_due` with the paid period not yet ended - the payment provider retries a failed card, and cutting a customer off mid-period refuses someone who has paid. A `past_due` record whose period **has** ended is not entitled, so the grace window is **finite**. Any other state, **including one this code does not recognise**, is not entitled: we do not guess in the paying direction.

**Live money only.** On production hosted, a record must be explicitly live-mode. A test-mode subscription costs nothing to create, so honouring one is a paywall bypass in the deny-**open** direction. **Null is refused exactly as false is** - a row written before that column existed, or by a webhook that forgot it, arrives null, and "we did not record whether this was real money" is not evidence that it was. This composes with the three outcomes rather than adding a fourth: the read succeeded and returned no *valid* entitlement, so it is absence, so it is the 402.

## Self-host

Unchanged, and structurally so: the hosted enrollment route is only mapped when hosted, so self-host never reaches this code at all. Passing no registry means no gate. That is the control test.

## The table is not ours to create

`EntitlementEntity` is mapped **excluded from migrations**. The payment side creates and writes it as the service role; this Gateway holds SELECT and nothing more. Verified rather than assumed: `dotnet ef migrations has-pending-model-changes` reports **no pending changes on both providers**, so there is no migration to add and the model stays in sync.

The tests create the table with raw SQL, which also **states the schema this code reads** - so a contract drift from the payment side shows up as a failing test rather than a runtime surprise.

## Proof

**Revert-prove**, run verbatim against the final file: delete the entitlement block from `Enroll`. Build succeeds and **7 of 10 redden** - both independent controls plus the state and grace-window cases. The 3 that stay green are exactly the right ones: the entitled case, the in-grace case, and the self-host control.

One note, because it nearly produced a false green: reverting with `if (false)` **broke the build** (unreachable code is an error here), and a test run after a failed build executes a **stale binary** and reports a pass. The revert that counts deletes the block outright.

**The two independent controls the design demands:**

- `ABSENCE_denies_with_402_and_mints_nothing` - table exists, no row. Knowledge, so it denies.
- `IGNORANCE_does_not_deny_and_does_not_mint` - the table does not exist at all, which is exactly what a lost SELECT grant or an un-migrated database looks like from this side. It is a **real** failed read, not a stub.

Every deny path asserts **both halves** of the refusal - the status **and** that no tenant was minted and no device key issued. A 402 that still minted a tenant would satisfy a status-only assertion while having given away the thing the gate protects.

The live-mode case is a `Theory` over **false and null**, with a control proving the same row *does* enroll when live money is not required - so the refusals are caused by the live-mode rule specifically and not by the row being rejected for something unrelated.

## One guard that fired during this work

`TenantScopeGuardTests` - which requires every mapped table to be tenant-scoped or explicitly allowlisted **with a reason** - caught the new entitlements table and forced a written justification. It is allowlisted alongside the tenant mapping table for the same reason: both are keyed by account subject and read **before** any tenant exists, so scoping them would be circular. That guard doing its job is why this table is documented rather than merely added.


thefrederiksen:
Review clearance for head 39776a8a: CLEAR. I verified the rebased branch against current main; range-diff shows both feature commits are patch-identical to the versions previously reviewed, and the branch diff remains limited to the ten entitlement-gate files. The gate runs after verified-subject extraction and before tenant/device mint; a successful read with no valid entitlement returns 402, a failed read returns 503, and both paths leave no tenant and no device key. Active and finite past_due policy holds, including the corrected strict now < CurrentPeriodEnd boundary and an exact-boundary refusal test. Hosted live-money enforcement requires Livemode == true, so false and null are denied; unrecognized states are denied. Hosted-only route wiring leaves self-host unaffected. The payment-owned table remains excluded from migrations; SQLite and PostgreSQL both report no pending model changes. On this rebased head, the focused entitlement and tenant-scope run passed 17/17. Merge remains gated on green CI for the rebased head.

PR #1903, opened 2026-07-19
fix(gateway): deny four cross-tenant content read families on the hosted gateway

thefrederiksen:
Four route families on the hosted Gateway served one account's CONTENT to any other
account's authenticated device key. Public signup on the backing account project is open,
so "any authenticated caller" means the public. This refuses all four on hosted, and leaves
self-host completely unchanged.

## The defect they share

Every one of them stands on a store with a **single on-disk root and no tenant anywhere** -
not in the path, not in the file name, not in the record - addressed by something that is
**not a tenant boundary**: a positional index, a turn id, or an identifier the caller
supplies in a header.

A caller-supplied identifier is not authorization. Secrecy of an identifier is not
authorization either, and upload ids travel through client logs, retries and
store-and-forward queues.

## What is denied, and why

| Family | Routes | What it hands over | Tracked as |
|---|---|---|---|
| Transcription analysis | `GET /transcription/turns`, `/stats`, `/terms`, `/words` | Up to 2000 records including full `rawText` and `cleanedText` from one shared daily log. **Needs no identifier at all** - one request returns everybody's speech. The other three aggregate the same unpartitioned records. | thefrederiksen/devthrottle_internal#896 |
| Wingman instructions | `GET`/`PUT` `/gateway/wingman/instructions`, `GET /versions`, `POST /revert`, `GET /default`, `GET /update`, `POST /switch-to-default`, `GET /records`, `POST /test` | Training records holding up to 20,000 characters of raw session **terminal output** plus the agent reply, addressed by a positional `<filename>#<lineindex>` id a caller can simply walk. Plus the single-owner wingman prompt every account's wingman speaks through. `/test` also spends up to 5 real model calls per request on shared infrastructure. | thefrederiksen/devthrottle_internal#876 (read side) |
| Wingman utterance upload | `POST /wingman/utterance/upload`, `PUT /{uploadId}/chunk/{index}`, `POST /{uploadId}/complete` | Staged audio and the assembled transcript, keyed solely by a caller-supplied `Idempotency-Key` under a shared root. `complete` returns the transcript; `chunk` lets a caller overwrite another account's recording. | thefrederiksen/devthrottle_internal#895 |
| Dictation upload | `POST /dictation/upload`, `PUT /{uploadId}/chunk/{index}`, `POST /{uploadId}/complete`, `POST /{uploadId}/ack`, `POST /{uploadId}/abandon` | Same `VoiceUploadStore` shape. After account A completes upload id X, account B posts `/dictation/upload` with `Idempotency-Key: X` and the terminal-register short-circuit hands B **A's transcript**. That leg looks no session up, so the fact that `/complete`'s session lookup already fails on hosted does not contain it. `ack` and `abandon` destroy another account's in-flight recording. | thefrederiksen/devthrottle_internal#889 |

The dictation family was **found while doing the other three**: a different route family standing
on the same store shape with the identical defect, live on hosted today. Denied here rather than
left for later.

## The shape of the deny

- **One endpoint-group filter per family**, not a guard repeated in every handler. The filter runs
  before every route in the group **including routes that do not exist yet**, so the refusal cannot
  rot as the group grows.
- **The routes are mapped where the unfiltered builder is not in scope.** Each family's production
  routes live in a private method that receives ONLY the guarded `RouteGroupBuilder`, so mapping a
  route around the filter does not compile. Copied from the key-vault deny, thefrederiksen/devthrottle#1904.
- **Gated on `GatewayHostedMode.IsHosted`** - the independent deployment signal - never on an
  optional boundary or tenant argument, which **fails open** the moment a caller omits it.
- **Deny, not partition.** The tenant is missing from the DATA, not from the query.
- **Refuse, not serve empty.** An empty payload is a false statement; an absent one is merely absent.
- **The body is exactly one `error` property**, asserted as an ALLOW-LIST by enumerating the property
  set, never a deny-list of known keys.
- **404, not 403.** On hosted these surfaces do not exist as a concept.

## Self-host is explicitly unaffected

`HostedContentReadSelfHostControlTests` boots the **same** `GatewayHost` with hosted mode off and
drives the **same** routes. Self-host is single-tenant, these features work there and the owner uses
them. Three untouched pre-existing suites are controls too, with real payloads:
`VoiceUploadLimitsTests`, `DictationSessionLockTests`, `DurableDictationDedupeTests`.

## Why the self-host controls were rebuilt

The first version of those controls asserted **status 200 plus the absence of the refusal strings**.
Both are satisfied by a masking route, so they would have passed on routes that had been deleted:

- `CockpitReactApp` maps `MapFallback("{*path}")` (CockpitReactApp.cs:125), matching **any**
  unclaimed path and **any** verb, answering 404 in a Debug test host but **200 with the HTML shell
  on a release host**. Its own not-found text says "release build only".
- `SessionWsProxyEndpoints` maps `app.Map("/sessions/{sid}/{**rest}")`
  (SessionWsProxyEndpoints.cs:153) - least-specific, every verb - answering **503 as
  `application/json`**, so a content-type check does not catch it. **None of the four families here
  lives under `/sessions/{sid}/`**, so this one cannot mask these routes; it is recorded because it
  is why "check the content type" is not a general defence on this Gateway.
- **No 405 is ever raised on these paths**, so a wrong verb does not announce itself. The old
  control drove `POST` at the dictation chunk route, which production maps `MapPut` only
  (GatewayDictationEndpoint.cs:176) - the request never reached the handler and nothing noticed.

All three facts were read on this branch rather than assumed.

Every read now asserts **seeded or route-specific JSON that only that handler could produce**, and
every write asserts **its own store receipt** - a state change read back, never a status:

| Route | Receipt |
|---|---|
| `GET /transcription/turns` | one telemetry record planted via the production writer, its `rawText` read back |
| `GET /transcription/stats` | `totalTurns == 1` exactly, in an isolated otherwise-empty log |
| `GET /transcription/terms` | the planted find/replace pair |
| `GET /transcription/words` | the planted word counted |
| `GET /gateway/wingman/instructions/records` | a training record planted on disk, its positional id and reply preview read back |
| `GET`/`PUT` `/gateway/wingman/instructions`, `GET /versions`, `GET /update` | a version saved through the route and round-tripped; the customized flag read back through a **different** route than the one that set it |
| `GET /gateway/wingman/instructions/default` | exact top-level property set `{version, hash, content}` plus non-empty content |
| `PUT /wingman/utterance/{id}/chunk/{i}` | the uploaded bytes located on disk **by content** |
| `PUT /dictation/{id}/chunk/{i}` | same, and the verb is `PUT` |
| `POST /dictation/{id}/abandon` | the tombstone read back at re-register |
| `POST /dictation/{id}/ack` | retires exactly once: first `retired=true`, second `retired=false` |

Ordering matters as much as the assertion: **the fingerprint is the first thing allowed to fail**,
and status is only ever asserted after it. A status check first would make a renamed route redden on
"404 != 200", proving a route changed rather than proving the canary detects it.

Every failure path is also an **assertion, never an exception**. `JsonDocument.Parse` threw on the
HTML shell and `GetProperty` threw when another handler answered; both now assert, naming the route
and showing what came back. A red that arrives as a crash does not prove the row it was aimed at.

## Why a run with no summary line is not a result

The mutation driver originally scored any run with no failures as a passing canary check. A Gateway
test run refused by the suite lock - or killed - produces **no summary line at all**, and the driver
read that as a test-quality verdict. This is the artifact, printed verbatim by the old driver:

```
=== MUTATION: GET /transcription/turns [rename] ===

  *** NO RED - THE CANARY CANNOT FAIL FOR THIS ROUTE ***
```

A blank summary line, then a conclusion about test quality invented from a **scheduling** event -
precisely inverted, and in the exact category this work exists to find. The rule now: **a run that
produced no real `Passed!`/`Failed!` summary line is UNPROVEN and is re-run individually - never
scored as "no red", in either direction.**

The driver was rebuilt around that: it retries a non-summary run, records it as unproven if it never
completes, restores the mutation from an `atexit` hook and signal handlers rather than only
`finally`, **refuses to start on a dirty production tree**, aborts a row whose anchor is not unique,
treats a failed build under mutation as unproven rather than trusting stale binaries, and checks each
run's counts reconcile against baseline.

The start-check is the load-bearing part. `finally` does not run through a kill - a killed sweep left
a production endpoint **renamed** in the working tree, which would have presented as a route
mysteriously missing rather than as anything obviously wrong. Handlers are a better convention; the
dirty-tree check is the backstop that does not depend on having anticipated the escape route.

## What the second review demanded, and what changed

Two blockers, and the first one is about the SHAPE OF THE CODE rather than the proof - so it is
answered first, because it changes what the proof has to cover.

### 1. Every route was still individually bypassable

All four families kept the unfiltered builder in scope at the mapping site alongside the guarded
group. `TranscriptionAnalysisEndpoint`, `WingmanInstructionsEndpoint` and `GatewayDictationEndpoint`
had both `outer` and the filtered `app`; `GatewayWingmanVoiceEndpoint` had the unfiltered `app` beside
the filtered `utterance`. Changing one mapping receiver compiles, passes every existing test, and
opens exactly that route on hosted while its siblings stay correctly denied. Four, nine, three and
five independently bypassable primitives.

**And the twenty-one rename canaries did not test that.** Renaming or re-verbing a route proves the
route is reachable through the filter. The bypass substitutes the BUILDER, not the route. Those are
different mutations against different axes, and a lot of work on the wrong axis leaves the question
exactly as open as it was.

Every family's production routes now live in a private method that receives ONLY its filtered
`RouteGroupBuilder`. Inside it there is nothing to map onto except the guarded group, so mapping
around the filter DOES NOT COMPILE. The bypass is inexpressible rather than unlikely.

This copies **the key-vault deny, pull request thefrederiksen/devthrottle#1904**, which is the reviewed instance of the pattern
and the version to compare this against.

**THE LIMIT OF THE PROPERTY, STATED SO IT IS NOT OVER-CLAIMED.** The compile-time guarantee holds
WITHIN ONE MAPPING SITE. It does not reach across sites. This unit has FOUR families, each creating
its own group, attaching its own filter and mapping from its own site - so any one site can be wrong
while the other three stay correct, and no compiler notices. The split removes the per-ROUTE bypass
inside a family; it does NOT merge four families into one primitive, and it is not an argument for
fewer arms. That is precisely why the registration below is four filter removals and four gate
inversions rather than one of each.

`GatewayWingmanVoiceEndpoint` is the sharpest case and worth reading on its own: the unfiltered
builder is LEGITIMATELY in that file, because the voice-turn, text-to-speech, transcribe, explain, ask
and menu routes are not part of the denied family and must keep serving on hosted. That is exactly
what made the bypass cheap there. Only the three utterance legs moved.

### 2. The proof omitted the gate direction

Each family's deny is a CONJUNCTION of two independent facts: the filter must be ATTACHED, and its
`GatewayHostedMode.IsHosted` condition must FACE THE RIGHT WAY. A filter-removal arm leaves the
condition untouched, so it cannot detect a gate written backwards - and a backwards gate serves every
denied route to every account on hosted while breaking the owner's own features on self-host. Nothing
in the previous twenty-five-run registration would have gone red for it.

Four gate-inversion arms are now registered beside the four filter removals. Each inversion predicts
BOTH the hosted rows and the self-host rows, because a backwards gate breaks both directions at once,
and each predicts that the outside-the-group scoping control stays GREEN - a gate inversion cannot
reach a route mapped outside the group, so if that control reddens the arm hit something wider than
the gate.

## The proof, re-registered against the new structure

**Ten full-suite runs: a clean baseline, eight independent arms, an exact restored green.** No arm is
ever combined with another - a combined arm mis-attributes rows.

| Arm | What it breaks | Must redden |
|---|---|---|
| 4 x filter removal | the filter is not attached at all | that family's hosted refusal rows, the future-route probe, the no-tenant row |
| 4 x gate inversion | the hosted condition faces backwards | the same hosted rows AND that family's self-host receipts |

The anchor preflight passes: eight arms, each anchor appearing exactly once, zero bad anchors.

### How each run is believed, and what refuses it

**The baseline must be GREEN AND COMPLETE.** Green alone is not a reference: a truncated run is green
by construction, because the tests that would have failed never ran, and every arm measured against
that too-small number would then look like it reconciles. The driver requires the baseline's own
`passed + failed + skipped == total` before anything is compared to it.

**Reconciliation happens BEFORE scoring, not after.** For each arm, `passed + failed + skipped` must
equal the baseline TOTAL. A SHORT TOTAL IS TRUNCATION, NOT A MUTATION THAT REDDENED FEWER TESTS THAN
PREDICTED - the two are indistinguishable from the red list alone, which is exactly why the count has
to be believed before the reds are read. A short total now makes the arm UNPROVEN and skips scoring
entirely, so a truncated run can no longer report "expected but stayed green" about tests that never
executed.

Three defects were fixed to get there, and each could have laundered a truncated run into a clean
result: the sum left SKIPPED out (it balanced only because the skip count happened not to move); it
compared against baseline PASSED rather than baseline TOTAL (but tests move from passed to failed
under an arm - that is the point of an arm - so TOTAL is the quantity that must hold); and it scored
first and reconciled afterwards.

**The run refuses to start unless the work under test is COMMITTED.** The clean-tree check covers the
whole `src/` tree, not just the Gateway project: the test project is where the assertions live, and a
lost test edit falsifies an arm exactly as thoroughly as a lost production edit. The head under test
is stamped into the log so every number is attributed to an exact commit.

**The pushed-head enforcement is deliberately NOT here** - it belongs to the repository-wide guard
unit (#1937), which has taken it into its own mechanism. Two implementations of one guarantee drift
until neither can be shown to be the one doing the work.

**And the reason it is worth having is not the one an earlier version of this body gave**, so the
corrected version is recorded rather than the flattering one. That version claimed a pushed-head check
CLOSES the stash gap. It does not. Nothing can know what a head was SUPPOSED to contain - a commit
missing a repair is indistinguishable, from inside, from a commit that never needed one. What it buys
is that the claim becomes CHECKABLE BY SOMEONE ELSE: the commit every number is attributed to can be
fetched and read, so a missing repair is visible in the record instead of invisible on one machine.

**That distinction matters because this driver has both kinds and they are not the same thing.** Some
controls PREVENT a failure: the truncation check is one - a short total means the arm cannot be
scored, full stop. Others cannot prevent anything and only EXPOSE: the head stamp is one - it stops
nothing, it moves a potential failure somewhere a second person can see it. Conflating them is how a
log full of reassuring lines comes to feel like a guarantee.

### Labelling every control, and what that audit found

Classifying two controls led to classifying all of them, and the exercise was not bookkeeping - it
found two defects in the load-bearing scoring path, both of the same shape: **a control I described
as PREVENTING was implemented as merely EXPOSING.**

**1. A crash-red was credited to the prediction.** The rule at the top of the driver says "a red only
counts if it is an assertion" - a red arriving as a `JsonException` or a `KeyNotFoundException` means
the request went somewhere unexpected and something threw upstream of the assertion, which is
laundering. The implementation added *every* red to the credited set and merely appended the crash to
a list. So a crash-red satisfied the prediction, kept the row out of the not-reddened list, and the
arm read as PROVEN. That is precisely the laundering the rule exists to stop, sitting inside the
mechanism that enforces it. Now only assertion-reds are credited, so a crash leaves the row uncredited
and the arm unproven.

**2. The sweep returned zero no matter what it found.** Every check appends to a problems list - crash
reds, rows that stayed green, truncated arms, non-unique anchors, builds that failed under mutation -
and the final line printed that list and then `return 0`. The whole apparatus was an exposing control
wearing the costume of a preventing one: anything reading the exit code would be told the sweep passed
while the log said otherwise, and the log is exactly what nobody re-reads once the exit code is green.
**A finding that does not change the verdict is not a finding, it is a comment.** The verdict now
depends on the findings, and the restored re-check is held to the same green-AND-complete bar as the
baseline.

| Control | Kind |
|---|---|
| Clean `src/` tree before starting | PREVENTS - nothing runs |
| Baseline green AND complete | PREVENTS - aborts |
| Anchor not unique / build fails under an arm | PREVENTS - arm not scored |
| No summary line (retry, then unproven) | PREVENTS - a non-run is never scored |
| Arm total short of baseline total | PREVENTS - truncation, not scored |
| Crash-red not credited to a prediction | PREVENTS - **was exposing until this audit** |
| Restore-on-exit via `atexit` and signal handlers | PREVENTS - no mutation survives the process |
| Restored re-check green AND complete | PREVENTS - **was exposing until this audit** |
| Exit code reflects the problems list | PREVENTS - **was exposing until this audit** |
| Head stamped into the log | EXPOSES - stops nothing; makes numbers attributable |
| Unexpected-red reporting | EXPOSES the anomaly, and now also fails the verdict |

Before this audit I would have said the sweep had nine preventing controls. It had six, and three of
the most important - is this red real, did the tree come back, did anything go wrong at all - were
printouts.

**The test that matters is not "what kind is this" but "would removing it change any verdict the
system can produce".** Six of the labels above had been verified by execution; four had been assigned
by READING the code, which is the method that has been wrong three times on this unit already. So
each of those four was then executed against the condition it claims to catch - a non-unique anchor, a
build that fails under an arm, an arm run that yields no summary line - and each does fail the
verdict, with a positive control confirming a fully correct sweep still returns zero.

**The last label was the interesting one, and it was a third of a guarantee.** Restore-on-exit was
labelled as preventing a leftover mutation. Measured against three real process deaths: an uncaught
exception restores, a `sys.exit` through the signal path restores, and a **hard kill does not - the
mutation survives on disk**. So for the death mode that actually happened to this unit once, restore
prevents nothing; the only thing standing between it and a later run measuring mutated source is the
clean-tree precondition, which was then confirmed to see exactly the leftover such a kill produces.
The driver's own note now records the measured split rather than the comfortable summary. All three harnesses were re-run after the change, and the run count re-measured, because
the fix altered control flow that every earlier check depended on.

**Every one of these guards was watched refusing, each with a positive control**, rather than being
trusted because it was written: clean-and-matching admits while dirty, untracked-in-`src` and
head-not-pushed all refuse; a green-and-complete baseline proceeds while a green-but-truncated one and
a complete-but-red one both abort; a short arm total is marked truncated and NOT scored while a
complete one is scored normally. A guard nobody has seen refuse is not a guard.

One correction found while doing this: the canary gate sat AFTER the slice baseline, so the default
path ran a baseline it never used and would have handed the lane an ELEVENTH run while the
registration said ten. A registered count the driver does not honour is worth nothing, and the error
was in the direction of quietly consuming more of a contended box.

**The registered count is now MEASURED, not counted by reading.** The driver's default path was
executed with its runners instrumented: **exactly ten full-suite runs and zero slice runs** - one
baseline, eight arms, one restored re-check. Measuring it also corrected an arithmetic claim of my
own about the opt-in canary path: I expected twenty-two slice runs (one baseline plus twenty-one
mutations) and it actually takes twenty-three, because that path has its own restored re-check at the
end. The driver was right and my arithmetic was wrong - which is the whole argument for executing the
path rather than counting it in your head. That path stays opt-in and is no part of the registration.

**What is NOT in the registration, and why that is the point.** The per-route builder-substitution
bypass has no arm, because it no longer compiles - twenty-one route-level primitives removed by
design. The twenty-one rename canaries are no longer part of the security proof either: they are
positive-handler canaries, a test-quality check on the filtered slice, now off by default behind a
flag. What did NOT shrink is the per-family work: eight arms for four families, because four mapping
sites and four gates are four independent things that can each be wrong on their own.

## WHICH failures this deny prevents, and which it does not

The per-failure-mode question was applied to the driver first and then, more usefully, to the thing
that actually ships. Every sentence anyone would write about this deny is true - "the four content
read families are refused on hosted" - and, like every comfortable summary, it is not exhaustive. The
coverage is stated here so nobody reads it as "this state is protected".

| Failure mode | Covered? | By what |
|---|---|---|
| An existing route in a guarded family serves cross-tenant content on hosted | YES | the group filter; proven by the hosted refusal rows |
| A NEW route added to a guarded family later | YES | the group filter runs before routes that do not exist yet; proven by the four future-route probes |
| An existing route mapped onto the ungrouped builder to escape the filter | YES | does not compile - the routes are mapped where the unfiltered builder is out of scope |
| The hosted gate written backwards | YES | the four gate-inversion arms |
| The deny accidentally reaching self-host | YES | the self-host control class, plus the outside-the-group scoping control |
| **A NEW ENDPOINT FILE reading the same store with no filter of its own** | **NO** | **nothing. No test fails.** |
| **Writes into these stores** | **NO for two families** | see the un-deny ledger below |

**The uncovered mode is not hypothetical, and one store makes it cheap.** Today each state has exactly
one reader and each sits inside a guarded group - that was checked, not assumed. But nothing prevents
a fifth surface: a new endpoint file that constructs the store and serves it, with no group and no
filter, would be cross-tenant on hosted and **no test in the repository would go red**. The group
filter defends the family it is attached to; it cannot defend a family nobody has written yet.

`VoiceUploadStore` sharpens it. Its PARAMETERLESS constructor binds the shared cross-tenant voice-turn
root (`VoiceUploadStore() : this(CcStorage.VoiceTurnUploads())`), so `new VoiceUploadStore()` written
in a new endpoint tomorrow silently reaches every account's staged audio - no argument, no warning,
nothing to notice in review. That constructor belongs to the tenant-keying unit for that store
(#1900), so it is reported rather than edited here: two units changing one file is how the drift this
mission keeps finding gets started.

The durable fix for this mode is not another deny - it is tenant-keying the stores, which is exactly
what the issues in the un-deny ledger below are for. Recorded here so the gap is visible in the
record rather than resting on my having noticed it once.

## The un-deny ledger: two questions, not one

A read-only deny is a DEFERRED LEAK - content keeps accumulating behind it, and the day it is lifted
it exposes a contaminated history rather than a clean start. The first version of this ledger asked
whether the DENIED ROUTES write. That is the wrong question, and it is the question all four denies
tonight got wrong. The right ones are two, asked separately:

**(a) Does ANYTHING still write this state?** Not "do the denied routes write it" - a deny closes one
door and says nothing about the others.
**(b) What is ALREADY SITTING THERE from before the deny?** That does not improve while the refusal
stands, and stopping accumulation never cleans a root.

**A CALL SITE PROVES CODE EXISTS, NOT THAT IT RUNS**, and **NO NEW WRITES IS A STATEMENT ABOUT THE
FUTURE, NOT EVIDENCE ABOUT THE PAST**. Both of those cost me a claim below.

Re-asking (a) properly changed an answer twice - once in each direction. I first wrote that the
wingman training records are written by the Gateway's own voice sweep timer with no request from
anybody. That came from reading call sites. The writers are real but LATENT: every one of them is
behind an opt-in setting that defaults to off. The claim is corrected in the table below rather than
quietly softened. The transcription claim was put through the same check and survived, and is now
recorded with the evidence instead of the inference.

The two upload families were re-swept at METHOD level: every store construction, every production
caller of every mutating method on the type, and any raw file writer into the root that bypasses the
type - not the routes that happen to touch the state.

| Family | (a) Anything still writing? | (b) What already exists | Un-deny condition |
|---|---|---|---|
| Transcription analysis (#1897) | **YES - re-checked through the gate, not the call sites, and it survives.** `TranscriptionTelemetryLog.Record` is UNCONDITIONAL: no enabled-check of any kind. The only flag near it decides whether the spoken text is included, defaults to TRUE, and has no configuration key wired to turn it off. So every turn appends a record with the text in it. The recording endpoints, the batch transcription endpoint and the wingman transcribe route all reach it and none is in this group. Read-only deny, same shape as the merged stats deny thefrederiksen/devthrottle#1888. | Every account's speech, from before the deny and growing the whole time. | **Removing the deny requires ALSO purging or partitioning what accumulated behind it.** Records carry no tenant, so they cannot be attributed afterwards: deletion or quarantine, never a later migration. |
| Wingman instructions (#1853) | **LATENT, not active - I had this wrong and corrected it.** The prompt half IS contained (save, revert, switch-to-default are all in this group). For the records I first said the voice sweep timer writes them, having read call sites. Following the gate instead: all three writers - the voice-turn route, the explain route, and the timer-driven `GenerateOnceAsync` - funnel through `WingmanTrainingStore.CaptureAsync`, whose first statement checks `wingman_training_capture`, which is opt-in and defaults to FALSE. On a box that never set it, nothing writes. Turn it on and three writers, one unattended, start appending - none covered by this deny. | Records from any period the setting was on, or carried in from a self-host box. | **Removing the deny requires ALSO purging or partitioning what accumulated behind it.** Plus the interim training-write deny, still not merged. The read deny is worth having even while the writers are dormant. |
| Wingman utterance upload (#1896) | **No.** Writers enumerated exhaustively, not inferred: every `VoiceUploadStore` construction in the repository; the only production one on this root is `GatewayHost._voiceTurnUploads`, handed to this endpoint alone, and every `uploads.` call in the file is inside the guarded group. `SweepAbandoned` has NO production caller, only a test. The one residual write is the constructor ensuring an empty directory. | Unknown, presumed contaminated: the shared root may hold pre-deny cross-tenant staged audio and assembled transcripts. | **The write is stopped - AND the un-deny still requires purging or quarantining the legacy root**, on top of tenant-keying the store. "The write is stopped" is not "the root is clean". |
| Dictation upload (#1884) | **No.** Same exhaustive enumeration: the only production writer on the dictation-uploads root is `GatewayHost._dictationUploads` into this endpoint, plus one read-only status query and Core's read-only `DictationLockReader`. Every mutation in the file is inside the group; the static `_completes` cache is filled only by the completion leg, refused too. | Unknown, presumed contaminated: pre-deny staged audio, delivery records and transcripts. | **The write is stopped - AND the un-deny still requires purging or quarantining the legacy root**, on top of tenant-keying the store, the record and the `_completes` cache. |

The boundary of the two "no" answers is deliberately narrow: they are about THAT STORE. A completed
dictation also writes a turn into the shared transcription telemetry log, which is the first row's
problem and is the accumulating kind.

Note what is deliberately NOT written anywhere: "blocked on partitioning" as a standalone phrase.
Someone reading a ticket later ticks that off in good faith and re-enables the route. Every row names
the specific store and the specific thing that has to be true first.

## The un-deny debt

**A deny is a tracked promise to come back, never a silent state.** When these features are turned on
for hosted, this is the explicit list of what to un-deny and what has to be true first.

| Route | File | Remove | Un-deny only when |
|---|---|---|---|
| `GET /transcription/turns` | `Api/TranscriptionAnalysisEndpoint.cs` | group filter calling `DenyOnHosted` | thefrederiksen/devthrottle_internal#896: the telemetry log is partitioned by tenant AND existing unattributed records are deleted or quarantined (they cannot be attributed after the fact) |
| `GET /transcription/stats` | same | same filter | same as above |
| `GET /transcription/terms` | same | same filter | same as above |
| `GET /transcription/words` | same | same filter | same as above |
| `GET /gateway/wingman/instructions` | `Api/WingmanInstructionsEndpoint.cs` | group filter calling `DenyOnHosted` | A per-tenant wingman prompt exists in the schema. Today a Gateway has exactly ONE, so there is no correct hosted answer |
| `PUT /gateway/wingman/instructions` | same | same filter | same as above |
| `GET /gateway/wingman/instructions/versions` | same | same filter | same as above |
| `POST /gateway/wingman/instructions/revert` | same | same filter | same as above |
| `GET /gateway/wingman/instructions/default` | same | same filter | same as above |
| `GET /gateway/wingman/instructions/update` | same | same filter | same as above |
| `POST /gateway/wingman/instructions/switch-to-default` | same | same filter | same as above |
| `GET /gateway/wingman/instructions/records` | same | same filter | thefrederiksen/devthrottle_internal#876: `WingmanTrainingStore` partitioned by tenant, the training WRITE deny lands, AND pre-existing records on the hosted box are dealt with |
| `POST /gateway/wingman/instructions/test` | same | same filter | thefrederiksen/devthrottle_internal#876 as above, PLUS a decision on who may spend model calls on shared infrastructure |
| `POST /wingman/utterance/upload` | `Api/GatewayWingmanVoiceEndpoint.cs` | group filter calling `DenyUtteranceOnHosted` | thefrederiksen/devthrottle_internal#895: store root AND record partitioned by tenant, tenant resolved and AUTHORIZED on all three legs, same-upload-id canaries proving chunks and transcript cannot cross in either direction, each with a positive control |
| `PUT /wingman/utterance/{uploadId}/chunk/{index}` | same | same filter | same as above |
| `POST /wingman/utterance/{uploadId}/complete` | same | same filter | same as above |
| `POST /dictation/upload` | `Api/GatewayDictationEndpoint.cs` | group filter calling `DenyOnHosted` | thefrederiksen/devthrottle_internal#889: the `VoiceUploadStore` root, the `DictationDeliveryRecord` and the static `_completes` cache all tenant-keyed, and the tenant authorized on every leg - not only `/complete` |
| `PUT /dictation/{uploadId}/chunk/{index}` | same | same filter | same as above |
| `POST /dictation/{uploadId}/complete` | same | same filter | same as above |
| `POST /dictation/{uploadId}/ack` | same | same filter | same as above |
| `POST /dictation/{uploadId}/abandon` | same | same filter | same as above |

## Two things deliberately NOT done here

- **The transcription audio archive has no HTTP read route on `main`.** thefrederiksen/devthrottle_internal#896 names it alongside the
  telemetry log, so it was searched for specifically: `TranscriptionAudioArchive` exposes `FileFor`
  and `TrySave` and nothing that reads, no route resolves it, and the Gateway serves no static files.
  The raw audio is still unpartitioned on disk - the rest of thefrederiksen/devthrottle_internal#896's job - but there is no read
  surface to deny.
- **The training-capture WRITE (#1853) is still open on `main`.** The interim hosted write deny that
  issue describes is not merged. This closes the READ, which a write deny would not have done anyway.

## Proof status

Rebased onto `origin/main` at f2028deb. `git range-diff` reports every commit identical across the
rebase - the verification is that range-diff, not a rebuild that happened to succeed.

There is no replacement proof on this head yet, and the old results stay retired. The registration
above is what will be run, in one drained sitting, and it is deliberately smaller than the one it
replaces because the code no longer needs the difference.

A run that produces no real `Passed!`/`Failed!` summary line is UNPROVEN and re-run individually,
never scored as "no red" in either direction. Full seven-project counts go up with the arms.

### DIFF
--- a/src/CcDirector.Gateway.Tests/HostedContentDenyGroupFilterTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedContentDenyGroupFilterTests.cs
@@ -1,0 +1,264 @@
+using System.Net;
+using System.Text.Json;
+using CcDirector.AgentBrain;
+using CcDirector.Core;
+using CcDirector.Core.Configuration;
+using CcDirector.Gateway.Tests.Data;
+using CcDirector.Gateway.Api;
+using CcDirector.Gateway.Discovery;
+using CcDirector.Gateway.Transcription;
+using CcDirector.Gateway.Voice;
+using CcDirector.Gateway.Wingman;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// PROVES THE PROPERTY THAT MADE THE GROUP FILTER THE RIGHT SHAPE, WHICH NOTHING ELSE HERE CAN SEE.
+///
+/// Each of the four denied families is guarded by ONE <c>AddEndpointFilter</c> on the route group rather
+/// than a <c>DenyOnHosted()</c> call repeated in every handler. The stated reason is that a per-route
+/// guard ROTS - the moment someone adds a route to that file it is undefended and nothing fails - whereas
+/// a group filter runs before every route in the group INCLUDING ROUTES THAT DO NOT EXIST YET.
+///
+/// That difference is completely invisible to any test that only drives the routes existing today: a
+/// per-route guard and a group filter behave identically on those. Which is exactly what makes the
+/// per-route shape dangerous, and exactly why the claim needs its own proof rather than an assurance in a
+/// pull request body. Until this file existed, the pull request asserted the future-route property and
+/// nothing tested it.
+///
+/// So each family maps a BRAND-NEW route onto its returned group, with NO deny written for it anywhere,
+/// and asserts:
+///
+///   1. HOSTED           - the probe route is refused, carrying that family's refusal and nothing else.
+///   2. SELF-HOST        - the SAME probe route is SERVED, over BOTH non-hosted forms that occur in
+///                         practice (the variable absent, and the variable present but "0").
+///   3. SCOPING CONTROL  - a route mapped OUTSIDE the group still SERVES on hosted.
+///
+/// All three are needed and none is redundant. Without (2) a filter that refused everything
+/// unconditionally would pass every hosted assertion in this file while having silently killed the route
+/// for self-host too - a brick is indistinguishable from a working gate if you only push on it from one
+/// side. Without (3) the hosted passes could be the host refusing everything rather than the filter doing
+/// its job.
+///
+/// The self-host legs also assert <see cref="GatewayHostedMode.IsHosted"/> is genuinely false rather than
+/// trusting the environment variable to have taken effect, so no leg can silently run in the mode it
+/// believes it is not in.
+/// </summary>
+[Collection("DirectorRoot")]   // serialized: every leg mutates the process-wide CC_GATEWAY_HOSTED and
+                              // CC_DIRECTOR_ROOT, so running beside another collection would let one
+                              // test's mode leak into another's - the exact silent-wrong-mode failure
+                              // the IsHosted assertions above exist to catch.
+public sealed class HostedContentDenyGroupFilterTests
+{
+    private const string ProbeBody = "probe-served-zqxjv";
+
+    private const string TranscriptionRefusal = "transcription analysis is not available on the hosted gateway";
+    private const string InstructionsRefusal = "the wingman instructions surface is not available on the hosted gateway";
+    private const string UtteranceRefusal = "the wingman utterance upload is not available on the hosted gateway";
+    private const string DictationRefusal = "dictation upload is not available on the hosted gateway";
+
+    /// <summary>Which family, its group-mapper, and the refusal its filter must produce.</summary>
+    public static TheoryData<string> Families() => new() { "transcription", "instructions", "utterance", "dictation" };
+
+    private static string RefusalFor(string family) => family switch
+    {
+        "transcription" => TranscriptionRefusal,
+        "instructions" => InstructionsRefusal,
+        "utterance" => UtteranceRefusal,
+        "dictation" => DictationRefusal,
+        _ => throw new ArgumentOutOfRangeException(nameof(family)),
+    };
+
+    // ===== 1. HOSTED: a route added to the group later is refused, with no deny of its own =====
+
+    [Theory]
+    [MemberData(nameof(Families))]
+    public async Task A_route_added_to_the_group_later_is_refused_on_hosted(string family)
+    {
+        using var env = new HostedEnv("1");
+        Assert.True(GatewayHostedMode.IsHosted, "the hosted leg must actually be in hosted mode");
+
+        await using var rig = await Rig.StartAsync(family);
+
+        var resp = await rig.Http.GetAsync("probe-added-later");
+
+        var root = await ContentFingerprint.AsJsonObjectAsync(resp, $"{family} probe on hosted");
+        Assert.Equal(new[] { "error" }, root.EnumerateObject().Select(p => p.Name).ToArray());
+        Assert.Equal(RefusalFor(family), ContentFingerprint.Text(root, "error", family));
+        Assert.Equal(HttpStatusCode.NotFound, resp.StatusCode);
+    }
+
+    // ===== 2. SELF-HOST: the same probe route still serves, over BOTH non-hosted forms =====
+
+    [Theory]
+    [InlineData("transcription", null)]
+    [InlineData("transcription", "0")]
+    [InlineData("instructions", null)]
+    [InlineData("instructions", "0")]
+    [InlineData("utterance", null)]
+    [InlineData("utterance", "0")]
+    [InlineData("dictation", null)]
+    [InlineData("dictation", "0")]
+    public async Task A_route_added_to_the_group_still_serves_on_self_host(string family, string? hostedValue)
+    {
+        using var env = new HostedEnv(hostedValue);
+        Assert.False(GatewayHostedMode.IsHosted,
+            $"the self-host leg must actually be in self-host mode (CC_GATEWAY_HOSTED={hostedValue ?? "absent"})");
+
+        await using var rig = await Rig.StartAsync(family);
+
+        var resp = await rig.Http.GetAsync("probe-added-later");
+        var body = await resp.Content.ReadAsStringAsync();
+
+        Assert.Equal(HttpStatusCode.OK, resp.StatusCode);
+        Assert.Equal(ProbeBody, body);
+    }
+
+    // ===== 3. SCOPING CONTROL: a route OUTSIDE the group still serves on hosted =====
+
+    [Theory]
+    [MemberData(nameof(Families))]
+    public async Task A_route_outside_the_group_still_serves_on_hosted(string family)
+    {
+        using var env = new HostedEnv("1");
+        Assert.True(GatewayHostedMode.IsHosted, "the hosted leg must actually be in hosted mode");
+
+        await using var rig = await Rig.StartAsync(family);
+
+        // Mapped on the application, NOT on the guarded group. If this were refused, the hosted passes
+        // above would be the host refusing everything rather than the filter being correctly scoped.
+        var resp = await rig.Http.GetAsync("outside-the-group");
+        var body = await resp.Content.ReadAsStringAsync();
+
+        Assert.Equal(HttpStatusCode.OK, resp.StatusCode);
+        Assert.Equal(ProbeBody, body);
+    }
+
+    /// <summary>Sets CC_GATEWAY_HOSTED for one test and restores whatever was there before.</summary>
+    private sealed class HostedEnv : IDisposable
+    {
+        private readonly string? _prior;
+
+        public HostedEnv(string? value)
+        {
+            _prior = Environment.GetEnvironmentVariable("CC_GATEWAY_HOSTED");
+            Environment.SetEnvironmentVariable("CC_GATEWAY_HOSTED", value);
+        }
+
+        public void Dispose() => Environment.SetEnvironmentVariable("CC_GATEWAY_HOSTED", _prior);
+    }
+
+    /// <summary>
+    /// A minimal application carrying ONE family's guarded group, plus a probe route mapped onto that
+    /// group and a control route mapped outside it. Deliberately not a whole GatewayHost: the point is to
+    /// hold the returned group and map onto it, which only the endpoint's own Map can hand back.
+    /// </summary>
+    private sealed class Rig : IAsyncDisposable
+    {
+        private readonly WebApplication _app;
+        private readonly string _root;
+        private readonly string? _priorRoot;
+        private readonly GatewayDbTestHarness? _db;
+        public HttpClient Http { get; }
+
+        private Rig(WebApplication app, HttpClient http, string root, string? priorRoot,
+            GatewayDbTestHarness? db)
+        {
+            _app = app;
+            Http = http;
+            _root = root;
+            _priorRoot = priorRoot;
+            _db = db;
+        }
+
+        public static async Task<Rig> StartAsync(string family)
+        {
+            var priorRoot = Environment.GetEnvironmentVariable("CC_DIRECTOR_ROOT");
+            var root = Path.Combine(Path.GetTempPath(), "ccd-probe-" + Guid.NewGuid().ToString("N"));
+            Environment.SetEnvironmentVariable("CC_DIRECTOR_ROOT", root);
+
+            var builder = WebApplication.CreateBuilder();
+            builder.Logging.ClearProviders();
+            var app = builder.Build();
+            app.Urls.Add("http://127.0.0.1:0");
+
+            GatewayDbTestHarness? db = null;
+            var group = MapFamily(app, family, root, ref db);
+
+            // The brand-new route, on the guarded group, with NO deny of its own.
+            group.MapGet("/probe-added-later", () => Results.Text(ProbeBody));
+            // The control, deliberately OUTSIDE the group.
+            app.MapGet("/outside-the-group", () => Results.Text(ProbeBody));
+
+            await app.StartAsync();
+            return new Rig(app, new HttpClient { BaseAddress = new Uri(app.Urls.First()) },
+                root, priorRoot, db);
+        }
+
+        private static RouteGroupBuilder MapFamily(WebApplication app, string family, string root,
+            ref GatewayDbTestHarness? db)
+        {
+            var brain = (WingmanModelRole _, CancellationToken _) =>
+                Task.FromException<IAgentBrain>(
+                    new InvalidOperationException("the brain must not be reached by a routing probe"));
+
+            switch (family)
+            {
+                case "transcription":
+                    return TranscriptionAnalysisEndpoint.Map(app);
+
+                case "instructions":
+                    db = new GatewayDbTestHarness();
+                    return WingmanInstructionsEndpoint.Map(
+                        app,
+                        new WingmanInstructionsStore(db.Open(), db.LegacyPath("probe-legacy.json")),
+                        new WingmanTrainingStore(() => false, Path.Combine(root, "training")),
+                        brain);
+
+                case "utterance":
+                {
+                    var vault = new KeyVault(Path.Combine(root, "probe.vault"));
+                    var voice = new WingmanVoiceService(brain, vault, Path.Combine(root, "voice.json"));
+                    return GatewayWingmanVoiceEndpoint.Map(
+                        app,
+                        new DirectorRegistry(Path.Combine(root, "instances")),
+                        brain,
+                        vault,
+                        voice);
+                }
+
+                case "dictation":
+                {
+                    var vault = new KeyVault(Path.Combine(root, "probe.vault"));
+                    return GatewayDictationEndpoint.Map(
+                        app,
+                        new DirectorRegistry(Path.Combine(root, "instances")),
+                        owners: null,
+                        token: "probe-token",
+                        transcription: new GatewayTranscriptionService(vault),
+                        transcribingSessions: new TranscribingSessions(),
+                        uploads: new VoiceUploadStore(Path.Combine(root, "uploads")),
+                        devices: new Pairing.DeviceRegistry(Path.Combine(root, "devices.json")));
+                }
+
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(family), family, "unknown family");
+            }
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            Http.Dispose();
+            await _app.StopAsync();
+            await _app.DisposeAsync();
+            Environment.SetEnvironmentVariable("CC_DIRECTOR_ROOT", _priorRoot);
+            _db?.Dispose();
+            try { if (Directory.Exists(_root)) Directory.Delete(_root, true); } catch { /* best effort */ }
+        }
+    }
+}
--- a/src/CcDirector.Gateway.Tests/HostedContentDenyGroupFilterTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedContentDenyGroupFilterTests.cs
@@ -1,4 +1,5 @@
 using System.Net;
+using System.Text;
 using System.Text.Json;
 using CcDirector.AgentBrain;
 using CcDirector.Core;
@@ -37,18 +38,28 @@ namespace CcDirector.Gateway.Tests;
 /// nothing tested it.
 ///
 /// So each family maps a BRAND-NEW route onto its returned group handle, with NO deny written for it
-/// anywhere, and asserts:
+/// anywhere. The probe route is a BODY-BOUND POST, not a parameterless GET, because a parameterless GET is
+/// the ONE shape the underlying endpoint-filter defect could never be seen through: a filter that answered
+/// AFTER model binding still refused a GET correctly while leaking every bound body, a malformed body to the
+/// framework's own 400, and a wrong media type to its own 415. The GET probe this file first shipped proved
+/// the future-route property but was blind to a regression back to that defect. A body-bound POST is not, so
+/// each family asserts:
 ///
-///   1. HOSTED           - the probe route is refused, carrying that family's refusal and nothing else.
-///   2. SELF-HOST        - the SAME probe route is SERVED, over BOTH non-hosted forms that occur in
-///                         practice (the variable absent, and the variable present but "0").
+///   1. HOSTED           - the future route is refused across the shapes that regression hides in: a
+///                         MALFORMED body (answered by the framework's own 400 under the old defect), a
+///                         WRONG MEDIA TYPE (answered by its own 415), and - the load-bearing one - a
+///                         NO-BINDING sentinel proving NO handler-bound code ran behind the refusal on any
+///                         shape. Each carries that family's refusal and nothing else.
+///   2. SELF-HOST        - the SAME body-bound route still BINDS and SERVES, over BOTH non-hosted forms that
+///                         occur in practice (the variable absent, and the variable present but "0"), and the
+///                         custom binder is proven to have actually run.
 ///   3. SCOPING CONTROL  - a route mapped OUTSIDE the group still SERVES on hosted.
 ///
-/// All three are needed and none is redundant. Without (2) a filter that refused everything
-/// unconditionally would pass every hosted assertion in this file while having silently killed the route
-/// for self-host too - a brick is indistinguishable from a working gate if you only push on it from one
-/// side. Without (3) the hosted passes could be the host refusing everything rather than the primitive
-/// doing its job.
+/// All three are needed and none is redundant. Without (2) a filter that refused everything unconditionally -
+/// or one that bricked binding on every deployment - would pass every hosted assertion in this file while
+/// having silently killed the route for self-host too; a brick is indistinguishable from a working gate if
+/// you only push on it from one side. Without (3) the hosted passes could be the host refusing everything
+/// rather than the primitive doing its job.
 ///
 /// The self-host legs also assert <see cref="GatewayHostedMode.IsHosted"/> is genuinely false rather than
 /// trusting the environment variable to have taken effect, so no leg can silently run in the mode it
@@ -83,6 +94,9 @@ public sealed class HostedContentDenyGroupFilterTests
 
     private static string ProbePath(string family) => PrefixFor(family) + "probe-added-later";
 
+    /// <summary>The second future route, whose parameter is a custom binder we can watch run (or not run).</summary>
+    private static string BindingProbePath(string family) => PrefixFor(family) + "probe-binding";
+
     private const string TranscriptionRefusal = "transcription analysis is not available on the hosted gateway";
     private const string InstructionsRefusal = "the wingman instructions surface is not available on the hosted gateway";
     private const string UtteranceRefusal = "the wingman utterance upload is not available on the hosted gateway";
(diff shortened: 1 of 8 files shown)


thefrederiksen (commented):
REJECT

1. All four filter reverts were run only against the 40-test deny/control slice. For each real production-line filter removal, record a successful build and the full 2900+ Gateway suite count with only the intended deny assertions red and controls green; then restore, rebuild, and rerun full.

2. `HostedContentReadSelfHostControlTests.Self_host_answers_an_unknown_upload_id_itself_not_with_the_hosted_refusal` sends `POST` to `/dictation/no-such-id/chunk/0`, but production maps that route only as `PUT`. The request never reaches the dictation handler; it receives framework 405, and the helper passes because it merely cannot find one of four refusal strings. Correct the verb and assert the store's exact expected status/body.

3. The explicit hosted-off class omits the four wingman-instructions writes and the upload completion legs, while its unknown-id helper accepts any non-refusal response, including a missing-route 404 or wrong-method 405. The cited untouched suites do not explicitly clear hosted mode. Add explicit hosted-off controls that prove real handler behavior, not merely absence of four strings.

4. The four group filters have no future-route probes. Prove a probe mapped through each group/filter seam is refused on hosted and served with hosted mode explicitly off.

The direct hosted-mode gates, group structure, exact refusal allow-lists for current hosted routes, and route-by-route un-deny ledger are otherwise sound. The restored CI red is the known unrelated account-status case.

thefrederiksen (commented):
Reviewer correction to my earlier false-green finding: the POST-vs-PUT case is still non-probative, but the specific claim that ASP.NET necessarily returns a framework 405 is superseded. Mutation testing elsewhere in this Gateway showed that the last-mapped Cockpit SPA fallback swallows unmatched paths/verbs; an unmatched non-GET can instead return the Gateway fallback 404 body, and an unmatched GET can even return the built Cockpit HTML shell with 200. The acceptance requirement is therefore stronger: every self-host real-endpoint test must assert a handler-unique receipt (seeded payload, route-specific content type, or a state change re-read afterward), never merely status or absence of hosted refusal text, and the canary itself must be mutation-proven by renaming or re-verbing the route and observing red. The current POST dictation/chunk row still fails that bar and can false-green without selecting the intended PUT handler.

thefrederiksen:
## Status after re-audit: proofs are being redone, do not read the old results

The re-audit found the self-host control class unsound, and it has been rewritten. The
deny code itself is unchanged and its four revert-proofs still stand, but **the test
results previously posted on this pull request no longer describe the code on the branch**
and should not be relied on. New results will be posted when the test lane is available.

### What was wrong

1. **The self-host rows proved nothing.** They asserted status 200 plus the *absence* of the
   four refusal strings. `CockpitReactApp` maps `MapFallback("{*path}")`
   (CockpitReactApp.cs:125), which matches any unclaimed path and any verb and answers with
   the single-page-app shell - 404 in a Debug test host, but **200 with the HTML shell on a
   release host**, and its own not-found text says "release build only". So those rows would
   have passed on routes that had been deleted.
2. **The dictation chunk row used the wrong verb.** Production maps that path with `MapPut`
   and nothing else (GatewayDictationEndpoint.cs:176). The test drove `POST`, so the request
   never reached the handler - and because **no 405 is ever raised on these paths**, nothing
   in a status-based assertion could notice. This was the original rejection finding and it
   is now fixed.
3. **A second masking route exists** and is recorded for completeness:
   `SessionWsProxyEndpoints` maps `app.Map("/sessions/{sid}/{**rest}")`
   (SessionWsProxyEndpoints.cs:153), least-specific and every verb, answering **503 as
   `application/json`** - so a content-type check does not catch it. None of the four
   families denied here lives under `/sessions/{sid}/`, so this one cannot mask *these*
   routes; it is written into the test file because it is why "check the content type" is not
   a general defence on this Gateway. All three facts were read on this branch, not assumed.

### What the controls assert now

Every read asserts **seeded or route-specific JSON that only that handler could produce**,
and every write asserts **its own store receipt** - a state change read back, never a status.

| Route | Receipt |
|---|---|
| `GET /transcription/turns` | one telemetry record planted via the production writer, its `rawText` read back |
| `GET /transcription/stats` | `totalTurns == 1` exactly, in an isolated otherwise-empty log |
| `GET /transcription/terms` | the planted find/replace pair read back |
| `GET /transcription/words` | the planted word counted |
| `GET /gateway/wingman/instructions/records` | a training record planted on disk, its positional `<filename>#<lineindex>` id and reply preview read back |
| `GET`/`PUT` `/gateway/wingman/instructions`, `GET /versions`, `GET /update` | a version is saved through the route and round-tripped, and the customized flag is read back through a **different** route than the one that set it |
| `GET /gateway/wingman/instructions/default` | exact top-level property set `{version, hash, content}` plus non-empty content |
| `PUT /wingman/utterance/{id}/chunk/{i}` | the uploaded bytes are located on disk **by content** under the isolated root |
| `PUT /dictation/{id}/chunk/{i}` | same, and the verb is now `PUT` |
| `POST /dictation/{id}/abandon` | the tombstone is read back at re-register (`terminal`, `dropped`) |
| `POST /dictation/{id}/ack` | retires exactly once: first ack `retired=true`, second `retired=false` |

The hosted deny rows already survived the fallback by construction - each parses the body and
requires the property set to be exactly one `error` field with an exact message, so an HTML
shell fails to parse and any other JSON fails the property-set check.

### Proofs still owed

- **Mutation-red on every row**: rename or re-verb each route, confirm the matching test goes
  RED, restore, confirm green. This is the only check that catches both masking routes,
  because it does not depend on predicting what the framework returns. The recipe is written
  into the test class.
- **Re-run the four deny revert-proofs** against the reworked file.
- **Full seven-project counts**, rebased.

Rebased onto `origin/main` at thefrederiksen/devthrottle#1916, so the `HostedAccountStatusTests.Enrolled_without_a_recorded_email`
allowance is retired - if that failure appears it is real and mine.

thefrederiksen:
REJECT on exact head `5263ce21` (static review; no lane consumed).

The un-deny split is substantially better: transcription telemetry and wingman training records have live writers outside the read deny and therefore require writer conversion plus purge/quarantine; the two upload filters stop new route-driven writes into their respective `VoiceUploadStore`s. Two blockers remain.

1. All four route families still have an independently bypassable mapping primitive per route. `TranscriptionAnalysisEndpoint`, `WingmanInstructionsEndpoint`, and `GatewayDictationEndpoint` keep both `outer` and filtered `app` in scope while mapping production routes; `GatewayWingmanVoiceEndpoint` keeps unfiltered `app` beside filtered `utterance`. Changing one mapping receiver compiles and opens only that route on hosted. A future-route probe proves inheritance only for routes actually put on the group. Move each family’s production mappings into a private method that accepts only its filtered `RouteGroupBuilder`, as the recording/cached-voice rework now does.

The twenty-one rename/re-verb arms are positive-handler canaries; they are not mutations of this gate bypass. In the current shape, every builder substitution is its own security arm and must run against the full Gateway suite. The scope split is the sensible way to remove those primitives.

2. The registered proof also omits the four independent hosted-gate-direction mutations. Each family is a conjunction: the filter must be attached, and its `GatewayHostedMode.IsHosted` condition must face the correct way. A filter-removal arm does not prove the condition. After the scope split, the minimum exact-head proof is a clean baseline, eight independent full-suite arms (four filter removals plus four gate inversions), and an exact restored green, with assertion reds and reconciliation.

One ledger correction is also required before merge: “write stopped” means no *new guarded-route* upload writes while the deny stands; it does not prove the existing shared upload roots are clean. They may contain pre-deny cross-tenant content. Future un-deny must tenant-key the store and quarantine/purge/migrate those legacy roots, even though the filters correctly stop further route accumulation.

The old proofs are correctly retired and the current head has no replacement proof. Rework and rebase onto current main before requesting the lane.

thefrederiksen:
STANDING: REJECTED
cb29e3f6df99796859a075718e6592fb0e2ad7df

The hosted denial is not uniform across all request shapes: part of the request pipeline can answer before the family-wide refusal boundary on several denied routes. The current valid-input coverage therefore cannot prove the stated exact-refusal guarantee.

Security details and the required proof direction were reported privately.

thefrederiksen:
STANDING: REJECTED
cb29e3f6df99796859a075718e6592fb0e2ad7df

The rework closes the prior family-local blockers. Each mapping helper now receives only its guarded group; four independent gate-direction arms are registered beside the four attachment arms; the transcription/training ledgers distinguish active, latent, and existing state; and both upload roots are correctly treated as potentially contaminated before un-deny.

The common boundary remains four endpoint-group filters, so it is still post-binding. All four future-route probes are parameterless GETs and therefore do not exercise the required pre-binding property. The outside-group controls use unrelated paths rather than same-namespace neighbours, so they do not test the over-refusal direction at the family boundary.

There is no replacement proof on this head. The clean complete baseline, eight independent full-suite arms, exact restored green, seven-project counts, and current-main reconciliation are outstanding.

Rework must adopt a cleared shared pre-binding primitive, convert every future canary to a body-bound POST with observable no-binding evidence, add meaningful family neighbours with positive controls, and then execute the registered serialized proof on the rebased final head.

PR #1904, opened 2026-07-19
fix(gateway): deny the whole key-vault route group on hosted

thefrederiksen:
## What this closes

The `/vault/keys` group on the hosted Gateway was a credential-theft and tampering path, open to every subscriber.

The store behind those routes is ONE global key vault file at the shared storage root. There is no tenant in the file format, no tenant in the store, and no tenant in the routes. The group sits behind only the host-wide authentication gate, and that gate admits ANY enrolled device key from ANY account. So on the hosted Gateway, any subscriber could:

- list every key name on the box (`GET /vault/keys`),
- read any other subscriber's provider credential in cleartext (`GET /vault/keys/{name}` returns the raw value),
- overwrite it and redirect that account's paid usage (`PUT /vault/keys/{name}`),
- delete it and silently disable that account's service (`DELETE /vault/keys/{name}`).

That is theft and tampering, not merely disclosure, which is why the write and the delete close alongside the read.

## CORRECTION: a false claim this pull request previously made

An earlier revision of this pull request stated, in the body, the document and the code, that **"the write is stopped"** - that the deny covered `PUT` and `DELETE` so nothing accrued behind the refusal, and that lifting it would need only a per-account namespace with **no purge**.

**That was false.** Reviewer 2 found the other writers. I verified the finding independently against `origin/main` rather than accepting it, and it is correct - and slightly worse than reported.

The framing error is the reusable part, so it is recorded rather than quietly dropped:

> The question asked was **"do the DENIED ROUTES write?"** - to which the answer really is no.
> The question that decides an un-deny condition is **"does ANYTHING write?"**, and separately **"what is ALREADY SITTING THERE from before the deny existed?"**
> A deny closes ONE door. It says nothing about the other doors, and nothing about what accumulated before it was hung.

### Every writer to this vault, enumerated exhaustively

The enumeration is over the complete mutating surface of `KeyVault` - `Set`, `SetIfAbsent`, `Delete` (`src/CcDirector.Core/KeyVault.cs` lines 47, 66, 83) - grepped across all of `src` on `origin/main`, production code only:

| Writer | Operation | When it fires | Denied by this change? | Hosted-gated? |
|---|---|---|---|---|
| `VaultEndpoints` PUT handler | `Set` | HTTP route | **Yes** | n/a |
| `VaultEndpoints` DELETE handler | `Delete` | HTTP route | **Yes** | n/a |
| `GatewayHost.SeedKeyVaultFromEnvironment` | `SetIfAbsent` | Host startup | **No** | **No** |
| `TranscriptionKeyAutoProvisioner.EnsureAsync` | `SetIfAbsent`, `Set` | **Every sign-in**, and again at startup | **No** | **No** |
| `TranscriptionKeyAutoProvisioner.RevokeMintedKeyAsync` | `Delete` x2 | **Every sign-out** | **No** | **No** |

Three of the five production write paths, and two entire components, write this vault outside the deny.

On the hosted gating specifically: both startup writers are called at `GatewayHost.StartAsync` lines 1264 and 1270 - **before** the `if (!GatewayHostedMode.IsHosted)` at line 1282, which guards the Tailscale provisioner and not them. Neither has a hosted check of its own.

So key material **still arrives in the global vault while this deny is in force**, by paths that never touch the denied routes.

## The un-deny condition, corrected - THE PURGE IS REQUIRED

The sentence is narrowed to what is actually true, and the condition is ordered:

> **This deny stops the HTTP route write and the HTTP route delete. It does not stop every write.**
> **Removing it REQUIRES purging or partitioning what accumulated behind it.**

In order:

1. **Tenant-partition every remaining producer** in the table above.
2. **Quarantine, purge or migrate the pre-existing global vault root.** Material predating this deny is already in the file, and this change never touched it.
3. Only then restore a tenant-scoped route, one at a time. **The raw-value `GET` is the last one back.**

| Route | What self-host still does with it | What hosted does now | Before it can be un-denied |
|---|---|---|---|
| `GET /vault/keys` | Lists the owner's key names for the Cockpit keys panel | `404`, one `error` property | Steps 1 and 2 complete, then scope the list to the caller's account |
| `GET /vault/keys/{name}` | Returns the raw value to the owner's own Director on demand | `404`, one `error` property | Steps 1 and 2 complete, plus a per-account read path. The value read is the theft, so this is **the last one back** |
| `PUT /vault/keys/{name}` | Sets or rotates the owner's key | `404`, one `error` property | Steps 1 and 2 complete, so a write can only touch the caller's own account |
| `DELETE /vault/keys/{name}` | Removes the owner's key | `404`, one `error` property | Steps 1 and 2 complete, so a delete can only touch the caller's own account |

Step 2 is a **storage** condition, not a deployment one. It is not satisfied by any amount of configuration, environment or rollout work, and it must not be ticked off on the grounds that the routes were denied. Do not treat it as optional.

The group filter also denies any route ADDED to this file later. That is intended: a new key-vault route should have to be un-denied deliberately, not inherit access by being written.

**This deny is not clearance for anything else.** It closes one route group; it does not make key material on the hosted Gateway safe. In particular it says nothing about the voice vault key, which stays UNCONFIGURED on hosted until the whole voice chain is partitioned. Configuring that key does not deliver voice - it arms a cross-tenant leak, because clips and the plaintext transcript beside each one land in a shared directory until that chain is partitioned. Nothing here changes that.

## Severe adjacent finding, NOT fixed here, flagged rather than folded in

While enumerating the writers I found that `TranscriptionKeyAutoProvisioner` is not merely an unfiltered writer - on a hosted multi-tenant Gateway it is a live cross-tenant defect in its own right. Its own code comments show it, written for a single-user desktop install and never revisited for hosted:

- `EnsureAsync` uses `SetIfAbsent`, commented "the first writer wins". On hosted that means the **first tenant to sign in** wins, and every later tenant finds "a DevThrottle key is already stored -> nothing to mint" and silently uses **that tenant's key**, spending their credits.
- `RevokeMintedKeyAsync` fires on sign-out and deletes the key globally, with the comment that leaving it "would let the NEXT account signed in on this machine reuse it". On hosted, "the next account on this machine" is another tenant, continuously - and one tenant signing out revokes and deletes the key **every other tenant is using**.

Both are reachable through the sign-in and sign-out routes, which this deny does not cover. This is a separate surface with its own guard question, its own mutation and its own proof run, and it belongs to a separate pull request. Flagged here rather than quietly widening this one.

## The bypass count is ONE, and it is one BY DESIGN

The number of proof runs owed is set by independent bypassability: how many things can be individually wrong while everything else stays correct and isolation still breaks. That number is one.

**What was reduced.** Previously all four routes were mapped in the same method that holds the ungrouped route builder. Each of the four could INDIVIDUALLY have been mapped onto that ungrouped builder instead of onto the guarded group - a one-character edit that bypasses the filter for that route alone while every other route stays correctly denied. That is four independently bypassable primitives on top of the filter itself, and each would have owed its own single-mutation full-suite run.

They are gone because the mistake is now **inexpressible**, not because it is unlikely. The routes live in `VaultEndpoints.MapRoutes`, which receives the guarded `RouteGroupBuilder` and never receives the ungrouped `IEndpointRouteBuilder`. Inside that method there is nothing to map onto except the guarded group. That is a compile-time property, which is why no test asserts it - a test could not.

**What remains, and why nothing else counts.**

- Exactly one `AddEndpointFilter`, on exactly one `MapGroup`, at exactly one `VaultEndpoints.Map` call site - `GatewayHost.cs`, one call, checked rather than assumed. No route carries a guard of its own, so deleting the filter fails all four routes together.
- Nothing vault-specific runs ahead of the filter. The only thing in front is the host-wide authentication middleware; `An_unauthenticated_caller_is_still_rejected` pins it and stays green in both legs.
- The write handler's body read is inside the handler, after the filter.
- Removing the hosted check from `DenyOnHosted` is not a second primitive. It makes the group refuse ALWAYS, which breaks the shipped self-host product rather than isolation, and the self-host controls are what redden on it.

**The correction above does not raise the count.** It changes claims - the body, the document and the code comment - and adds no guard, no branch and no new denied surface. The quarantine is a **precondition for un-denying later**, not a mechanism shipped here, so there is nothing new to mutate. The count stays at one mutation plus the restore: **two full-suite runs.**

## Format facts before every parse, and no claim resting on absence

On this Gateway a `404` is not necessarily JSON - the single-page-app fallback answers unmatched paths with something else - so a mutation that routed a denied path to the fallback would make `JsonDocument.Parse` THROW. That red is a crash, and a crash proves only that the mutation broke something upstream of the claim; it cannot say what was served in place of the refusal.

Every place that parses a body asserts **status and media type first**: the hosted refusal helper, the group-probe helper, and every self-host receipt. The same mutation now reddens as *"expected NotFound, got OK"* or *"expected application/json, got text/plain"*.

Assertions that rested on absence alone were replaced:

- the refused-read leak check pins the exact refusal FIRST and only then checks the secret is absent - on its own, "the body does not contain the secret" is satisfied by any body that simply is not the vault;
- the three refused-effect tests assert the full refusal rather than a bare status code;
- the two served-side probes assert media type and then the handler's own parsed payload rather than a substring, since `"true"` and `"served"` appear in plenty of bodies that are not the handler's answer;
- the four self-host write and delete receipts assert media type and the parsed body before reading the store back.

## Handler-positive receipts and destructibility controls

A `404` deny is indistinguishable from a route that was never mapped, so **all four verbs** carry a self-host receipt asserting the real payload and the real effect on the store - through the group in both non-hosted forms (`CC_GATEWAY_HOSTED` absent, and set to `"0"`), and through the production `GatewayHost` with the authentication gate on and hosted mode asserted off.

The two hosted survival assertions carry matching destructibility controls: the identical write **overwrites** the identical seeded key on self-host, and the identical delete **destroys** it. Without those, "the key survived the refusal" is satisfied just as well by an operation that was never capable of destroying anything.

Every class that asserts a mode sets that mode itself and asserts `GatewayHostedMode.IsHosted` took the value.

## The dual-sided future-route probe

`VaultEndpoints.Map` returns its `RouteGroupBuilder`, solely so this property is statable from outside the file. A per-route guard and a group filter behave identically on every route that exists today, so the property that justifies a group filter is precisely the one nothing else tests.

| Direction | Test | Asserts |
|---|---|---|
| Refused on hosted | `HostedVaultGroupFilterTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own` | A brand-new route mapped onto the group is already refused, with no deny written for it anywhere, and the probe sentinel appears nowhere in the body |
| Served on self-host, **both** non-hosted forms | `SelfHostVaultGroupControlTests.A_route_added_to_the_group_still_serves_on_self_host` | The **same** probe path returns 200, `application/json`, and its parsed payload |
| Outside-the-group control | `HostedVaultGroupFilterTests.A_route_outside_the_group_still_serves_on_hosted` | A route mapped outside the group still serves on hosted, so the hosted passes are the filter being scoped and not the host refusing everything |

One direction alone cannot tell a working gate apart from a brick: a filter that refused everything unconditionally would pass every hosted assertion in this file while having silently killed the routes for self-host too.

## Proof status - OUTSTANDING, and not claimed

**No Gateway test has been run on any head of this branch.** The build is clean (`0 Warning(s), 0 Error(s)`) and nothing further is claimed.

The earliest proof recorded on this pull request was a FILTERED run - ten deny tests plus four controls - and it does not count. It is struck rather than restated.

Continuous integration is **not** the revert proof, and is not offered as one. It runs the suite once on the unmutated head; it never removes the guard, so it cannot show that anything is pinned to it.

What is owed, on the final exact head, in order:

1. A complete **clean full-suite baseline**, green.
2. **ONE mutation** with a source and build preflight: in `src/CcDirector.Gateway/Api/VaultEndpoints.cs` delete the `guarded.AddEndpointFilter(...)` block outright, leaving `var guarded = outer.MapGroup("");` in place so the group still exists and the file still compiles, with NO per-route guard put back. Never `if (false)` - unreachable code is a build error here. Rebuild, confirm zero errors (a run after a failed build executes the previous binary and reports a false pass), then run the FULL Gateway suite and record every red BY NAME with the form it arrived in. A red only counts if it fails WITH THE SYMPTOM - an assertion naming what was served instead of the refusal. Crash-reds are UNPROVEN.
3. **Restore to the exact head**, rebuild, rerun the full suite green, and confirm the counts reconcile arithmetically with total and skipped unchanged.

The machine's Gateway test lane is single-occupancy and is held by another run. This is queued behind the holder.

## Does anything on the hosted path need these ROUTES?

Investigated, not assumed. **Nothing on the hosted path needs these routes, so denying the routes is correct - though, per the correction above, not sufficient to stop all writes.**

- **In-process, unaffected.** The Gateway reads the vault through the `KeyVault` object it owns. Those paths never go through HTTP.
- **No web caller.** Nothing in the cockpit or mobile bundles calls `/vault/keys`. The only mention outside the Gateway is the generated OpenAPI schema.
- **One real HTTP caller, and it is a DIRECTOR, not a hosted feature.** `OpenAiKeyResolver` does `GET {gateway}/vault/keys/{name}` when a Director is attached to a Gateway. On self-host that is the owner's own Director reading the owner's own key, and it keeps working. Against a HOSTED Gateway that call was itself part of the defect: the value it would receive is the single global key, which belongs to the operator or to another account, never to the caller. Its `404` branch is already the "key simply not set yet" path.

So no hosted feature loses a capability it legitimately had. The one behavior that stops is a desktop Director pulling a key it was never entitled to.

## Documentation

`docs/architecture/gateway/GATEWAY_KEY_VAULT.md` states, next to the route table, that the whole table is self-host only and refused on hosted, with the reason, the exact refusal body, the corrected and ordered un-deny condition including the required purge, the full table of other writers, and the note that this deny is not clearance for configuring the voice vault key.

### DIFF
--- a/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
+++ b/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
@@ -48,6 +48,16 @@ On the always-on Gateway box, in a store **outside git** (same principle as toda
 | `PUT /vault/keys/{name}` `{ value }` | set / update a key |
 | `DELETE /vault/keys/{name}` | remove a key |
 
+**SELF-HOST ONLY. Every route in the table above is refused on the hosted Gateway** (404, body
+`{ "error": "the key vault is not available on the hosted gateway" }`). The store is one global vault
+file at the shared storage root with no tenant in the file, the store, or the routes, and the routes sit
+behind only the host-wide authentication gate - which admits any enrolled device key from any account. So
+on shared hosted infrastructure the value read is credential theft, the write is tampering with another
+account's paid service, and the delete disables it. The refusal is gated on the hosted deployment signal
+(`GatewayHostedMode.IsHosted`), applied as one filter over the whole route group, so self-host behavior is
+byte-identical to what is documented here. The routes come back one at a time when the vault is
+partitioned per account.
+
 ## 5. How Directors get keys
 
 **Pull on demand.** When a Director feature needs a key (dictation -> `OPENAI_API_KEY`), it `GET`s it from the Gateway and **caches it in memory only** - never writes it to local disk. On a cache miss (or a provider rejecting the key), it re-fetches.
--- a/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
+++ b/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
@@ -55,8 +55,21 @@ behind only the host-wide authentication gate - which admits any enrolled device
 on shared hosted infrastructure the value read is credential theft, the write is tampering with another
 account's paid service, and the delete disables it. The refusal is gated on the hosted deployment signal
 (`GatewayHostedMode.IsHosted`), applied as one filter over the whole route group, so self-host behavior is
-byte-identical to what is documented here. The routes come back one at a time when the vault is
-partitioned per account.
+byte-identical to what is documented here.
+
+**The un-deny condition, stated so it cannot be ticked off by mistake: THE WRITE IS STOPPED.** The deny
+covers `PUT` and `DELETE` as well as the two reads, so no key material accumulates in the global vault file
+behind this refusal while it is in force. Removing the deny therefore requires only that a per-account key
+namespace exists in the vault - it does NOT additionally require purging a contaminated history, because
+there is no history being written. This is the opposite of a read-only deny such as the hosted stats deny
+(pull request #1888), where the data keeps accruing behind the refusal and lifting it would expose what
+piled up. Un-deny the routes one at a time once the namespace exists.
+
+**This deny does not make key material on the hosted Gateway safe, and must not be read as clearance for
+anything else.** It closes one route group. In particular it says nothing about the voice vault key, which
+stays UNCONFIGURED on hosted until the whole voice chain is partitioned: configuring it does not deliver
+voice, it arms a cross-tenant leak, because clips and the plaintext transcript beside each one land in a
+shared directory until that chain is partitioned. Nothing here changes that.
 
 ## 5. How Directors get keys
 
--- a/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
+++ b/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
@@ -57,13 +57,35 @@ account's paid service, and the delete disables it. The refusal is gated on the 
 (`GatewayHostedMode.IsHosted`), applied as one filter over the whole route group, so self-host behavior is
 byte-identical to what is documented here.
 
-**The un-deny condition, stated so it cannot be ticked off by mistake: THE WRITE IS STOPPED.** The deny
-covers `PUT` and `DELETE` as well as the two reads, so no key material accumulates in the global vault file
-behind this refusal while it is in force. Removing the deny therefore requires only that a per-account key
-namespace exists in the vault - it does NOT additionally require purging a contaminated history, because
-there is no history being written. This is the opposite of a read-only deny such as the hosted stats deny
-(pull request #1888), where the data keeps accruing behind the refusal and lifting it would expose what
-piled up. Un-deny the routes one at a time once the namespace exists.
+**What this deny stops, narrowly: the HTTP ROUTE write and the HTTP ROUTE delete - NOT every write to this
+vault.** An earlier revision of this document claimed "the write is stopped" outright. That was false, and
+the correction is recorded here rather than quietly dropped, because the mistake is the reusable lesson: the
+question asked was "do the DENIED ROUTES write?", and the question that decides an un-deny condition is
+"does ANYTHING write?" plus "what is ALREADY SITTING THERE from before the deny existed?". A deny closes one
+door; it says nothing about the other doors, nor about what accumulated before it was hung.
+
+The other writers to this same global vault, enumerated exhaustively (every `Set`, `SetIfAbsent` and
+`Delete` call in the codebase):
+
+| Writer | Operation | When it fires | Hosted-gated? |
+|---|---|---|---|
+| `GatewayHost.SeedKeyVaultFromEnvironment` | `SetIfAbsent` | Host startup | **No** |
+| `TranscriptionKeyAutoProvisioner.EnsureAsync` | `SetIfAbsent`, `Set` | Every sign-in, and again at startup | **No** |
+| `TranscriptionKeyAutoProvisioner.RevokeMintedKeyAsync` | `Delete` x2 | Every sign-out | **No** |
+
+Key material therefore **still arrives in the global vault while this deny is in force**, by paths that never
+touch the denied routes.
+
+**The un-deny condition, in order, and the purge is REQUIRED:**
+
+1. Tenant-partition every remaining producer in the table above.
+2. **Quarantine, purge or migrate the pre-existing global vault root** - material predating this deny is
+   already in the file, and this change never touched it.
+3. Only then restore a tenant-scoped route, one at a time. **The raw-value `GET` is the last one back.**
+
+Anyone lifting this deny needs the partition **and** the purge. Do not treat step 2 as optional; it is not
+satisfied by any amount of configuration or rollout work, and it cannot be skipped on the grounds that the
+routes were denied.
 
 **This deny does not make key material on the hosted Gateway safe, and must not be read as clearance for
 anything else.** It closes one route group. In particular it says nothing about the voice vault key, which
--- a/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
+++ b/docs/architecture/gateway/GATEWAY_KEY_VAULT.md
@@ -53,9 +53,14 @@ On the always-on Gateway box, in a store **outside git** (same principle as toda
 file at the shared storage root with no tenant in the file, the store, or the routes, and the routes sit
 behind only the host-wide authentication gate - which admits any enrolled device key from any account. So
 on shared hosted infrastructure the value read is credential theft, the write is tampering with another
-account's paid service, and the delete disables it. The refusal is gated on the hosted deployment signal
-(`GatewayHostedMode.IsHosted`), applied as one filter over the whole route group, so self-host behavior is
-byte-identical to what is documented here.
+account's paid service, and the delete disables it. The refusal is expressed through the shared hosted-refusal
+primitive (`HostedRouteDeny.ExclusiveGroup`, the one boundary every deny family on this Gateway adopts): on
+hosted the four handlers are never mapped and one verb-less catch-all refuses everything under `/vault/keys` -
+every request shape, and any route added under the prefix later. The claim that nothing else serves beneath
+`/vault/keys` is startup-validated (`HostedRefusalRouteSpace`), and the hosted decision reads
+`GatewayHostedMode.IsHosted` directly rather than an argument a caller can omit. Off hosted the primitive maps
+the real handlers and creates no refusal at all, so self-host behavior is byte-identical to what is documented
+here.
 
 **What this deny stops, narrowly: the HTTP ROUTE write and the HTTP ROUTE delete - NOT every write to this
 vault.** An earlier revision of this document claimed "the write is stopped" outright. That was false, and
@@ -69,12 +74,15 @@ The other writers to this same global vault, enumerated exhaustively (every `Set
 
 | Writer | Operation | When it fires | Hosted-gated? |
 |---|---|---|---|
-| `GatewayHost.SeedKeyVaultFromEnvironment` | `SetIfAbsent` | Host startup | **No** |
-| `TranscriptionKeyAutoProvisioner.EnsureAsync` | `SetIfAbsent`, `Set` | Every sign-in, and again at startup | **No** |
-| `TranscriptionKeyAutoProvisioner.RevokeMintedKeyAsync` | `Delete` x2 | Every sign-out | **No** |
+| `GatewayHost.SeedKeyVaultFromEnvironment` | `SetIfAbsent` | Host startup | **Yes** - no-ops on hosted |
+| `TranscriptionKeyAutoProvisioner.EnsureAsync` | `SetIfAbsent`, `Set` | Every sign-in, and again at startup | **Yes** - inert on hosted |
+| `TranscriptionKeyAutoProvisioner.RevokeMintedKeyAsync` | `Delete` x2 | Every sign-out | **Yes** - inert on hosted |
 
-Key material therefore **still arrives in the global vault while this deny is in force**, by paths that never
-touch the denied routes.
+All three writers are gated on `GatewayHostedMode.IsHosted` (each reads the deployment signal directly, so it
+cannot fail open by a caller omitting an argument), so **no new key material arrives in the global vault on
+hosted while this deny is in force.** Material that **predates** this deny is a separate concern that the
+un-deny condition below still owns: the deny closes the route door and the writer gates close the write doors,
+but neither touches what was already in the file.
 
 **The un-deny condition, in order, and the purge is REQUIRED:**
 
--- a/src/CcDirector.Gateway.Tests/HostedVaultDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedVaultDenyTests.cs
@@ -1,0 +1,334 @@
+using System.Net;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Text.Json;
+using CcDirector.Core;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// The whole <c>/vault/keys</c> group is DENIED on the hosted Gateway - reads, writes and delete alike.
+///
+/// The store behind those routes is ONE global key vault file at the shared storage root, with no tenant
+/// in the file, the store, or the routes, and the group sits behind only the host-wide authentication
+/// gate - which admits ANY enrolled device key from ANY account. So before this change any hosted
+/// subscriber could read every other subscriber's provider credentials in cleartext (the single-key GET
+/// returns the raw value), overwrite them, and delete them. That is credential THEFT and TAMPERING, not a
+/// disclosure, which is why the whole surface closes and not only the value-returning read.
+///
+/// THE GATE IS ON THE DEPLOYMENT SIGNAL. It reads <see cref="GatewayHostedMode.IsHosted"/> directly, never
+/// an optional boundary or tenant argument. A security branch that depends on an optional argument fails
+/// OPEN the moment a caller omits it.
+///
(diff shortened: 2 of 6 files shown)


thefrederiksen (commented):
REJECT

1. The production filter revert was run only against 10 deny tests plus 4 self-host controls. Remove the real filter after a successful build and run the full 2965-test Gateway suite, recording the intended nine deny reds and all controls/other tests green; then restore, rebuild, and rerun full.

2. The group filter has no future-route probe despite claiming routes added later are denied by default. Map a probe through the vault group/filter seam and prove hosted returns exactly one `error` property while explicit hosted-off reaches the probe.

The direct hosted-mode gate, one group filter, deny-closed/effect assertions, exact response allow-list, explicit hosted-off end-to-end controls for all four routes, route-by-route un-deny table, and restored clean suite are otherwise sound.

thefrederiksen (commented):
Re-audit against the measured 404/SPA masking bar finds an additional blocker at `85f7ab26`.

The self-host controls are genuinely positive—not absence-only: they assert seeded key names/values and verify write/delete through a fresh `KeyVault`. However, both sides parse JSON before asserting media type. `AssertBodyIsNothingButTheRefusal` calls `JsonDocument.Parse` without first requiring `application/json`, and the self-host list/read controls do the same. If a route is absent or a fallback answers HTML/empty content, those tests can redden as parser crashes rather than as statements that the expected handler/refusal did not answer. Add media-type-first assertions before every parse, retain the seeded/effect fingerprints, and add the same-path future-group served/refused probe. The final proof must start from a green complete baseline, mutate one bypassable primitive at a time in the full Gateway suite, and restore the exact head green. Verdict remains **REJECT**.

thefrederiksen:
## Rework pushed - rebased onto current main, three review findings answered

Rebased onto `712ef1ff`, so this branch now carries the machine-wide test lock (#1918), #1916 and v1.6.0. Head is `2ecee448`.

### 1. Parse ordering, fixed in BOTH directions

Every place that parsed a body now asserts **status and media type first**: the hosted refusal helper, the group-probe helper, and every self-host GET control. A 404 here is not necessarily JSON - the single-page-app fallback answers unmatched paths with something else - so a mutation that routed a denied path to the fallback would have made `JsonDocument.Parse` throw. That red is a crash, and a crash proves only that the mutation broke something upstream of the claim; it cannot say what was served in place of the refusal. Now the same mutation reddens as *"expected NotFound, got OK"* or *"expected application/json, got text/plain"*.

### 2. The dual-sided future-route probe

`VaultEndpoints.Map` now returns its `RouteGroupBuilder`, solely so this property is statable from outside the file.

| Direction | Test | Asserts |
|---|---|---|
| Refused on hosted | `HostedVaultGroupFilterTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own` | A brand-new route mapped onto the group is already refused, with no deny written for it anywhere, and the probe sentinel appears nowhere in the body |
| Served on self-host, **both** non-hosted forms | `SelfHostVaultGroupControlTests.A_route_added_to_the_group_still_serves_on_self_host` (absent, and `"0"`) | The **same** probe path returns 200 and its payload |
| Outside-the-group control | `HostedVaultGroupFilterTests.A_route_outside_the_group_still_serves_on_hosted` | A route mapped outside the group still serves on hosted, so the hosted passes are the filter being scoped and not the host refusing everything |

### 3. Handler-positive receipts and destructibility controls

A 404 deny is indistinguishable from a route that was never mapped, so **all four verbs** now carry a self-host receipt asserting the real payload and the real effect on the store - through the group in both non-hosted forms, and through the production `GatewayHost` with the authentication gate on and hosted mode asserted off.

The two hosted survival assertions gained their matching destructibility controls: the identical write **overwrites** the identical seeded key on self-host, and the identical delete **destroys** it. Without those, "the key survived the refusal" is satisfied just as well by an operation that was never capable of destroying anything.

Every class that asserts a mode now sets that mode itself and asserts `GatewayHostedMode.IsHosted` took the value, so no test can silently run in the mode it believes it is not in.

### Self-audit against the five standing rules

| Rule | State |
|---|---|
| 1. One primitive per full-suite run | **Outstanding.** The mutation is a single one - delete the `AddEndpointFilter` block, nothing else - but the run has not been redone yet. The earlier proof on this branch was a filtered run and does not count. |
| 2. A red only counts if it fails with the symptom | **Outstanding as evidence, fixed as mechanism.** The parse-ordering fix is exactly what makes the reds arrive as assertions; the run that records them by name and form is still to come. |
| 3. Format facts before parsing, plus a handler-positive receipt per denied path | **Now compliant.** Both halves done, described above. |
| 4. A survival assertion needs a destructibility control | **Now compliant.** Both survival tests have a same-operation, same-key control proving the operation destroys when the guard is not in the way. |
| 5. Confirm the tests executed, by name | **Outstanding.** Requires the run. |

Rules 1, 2 and 5 all need the same thing: the full-suite revert-proof re-run under the lock. Holding on that - the box is held by a sweep and two pull requests are queued ahead of this one. Build is clean at head (`0 Warning(s), 0 Error(s)`); no Gateway test has been run on this head, and none will be claimed until it has.

Where the #1888 template does not fit: that pull request books its un-deny debt as separate rows because each route's restore condition differs. Here all four routes share one condition - a per-account key namespace in the vault file - so the table in the body keeps four named rows for the four routes but states the same blocker on each, deliberately rather than by omission.

thefrederiksen:
## Why this pull request owes ONE mutation, stated so it cannot be misread

The un-deny condition and the proof-run count are two different axes here, and the two facts sit adjacent in this pull request, which invites exactly the wrong inference. So, in these words:

**The count is one because there is ONE GUARD, not because the four routes share a restore condition.**

The un-deny condition is genuinely shared - one per-account key namespace in the vault file unblocks all four routes - but that is not what sets the count. The count is set by independent bypassability: how many things can be individually wrong while everything else stays correct. Here that number is one.

- There is exactly one `AddEndpointFilter`, on exactly one `MapGroup`, created at exactly one `VaultEndpoints.Map` call site (`GatewayHost.cs`, one call).
- **No route carries a guard of its own.** Delete the filter and all four routes fail together, which is precisely why a single mutation attributes all four reds. Four routes cannot fail independently when none of them has anything of its own to fail.

### What I checked for "something before or outside the filter", since that is the candidate that would change the answer

1. **Nothing vault-specific runs ahead of the filter.** The only thing in front of it is the host-wide authentication middleware, which is shared infrastructure and not part of this deny. `An_unauthenticated_caller_is_still_rejected` pins it and stays green in both legs, which is what a control is for.
2. **The mapping site is singular.** If `VaultEndpoints.Map` were called a second time, that second call could map the same paths without the filter, and that WOULD be a second independently-wrong thing owing a second run. It is called once, and I checked that rather than assumed it.
3. **The write handler's body read is inside the handler, after the filter.** A short-circuited request never deserializes the body, so there is no pre-filter work on any of the four routes.

Nothing in the family is individually bypassable. One mutation plus the restore is the correct and complete count.

### One adjacent finding, deliberately NOT folded into this count

`AiModelsEndpoint` (`/gateway/ai/models`, `/gateway/ai/test-chat`) is handed the same `KeyVault` and reads the key server-side to call DevThrottle. It is **not** a second copy of this disclosure: it never returns the credential - the browser never sees it - and it is a different route family with different semantics, so denying it is not this pull request's guard failing.

But on hosted it does let any enrolled tenant cause the box to SPEND the operator's key. That is a separate surface with its own guard question, and it belongs to a separate pull request with its own mutation and its own run. I am flagging it here rather than quietly widening this one.

thefrederiksen:
REJECT on exact head `ac918164` (static review; no lane consumed).

The deny mechanics are now much better: the unfiltered builder is out of scope in `MapRoutes`, the hosted tests assert status/media type/exact refusal, and the self-host and future-route controls are positive. The un-deny statement is still false, though.

`PUT /gateway/vault/{name}` is not the only production writer to this global `KeyVault`. `GatewayHost.SeedKeyVaultFromEnvironment()` writes `DEVTHROTTLE_API_KEY` with `SetIfAbsent` at startup, and `Account/TranscriptionKeyAutoProvisioner.EnsureAsync()` writes the same vault with both `SetIfAbsent` and `Set`. The global root can also already contain material from before this deny lands. Therefore this change proves only that the denied HTTP PUT/DELETE handlers no longer mutate the vault; it does not prove that “nothing accumulates”, nor that later lifting needs no purge.

Please narrow the sentence to the HTTP route write, and make the un-deny condition explicit: tenant-partition all remaining producers, then quarantine/purge/migrate the pre-existing global vault root before any tenant-scoped GET can be restored. The raw-value GET remains the last route to un-deny.

Proof is also still outstanding. After this is corrected and rebased onto current main, the final exact head needs a complete clean full-suite baseline, the one filter-removal arm with assertion reds and source/build preflight, and an exact restored green. CI is not that revert proof.

thefrederiksen:
STANDING: HELD
07ee0cae91265e1c7c591f33cf910b7e32100371

The implementation and corrected un-deny ledger are statically sound, and CI is green. Clearance remains held on exact-head revert attribution.

The registered proof also needs one correction before the lane is spent: filter attachment and hosted-gate direction are independently wrongable. Run a clean complete baseline, one filter-deletion arm, one hosted-gate inversion arm, and an exact-head restored run. The filter arm must redden the hosted refusal/effect/probe rows while self-host controls stay green; the inversion arm must redden both hosted rows and self-host receipts while the outside-group control stays green. Reconcile complete totals before scoring and classify every red at its intended assertion.

thefrederiksen:
STANDING: REJECTED
07ee0cae91265e1c7c591f33cf910b7e32100371

This supersedes my preceding HELD clearance under the mission-wide ruling now applied to every deny.

The family-wide refusal does not establish the required guarantee for every future request shape, and the current future-route probe certifies only a narrower shape. Move the refusal to the required earlier family boundary, retain inherited future-route coverage, and add the required write-shaped future-route proof. Security details and the exact proof direction are recorded privately.

After that rework, the exact-head attribution must still include separate filter-removal and gate-direction arms as described in the preceding clearance.

thefrederiksen:
STANDING: REJECTED
07ee0cae91265e1c7c591f33cf910b7e32100371

The rework correctly withdraws the false “write stopped / nothing to purge” claim. It now inventories the non-route vault writers, treats the existing global root as contaminated, and requires producer partitioning plus quarantine/purge/migration before un-deny. The route helper also removes the per-route receiver bypass.

The deny boundary itself remains an endpoint-group filter, so it is still post-binding. The future-route canary is a parameterless GET and cannot exercise the property that now decides this class of deny; it must be a body-bound POST with observable binding. The outside-group control is a wholly unrelated path, not a same-namespace neighbour capable of detecting over-refusal.

No Gateway test has run on this head. The clean baseline, independent guard-removal arm, exact restore, final green, and current-main reconciliation are outstanding.

Rework must adopt a cleared shared pre-binding primitive, replace the future canary with the required body-bound shape, add a meaningful hosted neighbour plus self-host positive control, and then run the registered serialized proof on the rebased final head.

PR #1906, opened 2026-07-19
test(gateway): prove context-less database routes are tenant-isolated, not fail-loud

thefrederiksen:
Reworked after review, twice. The first version asserted status codes and called that a positive control, parsed responses before proving they were JSON at all, reverted a FIXTURE rather than a production mechanism, and let a six-route result stand as a sixty-six-route claim. Those are fixed and the narrowed conclusion has been accepted on inspection for the eight cross-probed operations. This revision answers the three gates that remain.

**This pull request is BLOCKED and must not be admitted to the test lane yet.** #1909 lands first, then this branch rebases onto it, takes a fresh baseline there, and runs. That is the only gate. Details under "Convergence and the blocking dependency" below.

---

## What this proves, and what it does not

A census found roughly sixty-six routes taking a path parameter but no `HttpContext`, and INFERRED that the database-backed ones must hit the deny-by-default throw in `AsyncLocalTenantContext.Current` — broken, but not leaking. That inference halved the remaining security work on this mission and had never been executed.

Executed, the inference is wrong in both halves: these routes are **neither broken nor leaking**. They are correctly isolated, by a mechanism the inference did not consider.

**It proves that for eight routes.** Not sixty-six. See "The inventory, and what is still open" below — the honest count of what remains unexamined is stated there rather than buried.

## A defect in the previous version, found while re-deriving the inventory

The earlier version listed `POST /lists/{name}/items` as one of its context-less routes. It is not: its handler is `async (string name, HttpContext ctx)` (`src/CcDirector.Gateway/Api/WorkListEndpoints.cs:64`) — it takes an `HttpContext` to read the request body. It never belonged in the claim. It has been replaced by `DELETE /lists/{name}/items/{source}/{id}` (`WorkListEndpoints.cs:110`), which is genuinely context-less AND destructive, so the write-path probe is now both honest and stronger.

This is why the inventory below is re-derived from the code rather than taken from the census's own list: re-deriving it is the only way to test whether the census's frame was right.

## Positive controls now carry exact seeded fingerprints

Every owner control asserts the exact values this test itself wrote, not a status code:

- work list — the exact list name AND its one seeded item (`source=github`, `id=77`, `area=alpha`)
- cron job — identifier, name, schedule kind, cron expression, time zone, target machine, action seed
- session spend — session identifier, agent kind, billing mode, `tokensCaptured`, `inputTokens=111`, `outputTokens=222`
- workflow — identifier, name, summary, when-to-use, version, and the exact step name and description; the versions list asserts `version=1` and `authoredBy=test-session`

**Status and media type are asserted BEFORE any parse.** Parsing is itself an assertion about format: the single-page-application catch-all serves HTML and the pipeline boundary serves a generic body, so a parse attempted first would throw and launder a wrong response into a crash that proves nothing. Every property access also goes through a typed accessor that asserts presence and value kind, so a missing field fails as a red the test raised rather than as a `KeyNotFoundException`.

## No assertion rests on absence any more

The old work-list write check was `Assert.DoesNotContain("999")` — which passes just as happily on an empty or wrong response. Every denied path now carries a served-side positive fact:

- after tenant A's failed delete, tenant B's list is **re-read** and asserted to hold exactly the seeded item, unchanged
- then **tenant B performs the same delete itself and succeeds**, so tenant A's 404 is a refusal and not an operation that was inert for everybody
- the same shape is applied to the cron job delete and the workflow archive: cross-tenant attempt, exact re-read unchanged, then the owner does it and the effect is re-read again

## The two mechanisms are now separated, not conflated

Binding both fixture device keys to one tenant showed only that the tests react when their identities are collapsed. The outcome actually rests on **three** production mechanisms, and an end-to-end test passes if any one of them is doing all the work. They are now told apart:

1a. **The auth boundary resolves.** A host-wide middleware registered before routing reads the authenticated device key and resolves its tenant (`GatewayHost.cs:1517`).
1b. **And it enters the scope it resolved**, around the whole downstream pipeline (`GatewayHost.cs:1520`, before `UseRouting()` at `:1546`). This is a SEPARATE step: resolution could be perfectly correct and simply never entered, so 1a and 1b cannot share a mutation arm.
2. **The global query filter filters.** `GatewayDbContext.cs:404`, applied to sixteen entity types at `:341-356`.
3. **No scope means no database.** `GatewayDatabase.cs:196` reads `AsyncLocalTenantContext.Current`, which throws by design rather than defaulting to a tenant.

`ContextLessRouteTenancyMechanismProofTests` exercises 2 and 3 **with no HTTP and no middleware at all** — the tenant is entered by the test itself:

- the query filter alone hides tenant B's row from tenant A over one shared database file, and the **same query re-run with the filter ignored finds the row on that very connection**, stamped with tenant B's identifier. That arm is what makes the emptiness a filter result rather than a seed that never landed.
- the deny-by-default throw is caught and pinned by **type and message** (`InvalidOperationException`, containing "No tenant is in scope for this hosted operation" and "deny-by-default"), with the identical call inside a scope as the positive control, carrying exactly the entered tenant into the context.
- the scope primitive is shown to nest and to restore the denial on the way out. This proves the primitive WORKS; it does NOT prove the middleware calls it, and no test that enters the scope itself could. Step 1b is established only causally, by arm 3 below.

Step 1a is probed directly against the running boundary with exact values: key A resolves to `tenant-alice`, key B to `tenant-bob`, and the shared token to nothing.

The 500 is no longer asked to carry causal weight by status alone. It is now a **three-way probe on one route with nothing else varied but the credential** — device key 200, shared token 500 with the exact body `{"error":"internal error"}`, garbage credential 401. The 401 arm is what stops the 500 being read as an authentication artifact. Which internal error produced it is pinned separately and without HTTP, as above.

---

## The inventory, and what is still open

Re-derived from the code, not from the census: the Gateway registers **263** routes; **118** take a path parameter; **66** of those take a path parameter and no `HttpContext`. The census's magnitude was right.

Of those 66, exactly **18 reach a tenant-scoped database store from their own handler**. The other **48 do not** — they reach in-memory registries, the Director command router, and file-backed stores.

| route | store | source | status |
|---|---|---|---|
| `GET /cron/jobs/{id}` | CronJobStore | src/CcDirector.Gateway/Api/CronJobEndpoints.cs:56 | PROVEN |
| `DELETE /cron/jobs/{id}` | CronJobStore | src/CcDirector.Gateway/Api/CronJobEndpoints.cs:91 | PROVEN |
| `GET /cron/jobs/{id}/runs` | CronRunHistoryStore | src/CcDirector.Gateway/Api/CronRunEndpoints.cs:35 | UNPROVEN |
| `GET /gateway/governance/session-spend/{sessionId}` | SessionSpendStore | src/CcDirector.Gateway/Api/GovernanceSpendEndpoints.cs:62 | PROVEN |
| `GET /lists/{name}` | WorkListStore | src/CcDirector.Gateway/Api/WorkListEndpoints.cs:56 | PROVEN |
| `DELETE /lists/{name}/items/{source}/{id}` | WorkListStore | src/CcDirector.Gateway/Api/WorkListEndpoints.cs:110 | PROVEN |
| `DELETE /lists/{name}/consumer` | WorkListStore | src/CcDirector.Gateway/Api/WorkListEndpoints.cs:145 | UNPROVEN |
| `GET /gateway/workflows/{id}` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:57 | PROVEN |
| `POST /gateway/workflows/{id}/publish` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:105 | owner path only, never probed cross-tenant |
| `POST /gateway/workflows/{id}/reset` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:114 | UNPROVEN |
| `DELETE /gateway/workflows/{id}` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:123 | PROVEN |
| `GET /gateway/workflows/{id}/versions` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:131 | PROVEN |
| `GET /gateway/workflows/{id}/versions/{version:int}` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:137 | UNPROVEN |
| `GET /gateway/workflows/{id}/instructions` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:145 | UNPROVEN |
| `POST /gateway/workflows/{id}/enable` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:158 | UNPROVEN |
| `POST /gateway/workflows/{id}/disable` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:163 | UNPROVEN |
| `GET /gateway/workflows/{id}/files/{fileName}` | WorkflowStore | src/CcDirector.Gateway/Api/WorkflowEndpoints.cs:168 | UNPROVEN |
| `GET /gateway/workflow-runs/{id:guid}` | WorkflowRunStore | src/CcDirector.Gateway/Api/WorkflowRunEndpoints.cs:57 | UNPROVEN |

**STILL OPEN, STATED PLAINLY.**

- **Ten of the eighteen database-backed context-less routes are unproven here.** They are named above. They are not covered by inference from the other eight, and nothing in this pull request should be read as retiring them.
- **All forty-eight non-database context-less routes are unproven, and the mechanism proven here cannot be their protection.** The isolation demonstrated in this pull request is the Entity Framework global query filter. A route reaching an in-memory registry, the Director command router, or a file-backed store is not touched by that filter at all, so whatever protects those forty-eight — if anything does — is a different mechanism that this unit did not examine. Among them: `GET /vault/keys/{name}`, `DELETE /vault/keys/{name}`, `GET /directors/{id}/fs/list`, `GET /directors/{id}/handovers/content`, `GET /ingest/recording/{id}/transcript`, `GET /sessions/{sid}/turnbriefs`. This gap has been raised as a finding in its own right and routed to the completeness audit; it is recorded here so that nobody reads this pull request's narrowed claim as the broad one, and it is NOT carried by this unit.
- **The census's framing is worth correcting for whoever picks this up.** It treated "database-backed" as the risky category. On the re-derived numbers the database-backed group is the *smaller* one — eighteen of sixty-six — and it is the group that turns out to be protected. The larger group is the one with no proven mechanism.

---

## Convergence and the blocking dependency

This commit parents `f5bb926f`; current main has already moved to `f2028deb`. The rebase is deliberately NOT done yet, because the base is still moving and two units must land first:

- **#1909 must land first, and it will break this branch on purpose.** It replaces a prose allowlist with a type whose identifier carries a private setter. `ContextLessRouteTenancyMechanismProofTests` currently does `Id = rowId` when it constructs a `WorkListEntity`, and that assignment will stop compiling. That is the guard doing its job: it caught a real caller-supplied key assignment in this unit that no prose rule and no reflection check had flagged. The fix is not to re-widen the setter — it is to construct the row WITHOUT assigning an identifier, capture `row.Id` after construction, and keep every exact-fingerprint assertion unchanged (the fingerprint already includes the name, the consumer token and the tenant identifier, so nothing is weakened by taking a base-minted identifier instead of a caller-chosen one).
**#1937 is NOT a gate on this unit, and an earlier revision of this body wrongly said it was.** That condition assumed #1937 would merge into the gap ahead of this unit; it is in rework and is no longer doing so, and blocking on it would park this unit behind one that is itself being rebuilt.

The rule that makes dropping it safe is general, and does not depend on #1937 at all: **every unit takes a fresh green-and-complete baseline at its OWN head and reconciles against THAT, never against a shared or quoted constant.** The lane lock serialises units, so merges land in the gaps BETWEEN them and never mid-unit. All this unit must guarantee is that its baseline and its five arms come from the same world. Whether #1937 has landed by then is irrelevant: if it lands before this unit's turn, the fresh baseline includes it; if after, it does not. The failure mode the rule exists to prevent — a baseline on one side of a merge and the arms on the other — is prevented by not merging mid-unit, not by waiting.

So: #1909 lands, this branch rebases onto it, takes a fresh baseline on that exact head, reconciles against itself, and runs. Nothing else gates it.

---

## Pre-registered mutations, and the reds predicted for each

One production primitive per full-suite run, never combined. Each is independently bypassable — each can be individually wrong while the others stay correct and isolation still breaks — so overlap does not exempt any of them from its own run. A predicted red that does not appear will be reported as a finding, not explained away. Every predicted red is an assertion the test's own logic raises; any red arriving as a crash will be classified as such and will not count.

The auth-bound mechanism is **two** separately bypassable steps, not one, and they now get separate arms:

- `GatewayHost.cs:1517` — resolving the authenticated device key to the caller's tenant.
- `GatewayHost.cs:1520` — actually entering `EnterScope(resolved)` around the downstream pipeline.

Resolution could be perfectly correct and simply never entered, so a single arm cannot say which is load-bearing. The direct `AsyncLocalTenantContext.Enter` test proves the primitive works; it does NOT prove the middleware calls it, and no test that enters the scope itself ever could. Only arm 3 below establishes that.

**Baseline.** Green and explicitly complete, taken fresh on the final rebased head, or the whole exercise aborts. Computed reconciliation on every arm: passed plus failed under the mutation equals passed under the restored run. Tests confirmed to have executed by name. A run with no summary line is not a result in either direction.

**Arm 2 — constant resolution** (`GatewayHost.cs:1517`): resolve every device key to one constant tenant. A scope is still entered and the query filter is untouched, but the scope no longer carries the CALLER's tenant.

- predicted red (5): `WorkListGetByName_...`, `WorkListDeleteItem_...`, `CronJobGetAndDelete_...`, `SessionSpendGetBySessionId_...`, `WorkflowReadVersionsAndArchive_...` — each at its cross-tenant arm, expecting 404 and receiving 200.
- predicted green: all three tests in `ContextLessRouteTenancyMechanismProofTests`, `TheAuthBoundary_ResolvesEachDeviceKeyToItsOwnTenant_...` (it probes the boundary, not the middleware), and — decisively — `OnOneRoute_TheCredentialAloneDecides...`, whose device-key arm still gets its 200.

**Arm 3 — middleware entry bypass** (`GatewayHost.cs:1520`): remove the `EnterScope(resolved)` call so the `await next()` runs outside any scope. Resolution at :1517 still executes and is still consulted by the branch; the query filter is untouched.

- **distinct signature: device-key requests take the no-scope path and 500 — they do NOT produce the cross-tenant 200s of arm 2.**
- predicted red (6): the five end-to-end tests fail at their SEED (a device-key write expecting 200 or 201 and receiving 500), **plus `OnOneRoute_TheCredentialAloneDecides...`, whose device-key control expects 200 and receives 500.** That last one is the discriminator: it is green under arm 2 and red here, so the two steps are told apart by a test rather than by argument.
- predicted green: `TheAuthBoundary_ResolvesEachDeviceKeyToItsOwnTenant_...` (resolution untouched) and all three mechanism-proof tests (no middleware involved).

**Arm 4 — the global query filter** (`GatewayDbContext.cs:404`): remove the filter. The middleware is untouched and still resolves and enters the correct caller tenant.

- predicted red: the same five cross-tenant arms, **plus** `TheGlobalQueryFilterAlone_HidesTheOtherTenantsRow_...`, plus `TenantScopeGuardTests.EveryTenantScopedEntity_HasTenantIdColumnAndQueryFilter`.
- predicted green: `WithNoTenantScope_...` and `EnteringAScope_...` (both upstream of the filter), `TheAuthBoundary_...`, `OnOneRoute_...`.
- **This is the arm that separates the filter from the middleware:** the isolated query-filter test reddens here and stays green under arms 2 and 3.

**Arm 5 — the deny-by-default** (`AsyncLocalTenantContext.Current`): return the local tenant instead of throwing.

- predicted red: `WithNoTenantScope_OpeningAScopedContextFailsClosed_...`, `EnteringAScope_..._RestoresTheDenialOnTheWayOut`, `OnOneRoute_TheCredentialAloneDecides...` (the shared token stops producing a 500), plus `AsyncLocalTenantContextTests` in the Core test project.
- predicted green: all five cross-tenant arms (a device-key request always had a scope) and `TheGlobalQueryFilterAlone_...`.

**Arm 6 — exact restoration, and restored green.**

**Run count: six** — baseline, constant resolution, middleware entry bypass, query-filter removal, deny-by-default removal, restored green. The count cannot be reduced by design: each arm has a distinct red signature and no two collapse into one another. This count is registered here by this unit, on its own reading of the call sites, and will be re-registered against the rebased head before any run is claimed.

**Admission conditions.** Continuous integration is read green on the rebased head before this unit declares itself ready — it is free, and it is an admission condition, not a formality. The six arms are re-registered against that head before any run is claimed.

**None of these runs has been performed.** The machine's single Gateway test lane is held by another unit and this one is not admitted; the reviewer has also ruled that an exact-head run would not be valid until the corrections above are in. What is claimed today is: the reworked tests, the code-derived inventory, the corrected count, and a clean build with zero warnings and zero errors. No run result is asserted.

Verification only — no production code is changed by this pull request.

### DIFF
--- a/src/CcDirector.Gateway.Tests/ContextLessDatabaseRouteTenancyTests.cs
+++ b/src/CcDirector.Gateway.Tests/ContextLessDatabaseRouteTenancyTests.cs
@@ -1,0 +1,528 @@
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Net.Sockets;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using CcDirector.Gateway;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// VERIFICATION, BY EXECUTION, of a claim about the hosted Gateway's context-less routes.
+///
+/// THE CLAIM UNDER TEST. A census found sixty-six routes that take a path parameter but no
+/// <c>HttpContext</c>, and INFERRED that the database-backed ones must hit the deny-by-default throw in
+/// <c>AsyncLocalTenantContext.Current</c> - broken, but not leaking - because a handler with no
+/// <c>HttpContext</c> cannot resolve the caller's tenant. That inference had never been executed.
+///
+/// WHAT THIS FILE DOES. It seeds real rows for TWO tenants over real HTTP through the real auth middleware
+/// on a real hosted <see cref="GatewayHost"/>, then has tenant A name tenant B's object id on context-less
+/// database-backed routes, and records exactly what comes back.
+///
+/// SCOPE - READ THIS BEFORE QUOTING THE RESULT. This file CROSS-PROBES EIGHT of the sixty-six context-less
+/// routes, across four stores: two work-list operations, two cron operations, one session-spend read, and
+/// three workflow operations. A ninth context-less route, POST /gateway/workflows/{id}/publish, is exercised
+/// on the OWNER path only, as a seed, and is NOT cross-probed - it is not part of the claim.
+///
+/// It does NOT retire the other fifty-eight by inference; that was the census's error and repeating it here
+/// would be the same mistake in the opposite direction. The full code-derived inventory, and exactly what
+/// remains unproven, is stated in the pull request. Eight samples cannot stand for sixty-six.
+///
+/// HOW A POSITIVE CONTROL IS BUILT HERE. Asserting a status code does not prove the intended handler
+/// returned the seeded row - an unrelated 200, a catch-all, or a failed seed answering empty would all pass.
+/// So every owner control asserts the exact SEEDED FINGERPRINT (the values this test itself wrote), and
+/// every write or delete is judged by an independent RE-READ of the resulting state, never by the absence of
+/// a substring. Status and media type are asserted BEFORE any parse, because parsing is itself an assertion
+/// about format: parse-first would turn a wrong response into a parser crash, which proves nothing.
+///
+/// WHAT EXECUTION FOUND. Nothing threw and nothing leaked. Every route answered its owner with the exact
+/// seeded row and answered the other tenant with an ordinary not-found; tenant B's rows survived tenant A's
+/// delete attempts byte-for-byte, and tenant B could then perform those same deletes itself - so the refusals
+/// are refusals, not inert operations. The claim's conclusion about the database half is wrong in BOTH
+/// halves: these routes are neither broken nor leaking.
+///
+/// THE MECHANISM, AND WHY IT IS PROVEN ELSEWHERE. The tenant scope is not entered by the handler, so a
+/// handler's signature cannot decide whether one exists. It is entered by a host-wide middleware registered
+/// before routing that binds the AUTHENTICATED device key's tenant for the whole pipeline
+/// (<c>GatewayHost.cs</c>, the device-key HTTP boundary). A context-less handler therefore runs inside the
+/// caller's scope, and the entity global query filter answers about the caller's tenant only.
+///
+/// Those are TWO SEPARATE production mechanisms, and an end-to-end test like this one cannot tell them
+/// apart - it would pass if either did all the work. They are therefore isolated from each other, without
+/// HTTP and without the middleware, in <see cref="ContextLessRouteTenancyMechanismProofTests"/>, which is
+/// the companion to this file and is where the causal claims are actually earned.
+/// </summary>
+public sealed class ContextLessDatabaseRouteTenancyTests : IAsyncLifetime
+{
+    private const string SharedToken = "test-token";
+    private const string TenantA = "tenant-alice";
+    private const string TenantB = "tenant-bob";
+
+    private readonly ITestOutputHelper _out;
+    private GatewayHost _gateway = null!;
+    private HttpClient _http = null!;
+    private string _keyA = "";
+    private string _keyB = "";
+
+    private readonly string _instancesDir =
+        Path.Combine(Path.GetTempPath(), "cc-ctxless-" + Guid.NewGuid().ToString("N"));
+    private string? _priorHosted;
+
+    public ContextLessDatabaseRouteTenancyTests(ITestOutputHelper output) => _out = output;
+
+    public async Task InitializeAsync()
+    {
+        _priorHosted = Environment.GetEnvironmentVariable("CC_GATEWAY_HOSTED");
+        Environment.SetEnvironmentVariable("CC_GATEWAY_HOSTED", "1");
+
+        _gateway = new GatewayHost(port: FreePort(), token: SharedToken, authEnabled: true,
+            instancesDirectory: _instancesDir,
+            workListsPath: Path.Combine(_instancesDir, "worklists", "worklists.json"),
+            cronJobsPath: Path.Combine(_instancesDir, "cron", "cronjobs.json"),
+            snoozePath: Path.Combine(_instancesDir, "snooze", "snooze.json"),
+            missionsPath: Path.Combine(_instancesDir, "missions", "missions.json"),
+            streamMode: true);
+        await _gateway.StartAsync();
+        _http = new HttpClient { BaseAddress = new Uri($"http://127.0.0.1:{_gateway.Port}/") };
+
+        _keyA = _gateway.Devices.Register("dev-a", "MA").DeviceKey;
+        _keyB = _gateway.Devices.Register("dev-b", "MB").DeviceKey;
+        _gateway.Devices.SetAccountBinding("dev-a", "sub-alice", TenantA);
+        _gateway.Devices.SetAccountBinding("dev-b", "sub-bob", TenantB);
+
+        Assert.True(_gateway.TenantBoundary.IsHosted, "The harness must be running the HOSTED tenant boundary.");
+    }
+
+    public async Task DisposeAsync()
+    {
+        _http.Dispose();
+        await _gateway.StopAsync();
+        Environment.SetEnvironmentVariable("CC_GATEWAY_HOSTED", _priorHosted);
+        try { if (Directory.Exists(_instancesDir)) Directory.Delete(_instancesDir, true); }
+        catch { /* best-effort */ }
+    }
+
+    // ================================================================= work lists (worklists table)
+
+    /// <summary>
+    /// GET /lists/{name} - context-less (WorkListEndpoints.cs:56, handler <c>(string name)</c>).
+    /// The owner control asserts the exact seeded list AND its exact seeded item, so an empty or unrelated
+    /// 200 cannot pass for a served row.
+    /// </summary>
+    [Fact]
+    public async Task WorkListGetByName_ServesTheOwnerTheExactSeededList_AndIsAPlainNotFoundForTheOtherTenant()
+    {
+        const string list = "bobs-backlog";
+
+        var seed = await Json(await Send("POST", "lists", _keyB, "{\"name\":\"" + list + "\"}"),
+            HttpStatusCode.OK, "SEED    POST /lists (tenant B)");
+        Assert.Equal(list, Str(seed, "name"));
+
+        var seedItem = await Json(
+            await Send("POST", $"lists/{list}/items", _keyB,
+                "{\"source\":\"github\",\"id\":\"77\",\"area\":\"alpha\"}"),
+            HttpStatusCode.OK, "SEED    POST /lists/{name}/items (tenant B)");
+        Assert.Equal(list, Str(seedItem, "name"));
+
+        // CONTROL: the exact seeded fingerprint, not merely a 200.
+        var control = await Json(await Send("GET", $"lists/{list}", _keyB, null),
+            HttpStatusCode.OK, "CONTROL GET /lists/{name} (owner B)");
+        AssertIsBobsBacklogWithTheOneSeededItem(control, list);
+
+        // CROSS: tenant A naming tenant B's list.
+        var cross = await Json(await Send("GET", $"lists/{list}", _keyA, null),
+            HttpStatusCode.NotFound, "CROSS   GET /lists/{name} (tenant A naming B's list)");
+        Assert.Equal("no such list", Str(cross, "error"));
+        Assert.Equal(list, Str(cross, "name"));
+    }
+
+    /// <summary>
+    /// DELETE /lists/{name}/items/{source}/{id} - context-less (WorkListEndpoints.cs:110, handler
+    /// <c>(string name, string source, string id)</c>). A destructive context-less route.
+    ///
+    /// The refusal is judged by an independent RE-READ of tenant B's exact list state, never by the absence
+    /// of a substring in tenant A's response - absence can pass on an empty or wrong body. And the owner then
+    /// performs the SAME delete successfully, which is what makes tenant A's 404 a refusal rather than an
+    /// operation that was inert for everyone.
+    /// </summary>
+    [Fact]
+    public async Task WorkListDeleteItem_ContextLess_LeavesTheOtherTenantsListExactlyAsItWas_AndTheOwnerCanStillDeleteIt()
+    {
+        const string list = "bobs-delete-target";
+
+        Assert.Equal(list, Str(await Json(await Send("POST", "lists", _keyB, "{\"name\":\"" + list + "\"}"),
+            HttpStatusCode.OK, "SEED    POST /lists (tenant B)"), "name"));
+        await Json(await Send("POST", $"lists/{list}/items", _keyB,
+                "{\"source\":\"github\",\"id\":\"77\",\"area\":\"alpha\"}"),
+            HttpStatusCode.OK, "SEED    POST /lists/{name}/items (tenant B)");
+
+        // CROSS: tenant A tries to delete tenant B's item by name, source and id.
+        var cross = await Json(await Send("DELETE", $"lists/{list}/items/github/77", _keyA, null),
+            HttpStatusCode.NotFound, "CROSS   DELETE /lists/{name}/items/{source}/{id} (tenant A)");
+        Assert.Equal("no such list", Str(cross, "error"));
+
+        // INDEPENDENT RE-READ: B's list must still hold exactly the seeded item, unchanged.
+        var after = await Json(await Send("GET", $"lists/{list}", _keyB, null),
+            HttpStatusCode.OK, "AFTER   GET /lists/{name} (owner B, after A's delete attempt)");
+        AssertIsBobsBacklogWithTheOneSeededItem(after, list);
+
+        // CONTROL: the owner performs the SAME operation - so the refusal above is a refusal, not inertia.
+        var ownerDelete = await Json(await Send("DELETE", $"lists/{list}/items/github/77", _keyB, null),
+            HttpStatusCode.OK, "CONTROL DELETE /lists/{name}/items/{source}/{id} (owner B)");
+        Assert.Equal(list, Str(ownerDelete, "name"));
+        Assert.Equal("github", Str(ownerDelete, "source"));
+        Assert.Equal("77", Str(ownerDelete, "id"));
+        Assert.True(Bool(ownerDelete, "removed"), "the owner's own delete reported nothing removed");
+
+        // And the effect of the owner's delete is independently re-read.
+        var emptied = await Json(await Send("GET", $"lists/{list}", _keyB, null),
+            HttpStatusCode.OK, "AFTER   GET /lists/{name} (owner B, after its OWN delete)");
+        Assert.Equal(list, Str(emptied, "name"));
+        Assert.Empty(Arr(emptied, "items").EnumerateArray());
+    }
+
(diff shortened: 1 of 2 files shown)


thefrederiksen:
REJECTED pending rework.

The execution is useful, but the current positive controls do not yet distinguish ?the intended handler returned the seeded row? from an unrelated 200/catch-all, and the submitted mutation does not prove the production mechanism claimed.

Most owner controls assert status only. Concretely: work-list GET never asserts the seeded name/items; cron GET and post-delete survival never assert the seeded job; session-spend never asserts the seeded token values; workflow GET and post-delete survival never assert the seeded workflow/version state. The work-list write check is only `DoesNotContain("999")`, so absence can pass on an empty or wrong response. JSON is parsed without first asserting the expected media type. For each exact path+verb, assert status, media type, and an exact handler-specific seeded fingerprint. For writes/deletes, assert the owner control can perform the same operation and then exact reread state/effect; the cross-tenant attempt must leave that exact state unchanged. Parse only after format assertions so a wrong response becomes an assertion failure, not a parser crash.

Binding both fixture device keys to one tenant is not a revert of the protection. It shows the tests react when their identities are deliberately collapsed, but it does not independently mutation-prove either production conjunct in the stated mechanism: (1) auth-bound middleware resolves/enters the caller tenant before routing, and (2) the EF global query filter applies that tenant. Mutate each independently bypassable production primitive in its own full Gateway run while holding the other fixed, and require named assertion-only reds. The shared-token `500` is likewise not causal by status alone; use an exact causal fingerprint or a direct middleware/scope probe so any unrelated internal error cannot satisfy it.

The branch merge-bases at old main `4fe2c3bb`. Rebase to current main and run the serialized proof protocol on the exact final head: green and complete full Gateway baseline, one production mutation per run, exact restoration, restored green, and current CI. Finally, constrain the conclusion to the six exercised operations unless the roughly 66-route census is supplied as an exact inventory mapping every claimed database-backed route/store to this same executed protection. Six samples cannot retire the remaining inventory by inference?the premise this PR is intended to replace.

thefrederiksen:
STATIC RE-REVIEW of `7dff5fee` ? the narrowed behavioral claim and positive controls now meet the inspection bar, but the production proof and convergence do not yet.

The original response-mask defects are fixed: owner controls assert exact seeded fingerprints; destructive paths use exact rereads plus owner destructibility; status and media type precede parsing; the generic 500 is no longer asked to identify its own cause; and the PR now names the ten database routes and forty-eight non-database routes that remain unproven. I accept the narrowed conclusion for the eight actually cross-probed operations, subject to the final mutations.

Remaining gates:

1. The claimed auth-bound mechanism still combines two independently bypassable primitives: resolving the authenticated key to the caller tenant, and actually entering that resolved scope around the downstream pipeline. The pre-registered constant-resolution mutation proves the first callsite. The direct `AsyncLocalTenantContext.Enter` test proves the primitive in isolation, but it does not prove the production middleware calls it. Add a separate full-suite mutation that removes/bypasses the middleware `EnterScope(resolved)` while leaving resolution and the query filter untouched. Its distinct expected signature is device-key requests taking the no-scope/500 path (including the device-key credential control), not the cross-tenant 200 signature of constant resolution. Under the established conjunction rule, resolution and entry cannot share one mutation arm.

2. This commit parents `f5bb926f`, while current main is already `f2028deb`; #1937 is also required to merge before lane admission. Rebase after that convergence point. More importantly, this test currently assigns `Id = rowId` on `WorkListEntity`. #1909?s accepted structural rework makes that setter private, so the two PRs cannot coexist unchanged. Land #1909 first, then rebase this branch, construct the row without assigning `Id`, capture `row.Id` after construction, and keep the exact fingerprint assertions.

3. Correct the source summary?s ?exercises SEVEN? statement. The code and PR title/body count eight cross-probed operations: two work-list, two cron, one session-spend, and three workflow operations. The owner-only publish seed is correctly listed as not cross-probed.

After those changes, final evidence is six serialized full Gateway suite units on one exact head: green complete baseline; constant-tenant resolution; middleware entry bypass; query-filter removal; deny-by-default removal; exact restoration and restored green. Classify assertion reds vs crashes, reconcile each arm, and confirm current CI. This remains rejected until that admitted evidence is posted.

thefrederiksen:
All three gates accepted. Head is now `01b3bcab`. Still rejected, still not admitted — correctly, and I am not asking for admission yet.

**1 — resolution and entry now get separate arms.** You are right, and it is the same error as the first round in a new place: my mechanism file entered the scope itself and I let that stand for the middleware entering it. No test that enters the scope can prove the middleware does. The call sites are separate and separately bypassable — `GatewayHost.cs:1517` resolves, `:1520` enters `EnterScope(resolved)` — so they now have separate arms, and the source comment says outright that the primitive test proves the primitive WORKS and not that production CALLS it.

The new arm removes only the `EnterScope(resolved)` call, leaving resolution executing and consulted and the query filter untouched. Its predicted signature is the one you named, and it is distinct from constant resolution: device-key requests take the no-scope path and 500 rather than producing cross-tenant 200s. The discriminator is `OnOneRoute_TheCredentialAloneDecidesWhetherATenantScopeExists`, whose device-key control expects 200 — **green under constant resolution, red under entry bypass**. The two steps are therefore told apart by a test rather than by argument. Six predicted reds for that arm, five of them the end-to-end tests failing at their device-key SEED.

**2 — convergence.** Not rebased, deliberately. #1909 lands first, then #1937, then this branch rebases onto that convergence point and takes a fresh green-and-complete baseline there, reconciled against itself and not against any count carried over from this head.

On `Id = rowId`: agreed, and I am not going to work around it. That is a caller-supplied key assignment sitting in a tenancy proof, which is precisely the thing this mission exists to stop, and no prose rule or reflection check had flagged it — your structural rework catching it on me before either unit reached main is the guard working. The fix is to construct the row without assigning an identifier and capture `row.Id` afterwards. The exact-fingerprint assertions survive untouched, because the fingerprint is the name, the consumer token and the tenant identifier; taking a base-minted identifier instead of a caller-chosen one weakens nothing. I have not pre-emptively made that edit, because the setter is still public on this base and dropping the assignment now would silently leave an empty identifier rather than a minted one.

**3 — the count.** Fixed precisely. The source said SEVEN; it is EIGHT, and it now states the breakdown you gave — two work-list, two cron, one session-spend, three workflow — and records that `POST /gateway/workflows/{id}/publish` is exercised on the owner path only and is not part of the claim. You are right that a count disagreeing with itself in a document about counts is exactly what a later reader cites.

**Run count.** I register six on my own reading of the call sites, matching yours: baseline, constant resolution, middleware entry bypass, query-filter removal, deny-by-default removal, restored green. It cannot be reduced by design — each arm has a distinct red signature and no two collapse into one another. I will re-register it against the rebased head before claiming any run.

Nothing is claimed today but the reworked tests, the corrected count, the code-derived inventory and a clean build with zero warnings and zero errors. No run has been performed and no result is asserted.

One thing I want left standing in the record and not buried in my narrowing: of the sixty-six context-less routes, only eighteen reach a tenant-scoped database store. The other forty-eight reach registries, the command router and file-backed stores, which the query filter proven here does not touch at all — so they were counted protected by a mechanism that never reached them. That is raised as a finding in its own right and routed to the completeness audit; this unit does not carry it, and this pull request's narrowed claim must not be read as the broad one.

thefrederiksen:
Correction to my previous comment, on sequencing only. No change to the proof, the arms or the head, which is still `01b3bcab`.

I wrote that this unit waits for #1909 and then #1937. **That was wrong: only #1909 gates this unit.** The #1937 condition assumed it would merge into the gap ahead of this unit. It is in rework, it is no longer merging into that gap, and blocking on it would park this unit behind one that is itself being rebuilt.

The rule that makes dropping it safe is general and does not depend on #1937: **every unit takes a fresh green-and-complete baseline at its OWN head and reconciles against THAT, never against a shared or quoted constant.** The lane lock serialises units, so merges land in the gaps BETWEEN units and never mid-unit. All this unit has to guarantee is that its baseline and its five arms come from the same world. If #1937 lands before this unit's turn, the fresh baseline includes it; if after, it does not — either way the reconciliation is internally consistent. The failure this condition was protecting against, a baseline on one side of a merge and the arms on the other, is prevented by not merging mid-unit rather than by waiting.

So the sequence is: #1909 lands, this branch rebases onto it, takes a fresh baseline on that exact head, re-registers the six arms there, and runs.

Adding one admission condition I did not previously state: **continuous integration is read green on the rebased head before this unit declares itself ready.** It is free and it is a condition, not a formality.

The body is updated accordingly. Still rejected, still not admitted, still blocked on #1909 alone.

thefrederiksen:
STANDING: HELD
01b3bcab3d1593f72ab7eec94eea49674b07474a

The narrowed eight-operation claim, exact owner/cross-tenant fingerprints, mechanism split, and explicit remaining inventory still meet the static bar. This head is not a clearance candidate yet by its own correct admission rule.

REASON-TO-TRUST CHECK: this proof is meant to be trusted because each independently bypassable production conjunct has a distinct causal mutation signature—not because eight ordinary requests pass. That justifying property is registered on this branch but has not been executed, so it cannot yet be cited as evidence.

#1909 remains open, and this branch still contains the caller-assigned identifier that #1909 intentionally makes unrepresentable. After #1909 lands, rebase onto the then-current main, construct that row through the new invariant and capture its minted identifier, read CI green, and re-register the six units against the new exact head before running them. No evidence from this head carries across that convergence point.

The hold is therefore the declared convergence dependency plus the entirely unexecuted exact-head production attribution, not a new defect in the narrowed test claim.

PR #1907, opened 2026-07-19
fix(gateway): deny the recording ingest and cached wingman voice surfaces on hosted

thefrederiksen:
Two live cross-tenant content leaks on the hosted gateway are closed by refusal. Self-host keeps both features entirely and is proven unchanged.

Rebased onto `origin/main` at `f2028deb`. The rebase was verified with `git range-diff`, not a rebuild: a rebuild proves it compiles, only a range-diff proves nothing crept in. Three of the four pre-existing commits came across byte-identical; the fourth differs in exactly two places, both intended and both forced by what landed on main in the meantime - see "What the rebase changed" below.

## What was wrong

Neither surface carries a tenant anywhere.

**Recording ingest.** `RecordingEndpoints` production-constructs `RecordingIngestService`, and the service builds a recording's durable directory from the CALLER-SUPPLIED recording id alone. Neither file mentions a tenant. So on the hosted gateway any authenticated device could enumerate every recording on the box, read any raw audio and any raw or cleaned transcript, overwrite chunks, titles, subtitles and summaries, promote another account's recording into a shared vault, or delete it. The id sanitiser only replaces invalid characters, so two distinct caller ids can also alias onto one directory. The same file's dictation dictionary routes read and write one shared glossary file with no tenant.

**Cached wingman voice.** `GatewayWingmanVoiceEndpoint` enumerates every ready session id, returns the spoken text and the agent's reply text and the mp3 by BARE session id, and deletes by bare session id.

The voice half needs restating after the per-tenant voice partition landed on main, because the obvious reading of that change is that this deny is now unnecessary, and it is not. `WingmanVoiceService` does now keep a per-tenant bucket and a per-tenant directory. But that change deliberately left these routes unconverted, in its own words: *"The voice routes and the background narration sweep are deliberately NOT converted here. They still pass the local tenant."* Every one of the four denied routes reads the state through the hardcoded constant `TenantId.Local`. **A partitioned store read through a constant is not a partitioned read** - it points every hosted caller at one shared partition, which is the same disclosure by a different mechanism.

## What the rebase changed

Only two hunks in the deny commit, and neither is new content:

1. The voice routes now call `voice.Get(TenantId.Local, sid)`, `ReadySessionIds(TenantId.Local)` and so on, because the partition change made the tenant a required first parameter. This is main's signature flowing through the group conversion.
2. The deny comment was rewritten to state the post-partition truth above, replacing the stale sentence "partitioning the voice state is separate, deferred work". The state is partitioned; the ROUTES are not. Leaving that sentence would have been a false statement pointing the next reader at work that is already done.

## Defects the rebase exposed, and the design change it prompted

**The rig was seeding a location nothing reads.** The cached-voice fixtures were written to the pre-partition path (`voice-sessions.json` and `voice-audio` directly under the storage root). Partitioning made that the LEGACY path, and the service now treats it as legacy at construction: on hosted it DELETES it, on self-host it MOVES it into the local partition. So:

- the hosted suite would have been proving a refusal in front of an EMPTY cache - absence proved twice, presence never - and would have stayed green regardless, because the deny answers before the handler ever looks;
- the self-host control would have passed by way of the migration, exercising a path its name does not mention.

The rig now seeds the partition directly, and asserts its path against the production `PartitionDirectoryFor` rather than against a second copy of the same arithmetic.

**The per-route bypass is now structurally impossible.** Both files kept the filtered and unfiltered builders in one scope. That makes every route-mapping line independently bypassable: one word - `outer` for `app`, or `app` for `cachedVoice` - opens exactly one route on hosted, compiles, and leaves every other route and every existing test green. Nineteen lines, nineteen chances, and nothing in the suite would show it.

Each family's routes now live in a method that receives ONLY the filtered group (`RecordingEndpoints.MapRoutes`, `GatewayWingmanVoiceEndpoint.MapCachedVoiceRoutes`). The unfiltered builder is not a parameter and is not in scope, so the bypass does not compile. The voice file needed this most, and it is exactly the coexistence the review flagged: most of that file is deliberately not denied, so `cachedVoice.MapGet` and `app.MapGet` genuinely sit a few lines apart and differ only in whether the route is reachable on hosted.

This is what takes the mutation plan from twenty-three runs to four, and the reduction is by construction rather than by argument.

## The shape of the deny

- ONE endpoint-group filter per surface, not a guard line per handler, so a route added next year is refused by default. The empty prefix keeps every route path written out in full, so the self-host surface is byte-identical.
- Gated on `GatewayHostedMode.IsHosted`, the INDEPENDENT deployment signal, never on an optional tenant or boundary argument. A security branch resting on an optional argument fails OPEN the moment a caller omits it.
- 404 rather than 403: on hosted these surfaces do not exist as a concept, and no credential can reach them.
- The refusal body is EXACTLY one `error` property, asserted by enumerating the property set as an allow-list, so any field added later reddens automatically.
- Refuses; never serves an empty list. An empty recording list is a FALSE statement about the box where an absent route is merely absent.

## The un-deny debt - and the two halves are NOT the same kind

A deny that stops only the READ is a DEFERRED LEAK: cross-tenant data keeps accumulating behind it, and the day the deny is lifted it exposes a contaminated history. So the entry is written per surface, now, in the pull request and at the guard itself.

### The frame: two questions, not one

Assessing the write-stop by looking at THE DENIED ROUTES is the wrong question, and it produced one unsupported claim per surface in the previous revision. A deny closes ONE door. It says nothing about the other doors, and nothing about what was already sitting behind it before it was hung. Each entry below answers two:

- **Q1** - does anything, ANYWHERE, still write this state on hosted?
- **Q2** - what is ALREADY sitting there, from before the deny existed?

### Recording ingest

**Q1 - no new writes.** `RecordingIngestService` is constructed in exactly one place in production (`BuildService`), reached only from a `Lazy` whose every `.Value` use is inside a handler mapped on the filtered group. No startup seed, no provisioner, no background job and no other endpoint constructs it. On hosted the filter answers before any handler runs, so the `Lazy` is never forced: the service is never built and its transcription worker never starts.

**Q2 - unknown, therefore assumed contaminated.** The previous revision argued straight from Q1 to "there is nothing to purge - the store is exactly as the deny found it". **That does not follow**, and it is withdrawn. What the deny FOUND is precisely what was never established. The recording root is ONE un-partitioned directory keyed on caller-supplied ids, it predates this refusal, and nothing in it records an owner - so anything written before the deny landed is still there, un-owned and un-attributable after the fact. "No new writes" is a statement about the future, not evidence about the past.

**Un-deny condition:** give recordings a per-tenant layout (keyed by tenant, not by caller-supplied id alone, which also closes the id-aliasing hazard), AND explicitly quarantine, purge or migrate the pre-deny root rather than adopting it - adopting it would silently assign un-owned material to whichever tenant happened to be inferred - and only then lift the refusal.

### Cached wingman voice

**Q1 - the writers are LATENT, not active. The previous revision's ongoing-write story was false and is withdrawn.** It claimed `voice-turn`, `explain`, `voice-mode/all` and the narration sweep keep filling the local partition on hosted. This code does not do that. Each of the four is gated behind a Local-scoped lookup that is empty on hosted and fails closed:

| Producer | Gate that stops it on hosted |
| --- | --- |
| `POST /sessions/{sid}/wingman/voice-turn` | `SessionVerbClient.ResolveAsync` locates with `TenantId.Local`, returns null; the route answers 404 before `OnSessionWorking`/`StoreSpokenAsync` |
| `POST /sessions/{sid}/wingman/explain` | same resolve, same 404, before `Mark`/`StoreSpokenAsync` |
| `POST /sessions/voice-mode/all` | targets come from `FleetByDirector(..., TenantId.Local)`; empty Local fleet = empty target list, no `Mark` |
| the narration sweep in `GatewayHost` | iterates `VoiceSessionIds(TenantId.Local)`; the turn-end path first tests `IsVoiceSession(TenantId.Local, sid)`. Both empty or false |

The error was reading call sites and concluding the loop runs - a call site proves the code EXISTS, not that it RUNS. The production comment on the sweep says it directly: *"on hosted a Local read is empty (never a wrong-tenant read)"*.

These writers are still wrong and must still be removed before un-deny, because converting the routes or otherwise populating the Local scope ARMS them. "Currently unreachable" is one merge away from reachable, and nothing would announce it.

**Q2 - this is the risk the deny actually closes, and it is a READ risk.** A pre-existing Local partition whose contents these routes hand to any hosted caller. It gets there with no hosted write at all: a box that ran self-host earlier, a build from before the routes were scoped, or migrated state. `WingmanVoiceService`'s constructor deleting the PRE-PARTITION legacy files on hosted is easy to misread as "the box starts clean" - it does not touch `tenants/local/`, and `tenants/local/` is exactly what these four routes read.

**Un-deny condition (unchanged - the conservative one is kept, only its justification is corrected):** convert the four latent writers off the hardcoded Local, AND dispose of the existing Local pile (quarantine, purge or attribute), and only then lift the refusal. Converting the routes alone is not sufficient: each account would read its own partition while the un-owned Local pile stayed on disk and the writers stayed aimed at it.

Neither entry is filed as "blocked on partitioning". The voice state IS already partitioned per tenant and this exposure survived it untouched, so that phrase would be ticked off in good faith by the next partitioning change while the pile stayed on disk.

### Route-by-route

| Route | Method | Tier | What it does on self-host | Before the deny is lifted |
| --- | --- | --- | --- | --- |
| `/ingest/recordings` | GET | 1 | Lists every recording on the gateway | Recordings owned by a tenant; the list filtered to the caller's |
| `/wingman/voice/ready` | GET | 1 | Lists every session with a ready clip | Writers converted, local pile purged, then filtered to the caller's |
| `/ingest/recording/{id}` | DELETE | 2 | Deletes the transient local recording | Ownership check before deleting |
| `/ingest/recording/{id}/chunk/{index}` | PUT | 2 | Stores one audio segment | Recording directory keyed by tenant; id aliasing closed |
| `/ingest/recording/{id}/meta` | PATCH, POST | 2 | Sets title, subtitle, summary | Ownership check before writing |
| `/ingest/recording/{id}/promote` | POST | 2 | Copies transcript and audio into the vault | A per-tenant vault target |
| `/ingest/recording` | POST | 2 | Registers a new recording | Recording directory keyed by tenant |
| `/ingest/recording/{id}/complete` | POST | 2 | Queues transcription | Ownership check plus per-tenant spend accounting |
| `/ingest/dictionary` | PUT | 2 | Replaces the whole shared glossary | A per-tenant glossary |
| `/ingest/dictionary/terms` | POST | 2 | Adds terms to the shared glossary | A per-tenant glossary |
| `/sessions/{sid}/wingman/voice/stop` | POST | 2 | Turns voice off and deletes the cached clip | Writers converted, local pile purged |
| `/ingest/recording/{id}/status` | GET | 3 | Reads one recording's status | Ownership check |
| `/ingest/recording/{id}/transcript` | GET | 3 | Reads the cleaned transcript | Ownership check |
| `/ingest/recording/{id}/audio/{index}` | GET | 3 | Streams one raw audio segment | Ownership check |
| `/ingest/dictionary` | GET | 3 | Reads the shared glossary | A per-tenant glossary |
| `/ingest/agent-info` | GET | 3 | An API guide naming the tailnet front door | Denied because it documents the routes above and names the owner's address |
| `/sessions/{sid}/wingman/voice` | GET | 3 | Returns spoken summary and agent reply text | Writers converted, local pile purged |
| `/sessions/{sid}/wingman/voice/audio` | GET | 3 | Streams the mp3 | Writers converted, local pile purged |

NOT denied, deliberately: the wingman generation and speech routes (`/wingman/tts`, `/wingman/transcribe`, `/wingman/utterance/*`, `/sessions/{sid}/wingman/voice-turn`, `/wingman/ask-direct`, `/wingman/ask-devthrottle`, `/sessions/{sid}/wingman/explain`, `/sessions/{sid}/wingman/menu`, `/sessions/voice-mode/all`). That is the judgement, and it is also precisely why the voice write is not stopped - those routes are the writers.

## Coverage

**Every denied route, both directions, from ONE table.** `RecordingAndVoiceDenyRig.DeniedRoutes` holds all nineteen routes with the verb, the body each needs, the refusal its group answers with, and the fingerprint its own handler produces. The hosted suite requires a refusal on every row; `SelfHostRecordingAndVoiceStillWorkTests.Every_denied_route_resolves_to_its_own_handler` requires every row to answer with its own fingerprint. A route added to a group and to that table is checked both ways automatically, so the suites cannot drift.

Handler-positive proof for the four routes the review named as uncovered:

| Route | Proof |
| --- | --- |
| `GET /ingest/recording/{id}/audio/{index}` | The owner uploads a segment and reads THOSE EXACT BYTES back, compared byte for byte, plus a second test against the seeded clip's bytes |
| `POST /ingest/recording/{id}/complete` | The completeness gate's own 409 receipt: `state` is `incomplete` and `missingOrBadIndices` is exactly `[0]`. Nothing is transcribed, so it needs no key and is deterministic |
| `POST /ingest/recording/{id}/promote` | The ingest service's own `is not ready to promote`, naming the recording id, with no vault write required |
| `POST /ingest/recording/{id}/meta` | Both verbs write, then the change is read back through DIFFERENT routes (the list and the status route) - a state change on the box, not the handler quoting its input |

**No status-shaped assertions.** On this Gateway "a missing route answers 404, a wrong verb answers 405" is false in two separately measured ways: `CockpitReactApp` maps `MapFallback("{*path}")` whose not-found page is release-build-only, and `SessionWsProxyEndpoints` maps an all-verb `/sessions/{sid}/{**rest}` catch-all that the cached-voice routes sit under. No 405 is produced on these paths in either condition. Every row matches a per-handler fingerprint instead.

**Future-route probe on BOTH groups.** `RecordingEndpoints.Map` and `GatewayWingmanVoiceEndpoint.Map` return the group they filtered; production callers ignore it. `HostedDenyGroupFutureRouteProbeTests` maps a brand-new route onto each group - written after the filter, carrying no guard, exactly like a future edit - and asserts hosted returns 404 with a body of exactly one `error` property and no probe fingerprint, while hosted-off serves 200 with the fingerprint. It runs on a bare host with neither the single-page-app fallback nor the session catch-all mapped, so nothing else can answer for the probe path. This is the only assertion that shows a group filter covers routes nobody has written yet, which is the entire reason the deny is one filter per family.

**The mode signal is set and read back.** Both rigs set `CC_GATEWAY_HOSTED` explicitly and then assert `GatewayHostedMode.IsHosted` agrees before driving a route - writing an environment variable is a request, not a result. The self-host arms run BOTH non-hosted forms, unset and explicitly `"0"`; the unset form alone cannot tell self-host apart from the runner's ambient default, which is what every runner supplies for free.

**Status and media type asserted BEFORE every parse.** Parsing is itself an assertion about format, and a parse failure launders a served route into a crash that names nothing. Removing a guard now reddens as "expected NotFound, got OK" and "expected application/json, got text/plain".

**Survival assertions carry destructibility controls.** "The recording survived a refusal" and "the clip survived a refusal" are worth nothing unless the operation demonstrably destroys them when permitted. Both controls are named at the assertion: the self-host life-cycle test watches the recording actually disappear, and the self-host voice test asserts those two exact files are gone after a permitted stop.

## Pre-registered mutation plan - six runs, and why not twenty-five

Written into the test class BEFORE any of these runs, with predicted reds AND predicted greens per mutation. Each mutation is applied ALONE in its own full-assembly run: a run with two mutations cannot attribute any red to either, and once one red is unassignable the rest is inference.

**Six complete units, not five.** An earlier revision listed only the four arms plus the restore. That cannot support what the arms claim: an arm's reds are attributable to its mutation only if the SAME assembly at the SAME head was green immediately before it. Without a baseline, a test already failing on this head is read as caused by the arm, and the restore at the end cannot substitute - it answers a different question (did everything come back) after the evidence was needed.

| # | Mutation | Individually wrong while everything else is correct? |
| --- | --- | --- |
| M0 | **Baseline - no mutation** | n/a - fixes the green total every arm reconciles against BY NAME |
| M1 | Delete the recording group filter | Yes - hosted serves recordings, self-host untouched |
| M2 | Delete the cached-voice group filter | Yes - hosted serves voice, self-host untouched |
| M3 | Invert the recording hosted gate | Yes - a DIFFERENT defect from M1: M1 asks whether the filter is attached, M3 asks whether it reads the signal in the right direction. Their signatures differ, which is what separates them - M3 also breaks self-host |
| M4 | Invert the cached-voice hosted gate | Yes, same reason |
| M5 | Restore everything | The whole assembly green at that exact head |

The nineteen route-mapping lines are NOT argued away - they are gone, removed by the scope split above, so there is no bypass left to mutate. Overlap does not exempt anything: the three table-driven tests redden under all four mutations, which is not a reason to merge runs but the reason not to - they are the reds a combined run could never attribute.

The self-host class states its own direction under each mutation rather than claiming to stay green throughout. Green throughout was true only of M1 and M2; under M3 and M4 the self-host tests for that surface MUST go red, because an inverted gate refuses the deployment that is supposed to be served, and green there would mean the control cannot see self-host being broken.

## Status of the runs - none have been made

Everything above is written and BUILDS CLEAN (`0 Warning(s), 0 Error(s)`, verified against a full non-incremental rebuild, with the new symbols confirmed present in the built assembly).

**NO test has been run on this head, and no claim of a passing run is made.** The single machine-wide Gateway test lane is held by another worker, and a concurrent run kills the test host and produces failures in untouched areas. The counts in earlier revisions of this body predate this rebase and the changes above and are therefore withdrawn rather than restated - they were measured against a rig that seeded a location nothing reads, so they were not measuring what they claimed.

When the lane opens, the whole proof runs in one sitting at this exact head and is posted here: the M0 baseline with its total, the four single-mutation full-assembly runs with their reds named individually and reconciled against that baseline by name, the restored green at the exact same head, and a clean seven-project count. Each unit is preceded by a source and build preflight - exactly one production file differing from pristine, and a build line reading 0 Error(s) - because a run after a failed build executes the stale binary and reports a false pass.

### DIFF
--- a/src/CcDirector.Gateway.Tests/HostedRecordingDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedRecordingDenyTests.cs
@@ -406,12 +406,16 @@ public sealed class HostedRecordingGroupFilterTests : IDisposable
     /// primitive's own body-bound <c>/family/future</c> in <see cref="Tenancy.HostedRouteDenyOnHostedTests"/>).
     /// So the future route takes a plain <see cref="RecordingProbeBody"/> the framework binds from JSON, and:
     ///
-    ///  * on hosted every shape meets the refusal - a valid body, a malformed body, and a verb the route never
-    ///    mapped;
+    ///  * on hosted every shape meets the refusal - a valid body, a malformed body, a WRONG MEDIA TYPE, and a
+    ///    verb the route never mapped;
     ///  * the FRAMEWORK-400 CONTROL fires the identical malformed body at an UNDENIED equivalent of the same
     ///    handler mapped OUTSIDE <c>/ingest</c>, which really reaches the framework and returns its own 400 -
     ///    so the denied route's refusal is a short-circuit BEFORE framework binding, not a route that merely
     ///    happens not to bind;
+    ///  * the FRAMEWORK-415 CONTROL fires the identical wrong media type at that same undenied equivalent,
+    ///    which reaches the framework's endpoint SELECTION and returns its own 415 - so the denied route's
+    ///    wrong-media refusal short-circuits BEFORE selection, the one shape a per-handler guard answering
+    ///    after binding could never intercept;
     ///  * the OBSERVABLE-BINDING SEAM (<see cref="RecordingProbeBinding"/>), a distinct custom-bound route
     ///    mapped alongside, carries the "no handler-bound code ran at all" assertion the framework record
     ///    cannot make on its own: its binder counter stays zero across every shape behind the refusal.
@@ -441,14 +445,18 @@ public sealed class HostedRecordingGroupFilterTests : IDisposable
         try
         {
             // The framework-bound future route is refused on hosted across every shape: a valid body, a
-            // malformed body (off hosted the framework's own 400), and a verb it never mapped (off hosted a
-            // 405 disclosing the route exists).
+            // malformed body (off hosted the framework's own 400), a WRONG MEDIA TYPE (off hosted the
+            // framework's own 415 - a shape endpoint SELECTION enforces ahead of any handler, so a per-handler
+            // guard that answered after binding would leak it), and a verb it never mapped (off hosted a 405
+            // disclosing the route exists).
             foreach (var resp in new[]
                      {
                          await http.PostAsync("/ingest/added-after-the-deny-was-written",
                              new StringContent("{\"text\":\"hello\"}", Encoding.UTF8, "application/json")),
                          await http.PostAsync("/ingest/added-after-the-deny-was-written",
                              new StringContent("{ not json", Encoding.UTF8, "application/json")),
+                         await http.PostAsync("/ingest/added-after-the-deny-was-written",
+                             new StringContent("hello", Encoding.UTF8, "text/plain")),
                          await http.GetAsync("/ingest/added-after-the-deny-was-written"),
                      })
             {
@@ -464,6 +472,15 @@ public sealed class HostedRecordingGroupFilterTests : IDisposable
                 new StringContent("{ not json", Encoding.UTF8, "application/json"));
             Assert.Equal(HttpStatusCode.BadRequest, undeniedMalformed.StatusCode);
 
+            // THE FRAMEWORK-415 CONTROL. A wrong media type reaches the framework's own 415 on the undenied
+            // equivalent, while the denied route returns the refusal above - so the denied route's wrong-media
+            // refusal is a short-circuit BEFORE endpoint SELECTION would 415, not a route that merely happens
+            // to reject the media type. This is the shape a per-handler guard answering after model binding
+            // could never intercept: selection would 415 first and the leak would be invisible.
+            var undeniedWrongMedia = await http.PostAsync("/undenied-equivalent",
+                new StringContent("hello", Encoding.UTF8, "text/plain"));
+            Assert.Equal(HttpStatusCode.UnsupportedMediaType, undeniedWrongMedia.StatusCode);
+
             // NOTHING behind the refusal bound an argument, on any shape - the assertion a parameterless GET
             // or a body-ignoring custom binder could never make, and the one that separates the exclusive
             // catch-all (refusal placed BEFORE binding) from the rejected per-handler filter (which ran the
@@ -474,6 +491,8 @@ public sealed class HostedRecordingGroupFilterTests : IDisposable
                              new StringContent("{\"text\":\"hello\"}", Encoding.UTF8, "application/json")),
                          await http.PostAsync("/ingest/added-after-the-deny-was-written-observed",
                              new StringContent("{ not json", Encoding.UTF8, "application/json")),
+                         await http.PostAsync("/ingest/added-after-the-deny-was-written-observed",
+                             new StringContent("hello", Encoding.UTF8, "text/plain")),
                          await http.GetAsync("/ingest/added-after-the-deny-was-written-observed"),
                      })
             {
@@ -672,6 +691,14 @@ public sealed class SelfHostRecordingGroupControlTests : IDisposable
             var malformed = await http.PostAsync("/ingest/added-after-the-deny-was-written",
                 new StringContent("{ not json", Encoding.UTF8, "application/json"));
             Assert.Equal(HttpStatusCode.BadRequest, malformed.StatusCode);
+
+            // And the SAME future route is capable of the 415 the hosted wrong-media refusal short-circuits: a
+            // wrong media type off hosted hits the framework's own endpoint-selection 415. Paired with the
+            // hosted wrong-media refusal, this proves the hosted 404 on text/plain is the deny firing before
+            // selection, not the route simply refusing the media type on both sides.
+            var wrongMedia = await http.PostAsync("/ingest/added-after-the-deny-was-written",
+                new StringContent("hello", Encoding.UTF8, "text/plain"));
+            Assert.Equal(HttpStatusCode.UnsupportedMediaType, wrongMedia.StatusCode);
 
             // The observable custom binder really ran off hosted - the positive twin of the hosted count of 0.
             var observed = await http.PostAsync("/ingest/added-after-the-deny-was-written-observed",
--- a/src/CcDirector.Gateway.Tests/HostedRecordingDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedRecordingDenyTests.cs
@@ -7,6 +7,7 @@ using System.Net.Http.Headers;
 using System.Net.Sockets;
 using System.Text;
 using System.Text.Json;
+using System.Threading;
 using System.Threading.Tasks;
 using CcDirector.Core.Recording;
 using CcDirector.Core.Storage;
@@ -160,6 +161,7 @@ public sealed class HostedRecordingDenyTests : IAsyncLifetime
     [InlineData("POST", "ingest/recording/" + SeededId + "/complete", "{}")]
     [InlineData("POST", "ingest/recording/" + SeededId + "/promote", null)]                        // promote into the vault
     [InlineData("PATCH", "ingest/recording/" + SeededId + "/meta", "{\"title\":\"hijacked\"}")]     // overwrite metadata
+    [InlineData("POST", "ingest/recording/" + SeededId + "/meta", "{\"title\":\"hijacked\"}")]      // /meta maps POST too - the omitted verb
     [InlineData("DELETE", "ingest/recording/" + SeededId, null)]                                   // destroy the record
     [InlineData("GET", "ingest/dictionary", null)]                                                 // the shared glossary
     [InlineData("PUT", "ingest/dictionary", "{\"vocabulary\":[],\"commonMistranscriptions\":{},\"profiles\":{}}")]
@@ -368,8 +370,6 @@ internal static class RecordingGroupProbeHost
 [Collection("DirectorRoot")]
 public sealed class HostedRecordingGroupFilterTests : IDisposable
 {
-    private const string ProbePayloadSentinel = "probe-payload-that-must-never-be-served-on-hosted";
-
     private readonly string _root;
     private readonly string? _prevRoot;
     private readonly string? _priorHosted;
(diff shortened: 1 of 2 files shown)


thefrederiksen (commented):
REJECT

1. Removing both real production filters was followed only by the 29 deny tests and 5 self-host controls. Run the successful-build revert through the full 2985-test Gateway suite, record the intended 28 deny reds with the unauthenticated and self-host controls green, then restore/rebuild/rerun full.

2. Neither the recording group nor cached-voice subgroup has a future-route probe. This matters especially in `GatewayWingmanVoiceEndpoint`, where the filtered `cachedVoice` builder and unfiltered `app` builder coexist. Map a probe through each filtered builder and prove hosted returns exactly one `error` property while explicit hosted-off reaches it.

The direct hosted-mode gates, all-current-route deny/effect coverage, exact response allow-list, explicit hosted-off end-to-end read/write/delete controls, and complete route-by-route un-deny debt are otherwise sound. The CI red is the known unrelated account-status case.

thefrederiksen (commented):
Re-review of a8f76499: the cached-audio canary is now handler-positive—exact seeded bytes plus audio/mpeg—and the branch includes #1916, so the prior account-status allowance is retired. Verdict remains REJECT under the raised route-coverage bar. The self-host controls still do not positively exercise GET /ingest/recording/{id}/audio/{index}, POST /ingest/recording/{id}/complete, POST /ingest/recording/{id}/promote, or the POST variant of /ingest/recording/{id}/meta; add a permanent every-denied-route table whose receipt/effect is unique to each real handler and mutation-prove it. Add a future-route probe to each of the two filtered groups, recording and cached voice. Finally, the PR body says its full counts predate this rebase/assertion change and the branch does not yet inherit #1918, so the serialized full-suite filter-removal named-red run and restored exact-head green are still owed.

thefrederiksen:
REJECT / proof hold on exact head `d247a7b0` (static review; no lane consumed).

The code-level deny shape now passes the important structural check: `RecordingEndpoints.MapRoutes` and `GatewayWingmanVoiceEndpoint.MapCachedVoiceRoutes` receive only the filtered group, so the per-route builder substitution is no longer expressible. The table-driven hosted/self-host controls, direct partition seeding, exact refusal checks, future probes, and destructive survival controls are also coherent on inspection.

The un-deny split is directionally right but its factual account needs correction:

- Recording ingest: the filter stops all new HTTP writes because the sole production service construction is inside the guarded mapping. But “nothing to purge” does not follow. The shared recording root can already contain cross-tenant material written before this deny lands. A future tenant-partitioned un-deny must explicitly quarantine/purge/migrate that legacy root; “the store is exactly as the deny found it” is not evidence that what it found was clean.
- Cached voice: requiring conversion of all hard-coded `TenantId.Local` producers plus disposal of the local pile is the safe condition. However the claim that those producers currently keep filling Local on hosted contradicts this same head’s production comments and flow: the hosted session stores are tenant-partitioned, `FleetByDirector(..., TenantId.Local)`/`VoiceSessionIds(Local)` see an empty Local fleet, and the no-scope down-channel fails closed. The risk that is proved here is the pre-existing/legacy Local cache being served, plus hard-coded Local writers that must be removed before un-deny—not demonstrated ongoing writes by every hosted caller. State that exact model.

No mutation proof exists on this head. After the ledger is made exact and the branch is rebased onto current main, the required sitting is six complete Gateway-suite units: clean exact-head baseline, four independent arms (filter removal and hosted-gate inversion for each family), and exact restored green, all with source/build preflight and reconciled counts. The current body lists only the four arms plus restore and does not establish the pre-mutation baseline.

thefrederiksen:
STANDING: REJECTED
cbb8e77aa6b4ba94e7f70a596cfed6a7fe698f56

The recording and cached-voice refusals do not establish the mission-wide guarantee for every future request shape, and both future-route probes certify only the narrower read-shaped case. Adopt the shared earlier family-boundary primitive, retain inherited scope for future routes, and add the required write-shaped future-route proof for each adopter.

Security details and the family-specific blast-radius check were recorded privately. I found no constraint in either family that prevents adoption of the proposed shared primitive.

thefrederiksen:
STANDING: REJECTED
cbb8e77aa6b4ba94e7f70a596cfed6a7fe698f56

The rework materially improves the unit: the recording legacy-root debt is now stated correctly, the cached-voice writers are correctly classified as latent rather than active, and the route-mapping scope removes the per-route receiver bypass.

The deny boundary itself still does not satisfy the standing contract. It remains an endpoint-group filter, so argument binding/content negotiation can answer before the refusal. The future-route probe is still a parameterless GET; it must be a body-bound POST to exercise the property that justifies the group mechanism. The proof set also lacks the required undenied-neighbour probe for over-refusal.

No test has run on this head and the serialized baseline, four independent arms, restore, and final seven-project green are explicitly outstanding. The branch is also behind current `main`.

Rework must adopt the cleared shared pre-binding primitive, replace the future-route canary with a body-bound POST plus observable no-binding assertion, add the hosted neighbour/self-host control, then execute the pre-registered full-suite units on the rebased final head with source/build/assembly identity and exact restoration.

PR #1908, opened 2026-07-19
perf(installer): move cc-* tools provisioning out of the installer to first launch

thefrederiksen:
Moves the `cc-*` Python tools provisioning out of the installer and onto first launch, so the Install step finishes in seconds instead of waiting on a bundle download plus an offline pip install.

This is finished, independently reviewed work that was never delivered. The branch was approved and then sat: no pull request had ever been opened for it.

## Correcting the numbers this change has been carried under

The slice brief describes the tools bundle as "~334 MB" taking "3-8 min", and the win has been described as removing that from install time. Both figures are stale, and the second describes the win incorrectly.

Measured on a real install log on a Windows machine on a 94 megabit line (`%LOCALAPPDATA%\cc-director\logs\setup\setup-20260719-023305.log`):

| Phase | Duration |
|---|---|
| Tools bundle: download 22.0 MB + 98.4 MB, stage Python, build the venv, install 95 wheels | **66.7 s** |
| Everything else: place 11 components, shortcut, 3 skills, launcher | 1.2 s |
| **Total, Install click to Complete** | **74.1 s** |

So the bundle on the wire is **120.4 MB, not 334 MB**, and the step took **67 seconds, not 3-8 minutes**.

Where the stale numbers came from:

- **334 MB** is a comment in `tools/cc-director-setup-engine/PythonToolsInstaller.cs:93-94` quoting an old progress string, `"Downloading 118.2 MB / 334.5 MB"`.
- **3 minutes** is `docs/install/INSTALLER_HANDOVER.md:125`, which dates from v0.6.3 and describes installing **23** tools. The release ships **9** today (`tools/registry.json`, entries with `"ship": true`). The bundle shrank and nobody re-measured.

**And the win is a move, not a removal.** The 67 seconds does not disappear - it relocates to first launch, where the app provisions the bundle in the background. The user still waits for it before any `cc-*` tool works. What actually improves:

- the Install step drops from about 74 seconds to about 7 seconds, so the wizard reaches Complete quickly and the window opens;
- the wait overlaps with the user's first actions instead of blocking a progress bar;
- an update no longer rebuilds the whole venv when the app is going to reconcile it anyway.

That is a real improvement and worth shipping. It is not several minutes off the clock, and nothing downstream should be planned as though it were.

## What is in the change

- The installer no longer runs Python tools provisioning. Both WPF and Avalonia runners call a common no-op seam (`InstallerToolsProvisioning`).
- The Install screen says so honestly rather than silently dropping the row.
- First-run provisioning from nothing is guaranteed: `ToolReconciler` treats an absent bundle record as drift and drives a full heavy provision, so a fresh install with no tools present provisions the whole bundle on first launch.
- The Gateway GUI's child `install --role gateway --component gateway` pass no longer provisions Python tools, via one component-scope decision used both to narrow the plan and to gate the tools install.
- Partial first-provision residue (a bare venv interpreter with no manifest and no sidecar) is now recognised and retried instead of being mistaken for a healthy install.

## Review status

Independently reviewed at `3d26c6d3` - `devthrottle_internal` `docs/install/codex-review-tools-out-v2.md`: **no blocking findings, both fixes PASS with revert-proofs reproduced against the production lines, verdict APPROVE.** Suites at review time: engine 329 passed, cli 27 passed, WPF setup 32 passed; release builds of the WPF setup, the Avalonia setup and the Director app all clean.

Rebased from `3d26c6d3` onto current `origin/main` for this pull request. The rebase is content-identical: the only differences in the patch are blob hashes and a 20-line offset in `Commands.cs`, where main has since added the hosted enrolment code above it.

## Not covered

Live network provisioning of the bundle and human GUI rendering were not exercised by the review. Those need a real install on a real machine, which is the human QA handoff.

### DIFF
--- a/tools/cc-director-setup-avalonia/Services/EngineInstallRunner.cs
+++ b/tools/cc-director-setup-avalonia/Services/EngineInstallRunner.cs
@@ -43,12 +43,10 @@ public sealed class EngineInstallRunner
         else dItem.SizeText = FormatSize(dAsset.Size);
         items.Add(dItem); byId["director"] = dItem;
 
-        var bItem = new ToolDownloadItem { Name = PythonToolsInstaller.ComponentId, AssetName = PythonToolsInstaller.ToolsAsset };
-        var pyAsset = release.Manifest.TryGetAsset(PythonToolsInstaller.PythonAsset);
-        var toolsAsset = release.Manifest.TryGetAsset(PythonToolsInstaller.ToolsAsset);
-        if (pyAsset is null || toolsAsset is null) { bItem.Status = "Skipped"; bItem.SizeText = "Not in release"; }
-        else bItem.SizeText = FormatSize(pyAsset.Size + toolsAsset.Size);
-        items.Add(bItem); byId[PythonToolsInstaller.ComponentId] = bItem;
+        // The shared-venv cc-* Python tools bundle is deliberately NOT an install-time row anymore: the
+        // installer no longer provisions it (that ~334 MB download + venv build was the dominant install
+        // time). The app provisions the bundle from nothing on first launch (ToolReconciler startup
+        // reconcile), so the Install screen shows an honest "finishes on first launch" note instead.
 
         // The launcher installs on macOS only in THIS wizard: the Windows wizard (the WPF one)
         // already installs it there. Older releases have no macOS launcher asset - skip cleanly.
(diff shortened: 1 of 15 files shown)


thefrederiksen:
Re-ran the three affected suites against the rebased base (this branch on current `origin/main`), one project per invocation:

- `CcDirector.Setup.Engine.Tests` - **340 passed**, 0 failed, 0 skipped
- `CcDirector.Setup.Cli.Tests` - **28 passed**, 0 failed, 0 skipped
- `CcDirectorSetup.Tests` (WPF setup) - **32 passed**, 0 failed, 0 skipped

Counts are slightly above the review's (329 / 27 / 32) because main has added tests since `3d26c6d3`.

Note for anyone re-running these: passing all three project files to a single `dotnet test` invocation fails with MSB1008 "Only one project can be specified" and still exits 0, which reads as a green run in a log. Run them one at a time.

PR #1909, opened 2026-07-19
fix(gateway): tenant-composite primary keys for caller-supplied keys (snoozes, session_spend, push_subscriptions)

thefrederiksen:
## The defect

Three tenant-scoped tables in `src/CcDirector.Gateway/Data/GatewayDbContext.cs` had a PRIMARY KEY built
entirely from CALLER-SUPPLIED input, with `tenant_id` present as a data column but not part of the key:

| table | old primary key | where the value comes from |
|---|---|---|
| `snoozes` | `SessionId` | the session id a Director presents on a snooze request |
| `session_spend` | `SessionId` | the session id on a record-spend request |
| `push_subscriptions` | `Endpoint` | the browser's Web Push URL |

Every store upserts the same way: read THROUGH the tenant query filter, add when the read finds nothing. So a
second tenant presenting an identifier the first tenant already holds reads nothing (the filter hides the other
tenant's row), takes the insert branch, and the database rejects it on the primary key. That is a cross-tenant
SQUAT - the second tenant cannot use that identifier at all, because another tenant has it - and an EXISTENCE
ORACLE, because the failure discloses that another tenant holds it. No row content crosses the boundary, but
the key space does.

Proven before the fix, driving the REAL stores (`SnoozeRegistry`, `SessionSpendStore`, `PushSubscriptionStore`)
with two tenants over one database. Three tests, three failures:

```
SQLite Error 19: 'UNIQUE constraint failed: snoozes.SessionId'
SQLite Error 19: 'UNIQUE constraint failed: session_spend.SessionId'
SQLite Error 19: 'UNIQUE constraint failed: push_subscriptions.Endpoint'
```

## The fix

Composite primary keys, following #1840 exactly (which did this for `cron_jobs`, `workflows`,
`workflow_versions` and `mission_notes`): `(tenant_id, SessionId)`, `(tenant_id, SessionId)`,
`(tenant_id, Endpoint)`. No store code changed - every lookup already goes through a filtered query, never
`Find(key)`.

## The sibling guard: the exemption is a TYPE, not a written allowlist

`TenantScopeGuardTests` asserted the tenant COLUMN and the query FILTER and is structurally blind to the KEY,
which is exactly how these three passed it. It now also requires every tenant-scoped table's primary key to
include `tenant_id` - unless the key is one the Gateway itself mints.

**The first version of that exemption was rejected, and was rejected correctly.** It was an allowlist of ten
table names, each with a sentence asserting "this store mints the key with `Guid.NewGuid()`", re-checked by
asserting the key was still a single `Guid`. That check could not fail for the change it existed to catch: a
store switching from minting its key to accepting one from the caller keeps the key a single `Guid`, so the
guard stayed green in precisely the dangerous case. A check that cannot fail for its own motivating case is
worse than no check, because it reads as protection and stops everyone downstream from looking.

The exemption is now the type `GatewayMintedKeyEntity`, whose `Id` has a **private setter** and is minted by
the base class itself. The ten entities that had a Gateway-minted key derive from it and no longer declare
their own `Id`, and the thirteen `Id = Guid.NewGuid()` assignments are gone - minting lives in the type.

**Assigning a caller-supplied value to one of these keys is now a compile error, not something a guard notices
afterwards:**

```
error CS0272: The property or indexer 'GatewayMintedKeyEntity.Id' cannot be used in this context
              because the set accessor is inaccessible
```

The compiler enumerated all thirteen existing assignment sites and every one was a fresh `Guid.NewGuid()`, so
the old prose claim was true *today* - it is now enforced rather than asserted.

The three ways to escape the type are each closed by construction, and each was verified by making the change
and watching it fail:

| Escape | What stops it - observed, not argued |
|---|---|
| A store's key switches from minted to caller-supplied | **Does not compile** (`CS0272`, single error, at the assignment) |
| Widen the setter so the above compiles again | `TheGatewayMintedKey_CanOnlyBeWrittenByTheBaseClassThatMintsIt` goes red: `setter.IsPrivate = False (public=True)` |
| Re-declare a settable `Id` on the derived entity, or drop the base class - still a single `Guid` key, the shape the old check accepted | `EveryTenantScopedTable_HasTenantIdInItsPrimaryKey_UnlessTheGatewayMintsTheKeyItself` goes red, naming it: `CronRunEntity (key: Id; declared on: CronRunEntity)` - the test requires the mapped key property to be DECLARED ON `GatewayMintedKeyEntity`, not merely to be a `Guid` |

Negative control: with every mutation reverted, both guards are green.

This already caught something real. #1906, developed in parallel, adds a `WorkListEntity { Id = rowId }`
construction site, which this change makes a compile error. No prose allowlist and no reflection guard had
flagged it. That branch rebases onto this one and lets the base type mint the row identifier - see the
convergence note at the bottom.

## Migrations, and what they do to existing rows

One migration per provider, generated from the same model:
`Data/Migrations/20260719201700_CallerSuppliedKeysScopedByTenant` (SQLite) and
`Migrations/20260719201740_CallerSuppliedKeysScopedByTenant` (Postgres). Each drops the three single-column
primary keys and adds the composite ones. `dotnet ef migrations has-pending-model-changes` reports no pending
changes on BOTH providers, re-run on the current head.

**Row preservation is established by running the upgrade, not by reading the operations.** That distinction
matters here: on SQLite a primary key cannot be altered in place, so EF turns each operation pair into a table
REBUILD - create a new table, copy every row, drop the original, rename - and whether that copy carries every
row and every column value is a property of the generated rebuild, not of the three lines anyone reads in the
migration file.

Two test classes, one per provider, both doing the same thing: migrate to `CompositeTenantKeys` (the migration
immediately before this one), insert rows through raw SQL the way a database in the field holds them (two
tenants, nullable columns exercised both null and populated), migrate the rest of the way, then assert every
row and every column value is unchanged. Both also assert the migration GENUINELY RAN - by name out of the
migrations history table, and by the key really being composite afterwards - so a migration that silently did
nothing cannot pass by leaving the rows undisturbed. A second test in each proves the upgrade delivers what it
is for: before it, a second tenant reusing an identifier is refused on the primary key; after it, both rows
coexist.

- `CallerSuppliedKeyUpgradePreservesRowsTests` (SQLite) - always runs, no external service.
- `CallerSuppliedKeyUpgradePreservesRowsPostgresTests` (Postgres) - gated on `CC_GATEWAY_TEST_PG_CONNECTION`
  exactly as `PostgresProviderProofTests` is, and reports SKIPPED when it is unset. Skipped is not passed: with
  no server configured it makes no claim rather than a false one. It drops and recreates, so it carries the
  same `ccpg`-prefix throwaway-database guard as the existing Postgres proofs.

Executed evidence, both providers, on this head. SQLite:

```
PRE-CHANGE KEYS:  snoozes=[SessionId] session_spend=[SessionId] push_subscriptions=[Endpoint]
ROWS BEFORE:      snoozes=2 session_spend=2 push_subscriptions=2
POST-CHANGE KEYS: snoozes=[tenant_id,SessionId] session_spend=[tenant_id,SessionId] push_subscriptions=[tenant_id,Endpoint]
APPLIED:          20260719201700_CallerSuppliedKeysScopedByTenant
ROWS AFTER:       snoozes=2 session_spend=2 push_subscriptions=2
CROSS-TENANT SAME SESSION ID: ACCEPTED (correct)
RESULT: PASS - every row and every column value preserved
```

Postgres 16, in a throwaway container, running the Postgres migration set through the `gateway` schema:

```
PRE-CHANGE KEYS:  snoozes=[SessionId] session_spend=[SessionId] push_subscriptions=[Endpoint]
ROWS BEFORE:      snoozes=2 session_spend=2 push_subscriptions=2
POST-CHANGE KEYS: snoozes=[tenant_id,SessionId] session_spend=[tenant_id,SessionId] push_subscriptions=[tenant_id,Endpoint]
APPLIED:          20260719201740_CallerSuppliedKeysScopedByTenant
OK  snoozes / session_spend / push_subscriptions: all rows and every column value preserved
OK  BEFORE the upgrade the second tenant is refused on the primary key (SqlState=23505)
OK  AFTER the upgrade the same insert is ACCEPTED
RESULT: PASS - all checks
```

The Postgres arm was also run with the upgrade suppressed, as a negative control: the four
migration-actually-ran assertions go red while the row-preservation assertions stay green - which is exactly
why those anti-vacuity assertions exist, since untouched rows survive trivially when nothing runs.

## Base, and what is still outstanding

Rebased onto current `origin/main`. `git range-diff` reports `=` across each rebase - content unchanged,
verified by range-diff rather than by a rebuild. The overlapping range was inspected: #1901 changed
`GatewayDbContext` and `TenantScopeGuardTests`, and its guard change (adding `EntitlementEntity` to
`allowedGlobalTables`) survived intact. It did NOT change either provider snapshot - `EntitlementEntity` is
mapped `ExcludeFromMigrations` and `entitlements` appears in neither snapshot, checked directly.

**The full-suite mutation proof is NOT yet run, and this is not mergeable until it is.** The suite lock is held
by another unit and this one is not admitted. Seven serialized full Gateway suite runs are owed, pre-registered
with their predicted reds by name in the durable comment below:

| Run | Mutation |
|---|---|
| baseline | none - a green AND explicitly complete run at the final head |
| M1 | `SnoozeEntity` key reverted to `HasKey(e => e.SessionId)` |
| M2 | `SessionSpendEntity` key reverted |
| M3 | `PushSubscriptionEntity` key reverted |
| M4 | `GatewayMintedKeyEntity.Id` setter widened to public |
| M5 | `public new Guid Id { get; set; }` re-declared on `CronRunEntity` |
| restored | exact restoration, green at the exact head |

One primitive per run, never combined - the three mappings are independently bypassable. Reconciliation will be
computed: passed plus failed under the mutation equals passed under the restored run. A predicted red that does
not appear is a finding, not something to explain away.

### Why every red is classified by WHERE it arrived

Each red in each arm is classified by where it arrived - **fixture, migration, or assertion** - and not merely
by whether the arm was red. That rule exists for one reason, and the reason must travel with it, because a rule
without its reason gets dropped by whoever is next and is sure they understand better:

> **The arm's colour is not evidence about the thing the arm was testing.**

Everything this branch has had to patch is an instance of that. A stale build gives a GREEN that is not
evidence of the mutation's absence - only that the mutation never reached the assembly. A fixture that throws
on a renamed column gives a RED that is not evidence the guard fired - only that the seed never ran. An
emptied fixture gives a GREEN that is not evidence anything survived - only that nothing was there to lose. A
migration that silently did nothing gives a GREEN row-preservation result that is not evidence the upgrade is
safe - untouched rows survive trivially.

In each case the colour is a statement about the HARNESS, and it was read as a statement about the SUBJECT.
That is why this proof asserts that the seed inserted rows, that the migration is recorded by name, and that
the migration under test is still the immediate successor of its named predecessor: each of those turns a
harness fact that would otherwise be invisible into something the arm has to state out loud before its colour
means anything.

Note the Postgres upgrade tests will report SKIPPED on the admission machine, so they contribute nothing to
those runs; their evidence is the executed container run above.

Proof-hygiene note for whoever runs the arms: restore each mutation with `git checkout --`, which stamps a
current modification time. Restoring by moving or copying an older backup over the file preserves that older
timestamp, MSBuild then treats the build as up to date, and the run silently executes the PREVIOUS binary. That
was observed on this branch while producing the evidence above - a restored source printed the mutation's
output because the assembly was never rebuilt - and in the direction that matters it would show a mutation
producing no red at all.

## Convergence with #1906

#1906 currently adds a `WorkListEntity { Id = rowId }` construction site and predates this private setter, so
the two branches cannot coexist unchanged. The order is: land this first, then rebase #1906 and let the base
type mint the row identifier before it is recorded or asserted. Final evidence for the two heads must not be
produced independently and then merged without reconciling that overlap.

### DIFF
--- a/docs/architecture/hosted-gateway-qa.md
+++ b/docs/architecture/hosted-gateway-qa.md
@@ -445,3 +445,93 @@ The hosted Gateway is validated end to end: a real Director authenticates and tu
 URL, a real operation persists through it into the isolated Supabase gateway schema, state survives an App
 Service restart, and the tunnel holds past the idle cap. The isolated test Director (pid 38432) and its
 scratch CC_DIRECTOR_ROOT were torn down after the run.
+
+## Tenant scoping - the PRIMARY KEY half (snoozes, session_spend, push_subscriptions)
+
+Date: 2026-07-19
+
+Issue #1840 made `cron_jobs`, `workflows`, `workflow_versions` and `mission_notes` tenant-composite. A
+persistence census afterwards found three tenant-scoped tables still keyed on CALLER-SUPPLIED input with
+`tenant_id` present only as a data column:
+
+- `snoozes` - primary key `SessionId` (the session id a Director presents on a snooze request)
+- `session_spend` - primary key `SessionId` (the session id on a record-spend request)
+- `push_subscriptions` - primary key `Endpoint` (the browser's push URL)
+
+### Why that is a defect, not a style point
+
+Every store upserts the same way: read THROUGH the tenant query filter, and add when the read finds nothing.
+So a second tenant presenting an identifier the first tenant already holds reads nothing (the filter hides the
+other tenant's row), takes the insert branch, and the database rejects it on the primary key. That is a
+cross-tenant SQUAT (the second tenant cannot use that identifier at all, because another tenant has it) and an
+EXISTENCE ORACLE (the failure discloses that another tenant holds it). No row content crosses the boundary,
+but the key space does.
+
+Reproduced before the fix by driving the REAL stores (`SnoozeRegistry`, `SessionSpendStore`,
+`PushSubscriptionStore`) with two tenants over one database - three tests, three failures, each
+`SQLite Error 19: 'UNIQUE constraint failed: <table>.<key>'`.
+
+### The fix
+
+`GatewayDbContext` now gives all three a composite primary key, following #1840 exactly:
+`(tenant_id, SessionId)`, `(tenant_id, SessionId)`, `(tenant_id, Endpoint)`. No store code changed - every
+lookup already goes through a filtered LINQ query, never `Find(key)`.
+
+### The guard that should have caught it
+
+`TenantScopeGuardTests` asserted the tenant COLUMN and the query FILTER, and is structurally blind to the KEY -
+which is exactly how these three passed it. A third assertion,
+`EveryTenantScopedTable_HasTenantIdInItsPrimaryKey_UnlessTheGatewayMintsTheKeyItself`, now requires every
+tenant-scoped table's primary key to include `tenant_id`, unless the key is one the Gateway itself mints.
+
+That exemption is a TYPE, not a written list. The first attempt was an allowlist of ten table names, each with a
+sentence asserting "this store mints the key with `Guid.NewGuid()`", re-checked by asserting the key was still a
+single `Guid`. That check could not fail for the change it existed to catch: a store switching from minting its
+key to accepting one from the caller keeps the key a single `Guid`, so the guard stayed green in precisely the
+dangerous case - which is worse than no guard, because it reads as protection and stops anyone downstream from
+looking.
+
+The exemption is now `GatewayMintedKeyEntity`, whose `Id` has a PRIVATE setter and is minted by the base class
+itself. The ten entities derive from it and no longer declare their own `Id`. Assigning a caller-supplied value
+to one of these keys is not a test failure found later - it does not compile. The three ways to escape that are
+each closed by construction:
+
+| Escape | What stops it |
+|---|---|
+| Assign a caller-supplied value to the key | Compile error `CS0272` - the setter is inaccessible |
+| Widen the setter so that compiles again | `TheGatewayMintedKey_CanOnlyBeWrittenByTheBaseClassThatMintsIt` turns red |
+| Re-declare a settable `Id` on the derived entity, or drop the base class | The primary-key test requires the mapped key property to be DECLARED ON `GatewayMintedKeyEntity`, so it turns red |
+
+Each of those three was verified by making the change and watching the failure, and both guards were confirmed
+green again once it was reverted.
+
+### The migration, and what it does to existing rows
+
+One migration per provider, generated from the same model:
+`Data/Migrations/20260719201700_CallerSuppliedKeysScopedByTenant` (SQLite) and
+`Migrations/20260719201740_CallerSuppliedKeysScopedByTenant` (Postgres). Each drops the three single-column
+primary keys and adds the composite ones. `dotnet ef migrations has-pending-model-changes` reports no pending
+changes on BOTH providers.
+
+Existing rows are preserved, and that is established by RUNNING the upgrade, not by reading the operations.
+`CallerSuppliedKeyUpgradePreservesRowsTests` migrates a database to the migration immediately before this one,
+inserts rows through raw SQL the way a database in the field holds them, migrates the rest of the way, and
+asserts every row and every column value is unchanged. It also asserts the migration actually ran - by name out
+of `__EFMigrationsHistory`, and by the key really being composite afterwards - so a migration that silently did
+nothing cannot pass by leaving the rows undisturbed. A second test shows the upgrade delivers what it is for:
+before it, a second tenant reusing a session id is refused on the primary key; after it, both rows coexist.
+
+This matters because the reasoning below is NOT sufficient on its own. On SQLite a primary key cannot be altered
+in place, so EF turns each operation pair into a table rebuild, and whether the copy carries every row and every
+column value is a property of that generated rebuild rather than of the three lines anyone reads in the
+migration file. The supporting reasons:
+
+- No row is deleted, inserted, or edited. On SQLite a primary-key change is a table REBUILD (EF creates the new
+  table, copies every row, drops the old, renames); on Postgres it is `ALTER TABLE ... DROP CONSTRAINT` plus
+  `ADD PRIMARY KEY`, with no table rewrite.
+- Uniqueness cannot be violated by existing data: the OLD key already made the second column unique on its own,
+  so `(tenant_id, <that column>)` is unique a fortiori. There is no possible duplicate to trip on.
+- `tenant_id` is already NOT NULL on all three tables (`ApplyTenantScope` marks it required), which a primary-key
+  column must be, so no backfill and no nullability change is needed.
+- On a single-tenant install every row is `tenant_id = 'local'`, so the composite key behaves identically to the
+  old one and the upgrade is invisible.
--- a/src/CcDirector.Gateway.Migrations.Postgres/Migrations/20260719201740_CallerSuppliedKeysScopedByTenant.Designer.cs
+++ b/src/CcDirector.Gateway.Migrations.Postgres/Migrations/20260719201740_CallerSuppliedKeysScopedByTenant.Designer.cs
@@ -3,42 +3,51 @@ using System;
 using CcDirector.Gateway.Data;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
 
 #nullable disable
 
-namespace CcDirector.Gateway.Data.Migrations
+namespace CcDirector.Gateway.Migrations.Postgres.Migrations
 {
     [DbContext(typeof(GatewayDbContext))]
-    partial class GatewayDbContextModelSnapshot : ModelSnapshot
+    [Migration("20260719201740_CallerSuppliedKeysScopedByTenant")]
+    partial class CallerSuppliedKeysScopedByTenant
     {
-        protected override void BuildModel(ModelBuilder modelBuilder)
+        /// <inheritdoc />
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
-            modelBuilder.HasAnnotation("ProductVersion", "9.0.2");
+            modelBuilder
+                .HasDefaultSchema("gateway")
+                .HasAnnotation("ProductVersion", "9.0.2")
+                .HasAnnotation("Relational:MaxIdentifierLength", 63);
+
+            NpgsqlModelBuilderExtensions.UseIdentityByDefaultColumns(modelBuilder);
 
             modelBuilder.Entity("CcDirector.Gateway.Data.Entities.AccountHostedAiSpendEntity", b =>
                 {
                     b.Property<Guid>("Id")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("uuid");
 
                     b.Property<long>("AmountMicros")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("bigint");
 
                     b.Property<string>("Kind")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<DateTime>("ObservedUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<string>("TenantId")
                         .IsRequired()
-                        .HasColumnType("TEXT")
+                        .HasColumnType("text")
                         .HasColumnName("tenant_id");
 
                     b.Property<DateTime>("TransactionCreatedUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.HasKey("Id");
 
@@ -48,108 +57,108 @@ namespace CcDirector.Gateway.Data.Migrations
 
                     b.HasIndex("TenantId", "Kind", "AmountMicros", "TransactionCreatedUtc");
 
-                    b.ToTable("account_hosted_ai_spend", (string)null);
+                    b.ToTable("account_hosted_ai_spend", "gateway");
                 });
 
             modelBuilder.Entity("CcDirector.Gateway.Data.Entities.CronJobEntity", b =>
                 {
                     b.Property<string>("TenantId")
-                        .HasColumnType("TEXT")
+                        .HasColumnType("text")
                         .HasColumnName("tenant_id");
 
                     b.Property<string>("Id")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<DateTime>("CreatedUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<string>("CronExpression")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<bool>("Enabled")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("boolean");
 
                     b.Property<DateTime?>("LastFiredUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<string>("LastStatus")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("Name")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<DateTime?>("NextRunUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<string>("NotifyOn")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("NotifyWebhookUrl")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<bool>("PreventOverlap")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("boolean");
 
                     b.Property<string>("RunAt")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("ScheduleKind")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("TimeZoneId")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.HasKey("TenantId", "Id");
 
                     b.HasIndex("TenantId");
 
-                    b.ToTable("cron_jobs", (string)null);
+                    b.ToTable("cron_jobs", "gateway");
                 });
 
             modelBuilder.Entity("CcDirector.Gateway.Data.Entities.CronRunEntity", b =>
                 {
                     b.Property<Guid>("Id")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("uuid");
 
                     b.Property<DateTime>("FiredUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<string>("InfraStatus")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("JobId")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("Machine")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<DateTime>("ScheduledUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<long>("Sequence")
-                        .HasColumnType("INTEGER");
+                        .HasColumnType("bigint");
 
                     b.Property<string>("SessionId")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("TargetDirectorId")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("TaskStatus")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("TenantId")
                         .IsRequired()
-                        .HasColumnType("TEXT")
+                        .HasColumnType("text")
                         .HasColumnName("tenant_id");
 
                     b.HasKey("Id");
@@ -158,44 +167,44 @@ namespace CcDirector.Gateway.Data.Migrations
 
                     b.HasIndex("JobId", "Sequence");
 
-                    b.ToTable("cron_runs", (string)null);
+                    b.ToTable("cron_runs", "gateway");
                 });
 
             modelBuilder.Entity("CcDirector.Gateway.Data.Entities.GovernanceAuditEventEntity", b =>
                 {
                     b.Property<Guid>("Id")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("uuid");
 
                     b.Property<string>("Actor")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("Category")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("Detail")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("EventType")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<DateTime>("OccurredUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<DateTime>("RecordedUtc")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("timestamp with time zone");
 
                     b.Property<Guid?>("RunId")
-                        .HasColumnType("TEXT");
+                        .HasColumnType("uuid");
 
                     b.Property<string>("SessionId")
                         .IsRequired()
-                        .HasColumnType("TEXT");
+                        .HasColumnType("text");
 
                     b.Property<string>("TenantId")
                         .IsRequired()
-                        .HasColumnType("TEXT")
+                        .HasColumnType("text")
                         .HasColumnName("tenant_id");
 
                     b.HasKey("Id");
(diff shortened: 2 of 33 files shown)


thefrederiksen:
REJECTED pending rework.

The three composite-key mappings and the real-store two-tenant scenarios are directionally sound, but the durable guard does not prove the exemption it claims to prove.

`EveryTenantScopedTable_HasTenantIdInItsPrimaryKey_OrIsAnAllowlistedGatewayMintedGuidKey` checks only that an allowlisted entity still has a one-property `Guid` PK. That is not equivalent to ?the Gateway mints this value and no caller can present it.? If any allowlisted construction path changes from `Guid.NewGuid()` to a caller-supplied `Guid`, this guard remains green. The comment?s ?not taken on trust? and ?keeps the stated reason honest? claims are therefore false for the exact property that justifies the exemption. The test can fail for a string/type change, but cannot fail for the dangerous same-type source change. Please either tenant-compose those keys too, or encode Gateway-only minting in a structural API/type boundary that caller input cannot satisfy and test that invariant. An audited prose allowlist may document the current state, but it is not an executable future guard.

The submitted revert proof also changes all three independent PK mappings together and runs only a filtered 11-test slice. Snooze, session-spend, and push-subscription mappings are independently bypassable primitives. Required proof is a green, complete, serialized full Gateway baseline on the final head, then three separate full-suite mutation runs?one mapping reverted per run while the other two remain fixed?with named assertion failures, followed by an exact-head restored green run. A pre-existing red baseline cannot be reconciled as proof; the environment-race test has since been fixed and the suite lock is now on main.

This branch also merge-bases at `4fe2c3bb`, while current main changed `GatewayDbContext` and both provider snapshots in #1901 (`712ef1ff`). Rebase onto current main, inspect the overlapping range-diff, regenerate/reconcile both provider migrations and snapshots, rerun both pending-model-change checks, and rerun the evidence on that exact head. Because this changes live keys, add an actual upgrade test from a pre-change database containing rows through the new migration and verify preservation (SQLite, and Postgres when the gated environment is available); reasoning over generated operations alone does not execute the migration.

thefrederiksen:
Reworked and re-based. This addresses all three parts of the rejection. Nothing here has been run against the Gateway suite yet - the suite lock is held by another unit and I have not queued behind it. This comment pre-registers the runs I owe, and the reds I predict, BEFORE the runs happen.

## 1. The guard could not fail for the dangerous change

Accepted in full. Your words: *"If any allowlisted construction path changes from `Guid.NewGuid()` to a caller-supplied `Guid`, this guard remains green... The test can fail for a string/type change, but cannot fail for the dangerous same-type source change."* That was exactly true, and it made the comment's "not taken on trust" claim false for the one property the exemption rests on.

The exemption is no longer a list of table names with a written reason. It is a type, `GatewayMintedKeyEntity`, whose `Id` has a private setter and is minted by the base class itself. The ten entities derive from it and no longer declare their own `Id`; the thirteen `Id = Guid.NewGuid()` assignments are gone, because minting now lives in the type. The compiler enumerated those thirteen sites, and every one of them was a fresh `Guid.NewGuid()` - so the old prose claim was true *today*, and is now enforced rather than asserted.

Three escapes, each closed by construction and each verified by making the change and watching it fail:

| Escape | Result, observed |
|---|---|
| A store's key switches from minted to caller-supplied (`Id = Guid.Parse(r.SessionId!)` in `CronRunHistoryStore.ToEntity`) | **Does not compile.** Single error: `CronRunHistoryStore.cs(119,9): error CS0272: The property or indexer 'GatewayMintedKeyEntity.Id' cannot be used in this context because the set accessor is inaccessible` |
| Widen the setter to public so the above compiles again | Build then **succeeds** - precisely the state where the old guard stayed green - and `TheGatewayMintedKey_CanOnlyBeWrittenByTheBaseClassThatMintsIt` evaluates **red**: `setter.IsPrivate = False (public=True)` |
| Re-declare a settable `Id` on the derived entity (`public new Guid Id { get; set; }` on `CronRunEntity`) - still a single `Guid` key, the shape the old check accepted | Builds, and the primary-key test evaluates **red**, naming it: `CronRunEntity (key: Id; declared on: CronRunEntity)`. The test requires the mapped key property to be DECLARED ON `GatewayMintedKeyEntity`, not merely to be a `Guid`. |

Negative control: with every mutation reverted, both guards evaluate green on the pushed head. Each mutation was reverted individually and the tree confirmed clean before the next.

Caveat, stated plainly rather than glossed: those two guard reds were produced by executing the assertion logic of the tests verbatim against the real EF model and real reflection over the built assembly, in a standalone console harness, because running them inside the Gateway test host would take the machine-wide suite lock I am not admitted to. They are executed evidence of the mechanism, not xUnit reds. Both are pre-registered below as mutations for the admitted run, so they are also demonstrated in the harness that actually gates the build.

## 2. Three separate one-primitive mutations - pre-registered

The three mappings are independent primitives, so one per full-suite run, never combined. Predicted reds, by name. A predicted red that does not appear is a finding, not something to explain away.

| Run | Mutation (one line, all else left fixed) | Predicted red |
|---|---|---|
| M1 | `SnoozeEntity` key back to `HasKey(e => e.SessionId)` | `CompositeTenantKeyTests.TwoTenants_SnoozeTheSameSessionId_BothSucceed_AndStayIsolated`, plus `TenantScopeGuardTests.EveryTenantScopedTable_HasTenantIdInItsPrimaryKey_UnlessTheGatewayMintsTheKeyItself`, plus both `CallerSuppliedKeyUpgradePreservesRowsTests` |
| M2 | `SessionSpendEntity` key back to `HasKey(e => e.SessionId)` | `CompositeTenantKeyTests.TwoTenants_RecordSpendForTheSameSessionId_BothSucceed_AndStayIsolated`, plus the same guard test, plus `UpgradingADatabaseThatAlreadyHasRows_KeepsEveryRow_AndMakesTheKeyComposite` |
| M3 | `PushSubscriptionEntity` key back to `HasKey(e => e.Endpoint)` | `CompositeTenantKeyTests.TwoTenants_RegisterTheSamePushEndpoint_BothSucceed_AndStayIsolated`, plus the same guard test, plus `UpgradingADatabaseThatAlreadyHasRows_KeepsEveryRow_AndMakesTheKeyComposite` |
| M4 | `GatewayMintedKeyEntity.Id` setter widened to public | `TenantScopeGuardTests.TheGatewayMintedKey_CanOnlyBeWrittenByTheBaseClassThatMintsIt` |
| M5 | `public new Guid Id { get; set; }` re-declared on `CronRunEntity` | `TenantScopeGuardTests.EveryTenantScopedTable_HasTenantIdInItsPrimaryKey_UnlessTheGatewayMintsTheKeyItself` |

Each of M1 to M3 also requires its migration reverted so the model and migrations stay consistent; that is part of the mutation, not a second primitive.

Every red will be classified by how it ARRIVED - an assertion the test raised itself, versus a crash. A crash is laundered and proves nothing. Reconciliation will be computed: passed plus failed under the mutation equals passed under the restored run. The baseline will be a green AND explicitly complete serialized full-suite run taken fresh on this exact head, or I abort rather than reconcile against a red or stale one. A run with no summary line is not a result in either direction.

## 3. Baseline and base

Rebased. Merge base is now `f2028deb` (current `origin/main`), not `4fe2c3bb`. `git range-diff` across both rebases reports `=` - the content is unchanged, verified by range-diff rather than by a rebuild.

I inspected the overlapping range. #1901 changed `GatewayDbContext` and `TenantScopeGuardTests`, and its `TenantScopeGuardTests` change - adding `EntitlementEntity` to `allowedGlobalTables` - survived the rebase intact. It did NOT change either provider snapshot: `EntitlementEntity` is mapped `ExcludeFromMigrations`, and `entitlements` appears in neither snapshot, which I checked directly rather than inferred.

Both providers, re-run on this exact head: `dotnet ef migrations has-pending-model-changes` reports "No changes have been made to the model since the last migration." The base-class refactor is model-neutral, as the checks confirm rather than as I assumed - the property maps identically, and the declaring type is not recorded in the snapshot. The full solution builds clean.

## 4. The upgrade actually executes

You were right that reasoning over generated operations proves nothing here. On SQLite a primary key cannot be altered in place, so EF turns each operation pair into a table rebuild, and whether the copy carries every row is a property of that rebuild.

`CallerSuppliedKeyUpgradePreservesRowsTests` runs the real upgrade: migrate to `CompositeTenantKeys` (the migration immediately before this one), insert rows via raw SQL the way a database in the field holds them - two tenants, nullable columns both null and populated - migrate the rest of the way, then assert every row and every column value is unchanged. It also asserts the migration genuinely ran, by name out of `__EFMigrationsHistory` and by the key really being composite afterwards, so a migration that silently did nothing cannot pass by leaving rows undisturbed. A second test shows the upgrade delivers what it is for: before it the second tenant's insert is refused on the primary key, after it both rows coexist.

Executed on this head, outside the test host, against a real SQLite database:

```
PRE-CHANGE KEYS:  snoozes=[SessionId] session_spend=[SessionId] push_subscriptions=[Endpoint]
ROWS BEFORE:      snoozes=2 session_spend=2 push_subscriptions=2
POST-CHANGE KEYS: snoozes=[tenant_id,SessionId] session_spend=[tenant_id,SessionId] push_subscriptions=[tenant_id,Endpoint]
APPLIED:          20260719201700_CallerSuppliedKeysScopedByTenant
ROWS AFTER:       snoozes=2 session_spend=2 push_subscriptions=2
CROSS-TENANT SAME SESSION ID: ACCEPTED (correct)
RESULT: PASS - every row and every column value preserved
```

Postgres alters a primary key in place and never rebuilds a table, so it has strictly less to lose; it stays covered by the gated live-Postgres proof tests, which need an environment I do not have.

The stale claims in `docs/architecture/hosted-gateway-qa.md` are corrected in the same change - both the allowlist description and the "cannot fail on data" reasoning, which now cites the executed test rather than standing on its own.

Head: `6539e472`. Awaiting admission to the test lane for the five runs above, plus a restored green at that exact head.

thefrederiksen:
STATIC RE-REVIEW of `6539e472` ? the central guard objection is materially addressed, but this is not ready for approval or lane admission yet.

The new `GatewayMintedKeyEntity` is the right structural answer to the same-type blind spot I rejected: ordinary store/entity construction can no longer replace `Guid.NewGuid()` with a caller-supplied `Guid`; that assignment does not compile, setter widening is reflected directly, and the EF-key guard pins the mapped property to the base declaration. The SQLite upgrade tests now execute the real pre-change-to-current migration with rows and verify both preservation and the delivered composite key. The three composite mappings and store probes remain sound on inspection.

Remaining gates:

1. The PR description is still the pre-rework body. It still describes the rejected prose allowlist as current, still claims its single-`Guid` shape check ?keeps the stated reason honest,? still documents the old combined mutation, old base, and red baseline. The durable rework comment does not make the primary description true. Update the description to the pushed type-based design, current base, executed upgrade evidence, and explicitly pending lane runs before this is citable or mergeable.

2. My migration request included executing the upgrade with existing rows on Postgres when that gated provider is available. The new always-on test executes SQLite only. Existing Postgres tests migrate an empty/fresh schema and therefore exercise SQL generation/constraint names, but not an upgrade carrying rows across this exact change. Add the equivalent environment-gated pre-change-row upgrade/preservation test for the Postgres migration set. It may remain skipped without the configured service; the proof must exist and run when that lane is available rather than replacing execution with the same DDL reasoning I rejected.

3. The five pre-registered mutations are still explicitly unrun. After the description/test correction and after the manager?s required #1937-first convergence, this needs a final-head serialized proof: green complete baseline; M1/M2/M3 separately for the three PK mappings; M4 setter widening; M5 derived-key shadowing; exact restoration; restored green. That is seven full Gateway suite units total, with assertion-vs-crash classification and reconciliation on every arm.

Cross-PR convergence note: #1906 currently adds a new `WorkListEntity { Id = rowId }` construction site. That branch predates this private setter and cannot coexist unchanged. The clean order is to finish/land #1909 first, then rebase #1906 and let the base type mint the row ID before recording/asserting it. Do not run final evidence for both independent heads and then merge them without reconciling this overlap.

thefrederiksen:
Two static gates closed at `32998b54`. The third (the lane runs) is untouched - still not admitted, and nothing has been run against the Gateway suite.

**Gate 1 - the description.** Rewritten. It described the rejected prose allowlist as current, still carried the "keeps the stated reason honest" claim you showed to be false, and still documented the combined mutation, the old base and the red baseline. You were right that the durable comment does not make the primary description true - the body is what gets verified at the gate, so a body describing a design that no longer exists is the defect, not untidiness. It now states the type-based exemption, says plainly that assigning a caller-supplied key is a compile error rather than something a guard notices afterwards, carries the current base and the executed upgrade evidence for both providers, and marks the seven lane runs as outstanding and blocking.

**Gate 2 - the Postgres pre-change-row upgrade.** Added as `CallerSuppliedKeyUpgradePreservesRowsPostgresTests`, gated on `CC_GATEWAY_TEST_PG_CONNECTION` the same way `PostgresProviderProofTests` is, reporting SKIPPED when unset, and carrying the same `ccpg`-prefix throwaway guard because it drops and recreates.

I took your warning about gated tests that never run as the main risk here, so I did not leave it merely written. Executed against Postgres 16 in a throwaway container, on this head, through the Postgres migration set and the `gateway` schema:

```
PRE-CHANGE KEYS:  snoozes=[SessionId] session_spend=[SessionId] push_subscriptions=[Endpoint]
ROWS BEFORE:      snoozes=2 session_spend=2 push_subscriptions=2
POST-CHANGE KEYS: snoozes=[tenant_id,SessionId] session_spend=[tenant_id,SessionId] push_subscriptions=[tenant_id,Endpoint]
APPLIED:          20260719201740_CallerSuppliedKeysScopedByTenant
OK  the migration under test is recorded in the history table by name
OK  snoozes / session_spend / push_subscriptions: all rows and every column value preserved
OK  BEFORE the upgrade the second tenant is refused on the primary key (SqlState=23505)
OK  AFTER the upgrade the same insert is ACCEPTED
RESULT: PASS - all checks
```

And I validated the detector rather than trusting the pass. Re-run with only the upgrade suppressed, the four migration-actually-ran assertions go red while every row-preservation assertion stays green:

```
POST-CHANGE KEYS: snoozes=[SessionId] session_spend=[SessionId] push_subscriptions=[Endpoint]
APPLIED:
FAIL the migration under test is recorded in the history table by name
FAIL snoozes key is now (tenant_id, SessionId)
FAIL session_spend key is now (tenant_id, SessionId)
FAIL push_subscriptions key is now (tenant_id, Endpoint)
OK   snoozes: all 2 rows and every column value preserved
OK   session_spend: all 2 rows and every column value preserved
OK   push_subscriptions: all 2 rows and every column value preserved
```

That green stripe is the whole reason those anti-vacuity assertions exist: rows survive trivially when nothing touches them, so row preservation on its own would have passed a migration that never ran.

Both executions were run outside the xUnit host, in a console harness over the same context wiring and the same migration assembly, because running them in the test host would take the machine-wide suite lock I am not admitted to. Stated as a limitation, not offered as equivalent: it is the test's body executed against a real server, not the test.

**Gate 3 - the lane.** Unchanged and unrun. Seven serialized full-suite units: baseline, M1 through M5, restored. One primitive per run, predicted reds by name in the earlier comment, assertion-versus-crash classification on every red, computed reconciliation. Standing by for admission and for the #1937-first convergence.

Two things worth flagging before those runs happen.

*The Postgres tests will report SKIPPED on the admission machine*, so they contribute nothing to the seven runs and must not be counted as coverage there. Their evidence is the container execution above. My earlier pre-registration named "both `CallerSuppliedKeyUpgradePreservesRowsTests`" as predicted reds for M1 - that means the SQLite class only, and no Postgres test should be expected to redden in any arm.

*A restore hazard that would silently corrupt the mutation reconciliation.* While producing the evidence above I restored a mutated source file by moving a backup over it, re-ran, and got the MUTATION's output back from a source that no longer contained it - the restored file carried the backup's older modification time, so MSBuild judged the build up to date and re-ran the previous assembly. Confirmed deliberately afterwards with a marker line: after the restore the binary still printed `MUTATED MARKER`, with the source timestamp twelve seconds behind the assembly. `git checkout --` stamps a current time and does rebuild, which I also verified. The dangerous direction is not the one I hit: if an arm's mutation is introduced by a path that preserves an old timestamp, the mutation never reaches the assembly, the run comes back green, and it reads as a predicted red that did not appear - a finding manufactured out of a stale build. Every arm should mutate by editing and restore by `git checkout --`, never by copying a backup back.

Head `32998b54`, local equals remote, full solution builds, both providers report no pending model changes.

thefrederiksen:
Correction to my previous comment, which is now false on one point.

I wrote "standing by for admission and for the #1937-first convergence." **That is wrong: #1937 is not a gate for this unit** and I should not have recorded it as one. It was found to have load-bearing bypasses of its own and no longer merges into the next gap; it lands whenever it is reworked and re-verified. Nothing about this unit waits on it. This unit is blocked on the test lane and nothing else, and #1906 is blocked on this one by the compile error.

The general rule is what makes that safe, and it is worth stating as a commitment rather than an understanding, because it constrains how I run the arms:

**Every unit takes a fresh green-and-complete baseline at its own head and reconciles against THAT, never against a shared constant or a number quoted from another unit.** The lock serialises, so merges land in the gaps between units, never mid-unit. The only thing I have to guarantee is that my baseline and my seven arms come from the same world.

So, concretely, for the seven runs:

- The baseline is taken fresh at whatever head this branch is on when I am admitted. It is not inherited from any earlier run of mine, including the ones already recorded here, and not from any other unit's numbers.
- **I will not rebase mid-proof.** Once the baseline is taken, that head is frozen for all seven arms. If `origin/main` moves while I am running, it moves without me; I finish the arms, then rebase, and if the rebase changes anything that the proof covered, the proof is re-run rather than carried over.
- A baseline that is not both green AND explicitly complete aborts the sequence rather than being reconciled against. A run with no summary line is not a result in either direction.

Whether #1937 has landed by my turn is irrelevant to this unit, and I will not treat its state as a precondition either way.

Continuous integration on `32998b54` is in progress; I will report it here when it resolves. Head unchanged, local equals remote, nothing has been run against the Gateway suite.

thefrederiksen:
Head is now `28aa59b6`, superseding `32998b54`. One change, described below, plus the re-run trigger I am holding myself to.

## The re-run trigger I had was too narrow

I had written that a rebase forces a re-run "if the rebase touches anything the proof covered." That is the wrong test. A merge can void a proof **without touching a single file the proof covers**: an unrelated change lands that alters what a fixture actually puts in place, and the proof then asserts against an emptier world while staying green. A clean `range-diff` proves the branch's TEXT is unchanged; it proves nothing about whether the proof still MEANS what it did.

The trigger I am holding is: **re-run when the merge changes anything my fixtures SEED or my assertions ASSUME**, not when it touches files I touched. In practice that means re-reading my own fixtures against the post-rebase code, rather than diffing my branch.

So I did that read now, ahead of time, and wrote down what this proof actually assumes. Classified by how each assumption would break, because the loud ones need no ceremony and the silent ones do:

| # | What the proof assumes about the world | If another unit breaks it |
|---|---|---|
| 1 | `CompositeTenantKeys` is the migration IMMEDIATELY before this one, per provider | **SILENT** - see below |
| 2 | The three tables' pre-change keys are the single columns | LOUD - the pre-change key assertions fail |
| 3 | The exact seeded column names exist (`SnoozeUntilUtc`, `MeteredCostMicros`, `P256dh`, ...) | LOUD, but arrives as a **crash**, not an assertion - see the note below |
| 4 | The tenant discriminator column is named `tenant_id` | LOUD (crash, same note) |
| 5 | Postgres tables live in schema `gateway` | LOUD (crash, same note) |
| 6 | The seed actually inserted rows - the empty-fixture case | **Already guarded**: `Assert.Equal(2, ...Before.Count)` on all three tables, both providers. Without it, an emptied fixture would compare nothing to nothing and pass. |
| 7 | The migration under test actually ran | Already guarded, by name out of the history table plus the key really being composite; validated by the suppressed-upgrade negative control recorded earlier |
| 8 | A nullable column added to these tables by someone else | Benign - the seed lists columns explicitly, and `SELECT *` carries the new column as null on both sides |

## The silent one is now a mechanism, not a checklist item

Assumption 1 was the only one that breaks quietly, and it breaks in exactly the shape described above. If another branch lands a migration between `CompositeTenantKeys` and this one, nothing in my test files changes, a range-diff of this branch is identical, and **every assertion still passes** - the rows are still preserved, the key still ends up composite. The test would simply have stopped isolating this change: it would be carrying rows across someone else's migration too, and reporting the result as evidence about this one.

Both upgrade proofs now assert it instead of assuming it, per provider, since the two migration sets are generated separately and can drift independently:

```
AssertThisProofStillIsolatesTheMigrationUnderTest()
```

It reads the ordered migration set and requires the entry immediately before `CallerSuppliedKeysScopedByTenant` to be `CompositeTenantKeys`, naming the actual predecessor when it is not, and telling the reader to re-read the fixture against the new predecessor rather than just re-point the constant.

Validated in both directions rather than assumed, against the real migration set:

```
MIGRATION SET (last 3): AddTenants -> CompositeTenantKeys -> CallerSuppliedKeysScopedByTenant

MigrationBeforeTheChange = "CompositeTenantKeys"
  GUARD GREEN: the proof still isolates the migration under test

MigrationBeforeTheChange = "AddTenants"
  GUARD RED: assumes 'AddTenants' is immediately before 'CallerSuppliedKeysScopedByTenant', but the
  migration before it is now '20260719064519_CompositeTenantKeys'. Another change has landed in between,
  so this test would be carrying rows across that other migration too while reporting the result as
  evidence about this one.
```

## A note that connects two of our rules, and cuts against me

Assumptions 3, 4 and 5 fail LOUDLY, but they fail as **exceptions, not as assertions** - a bad column name throws out of the fixture's `INSERT` before any assertion is reached. Under our own classification rule a crash is laundered and proves nothing, so if one of those breaks during an arm, the resulting red is **not** that arm's predicted red and must not be reconciled as one. It is a broken fixture wearing a red arm's clothes. I will classify every red in every arm by where it arrived - fixture, migration, or assertion - and not merely by whether the arm was red.

## Unchanged

Everything else stands: seven serialized arms, one primitive each, fresh green-and-complete baseline at my own head, head frozen across all seven, no rebase mid-proof, abort rather than reconcile against a baseline that is not both green and complete. The Postgres tests will report SKIPPED on the admission machine and contribute nothing there.

Continuous integration is now running against `28aa59b6`; `32998b54` is superseded and its result is not the one to read. Local equals remote, tree clean, full solution builds, both providers report no pending model changes. Nothing has been run against the Gateway suite.

thefrederiksen:
Continuous integration is **green on `28aa59b6`** - run [29728235106](https://github.com/thefrederiksen/devthrottle/actions/runs/29728235106), `headSha` confirmed to match the branch head, both jobs pass (`.NET` 18m45s, `web` 57s).

Green means nothing failed, not that anything ran, so I read the log rather than the badge. The `.NET` job runs the full suites:

```
CcDirector.Core.Tests      Total: 3320   Passed: 3312   Skipped: 8    Test Run Successful.
CcDirector.Gateway.Tests   Total: 3018   Passed: 3006   Skipped: 12   Test Run Successful.
```

Confirmed by name, not by total:

| Test | Verdict |
|---|---|
| `CallerSuppliedKeyUpgradePreservesRowsTests.UpgradingADatabaseThatAlreadyHasRows_KeepsEveryRow_AndMakesTheKeyComposite` | Passed |
| `CallerSuppliedKeyUpgradePreservesRowsTests.AfterTheUpgrade_TwoTenantsCanHoldTheSameCallerSuppliedIdentifier` | Passed |
| `CallerSuppliedKeyUpgradePreservesRowsPostgresTests.UpgradingAPostgresDatabaseThatAlreadyHasRows_KeepsEveryRow_AndMakesTheKeyComposite` | **Skipped** (as designed) |
| `CallerSuppliedKeyUpgradePreservesRowsPostgresTests.AfterThePostgresUpgrade_TwoTenantsCanHoldTheSameCallerSuppliedIdentifier` | **Skipped** (as designed) |
| `TenantScopeGuardTests.EveryTenantScopedTable_HasTenantIdInItsPrimaryKey_UnlessTheGatewayMintsTheKeyItself` | Passed |
| `TenantScopeGuardTests.TheGatewayMintedKey_CanOnlyBeWrittenByTheBaseClassThatMintsIt` | Passed |
| `CompositeTenantKeyTests.TwoTenants_SnoozeTheSameSessionId_BothSucceed_AndStayIsolated` | Passed |
| `CompositeTenantKeyTests.TwoTenants_RecordSpendForTheSameSessionId_BothSucceed_AndStayIsolated` | Passed |
| `CompositeTenantKeyTests.TwoTenants_RegisterTheSamePushEndpoint_BothSucceed_AndStayIsolated` | Passed |

**The skip count reconciles exactly.** All 12 Gateway skips are Postgres-gated and nothing else is skipped: 10 pre-existing (`GatewayDatabaseLivePostgresProofTests` x4, `PostgresProviderProofTests` x5, `GatewayHostBootSmokeTests` x1) plus the 2 added here. So the new gate behaves exactly like the established one - reported SKIPPED rather than passed - and it introduced no accidental skips anywhere else.

That also confirms the prediction I owed: the Postgres upgrade proofs contribute nothing on an unconfigured machine, so **no Postgres red should be expected in any of the seven arms**, and their absence is not a missing predicted red. Their evidence remains the executed container run recorded earlier.

## What this does and does not settle

It settles that the branch builds and the full Gateway suite passes at this exact head, on a clean machine, with the new tests executing by name.

It does **not** substitute for the seven serialized arms. Continuous integration proves the suite is green with the fix in place; it says nothing about whether the suite would go RED with the fix removed, which is the only thing that establishes these tests actually detect the defect. A test that passes both with and without the fix is the anti-coverage case this branch already had to remove once.

Still owed and unchanged: fresh green-and-complete baseline at my own head, M1 through M5 one primitive each, exact restoration, restored green - head frozen across all seven, no rebase mid-proof, every red classified by where it arrived.

Head `28aa59b6`, local equals remote, tree clean. Nothing has been run against the local Gateway suite.

thefrederiksen:
STANDING: REJECTED
28aa59b6b05d8074579a772ade4114d9c4bd9060

REASON-TO-TRUST CHECK: the single-Guid exemption is trusted only if no caller path can ever choose the key. The current guards prove the setter's accessibility and the mapped property's declaration, but not the complete construction/write surface that justifies that claim. A caller-fed write path can be added to the exempt base while both checked facts—and both guards—remain green.

Constrain and test the full key-creation surface, not only direct property assignment. Add a negative control that introduces a caller-fed base write path, proves ordinary production construction can use it, and requires the durable guard to redden; if the escape is not made structurally impossible, it also needs its own pre-registered arm. Security details and a concrete reproduction shape were reported privately.

The composite mappings, executed provider upgrades, current CI, and existing M1-M5 registration remain useful, but they cannot clear an exemption whose claimed boundary is still safe by today's implementation rather than enforced for the change it is meant to prevent.

PR #1910, opened 2026-07-19
fix(setup): stop blocking the install on three prerequisites that are not required

thefrederiksen:
A brand-new user on a clean Windows machine could not get past the Prerequisites screen. Four items were marked required and `NextButton.IsEnabled = allRequiredMet` stopped them dead until they left the wizard, installed three separate programs by hand, and came back. That is the single largest block in the path from downloading DevThrottle to running a first agent, and three quarters of it was never necessary.

## Only one of the four is genuinely required

| Prerequisite | What actually breaks without it | Required? |
|---|---|---|
| **.NET 10 Runtime** | The Director does not start. The Windows build publishes `--self-contained false`, so it needs the shared runtime. | **Yes** - and Setup already installs it |
| Python | **Nothing.** The `cc-*` tools ship their own relocatable CPython (`cc-python-win-x64.zip`, staged by `PythonToolsInstaller`) and never use the system interpreter. | No |
| Node.js | MCP servers and browser tools only. | No |
| Claude Code | Only the Claude agent. The Director runs eight agent command line tools (`src/CcDirector.Core/Agents/`). | No |

That last row was turning people away: a user who came for Codex or Gemini was stopped for not having a **different vendor's** tool installed.

## The order matters - please do not "simplify" it later

**Re-classify first, auto-install second, and leave the gate expression alone.**

Warn-and-continue *on its own* would be wrong: it would walk a user through to a Director that cannot start, because the .NET runtime really is load-bearing. Re-classification is what makes continuing safe - with only the runtime required, the gate is satisfied by the one item Setup can install itself, so nobody is ever stopped **and** the gate still protects the one real dependency. The expression at `MainWindow.xaml.cs` is unchanged in meaning; it simply has less to gate.

## What changed

- **Re-classification.** Claude Code, Python and Node.js become recommended. They are still checked and still shown; they no longer gate.
- **Auto-install, as data not logic.** The three gain the winget ids the existing generic path already knows how to drive - `Anthropic.ClaudeCode`, `Python.Python.3.12`, `OpenJS.NodeJS.LTS`. `RuntimeInstaller.InstallAsync` already takes any package id and already runs `--silent --accept-package-agreements --accept-source-agreements --disable-interactivity`, and `PrerequisiteInfo` already carries `CanAutoInstall` / `WingetId` / `ShowAutoInstall`. Each missing row now offers a one-click install with **per-row** progress ("Installing...") and a re-check on success - no single opaque wait.
- **The manual path survives.** Every row keeps its download link, because winget is absent on locked-down machines. A test asserts it on every row.
- **Degraded capabilities are reported at the END, not as a wall at the start.** New `CapabilityNotice` (shared, UI-free) computes one sentence naming each missing item and what it costs - "Claude Code: no coding agent is installed yet, so your board has nothing to run" - and the Complete screen renders it. The Prerequisites subtitle also stops claiming "All required prerequisites found" while three rows sit unfound.
- **One gate, one place.** `PrerequisiteChecker.AllRequiredMet` replaces the predicate that was duplicated in `MainWindow` and `PrerequisitesStep`, so the Next button and the screen's own subtitle cannot disagree.
- **macOS wizard** gets the same re-classification. It carries no runtime row at all, because that build is self-contained - so on macOS nothing is required, which is correct.
- **Docs** updated: the installation guide and the wizard walkthrough both said four were required.

## Verified

- `CcDirectorSetup.Tests` **40 passed** (was 32), `CcDirector.Setup.Engine.Tests` **345 passed** (was 340), `CcDirector.Setup.Cli.Tests` **28 passed**
- `CcDirector.Core.Tests` 3306 passed, `CcDirector.Avalonia.Tests` 247, `CcDirector.Engine.Tests` 62, `CcDirector.HostedAgent.Tests` 86, `CcDirector.Launcher.Tests` 27, `CcDirector.Terminal.Avalonia.Tests` 21
- Release builds of the WPF wizard, the Avalonia wizard and the Director: clean, zero warnings
- **Revert-proof on the production line:** setting `IsRequired = true` back on the real Claude Code row in `PrerequisiteChecker.CreateChecklist` turns two tests red - `CreateChecklist_OnlyDotNetIsRequired` and `AllRequiredMet_UserWithOnlyTheDotNetRuntime_CanContinue`. Production was restored before the final run.

## NOT verified

- **No winget install was run, and no installer was executed.** I confirmed all three package ids exist in the live winget catalogue (`Anthropic.ClaudeCode` 2.1.215 - the exact version the checker detects, `Python.Python.3.12` 3.12.10, `OpenJS.NodeJS.LTS` 24.18.0). Unattended install success, and the `PATH`-staleness behaviour that follows it, need a real machine and are for human QA.
- GUI rendering of the new Complete-screen panel was not seen; its content is covered by tests, its layout is not.

## Deliberately out of scope

The Anthropic account question. Installing Claude Code needs no account and no payment - the account is required only when an agent actually runs. That is a different problem on a different leg and should not be solved here.

### DIFF
--- a/docs/plans/install-autoupdate.md
+++ b/docs/plans/install-autoupdate.md
@@ -91,7 +91,8 @@ reinventing it.
 | D6 | Trigger | **Resident apps orchestrate** - Director (while open) and Gateway (always) run the CLI updater for all present components |
 | D7 | Gateway update privilege | **Self-updating service** - the LocalSystem service swaps + restarts its own build; no UAC after first install |
 | D8 | Update failure handling | **Keep `.old` backups + manual rollback** (`...setup rollback <component>`); no auto-rollback/health-check |
-| D9 | Framework bootstrap | **Detect + guide** (link to Claude Code / Codex official installer + Re-check); we never own their install mechanics |
+| D9 | Framework bootstrap | ~~**Detect + guide** (link to Claude Code / Codex official installer + Re-check); we never own their install mechanics~~ **SUPERSEDED (see D9a)** |
+| D9a | Framework bootstrap, revised | **Detect + offer + guide.** The graphical wizard offers a one-click `winget` install for a missing agent, and the download link stays as the fallback. Detect-and-guide alone was the single largest block in the path from download to a first agent: it stopped a new user on the Prerequisites screen and sent them off to install three programs by hand. We still do not *own* the install mechanics - we invoke the vendor's own published package (`Anthropic.ClaudeCode`), the user chooses to click, and the manual link is untouched. The headless CLI keeps detect-and-guide unchanged. |
 | D10 | Code signing | **Stay unsigned, document** the SmartScreen "More info -> Run anyway" step; revisit for external users |
 
 ---
@@ -237,10 +238,14 @@ read commands for agent consumption. No window in CLI mode.
 
 ### 4.8 Framework bootstrap (D9)
 
-`PrerequisiteChecker` detects Claude Code / Codex. If missing: the UI shows the
-official install link + a "Re-check" button; the CLI prints the link and exits
-with a distinct "prerequisite missing" code. The installer never runs the
-framework's installer itself.
+`PrerequisiteChecker` detects Claude Code / Codex. If missing: the graphical wizard
+shows an **Install automatically** action (the vendor's own `winget` package, run
+silently) **and** the official install link + a "Re-check" button. The CLI is
+unchanged - it prints the link and exits with a distinct "prerequisite missing" code.
+
+Superseded D9 (detect-and-guide only): the agent is no longer a required
+prerequisite either, so a missing agent never blocks the install. What the user gave
+up is reported on the Complete screen by `CapabilityNotice`.
 
 ---
 
--- a/docs/public/getting-started/02-installation.md
+++ b/docs/public/getting-started/02-installation.md
@@ -14,7 +14,7 @@ The DevThrottle **Setup** app checks for these on its Prerequisites screen. On W
 | [Node.js](#nodejs) | Recommended | 20+ | MCP servers and the browser tools |
 | [Tailscale](#tailscale-optional--remote-access) | Optional | latest | Reaching this gateway's Cockpit from a phone or another computer |
 
-Setup can install every one of these for you on Windows via `winget`; each row shows an **Install automatically** action while the tool is missing. If `winget` is unavailable, use the download link in the same row and click **Re-check**.
+On **Windows**, Setup can install every one of these for you via `winget`: each row shows an **Install automatically** action while the tool is missing. If `winget` is unavailable -- it is absent on some locked-down machines -- use the download link in the same row and click **Re-check**. On **macOS** there is no `winget`, so the links are the install path.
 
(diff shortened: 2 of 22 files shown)


thefrederiksen:
## Second pass: independent review found real defects, all fixed in 64926300

An adversarial review was run against the diff. It confirmed the central claim - only .NET is genuinely required (it traced the cc-* shims to the bundled relocatable CPython, confirmed nothing shells out to `node`/`npm`, and confirmed agent detection is a pure `File.Exists` PATH walk that spawns nothing) - and confirmed `AllRequiredMet` is exactly equivalent to the two predicates it replaced, empty list included.

It also found eight things wrong. Every one is fixed:

| # | Defect | Fix |
|---|---|---|
| A | **macOS was left half-done.** With nothing required there, the subtitle always took the "all required prerequisites found" branch - a mac user with all three rows reading Not found saw a green banner saying everything was found. It also rendered **no** capability notice, on the one platform with no winget. | Both wizards now share one gate and one notice; macOS subtitle tells the truth. |
| B | **The notice told Codex and Gemini users their board had nothing to run** - keyed purely on `claude` being absent, repeating in words the exact mistake this change removes. | New `AgentPresence` checks PATH for the seven other agent tools; wording follows. |
| C | **Every winget failure said "install .NET 10 manually"**, whichever row you clicked. | `RuntimeInstaller` names the row the user clicked. |
| D | **Tailscale swept into the notice** by `!IsRequired` - a gateway with Tailscale running but MagicDNS off would be told it was not installed. | Explicit `IsRecommended`; only recommended rows reach the notice. |
| E | **"Not installed" asserted for things that are installed** - `IsFound` is false for a Python 3.9 the row itself calls "Too old". | Prefix is now "Missing or out of date". |
| G | **Badge still read "Optional"** for every non-required row, making Claude Code indistinguishable from Tailscale. | Bound to `ImportanceLabel`: Required / Recommended / Optional. |
| H | **Docs and a recorded decision unreconciled.** | Install docs corrected (no winget on macOS; Claude Code package named). **Decision D9** - "detect + guide; we never own their install mechanics" - is explicitly **superseded by D9a**, in writing, rather than reversed in silence. |
| I | **Magic-string linkage untested** - a row rename would have silently degraded every user to "some features are unavailable", all tests green. | Row names are shared `PrerequisiteNames` constants; a test asserts every recommended name has its own consequence. |

### Verified after the fixes
setup engine **349 passed**, setup **44 passed**, setup cli **28 passed**, Avalonia 247, Engine 62. Release builds of both wizards and the Director clean, zero warnings. Revert-proof: removing the agent-aware branch reds `Describe_UserRunsAnotherAgent_DoesNotClaimTheyHaveNothingToRun`; putting `IsRequired = true` back on the Claude Code row still reds two classification tests.

### Still not verified
No winget install was run and no installer was executed - catalogue presence only.

### Known and deliberately not fixed (finding F)
After a successful winget install, `CheckPython`/`CheckNode` can still report Not found in the same session, because Windows resolves a child image name from the **parent's** PATH rather than the child's environment block. `CheckExecutable` (Claude Code) treats that same failure as found. That asymmetry is pre-existing, **Re-check** clears it, and changing detection semantics did not belong in this change. Flagged rather than silently carried.

PR #1915, opened 2026-07-19
fix(gateway): partition the voice state per tenant

thefrederiksen:
Partitions the voice STATE per tenant. Cut from `origin/main`, rebased onto lock-bearing main (`e1b8825c`, carrying #1918, #1916 and v1.6.0). Final head `4c2a8ebc`.

## What is partitioned

The voice state held one account's spoken narration, the reply text it was made from, and the audio clip itself, all keyed by bare session id in process-global dictionaries and one shared folder. On the hosted Gateway that is customer content reachable by whoever names the session id.

- **`WingmanVoiceService`** - one bucket of per-session dictionaries per tenant, plus one DIRECTORY per tenant on disk (`tenants/<id>/voice-sessions.json`, `tenants/<id>/voice-audio/`).
- **`VoiceTurnArchive`** - the same per-tenant directory for turn transcripts, summaries and reply audio.
- **`GatewayTurnJobStore`** - a per-tenant pair of dictionaries. Knowing an unguessable turn id is not owning the turn.

Every public read and mutate takes `TenantId` as a **required first parameter** - not optional, not nullable, not defaulted, never falling back to the local tenant. An unresolved tenant is refused, and a test passes `default` and asserts it throws. The tenant id is canonicalized before becoming a bucket key or path component; a shape this system does not mint - traversal, case variant, the reserved system identity - is refused rather than normalised.

**No entity model changed, so there is no migration in this pull request.**

## Legacy migration, asymmetric by deployment mode

Pre-partition state carries no tenant, so the mode is read from `GatewayHostedMode.IsHosted` **directly**, never from an argument a caller could omit - an optional argument fails open into "keep it".

- **Hosted: DELETE.** A clip whose owner cannot be established must not be handed to a guess. A cached narration is regenerable; a mis-attributed one is a disclosure.
- **Self-host: MOVE into the local partition.** One tenant, so attribution is unambiguous.

### Why not one combined run

The first attempt mutated all four boundaries together and produced a tidy table attributing each red to a primitive. **That table was not sound, and one row proved it:** `Clips_reload_into_the_tenant_that_owned_them` had to be listed as "B1+B2", because a run with everything mutated cannot say which mutation caused any given red. Once one red is unassignable the rest is inference too.

Separated, three of the four arms turned out **mis-attributed** and one turned out right.

> That arm happened to be right - but it was right by luck, and I had no way to know which arms were lucky until each was run alone.

Being right by luck is indistinguishable from being right by evidence, from the inside. A method that is correct on an unknown subset is not a method.

**And it could not have been argued instead of run:** `VoiceTurnArchive` and `WingmanVoiceService` **both** expose a method called `PartitionDirectoryFor`. Under a combined mutation a red in either could be attributed to "the directory primitive" with a completely straight face - the names agree, the story is coherent - and it would be wrong half the time.

### Results

Each arm preceded by a build confirmed succeeded, and by a harness check that **exactly one** production file differed from pristine (it aborts otherwise - a mechanism, not care, closing the stale-edit hole).

Restored baseline at final head: **2979 total, 2969 passed, 0 failed, 10 skipped.**

| Primitive mutated alone | Passed | Failed | Reconciliation | Crashes |
|---|---|---|---|---|
| **B1** `WingmanVoiceService.StateFor` (bucket) | 2960 | 9 | 2960+9=2969 OK | 0 |
| **B2** `WingmanVoiceService.PartitionDirectoryFor` (dir) | 2962 | 7 | 2962+7=2969 OK | 0 |
| **B3** `VoiceTurnArchive.PartitionDirectoryFor` (dir) | 2967 | 2 | 2967+2=2969 OK | 0 |
| **B4** `GatewayTurnJobStore.StateFor` (bucket) | 2968 | 1 | 2968+1=2969 OK | 0 |

**Four independent reconciliations, four chances to catch a dropped test, none dropped. Nineteen reds, every one an ASSERTION - not a single crash in any arm.**

### The sensitivity matrix

The durable copy lives in the `WingmanVoiceTenantPartitionTests` class doc, not only here: a pull request is read once by people who already have the context; that comment is read by whoever is standing there next, who has none.

| Test | B1 | B2 | B3 | B4 |
|---|---|---|---|---|
| `Ready_audio_stored_for_one_tenant_is_invisible_to_another` | RED | green | green | green |
| `Voice_session_marking_is_per_tenant` | RED | green | green | green |
| `Generating_unavailable_and_nothing_to_narrate_are_per_tenant` | RED | green | green | green |
| `Served_via_fallback_is_per_tenant` | RED | green | green | green |
| `Clearing_one_tenants_session_leaves_..._intact` | RED | green | green | green |
| `Regeneration_decision_reads_only_the_asking_tenants_cached_reply` | RED | green | green | green |
| `Each_tenants_clips_live_in_its_own_directory` | green | RED | green | green |
| `Clips_reload_into_the_tenant_that_owned_them` | RED | RED | green | green |
| `Voice_session_set_reloads_into_the_tenant_that_owned_it` | RED | RED | green | green |
| `Self_host_moves_the_pre_partition_voice_state_into_the_local_partition` | RED | RED | green | green |
| `Turn_archive_is_partitioned_and_a_turn_id_alone_does_not_read_it` | green | green | RED | green |
| `Self_host_moves_pre_partition_archived_turns_into_the_local_partition` | green | green | RED | green |
| `Turn_job_store_is_partitioned_and_a_turn_id_alone_does_not_read_it` | green | green | green | RED |
| `WingmanVoiceServiceTests.ReadyAudio_PersistsAndReloadsAcrossRestart` | green | RED | green | green |
| `WingmanVoiceServiceTests.ReadyAudio_ReloadsLegacyWavCacheWithDetectedContentType` | green | RED | green | green |
| `WingmanVoiceFallbackTests.ServedViaFallback_SurvivesAGatewayRestart` | green | RED | green | green |

Every cell is OBSERVED, with exactly one meaning: `RED` = reddened as an assertion (proof), `EXCL` = reddened as a crash (excluded, never proof), `green` = ran and passed, `NOTRUN` = no outcome line (evidence of nothing). **There is no EXCL cell and no NOTRUN cell.**

Every row has at least one RED, so no test here is decorative. B3 and B4 are perfectly isolated, which is what proves the archive and job store are guarded **separately** rather than incidentally.

### Pre-registered prediction, confirmed

Committed at 20:54:47 (`8c853a54`) with the B1 run verified still executing, so it cannot have been written afterwards. Predicted: the three pre-existing tests are sensitive to the **directory**, not the **bucket**. Observed: B1 alone leaves all three green; B2 alone reddens all three.

**If you move the voice on-disk layout: three tests in two other classes go red, and none of them has "tenant" in the name.** That is the durable fact this exercise produced.

## Two rules this proof paid for

**Rule one - when comparing observations from different builds, compare properties that CANNOT shift.** I nearly proved the straddle from line numbers: bucket failed at 306, directory at 322, therefore different assertions. Invalid - I had added seventeen lines of *comment* between the builds, so 306 maps to ~323, within one line of 322. **A change with no behaviour whatsoever was enough to invert the conclusion.** What settles it is the assertion TYPE (`Assert.False` vs `Assert.True`), which cannot shift.

**Rule two - a claim of the form "X and not Y" is not supported by an experiment that only varied X.** After B1 I published "these two are bucket-sensitive and NOT directory-sensitive". B2 showed both. Worse than an ordinary wrong guess: it generalises from a single arm while the refuting arm is unrun, and arrives dressed as a finding rather than a hypothesis. The honest form after one arm is **"X is sufficient; whether Y also suffices is UNTESTED."**

Both are recorded in the test file, along with the two straddling tests: **isolation rides on the bucket, reload rides on the directory**, and the failure kind is how you tell.

## Anti-laundering, structural

Every canary asks its cross-tenant question **first**, before any control, with every dereference guarded by an explicit null assertion. A sibling pull request's rule - assert status and media type before parsing - does not literally apply here (no HTTP calls), but what it is *for* does: an operation that throws on unexpected shape makes an unstated assertion, and when it throws you learn only that something upstream broke. The filesystem form is asserting `File.Exists` before `ReadAllBytes`. That is why nineteen reds are nineteen assertions.

## Tests confirmed EXECUTED, by name

A green suite does not tell you your tests ran; it tells you nothing failed, and those differ exactly when a harness drops a class. All **20** declared `[Fact]`/`[Theory]` methods produced outcome lines in the restored run, listed individually. The three pre-existing tests and the retired `Enrolled_without_a_recorded_email` allowance were each confirmed passing by name.

## Test-harness defects found and fixed

**Isolation only as deep as the deepest derived path component.** The harness randomised the persist **filename** in the shared temp directory - correct while the file was the deepest thing derived from. Partitioning made the service derive its per-tenant root from that file's **parent**, so every instance collapsed onto one shared `tenants/local` and a clip written by one test was loaded by the next test's constructor. That is the continuous-integration failure in `StoreSpokenAsync_WithFailingTts_DoesNotMarkReady`. Each instance now gets a unique **directory**; four restart pairs keep a deliberately shared root, documented at each call site.

**A test passing by coincidence.** `ReadyAudio_ReloadsLegacyWavCacheWithDetectedContentType` seeded the pre-partition location and only kept passing because the migration moved the file before the load ran - silently exercising migration, not detection, and it would have passed with the partition deleted. It now seeds the location the service reads.

**An assertion true by arithmetic.** `Each_tenants_clips_live_in_its_own_directory` compared two paths the *test* had built from the tenant ids - true regardless of production code. It now reads both paths back from the production method, and it fires on the directory mutation.

## The suite lock (#1918), including its negative control

| Run | Queued | Acquired | Wait |
|---|---|---|---|
| B1 | 20:45:00 | 20:50:07 | 5m07s |
| B2 | 20:58:05 | 21:04:45 | 6m40s |
| B3 | 21:13:53 | 21:18:51 | 4m58s |
| B4 | - | 21:26:28 | **none** |
| restored | - | 21:35:59 | **none** |

Every wait named the holder precisely (`fix-1848-stats`, session `c584f16f`). The last two are the **negative control**: once that worktree finished, the lock stopped blocking. Seven observations show it blocks when it should; these show it does not block when it should not. A lock that always blocked would have produced seven identical-looking successes and been catastrophically wrong.

## Deliberately left for the route and sweep worker

The `/wingman/voice*` route handlers, `/wingman/voice/ready`, the narration sweep and turn-end refresh all still pass `TenantId.Local` - correct on self-host, empty on hosted, so they degrade to a no-op rather than a wrong-tenant read. Converting a route before its state is partitioned arms a cross-tenant audio path instead of closing one. The per-tenant form of the sweep is a loop over `PartitionedTenants()`.

## Test counts

| Project | Result |
|---|---|
| Gateway | 2979 total, **2969 passed**, 0 failed, 10 skipped |
| Core | 3306 passed, 8 skipped |
| Engine | 62 passed |
| HostedAgent | 86 passed |
| Launcher | 27 passed |
| Avalonia | 247 passed |
| Terminal.Avalonia | 21 passed |

### DIFF
--- a/src/CcDirector.Gateway.Tests/GatewayTurnJobStoreTests.cs
+++ b/src/CcDirector.Gateway.Tests/GatewayTurnJobStoreTests.cs
@@ -1,5 +1,6 @@
 using CcDirector.Gateway.Voice;
 using Xunit;
+using CcDirector.Core.Tenancy;
 
 namespace CcDirector.Gateway.Tests;
 
@@ -15,8 +16,8 @@ public sealed class GatewayTurnJobStoreTests
         var store = new GatewayTurnJobStore();
         var uploadId = Guid.NewGuid().ToString("N");
 
-        var job = store.Create("sid-1", uploadId);
-        var found = store.FindTurnByUpload(uploadId);
+        var job = store.Create(TenantId.Local, "sid-1", uploadId);
+        var found = store.FindTurnByUpload(TenantId.Local, uploadId);
 
         Assert.NotNull(found);
         Assert.Equal(job.TurnId, found!.TurnId);
@@ -27,8 +28,8 @@ public sealed class GatewayTurnJobStoreTests
     public void FindTurnByUpload_UnknownOrEmpty_ReturnsNull()
     {
         var store = new GatewayTurnJobStore();
-        Assert.Null(store.FindTurnByUpload(Guid.NewGuid().ToString("N")));
-        Assert.Null(store.FindTurnByUpload(""));
+        Assert.Null(store.FindTurnByUpload(TenantId.Local, Guid.NewGuid().ToString("N")));
+        Assert.Null(store.FindTurnByUpload(TenantId.Local, ""));
     }
 
     [Fact]
@@ -36,22 +37,22 @@ public sealed class GatewayTurnJobStoreTests
     {
         var store = new GatewayTurnJobStore();
         var uploadId = Guid.NewGuid().ToString("N");
-        var job = store.Create("sid-1", uploadId);
+        var job = store.Create(TenantId.Local, "sid-1", uploadId);
 
         // Age the job past its TTL via the documented test seam; the next lookup must drop it.
         job.OverrideCreatedAtForTest(DateTime.UtcNow.AddMinutes(-11), GatewayTurnJobStore.Ttl);
 
-        Assert.Null(store.FindTurnByUpload(uploadId));
+        Assert.Null(store.FindTurnByUpload(TenantId.Local, uploadId));
     }
 
     [Fact]
     public void Create_WithoutUploadId_HasEmptyUploadIdAndNoIndex()
     {
         var store = new GatewayTurnJobStore();
-        var job = store.Create("sid-1");
+        var job = store.Create(TenantId.Local, "sid-1");
 
         Assert.Equal("", job.UploadId);
         // A retried completion with no upload context never collides with this job.
-        Assert.Null(store.FindTurnByUpload(""));
+        Assert.Null(store.FindTurnByUpload(TenantId.Local, ""));
     }
 }
--- a/src/CcDirector.Gateway.Tests/VoiceTurnArchiveTests.cs
+++ b/src/CcDirector.Gateway.Tests/VoiceTurnArchiveTests.cs
@@ -1,6 +1,7 @@
 using System.Text;
 using CcDirector.Gateway.Voice;
 using Xunit;
+using CcDirector.Core.Tenancy;
 
 namespace CcDirector.Gateway.Tests;
 
@@ -39,14 +40,14 @@ public sealed class VoiceTurnArchiveTests : IDisposable
     {
         var turnId = Guid.NewGuid().ToString();
         var audio = Encoding.UTF8.GetBytes("ID3-fake-mp3-bytes");
-        _archive.Save(Record(turnId, "sid-1"), audio);
+        _archive.Save(TenantId.Local, Record(turnId, "sid-1"), audio);
 
-        var got = _archive.Get(turnId);
+        var got = _archive.Get(TenantId.Local, turnId);
         Assert.NotNull(got);
         Assert.Equal("It is four.", got!.Summary);
         Assert.Equal("what is two plus two", got.Transcript);
 
-        var gotAudio = _archive.GetAudio(turnId);
+        var gotAudio = _archive.GetAudio(TenantId.Local, turnId);
         Assert.NotNull(gotAudio);
         Assert.Equal(audio, gotAudio);
     }
@@ -57,17 +58,17 @@ public sealed class VoiceTurnArchiveTests : IDisposable
         var turnId = Guid.NewGuid().ToString();
         var rec = Record(turnId, "sid-1");
         rec.HasAudio = false;
-        _archive.Save(rec, replyAudio: null);
+        _archive.Save(TenantId.Local, rec, replyAudio: null);
 
-        Assert.NotNull(_archive.Get(turnId));
-        Assert.Null(_archive.GetAudio(turnId));
+        Assert.NotNull(_archive.Get(TenantId.Local, turnId));
+        Assert.Null(_archive.GetAudio(TenantId.Local, turnId));
     }
 
     [Fact]
     public void Get_UnknownTurn_ReturnsNull()
     {
-        Assert.Null(_archive.Get(Guid.NewGuid().ToString()));
-        Assert.Null(_archive.Get("not-a-guid"));
+        Assert.Null(_archive.Get(TenantId.Local, Guid.NewGuid().ToString()));
+        Assert.Null(_archive.Get(TenantId.Local, "not-a-guid"));
     }
 
     [Fact]
@@ -75,14 +76,14 @@ public sealed class VoiceTurnArchiveTests : IDisposable
     {
         var turnId = Guid.NewGuid().ToString();
         var uploadId = Guid.NewGuid().ToString("N");
-        _archive.Save(Record(turnId, "sid-1", uploadId), null);
+        _archive.Save(TenantId.Local, Record(turnId, "sid-1", uploadId), null);
 
-        var found = _archive.FindByUpload(uploadId);
+        var found = _archive.FindByUpload(TenantId.Local, uploadId);
         Assert.NotNull(found);
         Assert.Equal(turnId, found!.TurnId);
 
-        Assert.Null(_archive.FindByUpload(Guid.NewGuid().ToString("N")));
-        Assert.Null(_archive.FindByUpload(""));
+        Assert.Null(_archive.FindByUpload(TenantId.Local, Guid.NewGuid().ToString("N")));
+        Assert.Null(_archive.FindByUpload(TenantId.Local, ""));
     }
 
     [Fact]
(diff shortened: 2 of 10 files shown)


thefrederiksen (commented):
REJECT on current head 51ae7f76. The CI failure is branch-caused and exposes a deterministic test-isolation regression: `WingmanVoiceServiceTests.NewService` randomizes only a persistence filename directly under `Path.GetTempPath()`, but the partitioned service now derives `_baseDir` from that file parent. Every supposedly isolated instance therefore reads and writes the same `%TEMP%/tenants/local` directory and the same `sid-1`; ready audio seeded by another test makes `StoreSpokenAsync_WithFailingTts_DoesNotMarkReady` observe `HasVoice == true`. Use a GUID-unique directory root per independent harness (while deliberately sharing one root only for restart round-trips) and clean that root. Then rebase onto fresh main carrying #1916 and the suite lock, and provide the full Gateway production-line revert with named reds plus restored exact-head green. The physical tenant directories, canonical tenant guards, required TenantId parameters, and directly `GatewayHostedMode.IsHosted`-gated hosted-delete/self-host-move migration otherwise look structurally sound in this pass.

thefrederiksen (commented):
A second proof blocker: the current four-guard revert recipe never mutates the core tenant partition. It deletes canonical-shape refusals and both legacy migration bodies while explicitly observing that all eleven A/B isolation tests stay green. That makes those eleven useful controls for those unrelated deletions, but it does not prove they are keyed to the production boundaries they claim to protect. Add individually attributable mutations for (1) WingmanVoiceService.StateFor in-memory tenant buckets, (2) WingmanVoiceService.PartitionDirectoryFor durable session/audio paths, (3) VoiceTurnArchive PartitionDirectoryFor/DirFor, and (4) GatewayTurnJobStore.StateFor. Each mutation must successfully build, run under the full serialized Gateway suite after the lock rebase, and redden its exact cross-tenant or physical-restart canaries for the distinguishing assertion—not merely through NullReferenceException or IOException—before the restored exact-head green. The existing migration/shape mutations remain necessary but are not substitutes for partition mutation.

thefrederiksen (commented):
Re-review at `aa2bec63`: the production partition and GUID-directory harness correction look sound, the branch is rebased onto lock-bearing main and includes #1916, and the reported arithmetic reconciles (`2953 + 16 = 2969`). The proof still does not meet the causal-attribution bar, though: B1 `WingmanVoiceService.StateFor`, B2 `WingmanVoiceService.PartitionDirectoryFor`, B3 `VoiceTurnArchive.PartitionDirectoryFor`/`DirFor`, and B4 `GatewayTurnJobStore.StateFor` were all mutated in one full-suite run. The agreed rule is one isolation primitive per run. A list assigning failures to lines after a collective mutation cannot prove that each primitive independently owns its canary; `Clips_reload_into_the_tenant_that_owned_them` is explicitly described as sensitive to both B1 and B2, demonstrating the ambiguity. Please run four separate locked full Gateway mutations on the final head—one exact boundary primitive at a time, each with a successful build and its own assertion-only red set—then restore the exact head and show the full Gateway suite green. The current .NET CI job is also still pending. Verdict remains **REJECT** until that evidence lands.

thefrederiksen (commented):
ACCEPT at exact head `4c2a8ebc`.

The implementation and harness corrections are sound, and the required proof is complete:

- lock-bearing/#1916 ancestry confirmed;
- four separate full-Gateway mutations, one boundary primitive per run;
- independent reconciliations against the green 2969-pass baseline: B1 2960+9, B2 2962+7, B3 2967+2, B4 2968+1;
- all 19 reds are distinguishing assertion failures, with zero crashes, no `EXCL`, and no `NOTRUN` cells;
- restored full Gateway run is green on the committed exact head;
- the changes from the earlier executable proof head to this head contain zero non-comment lines, so no behavior or test was removed while recording the matrix;
- exact-head CI is green for both .NET and web, converting the non-Gateway project counts from static unaffected judgment into observed green evidence.

The separated runs also corrected three claims the combined mutation could not validly attribute, confirming that the one-primitive proof requirement was load-bearing. ACCEPT.

PR #1918, opened 2026-07-19
test(gateway): serialize Gateway test runs machine-wide, automatically

thefrederiksen:
## The problem

Two concurrent runs of `CcDirector.Gateway.Tests` on one machine destroy each other. Overlapping runs report failures in areas nobody touched, and frequently end with no summary line at all - the shape of a crashed test host rather than a failed assertion. On one afternoon four runs overlapped and all four had to be discarded. Corrupted test evidence is worse than none, because somebody acts on it.

Asking people and agents not to overlap does not work, and cannot work. A runner can only observe its own run, so each one honours the rule individually while the aggregate still breaks it. A rule needing global knowledge cannot be obeyed by an actor holding local knowledge. A lock somebody has to remember to take is the same convention under a new name.

## What this does

Acquisition is automatic, inside the test process, before any test runs. A plain `dotnet test` serializes without the caller knowing this exists.

- **A module initializer** takes the lock on assembly load. Chosen over an assembly fixture because assembly-level fixtures are an xunit v3 feature and this project is on xunit 2.9; it is runner-independent and is the form `TestStorageRootRedirect` already uses here.
- **The operating system decides ownership, never metadata.** The lock is a file opened `FileAccess.Write` with `FileShare.Read`, held open for the process lifetime. A second run's open is refused; readers are still admitted, which is how a blocked run names its blocker.
- **A holder that dies needs no cleanup logic.** Crash, kill, or the known file-system-watcher dispose race - the kernel closes the handle and the next run acquires. No stale-detection code, so no way to wedge the lock permanently.
- **A holder alive but stuck makes the waiting run fail loudly and run no tests**, after forty-five minutes. It never breaks the lock on age: an age cap cannot distinguish "wedged" from "merely slow", so breaking on age would start a second suite beside a live first one - the exact corruption this prevents.
- **The path cannot be moved by the environment.** Derived from the Windows shell folder API, which ignores an overridden `LOCALAPPDATA` (measured); on Unix a fixed literal, because `GetTempPath` reads `TMPDIR` and the folder API reads `XDG_DATA_HOME` and `HOME`.
- **The derivation is a pure function of its inputs.** It reads no process-global state; one call site reads the real environment and passes it in.
- **Only sharing violations are treated as contention.** Permission failures, a read-only file, a directory at the path - all stop the run immediately with a message saying explicitly that this is not a holder.

### Scope: per-user-per-machine

Stated precisely, because someone will rely on the words. Two runs by the **same operating-system user** on one machine cannot overlap. Two runs by **different users** are not serialized, and on Windows cannot be, since the lock lives under per-user local application data.

That scope was chosen, not overlooked: every process that actually contends runs as the same user, verified rather than assumed. A machine-shared home would need somewhere writable by every account, bringing permission and cleanup failures that are themselves wedge risks - the class of problem this removes. **Revisit if a second operating-system user ever runs this suite** on one machine.

### Every run is serialized, filtered or not

Deliberate, and the first thing somebody will try to relax. A filtered run still loads the whole assembly, boots real hosts, touches shared machine state, and is exposed to the host-crash mode never shown to be intra-process. Per-class storage-root redirection is the only isolation actually demonstrated, and "no collision was found" is not "no collision exists". Acquisition sits in a module initializer precisely because it cannot see the filter, and so cannot be talked into an exception it has no basis for.

## Two defects found in review, both fixed

**The lock did not exclude.** The path came from `Path.GetTempPath()`, which reads `TEMP` and `TMP`. Two runs launched with different `TEMP` values computed different lock files, both reported acquiring, and ran concurrently - demonstrated, not theorised. The environments most likely to differ are exactly the ones this serializes: different agents, different shells, different working trees, a scheduled task against an interactive session.

**Setup faults were reported as contention.** Every `IOException` and `UnauthorizedAccessException` counted as "a live holder", so a read-only lock file or a directory at the path made the run wait the full timeout naming a holder that did not exist - then repeat forever. That is the permanent wedge, arriving through the error path. Measured before fixing: contention gives `IOException 0x80070020`; a directory at the path and a read-only file both give `UnauthorizedAccessException 0x80070005`. The diagnostic log also caught `IOException` alone, so an unwritable log would have thrown out of a diagnostic write after acquisition and aborted every later run.

## Evidence - all on head 3e28d771, one sitting, box verified drained first

**1. The lock's own tests - 8 passed.** Including both platform branches of the path derivation, exercised on Windows because the platform is an input rather than read from the running process. The Unix branch is therefore covered by a Windows run instead of being discovered by the `ubuntu-latest` continuous-integration leg.

**2. Two runs, two different `TEMP` values - the second waited.** The reviewer's exact experiment, which previously produced two simultaneous acquisitions:

```
19:45:26 pid 57920: Acquired the per-user Gateway test lock.     <- run A, TEMP=D:/two-temp-demo/temp-A
19:45:29 pid 40120: WAITING ... Holder: process 57920, started   <- run B, TEMP=D:/two-temp-demo/temp-B
                    2026-07-19T23:45:26Z, holding since ...
                    (3s ago), owner cc-director session c584f16f-...
19:45:35 pid 40120: Acquired the per-user Gateway test lock.     <- only after A exited
```

Both runs green (93 and 8 tests). Stated pass criterion met: B named A's process id and acquired only after A ended.

**3. A killed holder - the next run acquires cleanly.** Holder force-killed at 19:45:59; the waiter acquired the same second and passed. No stale logic involved - the kernel released the handle.

**4. A live holder past the timeout - loud refusal, zero tests run.** Exercised with the timeout temporarily shortened in a throwaway build; the shipped value is forty-five minutes, restored and rebuilt before the suite below. The refusal is visible at **default** verbosity, because the runner echoes the test host's standard error in the abort reason:

```
*** REFUSING TO RUN. A live process has held the per-user Gateway test lock ... NO TESTS WILL RUN.
This run stops instead of starting alongside it, because two concurrent runs of this suite corrupt
each other's results ... Holder: process 60368 ... ***
Test Run Aborted.   (exit 1, no tests executed)
```

**5. One isolated full Gateway suite - green.**

```
Passed! - Failed: 0, Passed: 2949, Skipped: 10, Total: 2959, Duration: 6 m 56 s
```

Started 19:47:19, finished 19:54:17, clean tree at `3e28d771`. The 10 skips are the environment-gated real-Postgres proofs. The box was independently re-censused immediately beforehand - zero `testhost` or `vstest` processes, zero non-PowerShell processes referencing `Gateway.Tests`, zero non-idle `dotnet` processes - because this is the last full suite that runs *without* the lock protecting it, so its cleanliness rests entirely on the box being drained.

## Notes for the reviewer

- No testability seam was added. An in-assembly test cannot spawn `dotnet test` children of its own assembly - the parent holds the lock, so the children would correctly wait forty-five minutes and the test would deadlock. The only way to make that shape work is a flag letting children target a different lock file, which is a bypass hole in the mechanism it would be testing. The property is pinned from inside the held lock instead, and the two-process demonstration lives outside the suite as evidence.
- Nothing in the tests mutates process-global state. An earlier draft proved environment-immunity by redirecting `TEMP`, `TMP` and `TMPDIR` and then recursively deleting the directory it redirected them to - which could have deleted a tree another test had been redirected into, introducing cross-class nondeterminism inside the fix for cross-process nondeterminism.

### DIFF
--- a/src/CcDirector.Gateway.Tests/GatewayTestSuiteLock.cs
+++ b/src/CcDirector.Gateway.Tests/GatewayTestSuiteLock.cs
@@ -56,6 +56,23 @@ namespace CcDirector.Gateway.Tests;
 /// They exist so a blocked run can name its blocker - a blocked run that cannot say who is blocking it is
 /// indistinguishable from a hang, and somebody will kill the wrong thing. They never decide ownership.
 ///
+/// THE GUARANTEE IS PER-USER-PER-MACHINE. Not machine-wide - say it precisely, because someone will
+/// eventually rely on the words. Two runs by the SAME operating-system user on the SAME machine cannot
+/// overlap. Two runs by DIFFERENT users on one machine are NOT serialized by this, and on Windows they
+/// cannot be, because the lock lives under the per-user local application data directory.
+///
+/// That scope was chosen, not overlooked. Every process that actually contends here runs as the same user:
+/// each agent is a subprocess of a session running under that account, and that was verified rather than
+/// assumed. A machine-shared home would have to sit somewhere writable by every account, which brings
+/// permission and cleanup failures of its own - and those failures are themselves wedge risks, which is the
+/// exact class of problem this mechanism exists to remove. Buying protection against a collision we do not
+/// have, at the cost of new ways to jam, is a bad trade.
+///
+/// REVISIT THIS if a second operating-system user ever runs this suite on one machine - a service account,
+/// a continuous-integration agent alongside an interactive login, a second person on a shared box. At that
+/// point the scope is genuinely too narrow and the home must move somewhere shared, with the permission
+/// model worked out deliberately. Until then, this is the observed contention boundary.
+///
 /// EVERY RUN OF THIS ASSEMBLY IS SERIALIZED, FILTERED OR NOT. That is a feature, not an oversight, and it
 /// is the first thing somebody will try to "optimise" - a filtered run looks small and harmless, so surely
 /// it can be let through. It cannot, because nobody can show that it is harmless. A filtered run of this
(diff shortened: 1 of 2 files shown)


thefrederiksen (commented):
REJECT on current head a180f318 pending three items. First, writer exclusion itself is sound: FileAccess.Write plus FileShare.Read refuses another writer while allowing diagnostic readers, and this head correctly fixes the demonstrated environment-relocatable TEMP path and setup-fault wedge. However, the chosen identity is per-user, not machine-wide: Windows uses the current user LocalApplicationData and Unix includes Environment.UserName, so two OS users on one machine take different locks. Either establish and document that same-user is the real contention boundary and stop claiming machine-wide scope, or move to a truly machine-shared identity. Second, the Gateway is cross-platform and Linux-hosted, but TheLockDoesNotLiveUnderTheEnvironmentSettableTemporaryDirectory fails on ordinary Unix: production deliberately returns /tmp/cc-director-..., while Path.GetTempPath ordinarily returns /tmp/. The environment-mutation tests are the valid invariant there, so make this assertion platform-correct. Third, the required bootstrap evidence is not yet tied to this exact post-fix head: provide the manually drained isolated full Gateway green and explicitly record the two-run wait, killed-holder release, live-timeout fail-loud with zero tests, and actual development-environment runner proof on a180 or its final successor. Do not infer that old demonstrations survived a code change that already fixed one inert-lock defect.

PR #1930, opened 2026-07-20
fix(gateway): authorize the Director that owns an up-stream (#1923)

thefrederiksen:
Fixes #1923.

## The defect

`DirectorHub.StreamUp` proved WHO the caller was - the Hello binding - but never that the caller OWNED the stream it named, and `GatewayStreamRegistry` recorded no owner at all: its entries were keyed on the bare stream id string. Any authenticated account that learned or guessed another account's live stream id could therefore write frames into that account's terminal, file, or screenshot sink, claim the stream ahead of the real Director, or tear it down.

The direction matters: this is INJECTION - a write into what another customer sees - plus a denial of service against their streams. It is not disclosure, and a fix that only prevented reading would not have touched it.

## The fix: authorize, do not deny

This is a live product path, and the stream is tenant-attributable, so the entry carries its owner rather than the path being denied.

`Register` now records a `StreamOwner` - the tenant whose request located the session, paired with the Director the open command is being sent to - and `ConsumeAsync` authorizes the calling identity against it and refuses on a mismatch.

**Why the pair, and why not the connection id.** The tenant alone authorizes nothing on self-host, where everything is the single local tenant. The Director id alone is not enough either: it is declared by the client in its Hello message, and the Director registry is keyed on tenant plus Director id, so two accounts can each own a Director calling itself the same thing. Separately, a Director id is re-usable across connections within one tenant - the Hello re-claim guard is per-connection - which means one of an account's own machines could name itself as another of that account's Directors. That is bounded to a single customer and is not a charge blocker, but it is exactly why the tenant half of the pair is load-bearing rather than decorative.

The SignalR connection id was considered and rejected: the Gateway does not hold it at registration time - it routes the open command by Director id - so recording it would have meant inventing a lookup rather than recording a fact.

**A refusal, never a silent drop.** A drop would be indistinguishable from the legitimate "the stream id is unknown because the browser already left" no-op, and it would hide a real misconfiguration from operators. The unknown-id no-op is preserved unchanged and a test pins the two outcomes apart.

## Every public registry member, checked or cleared

| Member | Check? | Why |
|---|---|---|
| constructor | No | No caller identity involved. |
| `LiveStreamCount` | No | Gateway-internal diagnostic count; not reachable by a Director, device, or browser. |
| `Register` | Records the owner | The source of the authorization fact. Fails loud on an invalid tenant or blank Director id - a stream whose owner is unknown must never be opened. |
| `ConsumeAsync` | Yes, added | The write leg, the claim leg, and a teardown leg all at once: the claim flag is set inside it and teardown runs in its `finally`, so one guard at the entry covers all three. On refusal nothing is claimed and nothing is torn down, so a wrong caller cannot deny the real owner service either. |
| `Close` | No, and documented in code | Every caller is the same leg that called `Register` for that id, in the same request; the id never leaves that local scope. The Director-facing surface is the hub, whose only stream method is `StreamUp`, and the Director's own close-stream travels Gateway to Director and never re-enters this registry. The doc comment tells the next person to carry and check an owner if a caller-supplied close is ever added. |

## Proof

Full Gateway suite, one primitive mutated, run on this exact commit.

- **Baseline:** 3005 passed, 0 failed, 10 skipped (Postgres-gated).
- **Reverted** (the ownership block deleted outright): 3003 passed, 2 failed. Both predicted reds appeared by name and as assertions, not crashes:
  - `Hosted_TenantB_StreamingIntoTenantAsStream_IsRefused_AndAsSinkReceivesNothing` - `Assert.Throws() Failure: No exception was thrown. Expected: typeof(Microsoft.AspNetCore.SignalR.HubException)`
  - `SameTenant_DifferentDirector_IsAlsoRefused` - `Assert.Throws() Failure: No exception was thrown. Expected: typeof(CcDirector.Gateway.Streaming.StreamOwnershipDeniedException)`
- **Restored:** 3005 passed, 0 failed. Reconciliation computed: 3003 + 2 = 3005.
- The other six test projects are green: Core 3306, Avalonia 247, HostedAgent 86, Engine 62, Launcher 27, Terminal.Avalonia 21.

The cross-tenant test is deliberately bracketed, because a refusal assertion over a stream that never worked would prove nothing:

1. **Positive control first** - tenant A streams to its own stream id over the real hub and the frames genuinely arrive, with the format facts asserted before the payload is read.
2. **The attack** - tenant B calls `StreamUp` with a stream id minted for tenant A: refused, tenant A's sink receives nothing, is not completed, and the stream is still live.
3. **Destructibility control after** - tenant A then streams into that very same refused stream and the frames land, proving the sink was alive and writable throughout. So the emptiness in step 2 was the ownership check, not a dead stream.

Both hosted and self-host modes are set explicitly in the tests and asserted, never inherited from the runner's default. Self-host is the control and its behaviour is unchanged.

### DIFF
--- a/src/CcDirector.Gateway.Tests/GatewayStreamRegistryTests.cs
+++ b/src/CcDirector.Gateway.Tests/GatewayStreamRegistryTests.cs
@@ -1,4 +1,5 @@
 using System.Runtime.CompilerServices;
+using CcDirector.Core.Tenancy;
 using CcDirector.Gateway.Contracts;
 using CcDirector.Gateway.Streaming;
 using Xunit;
@@ -13,6 +14,11 @@ namespace CcDirector.Gateway.Tests;
 /// </summary>
 public sealed class GatewayStreamRegistryTests
 {
+    // Issue #1923: every stream now records the identity that owns it. These lifecycle tests are all
+    // single-owner, so they register and consume as the same identity - the ownership CHECK itself is proved
+    // in StreamOwnershipTests.
+    private static readonly StreamOwner Owner = new(TenantId.Local, "dir-1");
+
     // A sink that records the frames it is handed, optionally blocking each write on a supplied gate so a test
     // can hold the pull and observe backpressure.
     private sealed class RecordingSink : IStreamSink
@@ -54,9 +60,9 @@ public sealed class GatewayStreamRegistryTests
     {
         var registry = new GatewayStreamRegistry();
         var sink = new RecordingSink();
-        registry.Register("s1", sink);
+        registry.Register("s1", Owner, sink);
 
-        await registry.ConsumeAsync("s1", SizeBinaryClosed("s1"), CancellationToken.None);
+        await registry.ConsumeAsync("s1", Owner, SizeBinaryClosed("s1"), CancellationToken.None);
 
         Assert.Equal(3, sink.Frames.Count);
         Assert.Equal(DirectorStreamFrameType.Size, sink.Frames[0].Kind);
@@ -75,7 +81,7 @@ public sealed class GatewayStreamRegistryTests
         var registry = new GatewayStreamRegistry();
         var gate = new TaskCompletionSource();
         var sink = new RecordingSink(gate: () => gate.Task);
-        registry.Register("bp", sink);
+        registry.Register("bp", Owner, sink);
 
         var yielded = 0;
         async IAsyncEnumerable<DirectorStreamFrame> Produce([EnumeratorCancellation] CancellationToken ct = default)
@@ -89,7 +95,7 @@ public sealed class GatewayStreamRegistryTests
             yield return Closed("bp", "eof");
         }
 
-        var consume = registry.ConsumeAsync("bp", Produce(), CancellationToken.None);
+        var consume = registry.ConsumeAsync("bp", Owner, Produce(), CancellationToken.None);
         await Task.Delay(150); // let the consumer pull the first frame and block on the held sink write
 
         Assert.Equal(1, Volatile.Read(ref yielded)); // only the first frame was pulled; the next pull is blocked
@@ -107,7 +113,7 @@ public sealed class GatewayStreamRegistryTests
     {
         var registry = new GatewayStreamRegistry();
         var sink = new RecordingSink();
-        registry.Register("inf", sink);
+        registry.Register("inf", Owner, sink);
 
         var started = new TaskCompletionSource();
         async IAsyncEnumerable<DirectorStreamFrame> Forever([EnumeratorCancellation] CancellationToken ct = default)
@@ -122,7 +128,7 @@ public sealed class GatewayStreamRegistryTests
             }
         }
 
-        var consume = registry.ConsumeAsync("inf", Forever(), CancellationToken.None);
+        var consume = registry.ConsumeAsync("inf", Owner, Forever(), CancellationToken.None);
         await started.Task;
         await Task.Delay(60); // let a few frames flow
 
@@ -140,10 +146,10 @@ public sealed class GatewayStreamRegistryTests
         // registered sink, so consume returns at once without touching any sink.
         var registry = new GatewayStreamRegistry();
         var sink = new RecordingSink();
-        registry.Register("gone", sink);
+        registry.Register("gone", Owner, sink);
         registry.Close("gone"); // sink torn down before StreamUp arrives
 
-        await registry.ConsumeAsync("gone", SizeBinaryClosed("gone"), CancellationToken.None);
+        await registry.ConsumeAsync("gone", Owner, SizeBinaryClosed("gone"), CancellationToken.None);
 
         Assert.Empty(sink.Frames); // nothing was pumped into the gone sink
         Assert.Equal(0, registry.LiveStreamCount);
@@ -156,7 +162,7 @@ public sealed class GatewayStreamRegistryTests
         // must not wait forever - the open timeout tears it down and fires the caller's token.
         var registry = new GatewayStreamRegistry(openTimeout: TimeSpan.FromMilliseconds(100));
         var sink = new RecordingSink();
-        var token = registry.Register("late", sink);
+        var token = registry.Register("late", Owner, sink);
 
         await Task.Delay(400);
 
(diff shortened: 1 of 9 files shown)


thefrederiksen (commented):
REJECT at `98a6bf2e`: the production check is placed correctly before claim/write/teardown and the refusal/destructibility ordering is strong, but the proof does not establish the tenant half of `StreamOwner`.

The hosted attack test changes **both** dimensions: owner `(tenant A, dir-a)` versus caller `(tenant B, dir-b)`. If the tenant equality conjunct is deleted from `StreamOwner.Matches`, the Director-id mismatch still refuses B, so that test stays green. `SameTenant_DifferentDirector_IsAlsoRefused` also stays green. In fact, every current test can pass while ownership has silently degraded to Director-id-only—the exact unsafe design the PR body rejects, because `DirectorId` is client-chosen in `Hello` and two tenants may both use the same value.

Add the collision canary the threat model requires: tenant A and tenant B each bind a Director with the **same** Director ID; A owns the stream; B is refused; the sink remains empty/live; then A writes to that same sink successfully. Pre/post positive controls should retain the current exact frame fingerprints. Under the bypassability rule, prove the owner pair with separate full-suite mutations on the rebased final head:

1. remove the registry ownership guard call-site (current arm);
2. remove only tenant equality from `StreamOwner.Matches`—the same-ID cross-tenant canary must red by assertion;
3. remove only Director equality—the existing same-tenant/different-Director canary must red by assertion.

Each arm needs a green+complete baseline, build success, full Gateway reconciliation, zero crash-reds, and exact-head restored green. Rebase first; current main has moved beyond `712ef1ff` (the intervening production change appears orthogonal, so inspect with range-diff). Exact-head .NET CI is also still pending. The code is statically sound, but the claim that both halves of the pair are independently load-bearing is currently unproved.

thefrederiksen:
## Rebased, canary added, mutation predictions pre-registered

Rebased onto `f5bb926f` (current origin/main). New head `0e19b5b4`. Verified by `git range-diff 712ef1ff..d4d2dfb1 f5bb926f..0e19b5b4`, not by a rebuild: both branch commits report `=`, so they are content-identical after the rebase. No conflicts arose.

The intervening main changes are orthogonal, checked rather than assumed. The only production file among them is `CcStorageMigration.cs` (storage root pinning); the rest are documentation, one new Core test file, and an assertion-message change in `HostedAccountStatusTests.cs`. Nothing under `Streaming/`, `Tenancy/`, or the hub is touched. The diff from the previously reviewed head `98a6bf2e` to `0e19b5b4` contains no Gateway production file at all, so the ownership check under review is byte-identical to the reviewed one; the only change on my side is the added test.

Because `HostedAccountStatusTests.cs` moved, the baseline count will be taken fresh on this head rather than carried across from the earlier run.

### The collision canary

`Hosted_TwoTenantsUsingTheSameDirectorId_CrossTenantStreamUp_IsRefused`. Tenant A and tenant B each bind a Director declaring the same id (`dir-shared`), and both Hellos are asserted to bind. A owns the stream; B is refused; the sink is asserted empty, not completed, and still live; then A writes to that same sink and the frames land.

Two premises were verified in the code rather than assumed. The `Hello` re-claim check reads `Context.Items`, which is per-connection, so two connections in different tenants can both bind the same Director id - if B's bind were rejected the refusal would prove nothing. And the caller tenant comes from `ResolveConnectionTenant()` via the authenticated device key, never from the Hello payload, so B's caller really is `(tenantB, "dir-shared")`.

The final owner write is not decoration. A test asserting only that B was refused cannot tell a refusal apart from an operation that was inert; the owner's successful write to the same sink is what makes "B wrote nothing" mean "B was refused" rather than "nothing was writable". Pre and post controls keep the exact frame fingerprints: count, frame kind, stream id, payload length, and the exact marker byte.

### Separability of the three mutations

`StreamOwner.Matches` has exactly one call site, `GatewayStreamRegistry.cs:124`. The three states are independently reachable and each compiles:

1. delete the guard block at `GatewayStreamRegistry.cs:124-132` - predicate intact, never consulted;
2. delete only the tenant equality conjunct in `Matches` - guard and Director conjunct intact;
3. delete only the Director equality conjunct in `Matches` - guard and tenant conjunct intact.

No two of these collapse into one another, so the run count cannot be reduced by design here. Arm 1 reddens a superset of arms 2 and 3, but overlap does not exempt a primitive from its own run.

### Pre-registered predictions

The suite contains exactly three ownership refusal assertions, all in `StreamOwnershipTests`, each isolating a different clause. Every predicted red should arrive as an xunit `ThrowsAsync` failure reporting that no exception was thrown - an assertion the test's own logic raised, not a crash.

Arm 1, guard call-site removed - predict exactly 3 reds:
- `Hosted_TenantB_StreamingIntoTenantAsStream_IsRefused_AndAsSinkReceivesNothing`
- `Hosted_TwoTenantsUsingTheSameDirectorId_CrossTenantStreamUp_IsRefused`
- `SameTenant_DifferentDirector_IsAlsoRefused`

Arm 2, tenant equality removed - predict exactly 1 red:
- `Hosted_TwoTenantsUsingTheSameDirectorId_CrossTenantStreamUp_IsRefused`

The other two stay green by construction, and that is the point of the arm: in the first test both halves of the owner differ, so the Director comparison alone still refuses it. That is precisely the gap this canary closes.

Arm 3, Director equality removed - predict exactly 1 red:
- `SameTenant_DifferentDirector_IsAlsoRefused`

`SameOwner_DirectorIdCaseDiffers_IsStillTheOwner` stays green under all three arms; it is a permit-path test and every arm only makes the predicate more permissive. For the same reason the permit-path tests in `GatewayStreamRegistryTests`, `TunnelStreamLegsTests`, and `TunnelMechanismProofTests` are predicted to contribute zero reds under every arm.

A predicted red that does not appear will be reported as a finding, not explained away.

Build on the rebased head succeeds with zero warnings and zero errors. The full-suite runs are queued behind the machine-wide test lock and are not claimed here.

thefrederiksen:
STANDING: HELD
Exact head: `0e19b5b486a3862dc58952f1ab6531fcd9988cc2`

The authorization implementation and the added same-identifier tenant collision canary pass static review. The owner pair is recorded at stream registration, checked before claim/write/teardown, and each equality clause now has an isolating assertion with a live permitted-path control.

Hold is proof-only plus final-head convergence. The pull-request body still presents the retired two-red proof as exact-head proof, while the newer comment correctly says no run is claimed on this rebased head; make the public proof status internally consistent. Rebase onto current `origin/main` (`df078618` at review time); the intervening changes are orthogonal on inspection, but no proof exists to carry. Then run and post one complete sitting on the exact final head: clean full-Gateway baseline, three separate arms (guard call-site, tenant equality, Director equality), and exact restored green, with build/source preflight, assertion reds, complete summaries, and reconciliation. Exact-head CI must remain green.

PR #1933, opened 2026-07-20
fix(gateway): partition car-mode telemetry by device credential

thefrederiksen:
Census row 40 / ingestion map work unit "Car-mode telemetry record with no storage key". `POST /carmode/telemetry` appended every caller's record to one process-global list, and both telemetry reads returned that whole list, so any authenticated caller could read every other caller's turn timings.

## The discriminator

Partitioned by DEVICE CREDENTIAL, as ruled. It is trusted (the Gateway derives it from the authenticated request, never the payload), the other Car Mode stores - conversation, pending action, subject - are already keyed the same way, so this is consistent rather than novel, and telemetry is per-device operational data. Telemetry following a person across their devices is a product question and is deliberately not built here.

## What I found in the existing records, and what I did with them

The stored document already carries a server-stamped `DeviceHash` on every record - it has been a server-filled field since the store shipped. I read the live store on this machine: **36 records, every one carrying a non-blank credential-derived hash across three devices (30 / 5 / 1), zero blank, zero anonymous.**

So there is nothing to migrate and nothing to invent: filtering the reads by the recorded hash attributes each existing record to exactly the device that wrote it. The purge ruling from the sibling diagnostic unit does not transfer here, because the attribution that unit lacked is present in this one.

The one case with no recorded attribution is a record with a blank hash. It belongs to no device, no partitioned read can return it, and guessing an owner would be exactly the half-partition we refuse - so those are purged on load rather than left as unreachable residue consuming the growth-guard budget. On today's file that purge removes nothing; it is proven against a crafted file.

## The change

- `CarModeTelemetryStore.Add(deviceHash, record)` stamps the trusted partition onto the record it stores, so a record can never be filed under a partition the caller chose. Blank throws rather than becoming a shared bucket.
- `Recent(deviceHash, limit)` and `Count(deviceHash)` both go through ONE private selection method, so the filter cannot be right in one read and wrong in another. The limit is applied after the partition, so a quiet device's slice never borrows from a busy one. `Count` is per-device: a process-wide total is other devices' activity disclosed as an aggregate.
- Both routes derive the partition from ONE helper, `DevicePartition(ctx)`, on the request credential. Never the posted body, never a query parameter. The caller-supplied turn id is not used as a discriminator anywhere.
- The HTML dashboard page is static and record-free; it fetches the data route, which is partitioned.
- The growth guard now evicts from the LARGEST device partition instead of the oldest record overall, so a busy device cannot push a quiet device's records out (the census completion predicate covers suppression and contention, not only disclosure).

Consequence the Architect should note: the dashboard now shows the records of the device you are viewing from, which is the ruled per-device behaviour.

## Proof design

Write-partition and read-filter are asserted as TWO SEPARATE FACTS. The write is proven from the stored JSON document directly, not through the read path, so a read-only filter could not satisfy it. Every cross-device assertion carries a positive control in the same test - each caller reading its OWN record back on the same route - so an empty answer cannot pass for isolation when it was really a failed seed. In the wire tests, status code and media type are asserted BEFORE any body is parsed.

New tests:
- `CarModeTelemetryEndpointPartitionTests` - the real `CarModeEndpoint` route map on an ephemeral port, two callers with different Bearer credentials: neither data read returns the other's records, the held count is per-device, the write records the partition (proven from the stored document), and a posted body claiming another device's hash still lands in the caller's own partition.
- `CarModeTelemetryPartitionTests` - store level: the trusted partition is stored and persisted, blank partition throws on write and on read, reads and counts are per-device, a limit never borrows across devices, existing records keep their recorded device, blank-hash records are purged from the stored document, and the growth guard evicts from the busiest device only.

## Pre-registered mutations (five runs, one primitive each)

| # | Mutation | Predicted reds |
| --- | --- | --- |
| 1 | `Add`: store `record` instead of `record with { DeviceHash = deviceHash }` | every test in `CarModeTelemetryPartitionTests` and all four endpoint tests; the single-device `CarModeTelemetryStoreTests` stay GREEN as controls |
| 2 | `MineNewestFirstLocked`: drop the device comparison, return every record | all cross-device tests in both new files; `Add_RawJsonOnDisk_CarriesTheWritingDevice` stays GREEN as a control (it does not use the read path) |
| 3 | `DevicePartition(ctx)`: return a constant instead of the credential hash | all four endpoint tests; every store test stays GREEN as a control |
| 4 | `Load`: remove the blank-hash purge | `Load_PurgesRecordsThatCarryNoDevice_RatherThanGuessingAnOwner` only |
| 5 | growth guard: evict the oldest record overall instead of from the largest partition | `GrowthGuard_EvictsFromTheBusiestDevice_NeverFromAQuietOne` only |

A predicted red that does not appear is a finding, not something to explain away.

## Not run

The Gateway test suite has NOT been run - the test lane is held by another worker. Full solution builds clean (0 warnings, 0 errors). Awaiting admission for the baseline plus the five mutation runs.

### DIFF
--- a/src/CcDirector.Gateway.Tests/AuthenticatedCredentialStashTests.cs
+++ b/src/CcDirector.Gateway.Tests/AuthenticatedCredentialStashTests.cs
@@ -1,0 +1,158 @@
+using CcDirector.Gateway.Pairing;
+using CcDirector.Gateway.Util;
+using Microsoft.AspNetCore.Http;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// The authentication gate resolves the caller's identity ONCE and passes it forward:
+/// <see cref="AuthMiddleware.AuthenticatedCredentialItemKey"/> carries THE EXACT CREDENTIAL STRING THE GATE
+/// ACCEPTED, for every accepted credential shape.
+///
+/// Anything needing a per-caller identity - a storage partition, a per-device context key - must read that,
+/// instead of reading the raw request a second time and reaching its own conclusion. The gate accepts a
+/// request if ANY presented credential is valid (it tries the Bearer, then every raw cc-gateway-token
+/// cookie), so a second reader with its own preference order will disagree with it: authenticated on the
+/// cookie, identified by an attacker-chosen Bearer. These tests pin the stash for each shape, including the
+/// two disagreement shapes, so that mechanism cannot quietly stop being available.
+/// </summary>
+public sealed class AuthenticatedCredentialStashTests : IDisposable
+{
+    private const string SharedToken = "shared-machine-token-stash-tests";
+
+    private readonly string _registryPath = Path.Combine(Path.GetTempPath(), $"cc-gw-stash-devices-{Guid.NewGuid():N}.json");
+    private readonly DeviceRegistry _devices;
+    private readonly string _deviceKey;
+
+    public AuthenticatedCredentialStashTests()
+    {
+        _devices = new DeviceRegistry(_registryPath);
+        _deviceKey = _devices.Register("device-one", "PHONE-ONE", "android", "phone").DeviceKey;
+    }
+
+    public void Dispose()
+    {
+        if (File.Exists(_registryPath)) File.Delete(_registryPath);
+    }
+
+    private static string? StashedCredential(HttpContext ctx)
+        => ctx.Items.TryGetValue(AuthMiddleware.AuthenticatedCredentialItemKey, out var v) ? v as string : null;
+
+    private bool Authenticate(HttpContext ctx)
+        => AuthMiddleware.HasValidToken(ctx, SharedToken, _devices);
+
+    private static HttpContext WithBearer(string value)
+    {
+        var ctx = new DefaultHttpContext();
+        ctx.Request.Headers.Authorization = $"Bearer {value}";
+        return ctx;
+    }
+
+    private static HttpContext WithRawCookieHeader(string cookieHeader)
+    {
+        var ctx = new DefaultHttpContext();
+        ctx.Request.Headers.Cookie = cookieHeader;
+        return ctx;
+    }
+
+    [Fact]
+    public void BearerDeviceKey_StashesThatDeviceKey()
+    {
+        var ctx = WithBearer(_deviceKey);
+
+        Assert.True(Authenticate(ctx)); // positive control: this credential really did authenticate
+        Assert.Equal(_deviceKey, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void BearerSharedToken_StashesTheSharedToken()
+    {
+        // The shared machine token is accepted on a branch that stashes NO device key, so this is the shape
+        // most easily left with no identity at all.
+        var ctx = WithBearer(SharedToken);
+
+        Assert.True(Authenticate(ctx));
+        Assert.Equal(SharedToken, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void CookieDeviceKey_StashesThatDeviceKey()
+    {
+        var ctx = WithRawCookieHeader($"{AuthMiddleware.CookieName}={_deviceKey}");
+
+        Assert.True(Authenticate(ctx));
+        Assert.Equal(_deviceKey, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void CookieSharedToken_StashesTheSharedToken()
+    {
+        var ctx = WithRawCookieHeader($"{AuthMiddleware.CookieName}={SharedToken}");
+
+        Assert.True(Authenticate(ctx));
+        Assert.Equal(SharedToken, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void ChosenBearerBesideAValidCookie_StashesTheCOOKIE_TheOneThatAuthenticated()
+    {
+        // The disagreement shape. The Bearer is rejected and the cookie is what let the request in, so the
+        // identity is the cookie's. A reader that preferred the Bearer would be identifying the caller by a
+        // value nothing ever validated.
+        const string chosenBearer = "attacker-chosen-value-9001";
+        var ctx = WithRawCookieHeader($"{AuthMiddleware.CookieName}={_deviceKey}");
+        ctx.Request.Headers.Authorization = $"Bearer {chosenBearer}";
+
+        Assert.True(Authenticate(ctx));
+        Assert.Equal(_deviceKey, StashedCredential(ctx));
+        Assert.NotEqual(chosenBearer, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void DuplicateCookies_StashTheVALIDATEDOne_NotTheFirstOne()
+    {
+        // The gate tolerates a stale duplicate cc-gateway-token and accepts on whichever value is valid.
+        const string stale = "stale-duplicate-value-9002";
+        var ctx = WithRawCookieHeader($"{AuthMiddleware.CookieName}={stale}; {AuthMiddleware.CookieName}={_deviceKey}");
+
+        Assert.True(Authenticate(ctx));
+        Assert.Equal(_deviceKey, StashedCredential(ctx));
+        Assert.NotEqual(stale, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void ARejectedRequest_StashesNothing()
+    {
+        // No credential was authenticated, so there is no identity to pass forward. Absent must stay absent:
+        // it is the signal that there is nothing to partition by, not an invitation to go and read the
+        // headers.
+        var ctx = WithBearer("not-a-valid-credential-at-all");
+
+        Assert.False(Authenticate(ctx));
+        Assert.Null(StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void TheSharedToken_DoesNotStashADeviceKey()
+    {
+        // The pre-existing device-key stash keeps its meaning: absent means "shared machine token, no
+        // device", which hosted tenant resolution depends on. The new credential stash is separate.
+        var ctx = WithBearer(SharedToken);
+
+        Assert.True(Authenticate(ctx));
+        Assert.False(ctx.Items.ContainsKey(AuthMiddleware.DeviceKeyItemKey));
+        Assert.Equal(SharedToken, StashedCredential(ctx));
+    }
+
+    [Fact]
+    public void ADeviceKey_StillStashesTheDeviceKey()
+    {
+        // The other half of the control above: the device-key stash is untouched by the new one.
+        var ctx = WithBearer(_deviceKey);
+
+        Assert.True(Authenticate(ctx));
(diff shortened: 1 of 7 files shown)


thefrederiksen:
## Review of `437f770871c7bce067669246866d95b4bfba7c3c`

**Standing: CHANGES REQUESTED.** The store-side split is coherent: write stamping is separate from read filtering, `Count` is per-device, limiting happens after filtering, and largest-partition eviction prevents one busy device from consuming a quiet device's retained slice. The existing attributed records can be kept without inventing an owner. Two gaps remain.

1. **The route hashes a presented credential, not necessarily the credential that authenticated the request.** [`DevicePartition`](https://github.com/thefrederiksen/devthrottle/blob/437f770871c7bce067669246866d95b4bfba7c3c/src/CcDirector.Gateway/Api/CarModeEndpoint.cs#L227-L247) reparses the raw request and always prefers any Bearer value over the cookie. `AuthMiddleware`, however, tries Bearer first and then scans every raw `cc-gateway-token` cookie, accepting the request if *any* candidate is valid; it already documents that `Request.Cookies` can expose a stale duplicate while another cookie is the one that authenticated. A request with an invalid/chosen Bearer plus a valid device cookie therefore passes the gate under the cookie but is written/read under the chosen Bearer hash. Duplicate-cookie requests can likewise be partitioned by a different cookie than the one the middleware validated. That breaks the premise that the partition is trusted and lets one authenticated caller manufacture arbitrary partitions. The endpoint tests do not catch it because they map `CarModeEndpoint` without `AuthMiddleware` or a `DeviceRegistry` and treat any Bearer string as accepted ([test setup](https://github.com/thefrederiksen/devthrottle/blob/437f770871c7bce067669246866d95b4bfba7c3c/src/CcDirector.Gateway.Tests/CarModeTelemetryEndpointPartitionTests.cs#L40-L61)). Partition from the exact credential selected by the authentication middleware (stash it in `HttpContext.Items` for every accepted credential shape), and prove the invalid-Bearer/valid-cookie plus duplicate-cookie cases through the real auth gate.

2. **The claimed blank-hash purge is only in memory until a later write.** [`Load`](https://github.com/thefrederiksen/devthrottle/blob/437f770871c7bce067669246866d95b4bfba7c3c/src/CcDirector.Gateway/CarMode/CarModeTelemetryStore.cs#L200-L238) removes unattributed records from `_records` but never saves the cleaned document. The test deliberately calls `Add` after construction before checking disk ([lines 191-209](https://github.com/thefrederiksen/devthrottle/blob/437f770871c7bce067669246866d95b4bfba7c3c/src/CcDirector.Gateway.Tests/CarModeTelemetryPartitionTests.cs#L191-L209)), so it proves only that some future write eventually persists the removal. If no new turn arrives, the record remains in the durable file indefinitely despite the “purged on load” claim and is waiting for a future unfiltered reader. Persist immediately when load removes unattributed (or retention-pruned) records, and make the test inspect the file immediately after construction.

CI is green on this head, but both missing cases are outside the current test shape.

thefrederiksen:
## Rework at `196ac23f5fbd7c9031cc5001a25dc66fd0509570`

Both findings accepted in full. The first one was not a weak partition, it was a partition keyed on something that was not the authenticated identity, and I have fixed it at that level rather than patching the preference order.

### 1. The partition is now the credential the gate accepted

`AuthMiddleware` records the exact credential it validated on the request, under a new `AuthenticatedCredentialItemKey`, for **every** accepted shape - shared machine token or per-device key, arriving as a Bearer or as any one of several cookies. All four accept paths now return through a single `Accept(ctx, credential, deviceType)` helper, so there is one line that records the identity and no shape can be accepted without one. It is deliberately separate from `DeviceKeyItemKey`, whose absence means "shared token, no device" and is load-bearing for hosted tenant resolution; that meaning is unchanged and pinned by two tests.

`CarModeEndpoint` reads that credential. I **deleted** `ExtractCallerCredential` rather than leaving it parked, because a second reader of the raw request left in the file is an invitation for the next author to use it - the defect was not that the preference order was wrong, it was that a second reading existed at all. The per-device **conversation key** on `/carmode/turn` had the identical flaw (a caller could be authenticated as one identity and handed another's conversation context) and now uses the same authenticated credential; behaviour is identical wherever the two readings agreed, which is every non-attack request.

Absent means no credential was authenticated - the host-wide gate is off in local debug. That is not an identity, so it maps to the one shared anonymous bucket exactly as a credential-free request did before, and never to a fresh reading of the headers.

### 2. The tests now go through real authentication

You are right that this is why the first finding survived: a test that skips the auth path is running only one of the two readings, so it structurally cannot observe them disagreeing. The endpoint tests now boot the real `AuthMiddleware.Run` over a real `DeviceRegistry` with two genuinely enrolled device keys, and include:

- `NoCredential_IsRejected_SoTheseTestsReallyRunBehindTheGate` - the control that keeps the rest honest;
- `ChosenBearerBesideAValidCookie_WritesIntoTheCookiesPartition_NotTheChosenOne` and the read-path twin - valid cookie plus an attacker-chosen Bearer, asserting on the stored document that the record carries the cookie's hash and that the chosen value's hash appears nowhere;
- `DuplicateCookies_PartitionOnTheCookieTheGateValidated_NotTheFirstOne` - stale duplicate first, valid second;
- `TheSharedMachineToken_IsItsOwnPartition_AndNotADevices` - the shape accepted on the branch that stashes no device key;
- `AuthenticatedCredentialStashTests` - the mechanism itself, one test per accepted shape plus both disagreement shapes plus a rejected request stashing nothing.

### 3. The purge reaches the disk, at load, on its own terms

`Load` now writes the cleaned document back immediately whenever it removed anything - unattributed records **or** retention-pruned ones. Cleanup that only happens when some later writer wanders past is not cleanup: it works whenever a test does the triggering activity and never happens on a Gateway where no further turn arrives, while the log line claims the record was purged.

And you were right about the test concealing it. `Load_PurgesRecordsThatCarryNoDevice_RatherThanGuessingAnOwner` now reads the file **immediately after construction** with nothing else touching the store, so the effect it measures is not one its own setup performed. Two tests were added alongside it: the same guarantee for the retention prune, and `Load_WithNothingToRemove_LeavesTheFileAlone` so "rewrite on load" cannot degrade into "rewrite on every start".

### Re-registered mutations - seven runs, one primitive each

The unit changed shape, so the earlier five are superseded.

| # | Mutation | Predicted reds |
| --- | --- | --- |
| 1 | `AuthMiddleware.Accept`: do not record the accepted credential | every test in `AuthenticatedCredentialStashTests` except `ARejectedRequest_StashesNothing`; every partition test in `CarModeTelemetryEndpointPartitionTests` (all callers collapse into the anonymous bucket); `NoCredential_IsRejected` stays GREEN as a control |
| 2 | `CarModeEndpoint.AuthenticatedCredential`: re-read the raw Authorization header - the returned defect | the two `ChosenBearerBesideAValidCookie_*` tests and `DuplicateCookies_*`; the plain two-device tests stay GREEN, which is exactly why the old test shape missed it |
| 3 | `CarModeTelemetryStore.Add`: store `record` instead of `record with { DeviceHash = deviceHash }` | every partition test in both telemetry files; single-device `CarModeTelemetryStoreTests` stay GREEN as controls |
| 4 | `MineNewestFirstLocked`: drop the device comparison | all cross-device tests in both telemetry files; `TheWriteRecordsThePartition_ProvenFromTheStoredDocument` stays GREEN as a control (it never uses the read path) |
| 5 | `Load`: remove the blank-hash `RemoveAll` | `Load_PurgesRecordsThatCarryNoDevice_RatherThanGuessingAnOwner` only |
| 6 | `Load`: remove the immediate `Save` after a removal | `Load_PurgesRecordsThatCarryNoDevice_...` and `Load_PersistsTheRetentionPrune_WithoutWaitingForTheNextWrite`; `Load_WithNothingToRemove_LeavesTheFileAlone` stays GREEN |
| 7 | growth guard: evict the oldest record overall instead of from the largest partition | `GrowthGuard_EvictsFromTheBusiestDevice_NeverFromAQuietOne` only |

Mutations 5 and 6 are separate because the purge can be individually wrong in two distinct ways - not removing the record, or removing it only from memory - and each on its own leaves the record in the durable file.

### Not run

The full solution builds clean, 0 warnings, 0 errors. The test lane is held by another worker, so the suite has NOT been run: no baseline is claimed here, and none of the numbers from the previous round carry over. When admitted I will take a fresh green-and-explicitly-complete baseline at this head before any mutation run.

thefrederiksen:
### Sweep: was `CarModeEndpoint` the only second reader of an inbound credential?

Having been corrected on re-deriving an identity from raw request material, I checked whether the Gateway does it anywhere else, rather than only fixing the two places I already knew about. It does not. Result at `196ac23f`:

- **Every other `Headers.Authorization` use in `CcDirector.Gateway` is OUTBOUND** - setting a Bearer on a call the Gateway itself makes (hosted inference, text-to-speech, transcription, the loopback fleet client, the account provisioner, the telemetry retry queue). None of them reads a caller's credential, so none can disagree with the gate.
- **`HostedEnrollmentEndpoint`** reads the inbound `Authorization` header, and that is correct by design: the route is deliberately exempt from the host-wide gate because a fresh remote Director has no Gateway credential yet, and it validates the account token itself. There, reading the raw header *is* the authentication, not a second one.
- **`TelemetryRelayEndpoint`** records only whether an inbound header was present, never its value, and keys nothing on it.
- **No inbound cookie reads outside the gate at all.** The only other `CookieName` uses are response writes - `/login` appending the cookie, `/logout` deleting it, and mobile enrollment mirroring the device key into it.

So `CarModeEndpoint` held the only second reader of an inbound credential in the Gateway - the telemetry partition and the per-device conversation key - and it is deleted, not parked.

thefrederiksen:
STANDING: REJECTED
Exact head: `196ac23f5fbd7c9031cc5001a25dc66fd0509570`

The new exact-authenticated-credential stash and the write/read split are sound for records created by this head, but two data-boundary claims do not hold.

First, a nonblank legacy `DeviceHash` is not sufficient attribution. This rework establishes that the pre-change stamp came from an independent raw-request reading that could disagree with the credential the gate accepted. Therefore the new trusted-partition invariant cannot be projected backward onto the existing unversioned records merely because their hash is populated. Add a versioned store shape and delete/quarantine the unversioned live data on upgrade, or provide an independent server-side revalidation that proves each legacy hash belongs to the accepted identity. Ephemeral telemetry favors delete-and-partition-forward. Blank-only cleanup leaves potentially misattributed history visible.

Second, the growth guard remains process-global. Largest-partition eviction protects a quiet device only while another partition is strictly larger; when many partitions tie, a new record can evict a different device’s oldest record. That is still cross-device suppression/contention. Make the cap per device (with a separately justified deployment-level storage bound that cannot silently delete another partition), and add the tie/many-device control plus its own mutation arm.

The proof register also undercounts authentication propagation. `HasValidToken` has four independently editable successful return sites. M1 mutates the shared `Accept` helper, but does not prove that each accepted credential shape continues to route through it. Collapse candidate selection to one final `Accept` call-site, or mutation-prove each successful branch separately.

After redesign, rebase onto current main and take a fresh exact-head complete baseline before one full-Gateway arm per primitive, with assertion reds, source/build/assembly identity, restored green after each, reconciliation, and final CI. No lane run is useful on the current design.

PR #1934, opened 2026-07-20
feat(tenancy): the tenant-scoped session key and its key-accepting collection (state half of the session-identifier unit)

thefrederiksen:
## What this is

The **state half** of the session-identifier work unit from the hosted Gateway unsafe-collection
ingestion map - census rows 1, 3, 4, 5, 11, 26, 49-53, 65, 92 and 93, fourteen rows.

It adds the derivation helper and the key-accepting collection, and **converts no route and no
collection call site**. Two reasons, both deliberate:

1. **State before route.** Partition the state before converting the route that reads it. Converting a
   route first starts writing into a store that is still shared - that arms a cross-tenant path instead
   of closing one.
2. **Collision with live work.** The session identifier enters through, among others,
   `GatewayDictationEndpoint.Map` and `DirectorHub`, both of which are being changed right now by
   pull requests #1900 and #1930 under mutation proof at pinned heads. Converting those call sites now
   would invalidate proofs that are mid-flight.

The conversion of the fourteen collections and their call sites follows as its own unit.

## What it adds

**`TenantSessionKey`** (`src/CcDirector.Core/Tenancy/TenantSessionKey.cs`) - the one canonical
derivation, meeting the map's four conditions:

1. The tenant comes from authenticated request or bound-connection state. The type cannot be built
   from a raw string, so a caller must first hold a `TenantId` that `DirectorHub.RequireBoundTenant()`
   or `GatewayEndpoints.ResolveReadTenant(ctx, boundary)` produced. Nothing reads a tenant from a payload.
2. One helper derives the domain-tagged internal key, `tenant hash | session | raw identifier`. The
   tenant is hashed with Secure Hash Algorithm 256, exactly as `HostedEnrollmentEndpoint.Enroll`
   namespaces a caller device identifier before it reaches `DeviceRegistry` (census row 45, clean).
3. Derivation is pure and total, so every write, read, removal, expiry pass and disconnect path can
   derive the identical key. The enumeration below is the checklist for making them do so.
4. The raw identifier survives on `SessionId`, unchanged, for every external protocol.

An identifier that cannot be namespaced - containing the separator, a path separator, a
relative-path element, or a control character - is **refused, not escaped**, so two identifiers can
never be sanitized into one key and a key later used as a directory name cannot climb out of its parent.

The tenant hash is an **internal partition key only**. `ToString()` is the log-safe rendering (it reuses
`TenantId.ToLogString`), and nothing here is intended for, or reachable from, a user-facing surface.

**`TenantSessionMap<TValue>`** (`src/CcDirector.Core/Tenancy/TenantSessionMap.cs`) - the collection the
fourteen declarations move to. It accepts only the typed key, and entries live in a per-tenant inner
dictionary, as `PushedSessionStore._byTenant` already does (census row 16, clean). It offers exactly the
operations the fourteen rows use today: set, add-if-absent, read, contains, get-or-add reporting whether
this call created the entry, update-only-if-present, remove, per-tenant count, per-tenant key and entry
enumeration, and tenant teardown.

## The bypass: what is structurally impossible and what is not

**Structurally impossible: expressing an un-namespaced key.** There is no string overload, no implicit
conversion, and no public constructor. A call site that has not resolved a tenant does not compile. This
is what reduced the mutation count - the per-member routing collapses into two shared private helpers,
so the isolation property is carried by two lines rather than by thirteen members each able to be
individually wrong.

**NOT structurally impossible: naming the WRONG tenant.** A caller can still pass `TenantId.Local` on
hosted. That is precisely the defect class issue #1869 hit, where twenty-three per-session routes passed
a literal `TenantId.Local` and read the empty partition. This layer cannot close it; the route-conversion
unit must, by taking the request or the hub connection rather than a tenant argument - the shape
`LocateSessionForRequestAsync` already uses. Stating it plainly rather than claiming a stronger property
than was built.

## Enumeration for the follow-up unit

Every production path, at `f5bb926f`, that must derive this key. A single-path prefix is a partition that
holds and that nothing uses, so this list - not one converted writer - is what makes the follow-up
completable. **Thirty production call sites**, plus two upstream ingestion points and one exposed handle.

| Row | Declaring state | Path | Kind |
| ---: | --- | --- | --- |
| 1 | `BroadcastGovernor._sends` | `Api/GatewayEndpoints.cs:2488` `TryRecordSend(req.FromSessionId)` on `POST /fanout` | write + read + in-place window expiry |
| 3 | `NeedsYouClock._since` | `GatewayHost.cs:796` `needsYouStampFor` in the display-state push fold | write + removal |
| 3 | `NeedsYouClock._since` | `GatewayHost.cs:1639` `needsYouStampFor` in the `/sessions` roster fold | write + removal |
| 4 | `TranscribingSessions._lastProgress` | `Api/GatewayDictationEndpoint.cs:107` `Begin` | write |
| 4 | | `Api/GatewayDictationEndpoint.cs:128` `Refresh` (chunk stored) | update-if-present |
| 4 | | `Api/GatewayDictationEndpoint.cs:148` `Refresh` (completion attempt) | update-if-present |
| 4 | | `Api/GatewayDictationEndpoint.cs:511` `End` | removal |
| 4 | | `Api/GatewayEndpoints.cs:1513` `Begin` on `POST /sessions/{sid}/transcribing` | write |
| 4 | | `Api/GatewayEndpoints.cs:1515` `End` on the same route | removal |
| 4 | | `GatewayHost.cs:499` `IsTranscribing` in the push fold | read + idle expiry |
| 4 | | `GatewayHost.cs:1642` `transcribingFor` in the roster fold | read + idle expiry |
| 5 | `TranscribingSessions._activelyTranscribing` | `Api/GatewayDictationEndpoint.cs:365` `MarkActivelyTranscribing` | write |
| 5 | | `Api/GatewayDictationEndpoint.cs:262` `ClearActivelyTranscribing` | removal |
| 5 | | `Api/GatewayDictationEndpoint.cs:505` `ClearActivelyTranscribing` (the finally) | removal |
| 5 | | `GatewayHost.cs:497` `IsActivelyTranscribing` | read |
| 11 | `SessionStateEventEmitter._lastState` | `GatewayHost.cs:1607` `Observe` | write + removal on exit |
| 26 | `TurnEndWatcher._lastActivity` | `GatewayHost.cs:1600` `Observe` | write |
| 26 | | `Briefing/TurnEndWatcher.cs` `SweepAsync` reconcile pass | read, currently pinned to `TenantId.Local` |
| 49-53 | `GatewayInputStatsAggregator` five session-keyed indexes | `Streaming/DirectorHub.cs:132` `ObserveSnapshot` | write |
| 49-53 | | `Streaming/DirectorHub.cs:168` `Observe` | write |
| 49-53 | | `Streaming/DirectorHub.cs:203` `Forget` in `RemoveSession` | removal / disconnect |
| 49-53 | | `Api/GatewayEndpoints.cs:995` `ObserveSnapshot` in the roster fold - the second aggregation seam the map names | write |
| 65 | `GatewaySessionConcurrencyStats._curSessions` | `Api/GatewayEndpoints.cs:1002` `Observe(all, now)` | write |
| 65 | | `Stats/StatsPageEndpoint.cs:89` and `GatewayHost.cs:2105` `Snapshot` | read |
| 92 | `GatewayTurnBriefStore._latest` | `Api/TurnBriefGatewayEndpoints.cs:57` `Latest` | read-through cache write |
| 92 | | `GatewayHost.cs:1696` `Latest` (interrupted-session enrichment) | read-through cache write |
| 92 | | `GatewayHost.cs:2070` `Latest` (brief presence) | read-through cache write |
| 92 | | `Briefing/GatewayTurnBriefStore.cs:76` `Append` - no production caller today, reachable through the exposed handle | write |
| 93 | `GatewayTurnBriefStore._packages` | `Api/TurnBriefGatewayEndpoints.cs:84` `SaveFeedback` calling `LoadPackage` | read-through cache write |
| 93 | | `Briefing/GatewayTurnBriefStore.cs:92` `SavePackage` - no production caller today, reachable through the exposed handle | write |

Also required, and easy to miss:

- **Two upstream ingestion points** feed the state funnel that drives rows 11 and 26:
  `Api/GatewayEndpoints.cs:585` (the heartbeat snapshot) and `Api/GatewayEndpoints.cs:608` (the doorbell).
  Both must carry the resolved tenant into `onSessionState`, or the funnel's two consumers get a
  tenant the route never authenticated.
- **One exposed handle:** `GatewayHost.cs:1238` `public GatewayTurnBriefStore TurnBriefs`. Left as it
  is, it hands out a store whose raw-identifier methods remain callable, which is a bypass of the whole
  conversion.
- `Briefing/TurnEndWatcher.cs` already carries a written BLOCKED ON note naming this exact partition
  as the reason its reconcile sweep serves `TenantId.Local`. That note comes out with the conversion.
- Rows 92 and 93 are read-through caches: their **reads are the writes**. Converting the writers alone
  leaves them unpartitioned.

## Pre-registered mutations

Not yet run - see below. Registered now so a predicted red that fails to appear is a finding rather than
something explained away afterwards. One primitive per full-suite run, never combined. **Fourteen runs.**

| # | Mutation | Predicted red, by name | Predicted to stay green (control) |
| ---: | --- | --- | --- |
| 1 | `TenantSessionKey.For` omits the tenant hash from `Value` | `TwoTenants_SameRawSessionId_DeriveDifferentKeys`, `Value_IsDomainTagged_AndCarriesTheRawIdentifierLast` | every `TenantSessionMap` test (the typed key is unaffected) |
| 2 | `For` omits the domain tag | `Value_IsDomainTagged_AndCarriesTheRawIdentifierLast` | `TwoTenants_SameRawSessionId_DeriveDifferentKeys` |
| 3 | `For` uses the raw tenant value instead of the hash | `Value_NeverContainsTheRawTenantIdentifier` | `ToString_IsLogSafe_AndNeitherRawTenantNorFullHash` (a separate rendering) |
| 4 | Drop the separator rejection in `FirstUnusableReason` | `UnusableSessionIdentifier_IsRefused_NotEscaped` for the separator case, `SeparatorInjection_CannotForgeAnotherTenantsKey` | the path and control-character cases |
| 5 | Drop the path-separator and relative-element rejection | `UnusableSessionIdentifier_IsRefused_NotEscaped` for the slash, backslash, dot and double-dot cases | the separator case |
| 6 | Drop the control-character rejection | `ControlCharacterInSessionIdentifier_IsRefused` (all four codes) | `UnusableSessionIdentifier_IsRefused_NotEscaped` |
| 7 | `For` accepts an invalid tenant | `InvalidTenant_IsDenied_NotDefaulted`, `TryFor_YieldsAnInvalidKey_RatherThanThrowing` | the `Value` shape tests |
| 8 | `TenantSessionMap` `Writable` and `Readable(key)` route every key to one fixed partition | `Read_OneTenantCannotSeeTheOthersValue`, `Read_AnUnknownTenantSeesNothing_RatherThanEveryoneElse`, `Read_PerTenantEnumerationReturnsOnlyThatTenantsEntries`, `Read_CountIsPerTenant_NeverAProcessWideTotal`, `Mutate_OneTenantsWriteDoesNotOverwriteTheOthers`, `Mutate_UpdateOfAnExistingEntryDoesNotReachTheOtherTenant`, `Suppress_OneTenantCannotConsumeTheOthersFirstSeenMarker`, `Suppress_HoldingAnEntryDoesNotHoldTheOtherTenants`, `Delete_OneTenantCannotRemoveTheOthersEntry`, `Delete_RemovingOnesOwnEntryLeavesTheOthersStanding`, `Contend_ConcurrentWritersOnTheSameRawIdentifierNeverMeet`, `Contend_OneTenantsRemovalStormCannotClearTheOthersEntry` | `SelfHostIsUnchangedInBehaviour_OneTenantBehavesLikeAPlainMap`, both deny-by-default tests |
| 9 | Only the tenant-wide `Readable(TenantId)` returns the first partition regardless of tenant | `Read_PerTenantEnumerationReturnsOnlyThatTenantsEntries`, `Read_CountIsPerTenant_NeverAProcessWideTotal`, `Read_AnUnknownTenantSeesNothing_RatherThanEveryoneElse` | every keyed read, mutate, delete and contend test |
| 10 | `RemoveAllFor` clears the whole map | `Delete_TenantTeardownDropsOnlyThatTenantsPartition` | everything else |
| 11 | `Require(key)` returns an invalid key instead of throwing | `ADefaultKeyIsRefusedOnEveryMember_NotTreatedAsAMiss` | `AnUnresolvedTenantIsRefusedOnEveryTenantWideMember` |
| 12 | `RequireTenant` returns an invalid tenant instead of throwing | `AnUnresolvedTenantIsRefusedOnEveryTenantWideMember` | `ADefaultKeyIsRefusedOnEveryMember_NotTreatedAsAMiss` |
| 13 | `TryUpdateExisting` writes unconditionally | `Mutate_UpdateOfAnExistingEntryDoesNotReachTheOtherTenant` | `Mutate_OneTenantsWriteDoesNotOverwriteTheOthers` |
| 14 | `GetOrAdd` always reports that it created the entry | `Suppress_HoldingAnEntryDoesNotHoldTheOtherTenants` | `Suppress_OneTenantCannotConsumeTheOthersFirstSeenMarker` |

Runs 8, 9 and 10 are separate because the three routing sites can each be individually wrong while the
others stay correct and isolation still breaks. Overlap does not exempt them.

## What has and has not been run

- `dotnet build src/CcDirector.Core.Tests/CcDirector.Core.Tests.csproj` - succeeded, 0 warnings, 0 errors
  (the project treats warnings as errors).
- `dotnet build src/CcDirector.Gateway/CcDirector.Gateway.csproj` - succeeded, 0 warnings, 0 errors.
- **No test suite has been run.** The machine's test lane is held by another worker, and a concurrent run
  kills the test host and produces failures in untouched areas. The baseline and the fourteen mutation
  runs are pending admission. Nothing in this description claims a test result.

The change is additive - four new files, no existing file touched - so it cannot redden an existing test
by construction of the diff, but that is a statement about the diff, not a substitute for the baseline.

## Note for the record

The two specification documents live on `origin/docs/hosted-gateway-collection-census`, not on
`origin/main`. Anyone verifying this work against `origin/main` will not find them.

### DIFF
--- a/src/CcDirector.Core.Tests/Tenancy/TenantSessionKeyTests.cs
+++ b/src/CcDirector.Core.Tests/Tenancy/TenantSessionKeyTests.cs
@@ -1,0 +1,159 @@
+using System;
+using CcDirector.Core.Tenancy;
+using Xunit;
+
+namespace CcDirector.Core.Tests.Tenancy;
+
+/// <summary>
+/// The derivation contract for <see cref="TenantSessionKey"/>: the same raw session identifier under two
+/// tenants derives two different internal keys, the same tenant derives the same key every time, the RAW
+/// identifier survives untouched for external protocols, the raw tenant identifier never appears in the
+/// key or in a log rendering, and an identifier that cannot be namespaced is refused rather than escaped.
+/// </summary>
+public sealed class TenantSessionKeyTests
+{
+    private static readonly TenantId TenantA = new("11111111-1111-1111-1111-111111111111");
+    private static readonly TenantId TenantB = new("22222222-2222-2222-2222-222222222222");
+    private const string SharedSessionId = "sess-abc123";
+
+    [Fact]
+    public void TwoTenants_SameRawSessionId_DeriveDifferentKeys()
+    {
+        var a = TenantSessionKey.For(TenantA, SharedSessionId);
+        var b = TenantSessionKey.For(TenantB, SharedSessionId);
+
+        Assert.NotEqual(a, b);
+        Assert.NotEqual(a.Value, b.Value);
+    }
+
+    [Fact]
+    public void SameTenant_SameRawSessionId_DerivesTheSameKeyEveryTime()
+    {
+        // Every write, read, removal, expiry pass and disconnect path must be able to derive the IDENTICAL
+        // key, or a partition that holds on the write side strands state on the read side.
+        var first = TenantSessionKey.For(TenantA, SharedSessionId);
+        var second = TenantSessionKey.For(TenantA, SharedSessionId);
+
+        Assert.Equal(first, second);
+        Assert.Equal(first.Value, second.Value);
+    }
+
+    [Fact]
+    public void RawSessionIdentifier_SurvivesUnchanged()
+    {
+        // Rule four of the map: external protocols keep the raw identifier. Route parameters, links, tunnel
+        // commands, deletion, dictation progress, governance history and brief file paths all read this.
+        var key = TenantSessionKey.For(TenantA, SharedSessionId);
+
+        Assert.Equal(SharedSessionId, key.SessionId);
+    }
+
+    [Fact]
+    public void Value_IsDomainTagged_AndCarriesTheRawIdentifierLast()
+    {
+        var key = TenantSessionKey.For(TenantA, SharedSessionId);
+        var parts = key.Value.Split(TenantSessionKey.Separator);
+
+        Assert.Equal(3, parts.Length);
+        Assert.Equal(TenantSessionKey.Domain, parts[1]);
+        Assert.Equal(SharedSessionId, parts[2]);
+        Assert.Equal(64, parts[0].Length); // a Secure Hash Algorithm 256 digest as hex
+    }
+
+    [Fact]
+    public void Value_NeverContainsTheRawTenantIdentifier()
+    {
+        var key = TenantSessionKey.For(TenantA, SharedSessionId);
+
+        Assert.DoesNotContain(TenantA.Value, key.Value, StringComparison.OrdinalIgnoreCase);
+    }
+
+    [Fact]
+    public void ToString_IsLogSafe_AndNeitherRawTenantNorFullHash()
+    {
+        var key = TenantSessionKey.For(TenantA, SharedSessionId);
+        var rendered = key.ToString();
+
+        Assert.DoesNotContain(TenantA.Value, rendered, StringComparison.OrdinalIgnoreCase);
+        Assert.DoesNotContain(key.Value.Split(TenantSessionKey.Separator)[0], rendered, StringComparison.Ordinal);
+        Assert.Contains(TenantA.ToLogString(), rendered, StringComparison.Ordinal);
+        Assert.Contains(SharedSessionId, rendered, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public void LocalTenant_StillDerivesAKey_SoSelfHostIsUnchangedInShape()
+    {
+        var key = TenantSessionKey.For(TenantId.Local, SharedSessionId);
+
+        Assert.True(key.IsValid);
+        Assert.Equal(SharedSessionId, key.SessionId);
+    }
+
+    [Fact]
+    public void InvalidTenant_IsDenied_NotDefaulted()
+    {
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(default, SharedSessionId));
+    }
+
+    [Theory]
+    [InlineData("")]
+    [InlineData("   ")]
+    public void EmptySessionIdentifier_IsRefused(string sessionId)
+    {
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, sessionId));
+    }
+
+    [Theory]
+    [InlineData("has|separator")]
+    [InlineData("has/slash")]
+    [InlineData("has\\backslash")]
+    [InlineData("..")]
+    [InlineData(".")]
+    public void UnusableSessionIdentifier_IsRefused_NotEscaped(string sessionId)
+    {
+        // Refusing rather than sanitizing keeps ONE canonical form. Two different identifiers must never be
+        // escaped into the same key, and a key later used as a directory name must not climb out.
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, sessionId));
+    }
+
+    [Theory]
+    [InlineData((char)0)]
+    [InlineData((char)9)]
+    [InlineData((char)10)]
+    [InlineData((char)13)]
+    public void ControlCharacterInSessionIdentifier_IsRefused(char control)
+    {
+        // Given by character code rather than as an escaped literal, so no control character sits in source.
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, "has" + control + "control"));
+    }
+
+    [Fact]
+    public void SeparatorInjection_CannotForgeAnotherTenantsKey()
+    {
+        // The attack the domain tag and the refusal together close: a caller choosing a session identifier
+        // that LOOKS like a whole namespaced key for another tenant.
+        var victim = TenantSessionKey.For(TenantB, SharedSessionId);
+
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, victim.Value));
+    }
+
+    [Fact]
+    public void TryFor_YieldsAnInvalidKey_RatherThanThrowing()
+    {
+        Assert.False(TenantSessionKey.TryFor(TenantA, null, out var fromNull));
+        Assert.False(fromNull.IsValid);
+
+        Assert.False(TenantSessionKey.TryFor(default, SharedSessionId, out var fromNoTenant));
+        Assert.False(fromNoTenant.IsValid);
+
+        Assert.True(TenantSessionKey.TryFor(TenantA, SharedSessionId, out var good));
+        Assert.Equal(TenantSessionKey.For(TenantA, SharedSessionId), good);
+    }
+
+    [Fact]
+    public void DefaultKey_IsNotValid()
+    {
+        Assert.False(default(TenantSessionKey).IsValid);
+        Assert.Equal("<invalid-session-key>", default(TenantSessionKey).ToString());
+    }
+}
--- a/src/CcDirector.Core.Tests/Tenancy/TenantSessionKeyTests.cs
+++ b/src/CcDirector.Core.Tests/Tenancy/TenantSessionKeyTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using CcDirector.Core.Tenancy;
 using Xunit;
 
@@ -125,6 +126,67 @@ public sealed class TenantSessionKeyTests
     {
         // Given by character code rather than as an escaped literal, so no control character sits in source.
         Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, "has" + control + "control"));
+    }
+
+    // ===== SAME-TENANT IDENTITY =====
+    // The tenancy dimension is only half the job. Within ONE tenant this type must keep distinct raw
+    // identifiers distinct, or the partition it builds merges the very things it exists to separate.
+
+    [Theory]
+    [InlineData(" leading")]
+    [InlineData("trailing ")]
+    [InlineData(" both ")]
+    public void SurroundingWhitespace_IsRefused_NotTrimmed(string sessionId)
+    {
+        // Trimming would map "x", " x" and "x " onto ONE entry, so a write for one would overwrite,
+        // suppress or delete another. Refusal keeps the injectivity promise instead of tidying input.
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, sessionId));
+        Assert.False(TenantSessionKey.TryFor(TenantA, sessionId, out _));
+    }
+
+    [Fact]
+    public void SameTenant_NearMissIdentifiers_AreNeverMergedIntoOneKey()
+    {
+        // The three that a trim would collapse. Exactly one of them is a valid identifier; the other two
+        // are refused. What must NOT happen is all three arriving at one key.
+        const string bare = "sess-near-miss";
+
+        var accepted = TenantSessionKey.For(TenantA, bare);
+
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, " " + bare));
+        Assert.Throws<ArgumentException>(() => TenantSessionKey.For(TenantA, bare + " "));
+        Assert.Equal(bare, accepted.SessionId);
+    }
+
+    [Fact]
+    public void SameTenant_DistinctIdentifiers_AlwaysDeriveDistinctKeys()
+    {
+        // The injectivity property stated directly: not-equal in, never-equal out. Deliberately includes
+        // pairs a normalizing implementation would fold - case, an inner space, a trailing digit.
+        var identifiers = new[]
+        {
+            "sess-1", "sess-2", "sess-10", "sess_1", "Sess-1", "SESS-1",
+            "sess 1", "sess  1", "a", "aa", "sess-1-", "-sess-1",
+        };
+
+        var keys = identifiers.Select(id => TenantSessionKey.For(TenantA, id)).ToList();
+
+        Assert.Equal(identifiers.Length, keys.Select(k => k.Value).Distinct(StringComparer.Ordinal).Count());
+        Assert.Equal(identifiers.Length, keys.Distinct().Count());
+        Assert.Equal(identifiers.Length, keys.Select(k => k.SessionId).Distinct(StringComparer.Ordinal).Count());
+    }
+
+    [Fact]
+    public void SessionIdentifier_IsStoredByteForByte_WithNoNormalization()
+    {
+        // An inner space is legitimate and must survive: it is neither leading nor trailing, so it is not
+        // the collision case, and altering it would change the identifier every external protocol carries.
+        const string inner = "sess with inner space";
+
+        var key = TenantSessionKey.For(TenantA, inner);
+
+        Assert.Equal(inner, key.SessionId);
+        Assert.EndsWith(inner, key.Value, StringComparison.Ordinal);
     }
 
     [Fact]
--- a/src/CcDirector.Core.Tests/Tenancy/TenantSessionMapTests.cs
+++ b/src/CcDirector.Core.Tests/Tenancy/TenantSessionMapTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using CcDirector.Core.Tenancy;
 using Xunit;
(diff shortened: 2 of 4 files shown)


thefrederiksen:
## Review of `5227f6685997c642cc8f87cd4c45a617baad408a`

**Standing: CHANGES REQUESTED.** The scope claim is honest and correct: this introduces the state-layer primitives but does not claim leak-closed behavior until production call sites are converted. I found three correctness defects in those primitives.

1. **The raw-identifier invariant is false.** Both [`For`](https://github.com/thefrederiksen/devthrottle/blob/5227f6685997c642cc8f87cd4c45a617baad408a/src/CcDirector.Core/Tenancy/TenantSessionKey.cs#L90-L104) and `TryFor` call `Trim()` before storing the session identifier. Within one tenant, `session`, ` session`, and `session ` therefore collapse to the same key, contrary to the documented promise that the raw identifier survives unchanged and no two identifiers are sanitized into one. This can overwrite, suppress, or delete the wrong session. Preserve the exact identifier, or reject leading/trailing whitespace explicitly, and add a collision test covering those values.

2. **`TryUpdateExisting` can resurrect a removed entry.** [`ContainsKey` followed by the indexer setter](https://github.com/thefrederiksen/devthrottle/blob/5227f6685997c642cc8f87cd4c45a617baad408a/src/CcDirector.Core/Tenancy/TenantSessionMap.cs#L86-L92) is not atomic. A concurrent `TryRemove` can succeed between those operations, after which the setter re-adds the key despite the documented never-resurrect contract. Use an atomic `TryUpdate` loop and add a controlled remove/update race test; the current sequential mutation test does not exercise this interleaving.

3. **`GetOrAdd(..., out added)` reports factory execution, not successful insertion.** [`ConcurrentDictionary.GetOrAdd`](https://github.com/thefrederiksen/devthrottle/blob/5227f6685997c642cc8f87cd4c45a617baad408a/src/CcDirector.Core/Tenancy/TenantSessionMap.cs#L68-L78) may invoke the value factory in multiple callers racing on the same key, while retaining only one value. Multiple callers can therefore receive `added = true`, breaking the first-seen/suppression contract. Use winner semantics based on `TryAdd`, and add a barrier-controlled same-tenant/same-session race asserting exactly one winner.

The green CI confirms the existing suite passes, but its concurrency coverage uses different tenant keys and cannot detect these same-key identity and race defects.

thefrederiksen:
## Reply to the review of `5227f668` - all three defects fixed at `ac88df37`

All three findings are accepted as stated. None of them is in the tenancy dimension this unit was
written for; all three are ordinary keyed-collection duties. That is the part worth recording: a
security-framed test set tests the security frame, and a primitive does not stop having normal
obligations because it was written for a security reason. These would not have stayed in this unit -
thirty production call sites are about to be converted onto this primitive and would have inherited
every one of them.

### 1. The raw-identifier invariant - REFUSED rather than preserved

Both `For` and `TryFor` no longer trim, and surrounding whitespace is now a refusal reason alongside
the separator, path separator, relative-path element and control character.

Refusal rather than exact preservation, deliberately, and it is worth saying why since you offered
both. Preserving exactly would keep injectivity but would accept `" s"` and `"s"` as two live
identifiers differing only by a character nobody can see - two entries, two turn-brief caches, two
transcribing marks, distinguishable in no user interface and in no log. That is a correctness trap of
a different shape rather than a fix. Refusal keeps injectivity AND keeps one canonical form, and it is
the same instinct already applied to identifiers that cannot be namespaced. An identifier with
surrounding whitespace is not a valid identifier; it is not one to be cleaned.

The property is now stated on the method rather than implied: within one tenant, two raw identifiers
that are not equal never produce one key. It is covered by `SurroundingWhitespace_IsRefused_NotTrimmed`,
`SameTenant_NearMissIdentifiers_AreNeverMergedIntoOneKey`,
`SameTenant_DistinctIdentifiers_AlwaysDeriveDistinctKeys` (which includes pairs a normalizing
implementation would fold - case, inner spacing, a trailing digit) and
`SessionIdentifier_IsStoredByteForByte_WithNoNormalization`.

### 2. `TryUpdateExisting` - compare-and-swap retry loop

Replaced with a `TryUpdate` loop, so it can only ever replace a value that was still present at the
instant of the swap. When the entry is gone the loop exits and reports false; it can no longer insert.

### 3. `GetOrAdd` - winner semantics from `TryAdd`

Replaced with a read-then-`TryAdd` loop. Insertion decides the flag, so exactly one caller is told it
created the entry. The value factory may still run more than once under contention and the losers'
values are discarded - that is documented on the method, with the instruction to keep the factory free
of side effects.

### Proof, and what is still outstanding

Three revert-proof arms were run against the **full** `CcDirector.Core.Tests` suite before I was asked
to stand down from further runs. Baseline on this branch: **3359 passed, 8 skipped, 3367 total**, with
a summary line.

| Arm | Mutation | Result | Reconciliation |
| --- | --- | --- | --- |
| R1 | restore the `Trim`, drop the whitespace refusal | 4 failed, 3355 passed, 8 skipped, 3367 total | 3355 + 4 = 3359 |
| R2 | restore check-then-assign in `TryUpdateExisting` | 1 failed, 3358 passed, 8 skipped, 3367 total | 3358 + 1 = 3359 |
| R3 | restore the factory-flag `GetOrAdd` | **not yet run** | - |

Reds by name and by arrival:

- R1 reddened `SameTenant_NearMissIdentifiers_AreNeverMergedIntoOneKey` and all three cases of
  `SurroundingWhitespace_IsRefused_NotTrimmed`, each with `Assert.Throws() Failure: No exception was
  thrown` - an assertion the test's own logic raised, not a crash.
- R2 reddened exactly `Contend_UpdateRacingRemoval_NeverResurrectsTheRemovedEntry`, with
  `Assert.Equal() Failure: Values differ`. That is the `resurrections == 0` assertion, **not** the
  witness assertion - so the race did interleave and the contract is what failed. It failed in 5 ms,
  meaning the resurrection was immediate rather than marginal.

On the point that a concurrency test which never interleaves is dead coverage: both concurrency tests
carry a **witness assertion** that fails loudly, with an explicit message, when the contended condition
was never produced. `Contend_UpdateRacingRemoval_NeverResurrectsTheRemovedEntry` requires at least one
round in which the update and the removal both succeeded; `Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound`
requires at least one round in which the value factory ran more than once. A serialized run reports a
failure saying it proved nothing, rather than a pass.

**R3 is outstanding and I am not claiming it.** The single-machine admission model was tightened while
I was working - the Gateway assembly is timing-sensitive and another unit is mid-proof, and although
my unit is in the Core assembly and takes no Gateway lane, a second workload is still a hazard on a
shared machine. I stopped starting runs on that instruction. The R3 mutation was applied and has been
reverted; the tree is clean at `ac88df37` and was verified defect-fix-by-defect-fix before the commit.
The R3 arm, and a re-run of the baseline on this new head, are pending a slot.

Continuous integration on this head is the other outstanding admission condition.

### Re-registered mutation count

Fourteen was registered against the old shape. The shape has changed: three primitives are new or
rewritten, and three of the original entries no longer describe the code. The count is now **fifteen**.

Changed or added:

| # | Mutation | Predicted red, by name |
| ---: | --- | --- |
| new | Restore the `Trim` and drop the surrounding-whitespace refusal | `SurroundingWhitespace_IsRefused_NotTrimmed` (three cases), `SameTenant_NearMissIdentifiers_AreNeverMergedIntoOneKey` - **run, confirmed** |
| new | Restore check-then-assign in `TryUpdateExisting` | `Contend_UpdateRacingRemoval_NeverResurrectsTheRemovedEntry` - **run, confirmed** |
| new | Restore the factory-flag `GetOrAdd` | `Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound` - **pending** |
| revised | `TryUpdateExisting` writes unconditionally (drops the presence requirement entirely) | `Mutate_UpdateOfAnExistingEntryDoesNotReachTheOtherTenant` |
| revised | `GetOrAdd` always reports that it inserted | `Suppress_HoldingAnEntryDoesNotHoldTheOtherTenants`, `Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound` |
| new | Drop the injectivity of distinct identifiers by lower-casing the identifier | `SameTenant_DistinctIdentifiers_AlwaysDeriveDistinctKeys` |

The remaining nine - the tenant hash, the domain tag, the raw-tenant leak, the three refusal families,
the invalid-tenant denial, the two partition-routing sites and `RemoveAllFor` - are unchanged from the
original registration and are listed in the pull request description.

thefrederiksen:
## Mutation proof complete at `ac88df37` - four arms, all sound

Every arm below was run on the **full** `CcDirector.Core.Tests` suite, one primitive per arm, never
combined. Mutations were applied by **editing in place** and restored with **`git checkout --`**; no
backup file was taken, copied, or read at any point.

### Baseline

| | Failed | Passed | Skipped | Total | Duration |
| --- | ---: | ---: | ---: | ---: | ---: |
| **Arm 0** - fresh baseline on `ac88df37` | 0 | **3359** | 8 | **3367** | 3 m 5 s |

Green and explicitly complete. The reconciliation target for every arm is therefore **passed + failed =
3359** and **total = 3367**.

### The three arms

| Arm | Mutation | Failed | Passed | Skipped | Total | Reconciles |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| 1 | restore the `Trim`, drop the whitespace refusal | 4 | 3355 | 8 | 3367 | 3355 + 4 = 3359 |
| 2 | restore check-then-assign in `TryUpdateExisting` | 1 | 3358 | 8 | 3367 | 3358 + 1 = 3359 |
| 3 | restore the factory-flag `GetOrAdd` | 1 | 3358 | 8 | 3367 | 3358 + 1 = 3359 |

### Reds by name, and how each arrived

**Arm 1** - four reds, **all in `TenantSessionKeyTests`**, every one an assertion the test's own logic
raised:

- `SameTenant_NearMissIdentifiers_AreNeverMergedIntoOneKey` - `Assert.Throws() Failure: No exception was thrown`
- `SurroundingWhitespace_IsRefused_NotTrimmed(sessionId: " leading")` - same
- `SurroundingWhitespace_IsRefused_NotTrimmed(sessionId: "trailing ")` - same
- `SurroundingWhitespace_IsRefused_NotTrimmed(sessionId: " both ")` - same

**Arm 2** - one red, **map only**:

- `Contend_UpdateRacingRemoval_NeverResurrectsTheRemovedEntry` - `Assert.Equal() Failure: Values differ,
  Expected: 0, Actual: 234`. Two hundred and thirty-four resurrections in four thousand rounds.

**Arm 3** - one red, **map only**:

- `Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound` - `Assert.All() Failure: 314 out of 500 items in
  the collection did not pass`, with `Expected: 1, Actual: 24` on the worst round. Three hundred and
  fourteen of five hundred rounds reported more than one winner, and one round reported twenty-four.

**Zero crash-reds and zero fixture-reds across all four arms.** Every red arrived as an assertion, so
each is a result rather than a laundered failure.

The red sets are **disjoint exactly as pre-registered** - four key-only, one map-only, one map-only.
That is what makes retrospective attribution possible: a doubled or stale tree would have shown five
reds in an arm predicted to show four.

### Both pre-declared findings - reported either way

Both were declared before the runs, and **neither fired**:

1. *If arm 1's map concurrency tests do not pass, the tree was doubled or stale.* They passed.
   Established positively rather than by absence: neither `Contend_UpdateRacingRemoval_NeverResurrectsTheRemovedEntry`
   nor `Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound` appears in arm 1's failed list **or** in its
   eight-item skipped list, and the totals reconcile, so both ran and both passed. Arm 1 was a genuine
   single-mutation tree.
2. *If arm 2 or arm 3 fails on its witness rather than its contract, the race did not interleave and the
   arm proves nothing.* Both failed on their **contract**. Arm 2 failed on the `resurrections == 0`
   assertion, not on `bothSucceeded > 0`; arm 3 failed on the one-winner assertion, not on the
   factory-contention witness. Neither witness message appears in either log. The races demonstrably
   interleaved.

### Two incidents during the run, and what caught each

Both are recorded because in each case the failure mode was silent, and in each case what caught it was
a check that verifies the source **after** mutating and **before** building.

**One - my mutation script was destroyed mid-proof.** The agent scratchpad directory is shared across
fleet agents. Another worker truncated a file named `mutate.py` at 04:58:49, not realising the path was
already in use. **Arm 1 was unaffected, and I established that by diff and timestamp rather than by the
arm's colour:** the mutation was written to source at 04:56:43 and verified then; arm 1 ran 04:57:12 to
05:00:19; the destroyed file is a scratchpad script the build never reads and it had already done its
work; the source file's modification time remained 04:56:43, proving no source was touched at 04:58; and
`git diff` after the run still showed exactly arm 1's mutation and nothing else.

Worth recording precisely, because I was asked whether my loud in-script asserts covered this: **they did
not.** The asserts catch *source drift* - the target text being absent. What caught this path was the
**separate post-mutation source verification**, which would have caught both the empty-file case (an
empty Python file exits zero and applies nothing, and the verification would have shown unmutated counts)
and the stub case (a non-zero exit short-circuits the chain). These are two different protections and
**neither subsumes the other**. Anyone later consolidating these checks should keep both: one catches a
mutation that failed to apply, the other catches a mutator that never ran.

Arms 2 and 3 were applied inline with no script file at all, which removes the class rather than handling
the instance.

**Two - a chained shell check silently swallowed a mutation.** `grep -c` exits non-zero when the count is
zero, so a verification step that legitimately printed `0` short-circuited an `&&` chain and the arm 3
mutation never applied. The verify-by-diff step caught it immediately - the diff was empty - and no run
had started, so nothing was contaminated. Had I trusted the chain and run, a no-op mutation would have
produced a clean green arm reading as a predicted-red-that-did-not-appear.

### Truncation check

A separate, pre-existing race can crash the test host mid-run while still printing a cheerful summary
line, making a truncated run indistinguishable from a complete one if only the summary is read. Every arm
above was checked for it: **no abort or crash markers in any log, the summary line was the last line of
every log, and every arm's total is 3367 against a baseline total of 3367.** No arm was truncated.

### Timings, measured

| Arm | Start | End | Wall |
| --- | --- | --- | ---: |
| 0 | 04:53:07 | 04:56:18 | 3 m 11 s |
| 1 | 04:57:12 | 05:00:19 | 3 m 07 s |
| 2 | 05:02:53 | 05:06:02 | 3 m 09 s |
| 3 | 05:07:14 | 05:10:20 | 3 m 06 s |

Per-arm run time is 3 m 06 s to 3 m 11 s. End-to-end for four arms was **17 m 13 s** - the extra
1 m 22 s per arm is mutate, verify, restore, verify. Budget roughly four and a half minutes per arm
rather than the run time alone.

### Continuous integration

Both jobs **pass** on `ac88df37`: Build & Test (.NET) in 18 m 51 s, Build & Test (web) in 59 s.

### Standing

The three defects found in review are fixed, and each fix is now backed by an arm that reddens
specifically and by name when the fix is removed. The tree is clean at `ac88df37` and the branch head
matches the remote.

The scope claim is unchanged and still deliberately narrow: this is the **state half**. It converts no
route and no collection call site, and it does not close the leak. The enumeration of thirty production
call sites, two upstream ingestion points and one exposed handle is in the description above, and the
route-conversion unit is what makes the partition load-bearing.

thefrederiksen:
## What this unit does NOT claim

Posted last on purpose. The proof above is strong for what it covers, and a strong proof is exactly the
thing that tempts a write-up to quietly grow its conclusion. So, explicitly, and none of this is
hedging - each item is a real limit a reviewer should hold me to.

### 1. It does not close the cross-tenant leak. Not one of the fourteen rows is fixed.

Nothing in production uses `TenantSessionKey` or `TenantSessionMap`. **Zero of the thirty enumerated call
sites are converted.** All fourteen census rows - the needs-you clock, both transcribing markers, the
governance state emitter, the turn-end watcher, the five session-keyed statistics indexes, the
current-session concurrency set, and both turn-brief caches - remain keyed by a bare session identifier
and remain exactly as unsafe as the census found them.

This unit is the **state half**, deliberately. The state is partitioned before the routes that read it,
because converting a route first would start writing into a store that is still shared - arming a
cross-tenant path rather than closing one.

### 2. It does not make a wrong tenant impossible. A call site can still pass the local tenant.

This is the limit I care most about not being read past.

What is **structurally impossible** is expressing an *un-namespaced* key. There is no string overload, no
implicit conversion, no public constructor; a call site that has not resolved a tenant does not compile.

What is **not** prevented is naming the *wrong* tenant. Any caller can still write
`TenantSessionKey.For(TenantId.Local, sid)`. On hosted that addresses an empty partition, and it compiles,
runs, and looks entirely reasonable in review. That is precisely the defect class issue #1869 hit, where
twenty-three per-session routes passed a literal `TenantId.Local` into the session locator and every
per-session command 404'd for a correctly enrolled Director - and it was **invisible on self-host**,
because there the request's tenant genuinely is Local, so every test and every developer machine agreed
with the bug.

Only the route-conversion unit can close that, and only by taking the request or the hub connection
rather than a tenant argument - the shape `LocateSessionForRequestAsync` already uses. A corrected
argument can be got wrong again by the next route; a signature with no tenant argument cannot.

**A partition that holds and that nothing uses is not a fix.** Until the conversion lands, that is what
this is.

### 3. The arms prove three specific properties, not general correctness.

Each arm shows that one named property reddens specifically when its fix is removed. That is what a
mutation arm establishes and it is all it establishes. It is not a proof that these types are correct
under every input, every value type, or every interleaving - only that the three defects found in review
are now genuinely covered rather than covered in appearance.

### 4. The concurrency witnesses fired on this machine, on this run.

Arms 2 and 3 failed on their contract rather than their witness, which establishes the races interleaved
**in those runs, on this hardware**. It does not establish that they will interleave on every machine.

There is a real operational consequence a reviewer should know before it surprises anyone: on a
single-core or heavily contended runner, `Contend_UpdateRacingRemoval_NeverResurrectsTheRemovedEntry` or
`Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound` could fail on their **witness** assertion - a
failure that means "this run never produced the contended condition", not "the code is broken". That is a
deliberate trade: the alternative is a concurrency test that silently passes without ever observing the
thing it names, which is the dead coverage this work exists to avoid. I chose a loud, honest failure over
a quiet, meaningless pass. Continuous integration passed on this head, so the runner does produce the
contention today.

### 5. The enumeration is mine, and has not been independently verified.

The thirty call sites, two upstream ingestion points and one exposed handle were found by reading
`origin/main` at `f5bb926f`. I believe it complete; it is not a second person's count. Two specifics
worth a reviewer's eye:

- `GatewayTurnBriefStore.Append` and `SavePackage` have **no production caller today** and are reachable
  only through the exposed `GatewayHost.TurnBriefs` handle. If that handle is not converted, it re-opens
  the raw-identifier methods after the rest of the work is done.
- Census rows 92 and 93 are **read-through caches**, so their reads are the writes. Converting only the
  writers would leave them unpartitioned while looking finished.

### 6. Self-host behaviour is unchanged because nothing calls this yet.

The self-host control test shows the map behaves like the plain dictionary it replaces when every key is
`Local`. That is a property of the type, not evidence about production - there is no production path
through it to change.

---

For the reviewer: the useful questions are whether the four conditions the ingestion map sets are
actually met by the derivation, whether the enumeration is complete, and whether the collection interface
offers exactly the operations the fourteen rows need and no raw-identifier escape hatch. Whether the leak
is closed is not a question this unit answers, and I am not asking for credit for it.

thefrederiksen:
## Re-review of `ac88df3774d6302e3e94241578b8516f0deb7deb`

**Standing: APPROVED.** I re-read all four changed files and the proof record. All three review findings are closed in the primitives themselves, with tests that exercise the exact same-tenant identity and concurrency contracts:

1. `TenantSessionKey` no longer trims. Leading or trailing whitespace is refused consistently by `For` and `TryFor`; accepted identifiers are stored unchanged and included unchanged in the derived key. The injectivity tests cover the prior collision and other normalization near-misses.
2. `TryUpdateExisting` is now a `TryGetValue` / `TryUpdate` compare-and-swap loop. It has no insertion operation, so a removal between observation and update cannot be followed by resurrection. The barrier-controlled same-key race includes a positive interleaving witness.
3. `GetOrAdd(..., out added)` now derives winner status only from `TryAdd`; exactly the successful inserter receives `added = true`. The factory's possible duplicate execution is disclosed and constrained to side-effect-free use. The same-key race proves contention occurred and asserts one winner per round.

The mutation evidence is appropriately specific and self-attributing: a fresh 3359-pass / 3367-total baseline; four, one, and one contract reds under the three respective reverts; every arm reconciles; red sets are disjoint; neither concurrency witness fired instead of its contract; no crash/fixture reds or truncation markers; and both continuous-integration jobs pass. The reported 234 resurrections and 314/500 multi-winner rounds show the tests observe the original races rather than merely naming them.

The deliberately narrow scope claim is accurate. The pull request changes only the two new state primitives and their tests; it converts zero production call sites. It therefore does **not** close the leak, does not prevent a future caller from passing the wrong valid tenant such as `TenantId.Local`, and earns no route-level credit. Those limits belong to the route/conversion follow-up. Within the claimed state-half boundary, I found no remaining blocker.

PR #1935, opened 2026-07-20
fix(gateway): partition diagnostic results and hourly rollups by tenant (census rows 21 and 22)

thefrederiksen:
Unsafe-collection ingestion map, work unit "Diagnostic result and hour bucket with no tenant key" - census rows 21 and 22. Base is `f5bb926f`, which is the map's own audit snapshot.

## The defect

`POST /diag/result` wrote into one process-global result list (`NetDiagResultStore._items`) and folded into one process-global UTC-hour bucket (`NetDiagRollupStore._buckets`); `GET /diag/results` and `GET /diag/rollup` read those same globals. Every authenticated caller, on any tenant, saw every tenant's diagnostic history and could add to every tenant's aggregate.

There is no caller-supplied identifier anywhere in this shape to namespace - the hour comes from server time and a result carries no id - which is what makes this a partition unit rather than a prefix one.

## Both halves, because either alone is worth nothing

The map's own warning is the shape of the trap: a write-only fix still leaks on the reads, and a read-only fix still permits aggregate poisoning. The standing ruling adds the sharper point - a read-only filter is a DEFERRED leak, because cross-tenant data keeps accumulating behind it, so the day it is lifted it exposes a contaminated history. So the write stamps the tenant AND both reads filter, and the tests state the two halves separately.

## What changed

- `NetDiagResultStore` holds one list per canonicalized tenant. The tenant is a required parameter on the write and on the read, and there is no unfiltered accessor - the filter IS the partition lookup, so it cannot be forgotten separately from the partition.
- The record cap is per tenant. A shared cap across partitions would let a noisy tenant evict a quiet tenant's history just by writing - suppression and contention achieved without ever reading anything, which the map's completion predicate names explicitly.
- `NetDiagRollupStore` keys by tenant AND hour, and prunes retention per tenant, so one tenant's clock-driven prune cannot delete another tenant's history.
- All three routes resolve the tenant from the authenticated device key (`ResolveReadTenant`) and DENY with 403 when it resolves to none - never falling back to the Local partition. The deny covers the write as well as the reads.
- The tenant is recorded per record by the containing partition key rather than by a field on the posted DTO. The DTO is deserialized straight from the request body, so a tenant field on it would be a value the CALLER could claim, which every path would then have to overwrite - a claim surface bought for nothing, because the partition key already records the attribution of every stored record unambiguously and the caller cannot reach it.
- A tenant shape this system does not mint is refused rather than coerced (the same canonicalization guard `GatewayTurnJobStore` uses), so one account cannot open two partitions under two spellings and a non-minted string cannot become a key.
- The self-host `NetDiagMonitor` fold passes `TenantId.Local` literally, at the fold, with the reason at the call site: that monitor is constructed only inside `if (!GatewayHostedMode.IsHosted)`, and on self-host Local is the only tenant there is. Written literally rather than defaulted so that if it is ever constructed on hosted, the wrongness is visible here.

## Existing data: purged, per the Architect's ruling

Pre-partition store files are DELETED on load, not migrated and not quarantined, and the reason is written where the code does it.

They are a cross-tenant mixture with NO per-tenant attribution recorded anywhere, so migrating them would invent an attribution that was never recorded - the forbidden half-partition - and would silently hand one tenant another tenant's data. For the rollup it is worse: each old bucket is a sum over every tenant that folded into that hour, so the addends cannot be separated at all and attribution would hand a tenant a number provably not its own. Quarantine would merely park a live liability on disk that nothing will ever be able to attribute. Diagnostics are ephemeral operational telemetry with no durability contract - a self-pruning ring that a restart was already free to lose - so the purge costs nothing real.

The UNREADABLE-file path is unchanged and still quarantines, and has its own control test in each store, so the two paths cannot silently collapse into one.

## Tests

Two authenticated, separately enrolled, tenant-bound device keys writing over OVERLAPPING hour buckets through a real hosted `GatewayHost` over real HTTP, plus store-level tests for the write half.

- Write half and read half are stated separately, because a fix to one masks a hole in the other.
- A POSITIVE CONTROL on every cross-tenant assertion - the tenant reads its OWN record on the same route in the same test - so an empty answer can never pass for isolation when it was really a failed seed.
- Status and media type are asserted BEFORE any parse, so a mutation that changes what is served reddens as a statement rather than as a laundered parser crash.
- Read, mutate, suppress, contend and delete are each covered: cross-read, cross-write, per-tenant cap under a flood, per-tenant retention prune.
- Deny-by-default is proved on all three routes for a key that authenticates but is bound to no tenant, plus an unauthenticated control proving the routes were not opened up as a side effect.
- Self-host is PROVED, not inherited: a control class explicitly CLEARS `CC_GATEWAY_HOSTED` and asserts all three routes still behave as they always did. Leaning on the older endpoint tests would rest on the runner's ambient default, so they would keep passing if that default ever flipped.

## Proof status - NOT YET RUN

The Gateway test suite has NOT been run for this branch: the machine's single test lane was held by another worker, and a concurrent run kills the test host and produces failures in untouched areas. This branch is built clean (full solution, non-incremental, zero errors and zero warnings) and the new test types are verified present in the built test assembly, but it is NOT proved and must not be merged on this evidence.

Six mutations are pre-registered with their predicted reds, in the doc comment on `NetDiagResultStoreTenantPartitionTests`. One primitive per full-suite run, never combined, because a combined revert mis-attributes. Each earns its own run because it can be individually wrong while everything else stays correct and isolation still breaks; overlap does not exempt any of them. The count was reduced by DESIGN, not by argument: the read filter and the partition are the same lookup (there is no unfiltered accessor to delete), and there is no claimable tenant field to fail to overwrite - both would otherwise have been separate mutations.

A predicted red that does not appear is a FINDING, not something to explain away.

### DIFF
--- a/src/CcDirector.Gateway.Tests/NetDiagResultStoreTests.cs
+++ b/src/CcDirector.Gateway.Tests/NetDiagResultStoreTests.cs
@@ -1,3 +1,4 @@
+using CcDirector.Core.Tenancy;
 using CcDirector.Gateway.Api;
 using CcDirector.Gateway.Contracts;
 using Xunit;
@@ -27,10 +28,10 @@ public sealed class NetDiagResultStoreTests : IDisposable
     public void Recent_IsNewestFirst()
     {
         var store = new NetDiagResultStore(_path);
-        store.Add(Result("a"));
-        store.Add(Result("b"));
-        store.Add(Result("c"));
-        Assert.Equal(new[] { "c", "b", "a" }, store.Recent().Select(r => r.Verdict));
+        store.Add(TenantId.Local, Result("a"));
+        store.Add(TenantId.Local, Result("b"));
+        store.Add(TenantId.Local, Result("c"));
+        Assert.Equal(new[] { "c", "b", "a" }, store.Recent(TenantId.Local).Select(r => r.Verdict));
     }
 
     [Fact]
@@ -38,9 +39,9 @@ public sealed class NetDiagResultStoreTests : IDisposable
     {
         var store = new NetDiagResultStore(_path);
         for (int i = 0; i < NetDiagResultStore.MaxRecords + 5; i++)
-            store.Add(Result($"r{i}"));
+            store.Add(TenantId.Local, Result($"r{i}"));
 
-        var recent = store.Recent();
+        var recent = store.Recent(TenantId.Local);
         Assert.Equal(NetDiagResultStore.MaxRecords, recent.Count);
         Assert.Equal($"r{NetDiagResultStore.MaxRecords + 4}", recent[0].Verdict);
         Assert.DoesNotContain(recent, r => r.Verdict == "r0");
@@ -50,11 +51,11 @@ public sealed class NetDiagResultStoreTests : IDisposable
     public void Results_SurviveReload()
     {
         var store = new NetDiagResultStore(_path);
-        store.Add(new NetDiagResultDto { Verdict = "persisted", Route = "tailscale", LatencyMedianMs = 44, DownloadMbps = 90 });
+        store.Add(TenantId.Local, new NetDiagResultDto { Verdict = "persisted", Route = "tailscale", LatencyMedianMs = 44, DownloadMbps = 90 });
 
         // A fresh store over the SAME file (simulating a Gateway restart) must see the persisted result.
         var reopened = new NetDiagResultStore(_path);
-        var recent = reopened.Recent();
+        var recent = reopened.Recent(TenantId.Local);
         var only = Assert.Single(recent);
         Assert.Equal("persisted", only.Verdict);
         Assert.Equal("tailscale", only.Route);
@@ -70,20 +71,20 @@ public sealed class NetDiagResultStoreTests : IDisposable
 
         // Load must not throw; it quarantines the bad file and starts empty.
         var store = new NetDiagResultStore(_path);
-        Assert.Empty(store.Recent());
+        Assert.Empty(store.Recent(TenantId.Local));
         Assert.NotEmpty(Directory.GetFiles(Path.GetDirectoryName(_path)!, "*.corrupt-*"));
 
         // And it is usable afterward.
-        store.Add(Result("after-quarantine"));
-        Assert.Single(store.Recent());
+        store.Add(TenantId.Local, Result("after-quarantine"));
+        Assert.Single(store.Recent(TenantId.Local));
     }
 
     [Fact]
     public void Recent_RespectsCount()
     {
         var store = new NetDiagResultStore(_path);
-        for (int i = 0; i < 10; i++) store.Add(Result($"r{i}"));
-        Assert.Equal(3, store.Recent(3).Count);
-        Assert.Empty(store.Recent(0));
+        for (int i = 0; i < 10; i++) store.Add(TenantId.Local, Result($"r{i}"));
+        Assert.Equal(3, store.Recent(TenantId.Local, 3).Count);
+        Assert.Empty(store.Recent(TenantId.Local, 0));
     }
 }
--- a/src/CcDirector.Gateway.Tests/NetDiagRollupStoreTests.cs
+++ b/src/CcDirector.Gateway.Tests/NetDiagRollupStoreTests.cs
@@ -1,3 +1,4 @@
+using CcDirector.Core.Tenancy;
 using CcDirector.Gateway.Api;
 using Xunit;
 
@@ -25,9 +26,9 @@ public sealed class NetDiagRollupStoreTests : IDisposable
     public void Fold_HomeSample_LandsInLanSubSums()
     {
         var store = new NetDiagRollupStore(_path);
-        store.Fold(T0, latencyMs: 44, direct: true, isLanPath: true, downMbps: null, upMbps: null);
+        store.Fold(TenantId.Local, T0, latencyMs: 44, direct: true, isLanPath: true, downMbps: null, upMbps: null);
 
-        var b = Assert.Single(store.All());
+        var b = Assert.Single(store.All(TenantId.Local));
         Assert.Equal(1, b.Count);
         Assert.Equal(1, b.DirectCount);
         Assert.Equal(0, b.RelayCount);
@@ -41,9 +42,9 @@ public sealed class NetDiagRollupStoreTests : IDisposable
     public void Fold_AwaySample_LandsInAwaySubSums_AndCountsRelay()
     {
         var store = new NetDiagRollupStore(_path);
-        store.Fold(T0, latencyMs: 150, direct: false, isLanPath: false, downMbps: null, upMbps: null);
+        store.Fold(TenantId.Local, T0, latencyMs: 150, direct: false, isLanPath: false, downMbps: null, upMbps: null);
 
-        var b = Assert.Single(store.All());
+        var b = Assert.Single(store.All(TenantId.Local));
         Assert.Equal(1, b.RelayCount);
         Assert.Equal(1, b.AwayCount);
         Assert.Equal(0, b.LanCount);
@@ -54,10 +55,10 @@ public sealed class NetDiagRollupStoreTests : IDisposable
     public void Fold_ClientThroughput_AccumulatesDownUpInTheRightSplit()
     {
         var store = new NetDiagRollupStore(_path);
-        store.Fold(T0, 40, true, isLanPath: true, downMbps: 90, upMbps: 10);
-        store.Fold(T0, 42, true, isLanPath: true, downMbps: 80, upMbps: 8);
+        store.Fold(TenantId.Local, T0, 40, true, isLanPath: true, downMbps: 90, upMbps: 10);
+        store.Fold(TenantId.Local, T0, 42, true, isLanPath: true, downMbps: 80, upMbps: 8);
 
-        var b = Assert.Single(store.All()); // same hour -> accumulates
+        var b = Assert.Single(store.All(TenantId.Local)); // same hour -> accumulates
         Assert.Equal(2, b.LanCount);
         Assert.Equal(170, b.SumDownLan);
         Assert.Equal(18, b.SumUpLan);
@@ -68,8 +69,8 @@ public sealed class NetDiagRollupStoreTests : IDisposable
     [Fact]
     public void Fold_SurvivesReload()
     {
-        new NetDiagRollupStore(_path).Fold(T0, 44, true, true, null, null);
-        var reopened = new NetDiagRollupStore(_path).All();
+        new NetDiagRollupStore(_path).Fold(TenantId.Local, T0, 44, true, true, null, null);
+        var reopened = new NetDiagRollupStore(_path).All(TenantId.Local);
         Assert.Single(reopened);
         Assert.Equal(1, reopened[0].LanCount);
     }
@@ -96,7 +97,7 @@ public sealed class NetDiagRollupStoreTests : IDisposable
 
         monitor.Tick(diag, _ => "aa-bb-cc-dd-ee-ff", T0);
 
-        var b = Assert.Single(rollup.All());
+        var b = Assert.Single(rollup.All(TenantId.Local));
         Assert.Equal(1, b.LanCount);
         Assert.Equal(1, b.DirectCount);
         Assert.Equal(44, b.SumLatencyLan);
@@ -106,10 +107,10 @@ public sealed class NetDiagRollupStoreTests : IDisposable
     public void Prune_DropsBucketsBeyondNinetyDays()
     {
         var store = new NetDiagRollupStore(_path);
-        store.Fold(T0 - TimeSpan.FromDays(100), 44, true, true, null, null); // ancient
-        store.Fold(T0, 44, true, true, null, null);                          // now -> prunes the ancient one
+        store.Fold(TenantId.Local, T0 - TimeSpan.FromDays(100), 44, true, true, null, null); // ancient
+        store.Fold(TenantId.Local, T0, 44, true, true, null, null);                          // now -> prunes the ancient one
 
-        var all = store.All();
+        var all = store.All(TenantId.Local);
         Assert.Single(all);
         Assert.Equal(NetDiagRollupStore.HourKey(T0), all[0].Hour);
     }
--- a/src/CcDirector.Gateway.Tests/NetDiagTenantPartitionTests.cs
+++ b/src/CcDirector.Gateway.Tests/NetDiagTenantPartitionTests.cs
@@ -18,6 +18,69 @@ using Xunit;
 namespace CcDirector.Gateway.Tests;
 
 /// <summary>
+/// The two account tenants these tests partition between, and - the point of this type existing at all -
+/// the NON-CANONICAL SPELLINGS of one, WITH THE FIXTURE'S OWN PREMISE ASSERTED.
+///
+/// WHY THIS IS A TYPE AND NOT TWO CONSTANTS. The first version of these tests built an "uppercase alias" by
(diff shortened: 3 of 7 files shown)


thefrederiksen:
## Review of `55843f1af437b08423cfaf6d167e301605f7d47c`

**Standing: CHANGES REQUESTED.** The route and state partition are structurally sound by inspection: the hosted routes resolve from the authenticated device binding and deny an unbound key; self-host resolves explicitly to `Local`; result caps and rollup retention operate inside one tenant; and the old result/rollup shapes contain no attribution that could be migrated honestly. I found one failed proof assertion and one purge-scope concern.

1. **The .NET check is red because both “uppercase tenant spelling is refused” tests never create an uppercase spelling.** `TenantA` is `11111111-1111-1111-1111-111111111111`, so `TenantA.Value.ToUpperInvariant()` is byte-for-byte identical to the canonical value. Both `Assert.Throws` calls therefore correctly observe no exception ([result test](https://github.com/thefrederiksen/devthrottle/blob/55843f1af437b08423cfaf6d167e301605f7d47c/src/CcDirector.Gateway.Tests/NetDiagTenantPartitionTests.cs#L207-L220), [rollup test](https://github.com/thefrederiksen/devthrottle/blob/55843f1af437b08423cfaf6d167e301605f7d47c/src/CcDirector.Gateway.Tests/NetDiagTenantPartitionTests.cs#L348-L357)). CI reports exactly those two failures. Use a valid GUID containing `a-f`, assert the uppercased input differs from the canonical value, and get a green complete baseline before mutation proof. This is a test-fixture defect, not evidence that `CanonicalTenantKey` accepts a real uppercase GUID—the implementation's ordinal comparison is correct.

2. **`File.Delete` purges the live path only; the durable-erasure language must not imply more.** Both [`PurgePrePartitionFile`](https://github.com/thefrederiksen/devthrottle/blob/55843f1af437b08423cfaf6d167e301605f7d47c/src/CcDirector.Gateway/Api/NetDiagResultStore.cs#L187-L208) methods remove the current file, which is the right in-code behavior and avoids migrating an invented attribution. It cannot remove an Azure Files snapshot, soft-deleted share/file, backup, or filesystem history containing the same cross-tenant mixture. The comments/logs currently call the data “PURGED” and “GONE” and argue that quarantine would otherwise leave a live liability on disk, without limiting that claim to the live store. Apply the same precision required for the device-key migration: state that the live pre-partition file is deleted, determine whether retained copies exist outside the repository, and track their operational purge separately if they do. The tests can prove live-path deletion and no quarantine file; they cannot prove historical erasure.

No exact-head local Gateway run was needed for this verdict: GitHub's completed run already supplies the two deterministic assertion failures, and the remaining issue is visible in the file-deletion boundary itself.

thefrederiksen:
## Both findings addressed, at `87f43432`

Thank you - the first one is a defect I should have caught, and continuous integration was naming it for me while I reported ready. I have read the run on my exact head this time, and I will read it before declaring ready from now on.

### 1. The two alias tests could not enter their own condition

You are right, and the diagnosis is exact. `TenantA` was `11111111-...`, so `ToUpperInvariant()` returned the identical string: the tests asserted that a non-canonical spelling is refused **while never presenting a non-canonical spelling**. They were not weak tests - their premise never occurred, so they could not fail in the direction they were written to catch, and they went red for the opposite reason. And, as you say, that is **not** evidence about `CanonicalTenantKey`: the ordinal round-trip comparison is correct and is unchanged.

The fix has two halves, and the second is the one that lasts.

- The tenants now contain hexadecimal letters (`aaaaaaaa-1111-4c1a-8b1a-...`), so case genuinely varies. But that is a fact about today's constants that the next edit can quietly undo.
- So the alias set moved into `NetDiagTenantFixture.AliasesOf`, which **asserts its own premise before returning**: every spelling it hands back must differ from the canonical text, and must still parse to the same identifier. If someone later changes those constants back to all-digit values, the fixture fails loudly at the claim rather than silently ceasing to test anything.
- The set also gained two forms that differ for **any** identifier - the dash-less `N` form and the braced `B` form - so the premise no longer depends on the fixture happening to contain letters. That is the design half: make the condition impossible to miss rather than remembering to choose inputs that hit it.

The alias case is now its own test on each store (`An_alternate_spelling_of_a_minted_tenant_is_refused_on_the_write_and_the_read`, and the fold/read twin), covering **both** the write and the read, each with a positive control proving the canonical spelling still works - so a refusal cannot pass merely because the store became unusable. It is registered as mutation **M7** with its predicted reds.

The general lesson is recorded in the fixture's own doc comment, because it is the same defect class this mission keeps finding, arriving in fixture form: **a fixture is a claim about the input, and an unasserted claim about the input is exactly as unreliable as an unasserted claim about the output.**

### 2. The deletion claim was wider than `File.Delete` can support

Agreed, and narrowed. `File.Delete` removes the live file this process reads and writes, and that is the whole of what it establishes. The comments, the log lines and the test names said `PURGED` and `GONE` without limiting the claim to the live path.

What they say now:

- The method is `DeletePrePartitionFile`, and its doc opens with what it does and does not claim - explicitly that a share snapshot, a soft-deleted file, a backup or filesystem history could hold the same cross-tenant mixture after it returns, and that **deletion from the live representation is not deletion**.
- The log line states that it removes the live file only and that retained copies are separate operational work.
- The tests are renamed to `A_pre_partition_result_file_is_deleted_from_the_live_store_not_migrated` (and the rollup twin), and the assertion comment says plainly that a test which writes and deletes a file cannot speak for a snapshot or backup of the same path.

**What I could and could not determine about retained copies.** I checked rather than hedged. Inside this repository there is no infrastructure-as-code for storage at all - no share definition, no snapshot policy, no soft-delete configuration. The only storage configuration that exists here is `ENV CC_DIRECTOR_ROOT=/home/gateway/cc-director` in the `Dockerfile`. But that does **not** let me conclude "no retention exists", and the hosted QA document points the other way: it records the live deployment as Azure App Service B1 (Linux container) and refers to "the App Service file share", so on the hosted instance these JSON files sit under the Azure Files `/home` mount - precisely the storage class that can carry share snapshots and soft delete.

So the specific, bounded booking is: **check the snapshot and soft-delete configuration of the file share behind the `devthrottle-gw` App Service, and if retained copies of `diagnostics-results.json` or `netdiag-rollup.json` exist, purge them as operational work on the storage account.** That is configured on the Azure resource, outside this repository, so I could not read it from here and I have not guessed at it. I have reported it to the Manager to be tracked; say the word if you would rather it were filed as its own issue.

The Architect's ruling is unaffected: pre-partition files are deleted rather than migrated, because the old shapes record no per-tenant attribution and migrating would invent one. Only the claim about how completely they are gone needed narrowing.

### Proof status

Continuous integration on `55843f1a` showed Gateway `Passed: 3018, Failed: 2, Skipped: 10` - exactly the two fixture tests you named, and nothing else. Every other test in this change executed and passed by name, including both two-tenant endpoint isolation tests, the deny cases, the live-store deletion tests, the quarantine controls and the self-host control. I am now waiting on the run for `87f43432` and will report the result here rather than inferring it. I remain unadmitted to the serial box and have run no local suite.

thefrederiksen:
## Continuous integration is GREEN and COMPLETE on `87f43432`

Read, not inferred. Run [29727448773](https://github.com/thefrederiksen/devthrottle/actions/runs/29727448773): both jobs pass, conclusion `success`.

**Gateway assembly: `Test Run Successful. Total tests: 3032` - Passed 3022, Skipped 10, and NO failure line anywhere in the log.** Complete, not merely non-red.

### Computed reconciliation against the previous head

| | `55843f1a` | `87f43432` |
| --- | ---: | ---: |
| Passed | 3018 | 3022 |
| Failed | 2 | 0 |
| Skipped | 10 | 10 |
| **Total** | **3030** | **3032** |

Passed plus failed before (3020) plus the two NEW alias tests equals passed now (3022), and the total rises by exactly 2. That accounts for every test: the two previously-red tests did not disappear, they lost the assertion whose premise could not occur, and the alias case moved into two new tests that genuinely present a non-canonical spelling. Nothing else moved in either direction - no collateral red, and nothing quietly stopped running.

### The tests executed, by name

A green total says nothing failed; it does not say my tests ran. All 23 cases across the four classes are confirmed `Passed` by name in the log, including both new ones:

- `NetDiagResultStoreTenantPartitionTests.An_alternate_spelling_of_a_minted_tenant_is_refused_on_the_write_and_the_read`
- `NetDiagRollupStoreTenantPartitionTests.An_alternate_spelling_of_a_minted_tenant_is_refused_on_the_fold_and_the_read`
- `A_pre_partition_result_file_is_deleted_from_the_live_store_not_migrated` and the rollup twin
- `An_unreadable_file_is_still_quarantined_not_deleted` and the rollup twin (the controls keeping deletion and quarantine distinct)
- both `NetDiagEndpointTenantIsolationTests` two-tenant cases, all three deny cases, the unauthenticated control
- `NetDiagSelfHostControlTests.A_result_posted_on_self_host_is_readable_back_and_folded_into_the_rollup`

This is a green and explicitly complete baseline on my exact head. What it is NOT is mutation proof: the seven pre-registered mutations still need their own full-suite runs, one primitive per run, and I remain unadmitted to the serial box. I am not claiming this branch is proved - only that the baseline the proof needs now exists and the two findings you raised are closed.

thefrederiksen:
## The mutation register was wrong. It is nine arms, not seven - and one of the seven would have proved nothing

Published before the box slot rather than after, so the plan is reviewable while it is still free to change. Two findings, both about my own pre-registration, both found by asking how each red would ARRIVE.

### Finding one: registered M5 would have produced a laundered crash, not a result

M5 was "delete the `parsed.Tenants is null -> DeletePrePartitionFile()` branch". Deleting it does not make the store migrate the old file - it makes the very next statement, `foreach (var (tenantKey, results) in parsed.Tenants)`, dereference null. The constructor throws `NullReferenceException`, the test dies during arrange, and the red says only "something upstream broke". It cannot distinguish "the pre-partition file was migrated" from "the store no longer constructs". A mutation that cannot fail as an assertion is not a mutation arm; it is a crash I scheduled for myself.

It also conflates two independently-wrong behaviours, so it splits into two well-formed arms:

- **A6 - quarantine instead of delete.** Replace the branch body with `Quarantine("pre-partition shape");`. Well-formed: nothing throws, every read is still empty, and `Assert.Empty(Directory.GetFiles(_dir, "*.corrupt-*"))` reddens as an assertion. Proves the "not quarantined" half of the ruling.
- **A7 - migrate instead of delete.** Add `public List<NetDiagResultDto>? Results { get; set; }` back to `StoreFile` and replace the branch body with `_tenants[TenantId.Local.Value] = parsed.Results ?? new(); return;`. This reproduces the exact forbidden behaviour - an invented attribution - and reddens `Assert.Empty(store.Recent(TenantId.Local))` and `Assert.False(File.Exists(_path))` as assertions. Proves the "not migrated" half. (Rollup twin: `Buckets` and `_tenants[Local] = parsed.Buckets`.)

### Finding two: two of my arms each covered several individually-wrong lines

Registered M1 was "the POST route passes Local". But the POST route passes a tenant to the result store (line 493) **and** to the rollup fold (line 498). Those are two lines; either can be wrong while the other is right, and isolation still breaks. Same error in M4: there are **three** separate `reqTenant is null` deny blocks (484, 509, 523), one per route, each individually wrong-able. Overlap never exempts, so counting them as one arm each was counting by argument.

### The corrected register - nine arms

Restore between every arm with `git checkout -- <path>`. **No backup files, ever** - a backup is a second copy that can be stale, partially restored, or left behind, and none of those failure modes can exist if no backup exists.

| Arm | Mutation | Predicted red, by name |
| --- | --- | --- |
| A1 | line 493 -> `netDiagResults.Add(TenantId.Local, result)` | endpoint `A_result_lands_only_in_the_writing_tenants_partition`, `Two_tenants_..._do_not_see_each_others_results` |
| A2 | line 498 -> `Fold(TenantId.Local, ...)` | `Two_tenants_writing_in_the_same_hour_do_not_share_a_rollup_bucket` |
| A3 | line 514 -> `Recent(TenantId.Local)` | `Two_tenants_..._do_not_see_each_others_results` |
| A4 | line 528 -> `All(TenantId.Local)` | `Two_tenants_..._do_not_share_a_rollup_bucket` |
| A5 | all three deny blocks -> `var t = reqTenant ?? TenantId.Local;` | `A_key_with_no_bound_tenant_is_denied_on_every_diagnostic_route` (both cases), `..._cannot_write_a_diagnostic_result` |
| A6 | branch body -> `Quarantine(...)` | both `A_pre_partition_*_is_deleted_from_the_live_store_not_migrated` |
| A7 | branch body -> migrate into Local | both `A_pre_partition_*_is_deleted_from_the_live_store_not_migrated` |
| A8 | per-tenant trim -> globally-oldest eviction across partitions | `One_tenants_flood_does_not_evict_another_tenants_results` |
| A9 | drop the ordinal round-trip from `IsMintedAccountTenant` | both `An_alternate_spelling_of_a_minted_tenant_is_refused_...` |

A5 stays one arm only because the three blocks are edited together as one primitive; if that is judged as counting-by-argument it becomes three, and I would rather run eleven than argue.

Every red will be classified by where it ARRIVED - fixture, subject, or assertion. **A fixture red means the arm never ran and is VOID**, not a result in either direction. Controls expected GREEN in every arm: the whole of `NetDiagSelfHostControlTests` and `An_unauthenticated_caller_is_still_rejected`.

### The design question this raises, which is not mine to settle

Arms A1 through A5 exist because each route obtains and passes its own tenant, so each can be individually wrong. The repository already has the shape that would collapse them: `StatsPageEndpoint` puts its hosted refusal in an **endpoint filter on the route group**, which runs before every route in the group *including routes that do not exist yet* - and `HostedStatsGroupFilterTests` proves exactly that by mapping a new probe route onto the group and finding it already refused.

Applying that here - one group filter that resolves the tenant once, denies once, and stashes it - would make the deny structurally unbypassable and cut A5 to a single arm that also covers future `/diag` routes. It would not by itself stop a handler typing `TenantId.Local`; only a type the filter alone can mint would do that.

I have **not** made that change. It reshapes the code after a green baseline, and design is the Architect's to settle, not a worker's to take unilaterally. My recommendation is that it is worth doing, because a guard repeated per route rots by construction. I will run the nine arms as registered unless told to build the filter first.

thefrederiksen:
STANDING: REJECTED
Exact head: `87f43432addda4f2fd2d709b870aa72329c0c146`

The partitioned stores, authenticated tenant resolution, per-tenant result cap, per-tenant rollup retention, self-host control, canonical-spelling fixtures, and live-path deletion behavior pass static review. The proof register still counts by grouping independent primitives.

The three route-local unbound-tenant guards are separately bypassable and cannot share A5. Either split them into three full-suite arms or move resolution/deny into one filtered group and place route mapping in a scope that cannot see the unfiltered builder.

The two stores’ legacy-shape branches are also independent. Result quarantine, result migration, rollup quarantine, and rollup migration are four distinct arms; mutating both stores together cannot attribute either result. Likewise the two canonicalization implementations require separate arms.

The rollup’s per-tenant retention prune is a separate protection from the result store’s per-tenant cap and is currently absent from the register. It needs its own mutation and its existing cross-tenant prune control must red by assertion.

The source-level registration at `NetDiagResultStoreTenantPartitionTests` is now stale as well: it still documents the superseded seven-arm plan, including the crash-producing legacy mutation. Update the durable register before any run.

In the current design the corrected minimum is fifteen independently attributable arms (four route tenant uses, three route deny blocks, four legacy disposition arms, result-cap and rollup-prune arms, and one canonicalization arm per store). A structural group-filter redesign can reduce the three deny arms, but only with an unfiltered-builder scope split and a fresh baseline. Rebase onto current main, then run the exact-head clean baseline, each full-Gateway arm separately with assertion reds and source/build identity, restored green after each, complete summaries, reconciliation, and final CI.

PR #1936, opened 2026-07-20
docs: land the hosted Gateway collection census and its remedy map

thefrederiksen:
Puts the hosted Gateway collection census and its remedy map onto main.

## Why this needs to land before more work is cut against it

These two documents are the specification for the eleven remedy units covering the
forty active cross-tenant unsafe rows. They have been driving worker briefs all
night. They have also been sitting on an unmerged branch the whole time.

Three separate workers independently reported the same problem: they were told to
read the specification from `origin/main`, and it is not there. Each had to go and
find it on the branch. One noted plainly that anyone verifying against main will
not find it.

That is the failure this repository has already learned once: an uncommitted design
is a question, not a specification. A worker cannot verify its work against a
document that is not on the branch it is told to trust, and a reviewer checking a
unit against "the map" has no single version to check against. Merged is the only
version a builder can rely on.

## What is in here

- **The census** (`a294d770`) - the finite enumeration, closed at 94 rows: 46 clean,
  7 latent, 40 active cross-tenant unsafe, 1 policy.
- **The ingestion map** (`ff63f85f`) - the regrouping of those 40 rows into eleven
  identifier or no-identifier work units, with the remedy for each chosen by the
  shape of the key: prefix at ingestion for a caller-supplied identifier, owner-bind
  for a server-minted one, explicit tenant partitioning where there is no caller
  identifier at all.

The map's own coverage reconciliation shows the union is exact: every one of the 40
active unsafe row numbers appears in exactly one work unit, with no duplicate
assignment and no omission.

## What this does not claim

Documentation only. No code changes. Neither document asserts that any remedy is
implemented, and the map says so about itself - it is a fix plan, and no row is
declared fixed by it. Landing this makes the specification citable; it does not
advance any unit.

### DIFF
--- a/docs/architecture/hosted-gateway-process-global-collection-census-2026-07-20.md
+++ b/docs/architecture/hosted-gateway-process-global-collection-census-2026-07-20.md
@@ -1,0 +1,187 @@
+# Hosted Gateway process-global collection census
+
+**Audit snapshot:** `f5bb926f18c0023e86fc7b212ab02d425ff9cc50` (`origin/main` on 20 July 2026)  
+**Scope:** retained collection state declared by the Gateway, including singleton instance fields and static fields  
+**Status:** enumeration audit only; this document does not claim that the findings are fixed
+
+## Headline
+
+**Forty independently addressable rows are actively unsafe across hosted tenants.**
+
+This is a finite census, not a sample. It found 102 retained collection declarations in 54 declaring
+types. Applying the collapse rule below produces 94 independently addressable rows: 46 clean, seven
+latent, 40 actively cross-tenant unsafe, and one separately unsafe authorization-policy row.
+
+| Reconciliation | Count |
+| --- | ---: |
+| Raw retained collection declarations | 102 |
+| One-row reductions from the eight named collapses | -8 |
+| Independently addressable rows | 94 |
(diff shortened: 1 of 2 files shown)


thefrederiksen:
STANDING: REJECTED
ff63f85f0f55359727f128cbb991641dc4f6b62c
REASON-TO-TRUST CHECK: The claimed trust property is not merely that the arithmetic reconciles; it is that 94 rows are a finite, complete denominator and that the 40-row remedy union is therefore exact. I tested the census question set against independent tenant-isolation dimensions. It omits keyed-versus-listed operation semantics, retained composite objects outside the declaration population, and shared capacity/eviction ownership.

Consequently, the 94-row total is not a complete denominator of tenant-isolation surfaces, and the ingestion map's exact-union proof establishes exactness only relative to an incomplete parent population. Green CI and internally consistent row arithmetic cannot repair that semantic gap.

Before landing as a citable specification, either (a) explicitly downgrade 94/40 to a lower-bound, collection-declaration sample and remove finite/closed/exact-denominator language, or (b) rerun and reconcile the census across direct plus composite retained state, each access operation (including enumeration), and shared resource admission/capacity/eviction behavior, then regenerate the remedy map from that corrected population.

PR #1937, opened 2026-07-20
test: refuse to start a mutation-proof baseline or arm on a contaminated tree

thefrederiksen:
## What this stops

A mutation proof works by mutating one primitive, running the whole suite, and reconciling the arithmetic: passed plus failed under the mutation must equal passed under the restored run. If they add up, the suite is held to have observed the primitive.

There is exactly one failure that reconciliation cannot see, and it is the one that matters most.

**If the baseline is taken on a tree where a security guard block has already been silently deleted and left uncommitted, the tree still compiles, the baseline still looks green and complete, and the mutation arm then reconciles PERFECTLY against it.** The two runs agree with each other rather than measuring anything. The cheapest detector we own passes while measuring nothing, and every conclusion drawn from it is worthless with nothing anywhere in the process saying so.

That is not hypothetical. On 2026-07-19 a worker found exactly that on disk in its own worktree: a guard block that a **killed** run had deleted and never restored, because a killed process skips its cleanup. A sweep of the mission's working trees that night found four dirty trees.

## Why an instruction could not close it

Telling every worker "check your tree is clean before you take a baseline" does not work. The worker who **inherits** a contaminated tree is, by definition, the one who does not know a mutation is there - it compiles, and the tests pass.

This is the same shape as the concurrent-suite problem the per-user test-suite lock was built for. Four workers were each told not to run two suites at once, each complied with everything it could observe, and four suites ran concurrently anyway. A rule that requires global knowledge cannot be obeyed by an actor holding only local knowledge. That one was solved by a mechanism rather than a briefing, and so is this: the guard acquires inside the test process, fires regardless of who started the run or what they knew, and names exactly what is wrong.

## It is gated, and that is deliberate - do not widen it

**With no pin, the guard has no opinion at all.** A worker mid-rework is legitimately dirty and must be able to run whatever it likes; only a baseline and a mutation arm carry a meaning that depends on the tree being exactly the pinned head.

If you are reading this in six months wondering whether to make it a blanket refusal on any dirty tree: **do not.** A blanket guard would be switched off within a day, and a guard that gets switched off protects nothing.

## The pinned head never moves, and that is enforced twice

A reviewer found that the first version broke this in the most dangerous possible way: every `set` recomputed `git rev-parse HEAD`, so the documented baseline-to-arm transition **silently re-pinned** whenever HEAD had moved between the two runs. The guard then compared the arm against the new head, matched exactly, and admitted it - the supported happy path reaching the exact event the guard exists to refuse.

Worth naming plainly, because it is the same shape this guard was built to catch: **that the pin is fixed is the property the whole design rests on, and it was the property nothing exercised** - precisely because it is the premise. I proved the guard fires on a contaminated tree and never asked whether the pin itself could move.

Two mechanisms now hold it, because fixing only the script fixes only the instance:

1. `set -Phase arm` carries the baseline's head forward and never recomputes it, refusing if the tree has moved off it and naming both heads. An arm with no baseline refuses; a baseline over a live pin refuses. Re-pinning requires an explicit `release`, which is loud and discards the old proof's numbers with its identity.
2. The pin carries a **proof identity**, recorded to the ledger with the head each run measured. A run whose pin names a head different from an earlier run of the same proof is refused - however the pin came to say what it says: a hand edit, a copied file, a future script change.

## How a run declares itself, and whether it can be forgotten

The declaration is made **once per proof, not once per run**, by pinning the working tree (`scripts/mutation-proof-pin.ps1 set -Phase baseline|arm`). From then until the pin is released, **every** run in that tree is checked - the baseline, the arm, and every re-run of either. So the only moment anybody can forget is before the proof starts, and at that moment forgetting means having no pinned head written down anywhere, which is not a proof anybody can write up.

Being honest about the limit: it is not impossible to forget. Two things narrow it. The pin is the only place the pinned head is recorded. And the ledger below records unpinned runs anyway, so a forgotten pin leaves evidence rather than a hole.

The arm is checked against a **tighter** rule than the baseline, because an arm is supposed to be dirty. Its pin declares which paths the mutation touches, and the tree must carry exactly those - no more (a contaminated arm) and no less. The "no less" half matters as much: an arm with no mutation in it is a second baseline wearing an arm's name, and it reconciles perfectly against the first while proving nothing.

## The second job: the record outlives the tree

Somebody was asked whether four **already-merged** security proofs had been taken on contaminated trees. The answer could not be given - not because the evidence was hard to find, but because it no longer existed. Those proofs ran in worktrees; the worktrees were removed after merging. That removal is correct hygiene, and it destroyed the only artifact that could have answered the question. Those four are recorded as unknown and will stay unknown.

So every run appends a line to a ledger under the per-user local application data directory, **outside every working tree** - beyond the reach of `git worktree remove`, `git clean -xdf`, and deleting the checkout.

The ledger writes on **admission** as well as refusal, and an admission states which head was verified and that the tree matched it. A log written only when something goes wrong cannot afterwards distinguish "verified clean" from "the guard never ran" - both are silence, and silence reads as success.

## Proof that the detector FIRES, not merely that it exists

A guard that never fires looks exactly like a guard that works: both are silent, and both let every run through. So every refusal below is paired with a control that must be admitted.

**Unit level (33 tests, real temporary git repositories, run in CcDirector.Core.Tests):**

- The 2026-07-19 event replayed: a guard block deleted and left uncommitted is REFUSED and the refusal names the file. The same repository is checked **before** the contamination in the same test and is admitted, so nothing passes because the two arms were set up differently.
- The mutation is asserted to leave a brace-balanced file. A detector that could only see a syntactically broken tree would not have caught the real event - a broken tree announces itself with a build failure and needs no guard.
- A head that has moved off the pinned head is REFUSED, naming both heads.
- CONTROL: a clean tree at the pinned head is ADMITTED.
- CONTROL FOR SCOPE: a mid-rework run with no pin is ADMITTED with three kinds of change in the tree.
- Plus: contaminated arm refused; arm with the mutation missing refused; a pin that cannot be parsed REFUSES rather than degrading into "no pin"; a pinned run that cannot read its tree refuses while an unpinned one still runs.

**End to end, against the real mechanism, in this worktree:**

| Arm | Tree | Pin | Result |
|---|---|---|---|
| Control | clean at the pinned head | baseline | ADMITTED, 33 passed; ledger records `headVerified=yes` and `VERIFIED ... MATCHED it` |
| **Firing** | a real cross-tenant partition filter deleted from `DirectorRegistry.ListDirectors`, uncommitted, **still compiles** | baseline | **REFUSED, exit code 1, "NO TESTS WILL RUN"**; ledger names the file |
| Scope control | the same contaminated tree | released | ADMITTED, 33 passed |

**The pinned head, end to end against the real script and the real guard:**

| Arm | Result |
|---|---|
| Pin baseline at A, commit B, `set -Phase arm` | **REFUSED**, naming A and B; the pin file still reads A afterwards |
| Run the suite at B with that pin | **REFUSED** by the guard |
| Hand-edit the pin's head to B (what a future script regression would do), tree clean and matching it exactly | **REFUSED** by the ledger's memory of this proof at A - the ledger line reads `headVerified=yes  verdict=refused`, so every other check passed and it was still stopped |
| Negative control: delete the `DetectMovedPin` call | exactly the moved-pin tests redden; their controls - same proof re-run at its own head, and an unrelated earlier proof at another head - stay green |
| Negative control: reintroduce the script's re-pin | exactly the transition test reddens; its control (arm accepted at the baseline head) stays green |

Those sequences are **tests**, not a walkthrough someone drove once: they run the real pin file, the real working tree, the real ledger and the real script on every push. Hand-run evidence is prose in a pull request and cannot notice a regression. The missing-PowerShell case fails rather than skipping, because a skipped test is invisible in a summary line and looks exactly like a test that ran.

The firing arm used a genuine security deletion, not a token edit: with that `.Where` gone, every caller receives the fleet-global director list. It was built and confirmed to compile first, because a mutation that fails the build is caught by the compiler and proves nothing about this guard.

This exercise also found a real defect in the pin script - PowerShell's comma binds tighter than `+`, so the timestamp landed on a line of its own and the guard refused the whole proof as a malformed pin. Correct behaviour, wrong cause; fixed in the second commit. It would have stopped the first proof anybody tried to run.

## One location, asked rather than derived

The pin's location used to be derived twice - once in PowerShell to write it, once in C# to read it - and a reviewer found the two disagreed off Windows: the tool printed `PINNED` while the guard read nothing and admitted contaminated baselines. Continuous integration runs on Linux, so the platform where it was dead is the one nobody watches.

Aligning the two would have fixed the instance and kept the class. There is no second derivation left: both sides ask git `rev-parse --absolute-git-dir`. That removes the hash, the per-user directory, the platform branch and the golden filename test, and brings four properties with it - per-worktree by construction, invisible to `git status` so the pin cannot dirty the tree it guards, beyond `git clean -xdf`, and not movable by the environment. The ledger had the same branch and now uses one rule on every platform, since it must survive `git worktree remove`.

New checks run on whatever platform executes them, so the Linux answer is checked by the Linux run: a pin written where the script writes it must be found and parsed by the guard here; two real worktrees get different pins; the pin does not dirty the tree; and the script's own text is read and compared against the guard's constant.

## Wiring

`Directory.Build.props` links the guard and its tests into every project whose name ends in `.Tests`, so a test project added next year is covered without anybody remembering. It keys on the project name rather than the `IsTestProject` flag deliberately: one of the seven existing test projects does not set that flag, and a guard that silently covers six of seven assemblies looks exactly like a guard that covers all of them. All seven build.

Documentation: `docs/testing/mutation-proof-pin.md`.

## It failed open, in three places, and one of them was documented as a design choice

A reviewer found that the guard **admitted a run whenever it failed to look**. Resolving the pin's location
launched git; any failure - git missing from the path, a timeout, a broken install - became "no pin found",
which admitted. A contaminated baseline would run to completion on such a machine while the pin file sat in
the git directory the whole time.

Asked whether that was one site or a class, I went and looked. It was three. Two were mine and unreported:
a failed working-tree-root lookup that admitted "loudly", and a null root that returned before the decision
ran at all. The same defect each time - **"I could not look" spelled as "there is nothing there."**

The detail worth keeping is the first one. My own comment at that site read:

> *"There is no pin to find and no way to know whether one exists, so nothing is claimed - the run is
> admitted."*

That is the defect written down, in a rationale, and not recognised as a defect. It is worse than an
undocumented bug: a comment confers authority, so the next reader finds a reason and stops looking. If you
are reading this file and a comment explains why something permissive is fine, that is exactly where to
look hardest.

**The two failure directions of a guard are not symmetric**, and this is the sentence to keep if only one
survives: refusing wrongly is loud, visible and self-correcting - somebody complains within the hour.
Admitting wrongly is silent and self-concealing, and it is worse precisely because the guard's existence
stops everyone else checking. This guard has now been found to admit wrongly twice, and both times the
mechanism was silence.

**Every ambiguous outcome in this guard now resolves to REFUSE, and that is structural rather than a
claim:**

- `PinOutcome` has three states, not two. `NoPinPresent` is only constructible from a *completed* search
  and is the only state that admits; a lookup failure is `CouldNotDetermine`, which refuses and cannot be
  spelled as an absence.
- `Decide` has exactly one admit, reachable from that one state, and an unrecognised state refuses - so
  adding a state cannot silently reuse the admit path, which is how this arose.
- Every path through the check ends at one exit that records and enforces. An early return was an admission
  that never reached the ledger either: invisible twice over.
- The lookup no longer launches git. It reads the `.git` marker git has already written and requires the
  resolved directory to contain `HEAD`, so resolving somewhere wrong refuses instead of finding no pin
  there and admitting. The safety-critical question is no longer answered by the least reliable mechanism
  in the file.

Controls, each paired so the fix cannot be satisfied by refusing everything: a real pin with git absent is
refused, while an unpinned run with git absent is still admitted; an unreadable marker and a marker naming
a non-git directory refuse, while outside a repository the answer is a settled absence and admits; and
`OnlyASettledAbsenceAdmits` enumerates every state except `NoPinPresent` and requires each to refuse.

Proven to fire: reproducing the reviewer's defect faithfully reddens exactly the git-absent test while its
control stays green. A first, weaker mutation reddened only the marker test - which told me the git-absent
test was not yet exercising the reported path. That red was available to write up and would have looked
like the guard firing; a red that reddens the wrong test is a proof you do not have.

## Two kinds of mechanism in here, and one is easy to delete by mistake

Some of this **prevents** a failure: the refusal on an undeclared change, the refusal on a moved pin. Those either work or they do not, and a test shows which.

The rest **cannot prevent anything**, and is not meant to. The ledger does not stop a bad proof; it records what each run verified so the question can be answered later. Requiring the pinned commit to exist on a remote does not stop a stashed repair; it makes the commit fetchable so a reader can see what was in it.

The honest sentence is that they **convert a private failure into a public one**. A contaminated proof nobody can detect afterwards lives on one machine and dies with it; the same failure recorded in a ledger, or attributed to a commit anyone can fetch, is still a failure but one a second person can find without being told where to look. That matters here more than it would elsewhere, because this whole proof structure rests on somebody else being able to re-derive a claim - and the four already-merged proofs that cannot now be shown to have been taken on clean trees were not a preventer failing, but nothing having made the failure visible beyond the machine it happened on.

If you are deciding whether to remove the ledger or the publication requirement because they do not stop anything: they are not supposed to. The question to answer first is whether the failure each one makes visible has any other way of being seen.

## What this guard promises, and the sentence a reader will hear instead

This is the limitation I most want on the record, because the two sentences are close enough to be
mistaken for each other.

**What it verifies:** this working tree is exactly the pinned head.
**What a reader will hear:** this arm measured the code under review.

The gap between them is a **stash**. A worker who stashes an uncommitted repair presents a perfectly clean
tree at exactly the pinned head — this guard admits, the arm runs on unrepaired source, and every count
reconciles beautifully. Another unit on this mission nearly lost its tree to a stash the same night, which
is how the case surfaced; it had independently built the same precondition and flagged the overlap rather
than shipping a duplicate, so it lands here once instead of drifting in two places.

Nothing here can close that gap, because nothing here can know what a head was *supposed* to contain. What
it now does is make the claim checkable by somebody else: **a proof must be pinned to a commit that exists
on a remote.** The head every number is attributed to can then be fetched and read, so a missing repair is
visible in the record rather than invisible on one machine — and it catches the common shape directly,
since a stash usually leaves the tree on a local commit that was never pushed. A repository with no remote
has no publication question and is not held to one.

That also completes a family worth naming: a **copy** restores a snapshot, a **checkout** restores the last
commit, a **stash** restores an index state, and none of them restores the state before one mutation.
Committing first closes all three — and it is what makes an admission here mean anything.

The whole source tree is checked, not just the project under test, for the reason that unit gave: a lost
test edit falsifies an arm as thoroughly as a lost production edit.

## The limit I am not claiming past

Every property of this guard is checked by tests on every push **on Windows only**. `.github/workflows/ci.yml` runs `dotnet test` exactly once, on `windows-latest`; the `ubuntu-latest` job is the JavaScript one and never touches .NET. So nothing executes this guard on Linux or macOS - the pin-path parity test would catch a recurrence of the platform divergence *if it were ever run there*, and today nothing runs it there.

That also corrects how finding 2 was framed: the divergence was real and the fix stands, but the exposure was to a person or agent working on Linux or macOS, not to CI, which never ran the guard there at all.

Closing it properly means a job on `ubuntu-latest` running these tests. I have not folded that into this change because I cannot verify from here that `CcDirector.Core.Tests` and its reference chain build on Linux - the product is a Windows application, and that chain is exactly where a Windows-only dependency would sit. That is a shared-CI change and deserves its own decision rather than being smuggled in here.

And a worker can still forget to pin at all. Unchanged, still recorded rather than refused.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -40,6 +40,26 @@
   <ItemGroup Condition="$(MSBuildProjectName.EndsWith('.Tests'))">
     <Compile Include="$(MSBuildThisFileDirectory)src\TestInfrastructure\MutationProofPinGuard.cs"
              Link="TestInfrastructure\MutationProofPinGuard.cs" />
+    <!--
+      The per-assembly sentinel, and ONLY the sentinel. Every assembly must prove its own copy of the
+      module initializer woke up, because one that silently does not run leaves no trace and every other
+      test in the assembly still passes.
+
+      The guard's BEHAVIOURAL suite is deliberately NOT here - it builds real git repositories and launches
+      real processes, and running that seven times buys no coverage while changing the load and the test
+      scheduling in assemblies that have nothing to do with this mechanism. See
+      MutationProofPinGuardArmedTests for the incident that prompted the split.
+    -->
+    <Compile Include="$(MSBuildThisFileDirectory)src\TestInfrastructure\MutationProofPinGuardArmedTests.cs"
+             Link="TestInfrastructure\MutationProofPinGuardArmedTests.cs" />
+  </ItemGroup>
+
+  <!--
+    The guard's behavioural suite, compiled into exactly one assembly. Core.Tests because it is the
+    fastest of the seven to build and run, has no user-interface threading requirements, and is where
+    this unit's proofs have been driven throughout.
+  -->
+  <ItemGroup Condition="'$(MSBuildProjectName)' == 'CcDirector.Core.Tests'">
     <Compile Include="$(MSBuildThisFileDirectory)src\TestInfrastructure\MutationProofPinGuardTests.cs"
              Link="TestInfrastructure\MutationProofPinGuardTests.cs" />
   </ItemGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -18,4 +18,29 @@
   <PropertyGroup>
     <Version>1.8.4</Version>
   </PropertyGroup>
+
+  <!--
+    THE MUTATION-PROOF PIN GUARD, linked into every test assembly in the repository.
+
+    It refuses to START a run that has been declared part of a mutation proof - a baseline or a mutation
+    arm - when the working tree is not exactly what that proof pinned, and it records what it verified to a
+    ledger that outlives the working tree. Every other run is admitted untouched. The full reasoning, and
+    the incident that produced it, are at the top of MutationProofPinGuard.cs; read that before changing
+    anything here.
+
+    WHY IT IS LINKED FROM HERE RATHER THAN LISTED IN EACH TEST PROJECT: so a test project added next year
+    is covered without anybody remembering. The guard exists because a rule that has to be remembered is
+    not a mechanism, and wiring that has to be remembered is the same mistake one level down.
+
+    WHY THE CONDITION IS ON THE PROJECT NAME AND NOT ON IsTestProject: one of the seven existing test
+    projects (CcDirector.Terminal.Avalonia.Tests) does not set IsTestProject, so keying on that flag would
+    have silently skipped it - and a guard that silently covers six of seven assemblies looks exactly like
+    a guard that covers all of them.
+  -->
+  <ItemGroup Condition="$(MSBuildProjectName.EndsWith('.Tests'))">
+    <Compile Include="$(MSBuildThisFileDirectory)src\TestInfrastructure\MutationProofPinGuard.cs"
+             Link="TestInfrastructure\MutationProofPinGuard.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)src\TestInfrastructure\MutationProofPinGuardTests.cs"
+             Link="TestInfrastructure\MutationProofPinGuardTests.cs" />
+  </ItemGroup>
 </Project>
--- a/docs/testing/mutation-proof-pin.md
+++ b/docs/testing/mutation-proof-pin.md
@@ -41,7 +41,8 @@ working tree is checked - the baseline, the arm, and any re-run of either.
 
 #    Run the suite. Record the numbers against the pinned head it printed.
 
-# 2. Apply the mutation, then re-declare as the arm, naming what it touches.
+# 2. Apply the mutation, then declare the arm, naming what it touches. This does NOT re-pin: it carries
+#    the baseline's head forward, and refuses if the tree has moved off it.
 ./scripts/mutation-proof-pin.ps1 set -Phase arm -Mutates src/CcDirector.Gateway/Api/GatewayEndpoints.cs
 
 #    Run the suite again. Reconcile.
@@ -65,10 +66,38 @@ declared changes:
 
 A baseline is simply an arm that declares no mutations, so one rule runs both phases.
 
-**With no pin, the guard has no opinion at all.** A worker mid-rework is legitimately dirty and must be able
-to run whatever it likes. This is deliberate and is not an oversight to be tidied up later: a blanket
-refusal on any dirty tree would be switched off within a day, and a guard that gets switched off protects
-nothing.
+**With no pin, no run is ever refused.** A worker mid-rework is legitimately dirty and must be able to run
+whatever it likes. This is deliberate and is not an oversight to be tidied up later: a blanket refusal on
+any dirty tree would be switched off within a day, and a guard that gets switched off protects nothing.
+
+### The pinned head never moves, and that is enforced twice
+
+A proof's pinned head is its identity. A baseline and its arm must be taken at the same commit, or the
+reconciliation compares two different programs and the arithmetic means nothing while still adding up.
+
+The first version of this tool got that wrong in the most dangerous possible way: every `set` recomputed
+`git rev-parse HEAD`, so the documented baseline-to-arm transition silently **re-pinned** to the new head
+whenever HEAD had moved between the two runs. The guard then compared the arm against the new head, found
+an exact match, and admitted it. The supported happy path walked into the exact event the guard exists to
+refuse.
+
+Two mechanisms now hold it:
+
+1. `set -Phase arm` **carries the baseline's head forward** and never recomputes it. If the tree has moved
+   off the pinned head it refuses, naming both. Setting an arm with no baseline refuses too, and pinning a
+   baseline over an existing pin refuses - re-pinning requires an explicit `release`, which is loud and
+   discards the old proof identity along with its numbers.
+2. The pin carries a **proof identity**, and every run records it to the ledger with the head it measured.
+   If any earlier run of the same proof was measured against a different head, the run is refused - however
+   the pin file came to say what it says. That covers a hand-edited pin, a copied file, and a future change
+   to this script. Fixing the script fixes the instance; this covers the class.
+
+### What an unpinned run costs
+
+It is not free, and an earlier version of this document wrongly said it was. Every run of every test
+assembly invokes git twice and appends one line to a log outside the tree - roughly a fifth of a second per
+assembly against a suite measured in minutes. That is the deliberate price of the ledger below: it is the
+entire reason a **forgotten** pin is still answerable afterwards.
 
 ## The second job: the record outlives the tree
 
@@ -80,7 +109,15 @@ Those proofs were run in worktrees, and the worktrees were removed after merging
 hygiene, and it destroyed the only artifact that could have answered the question. Those four are recorded
 as unknown and will stay unknown.
 
-So every run appends a line to a ledger that lives **outside every working tree**, under the per-user local
+The **pin** lives in the repository's git directory, found by asking `git rev-parse --absolute-git-dir` -
+the same question the guard asks, so there is no second derivation to disagree with. (There used to be one,
+computed from the working tree path in both PowerShell and C#, and the two copies did not agree off
+Windows: the tool printed `PINNED` while the guard read nothing at all. Continuous integration runs on
+Linux.) That location is per-worktree by construction, is invisible to `git status` so the pin cannot dirty
+the tree it guards, and survives `git clean -xdf`.
+
+The **ledger** has the opposite requirement - it must outlive the tree - so every run appends a line to a
+file **outside every working tree**, under the per-user local
 application data directory alongside the suite lock's own state. `git worktree remove`, `git clean -xdf`,
 and deleting the whole checkout cannot reach it.
 
--- a/docs/testing/mutation-proof-pin.md
+++ b/docs/testing/mutation-proof-pin.md
@@ -99,6 +99,25 @@ assembly invokes git twice and appends one line to a log outside the tree - roug
 assembly against a suite measured in minutes. That is the deliberate price of the ledger below: it is the
 entire reason a **forgotten** pin is still answerable afterwards.
 
+## Two kinds of mechanism here, and one is easy to delete by mistake
+
+Some of this **prevents** a failure: the refusal on an undeclared change, the refusal on a moved pin. Those
+either work or they do not.
+
+The rest **cannot prevent anything**, and is not meant to. The ledger does not stop a bad proof; it records
+what each run verified so the question can be answered later. Requiring the pinned commit to exist on a
+remote does not stop a stashed repair; it makes the commit fetchable so a reader can see what was in it.
+
+The honest sentence is that they **convert a private failure into a public one**. A contaminated proof
+nobody can detect afterwards lives on one machine and dies with it. The same failure recorded in a ledger,
+or attributed to a commit anyone can fetch, is still a failure — but one a second person can find without
+being told where to look. That matters here more than it would elsewhere, because this whole proof
+structure rests on somebody else being able to re-derive a claim.
+
+So do not remove the ledger or the publication requirement on the grounds that they "do not stop
+anything". They are not supposed to. The question to answer first is whether the failure each one makes
+visible has any other way of being seen.
+
 ## The second job: the record outlives the tree
 
 Refusing a bad run is only half of it.
--- a/docs/testing/mutation-proof-pin.md
+++ b/docs/testing/mutation-proof-pin.md
@@ -1,0 +1,113 @@
+# The mutation-proof pin: how to take a baseline and a mutation arm that mean something
+
+## What a mutation proof is, and the one way it lies
+
+We prove a security guard is real by mutating the one primitive it depends on, running the whole suite
+again, and reconciling the arithmetic: the tests that pass plus the tests that fail under the mutation must
+add up to the tests that passed under the restored run. If they add up, the suite is held to have observed
+the primitive.
+
+There is exactly one failure that reconciliation cannot see, and it is the one that matters most.
+
+Suppose the baseline is taken on a working tree where a security guard block has **already been silently
+deleted and left uncommitted**. That is not a hypothetical - it is what a killed run leaves behind, because
+a process that is killed skips its cleanup. The tree still compiles. The suite still runs. The baseline
+still looks green and complete. Then the mutation arm runs on that same contaminated tree, and the two runs
+reconcile **perfectly** - because they agree with each other, not because either one measured anything.
+
+The cheapest detector we own passes while measuring nothing, and nothing anywhere in the process says so.
+
+On 2026-07-19 a worker found exactly that on disk in its own worktree. A sweep of the mission's working
+trees that night found four dirty trees.
+
+## Why telling people to check their tree does not work
+
+The worker who **inherits** a contaminated tree is, by definition, the one who does not know there is a
+mutation in it. It looks clean to them: it compiles, and the tests pass.
+
+This is the same shape as the concurrent-suite problem the per-user test-suite lock was built for. Four
+workers were each told not to run two suites at once; each complied with everything it could observe; four
+suites ran concurrently anyway. A rule that requires global knowledge cannot be obeyed by an actor holding
+only local knowledge. That one was solved by a mechanism rather than a briefing, and so is this.
+
+## How to run a proof
+
+Set the pin once, before the baseline. From that moment until you release it, **every** test run in that
+working tree is checked - the baseline, the arm, and any re-run of either.
+
+```
+# 1. Before the baseline. Refuses if the tree is already dirty.
+./scripts/mutation-proof-pin.ps1 set -Phase baseline -Note "tenant isolation guard, issue 1901"
+
+#    Run the suite. Record the numbers against the pinned head it printed.
+
+# 2. Apply the mutation, then re-declare as the arm, naming what it touches.
+./scripts/mutation-proof-pin.ps1 set -Phase arm -Mutates src/CcDirector.Gateway/Api/GatewayEndpoints.cs
+
+#    Run the suite again. Reconcile.
+
+# 3. Restore the mutation and release the pin.
+./scripts/mutation-proof-pin.ps1 release
+```
+
+`status` shows the current pin and the tree beside it. `ledger` prints every proof run ever recorded on
+this machine.
+
+## What the guard refuses, and what it never touches
+
+With a pin active it requires the head to be the pinned head, and the working tree to carry **exactly** the
+declared changes:
+
+- **more** than was declared is a contaminated run, and is refused naming each undeclared file;
+- **less** than was declared is refused too, and that half matters as much: an arm with no mutation in it
+  is a second baseline wearing an arm's name, and it reconciles perfectly against the first while proving
+  nothing.
+
+A baseline is simply an arm that declares no mutations, so one rule runs both phases.
+
+**With no pin, the guard has no opinion at all.** A worker mid-rework is legitimately dirty and must be able
+to run whatever it likes. This is deliberate and is not an oversight to be tidied up later: a blanket
+refusal on any dirty tree would be switched off within a day, and a guard that gets switched off protects
+nothing.
+
+## The second job: the record outlives the tree
+
+Refusing a bad run is only half of it.
+
+Somebody was asked whether four **already-merged** security proofs had been taken on contaminated trees. The
+answer could not be given - not because the evidence was hard to find, but because it no longer existed.
+Those proofs were run in worktrees, and the worktrees were removed after merging. That removal is correct
+hygiene, and it destroyed the only artifact that could have answered the question. Those four are recorded
+as unknown and will stay unknown.
+
+So every run appends a line to a ledger that lives **outside every working tree**, under the per-user local
+application data directory alongside the suite lock's own state. `git worktree remove`, `git clean -xdf`,
+and deleting the whole checkout cannot reach it.
+
+The ledger writes on **admission** as well as refusal, and an admission line states which head was verified
+and that the tree matched it. That is the point: a log written only when something goes wrong cannot
+afterwards distinguish "verified clean" from "the guard never ran". Both are silence, and silence reads as
+success.
+
+Unpinned runs are recorded too - never refused. That is what makes a **forgotten** pin recoverable: when a
+proof's numbers are questioned later, the record still says whether the tree was dirty at the moment the
+baseline ran.
+
+## Where it lives
+
+- `src/TestInfrastructure/MutationProofPinGuard.cs` - the mechanism. Read the comment at the top before
+  changing it.
+- `src/TestInfrastructure/MutationProofPinGuardTests.cs` - the proof that it fires, with both controls.
+- `Directory.Build.props` - links both into every project whose name ends in `.Tests`, so a test project
+  added next year is covered without anybody remembering. It keys on the project name rather than the
+  `IsTestProject` flag because one of the seven existing test projects does not set that flag.
+- `scripts/mutation-proof-pin.ps1` - set, status, release, ledger.
+
+## The one thing that is not automatic
+
+A worker can still forget to set the pin at all, and then nothing is checked. That limit is the price of
+the scope gate, and it is stated here rather than papered over.
+
+Two things narrow it. The pin is the only place the pinned head is written down, so a proof that skipped it
+has no pinned head to cite and is not a proof anybody can write up. And the ledger records unpinned runs
+anyway, so a forgotten pin leaves evidence instead of a hole.
--- a/scripts/mutation-proof-pin.ps1
+++ b/scripts/mutation-proof-pin.ps1
@@ -155,6 +155,7 @@ switch ($Verb) {
                 throw "A baseline declares no mutations. A run that carries a mutation is an arm."
             }
             if ($null -ne $existing) {
+                # CONTROL (preventing). Test: TheScriptRefusesToPinABaselineOverAnActivePin.
                 Write-Host "CANNOT PIN A BASELINE: this working tree already has an active pin." -ForegroundColor Red
                 Write-Host ("    proof:       " + $existing['proofId'])
                 Write-Host ("    phase:       " + $existing['phase'])
@@ -169,6 +170,7 @@ switch ($Verb) {
             if ($changes.Count -gt 0) {
                 # Refused HERE as well as at run time, because pinning a contaminated tree would record a
                 # false starting point and the run-time refusal would then look like a mystery.
+                # CONTROL (preventing). Test: TheScriptRefusesToPinABaselineOnADirtyTree.
                 Write-Host "CANNOT PIN A BASELINE: this working tree already carries uncommitted changes." -ForegroundColor Red
                 $changes | ForEach-Object { Write-Host "    $_" -ForegroundColor Red }
                 Write-Host ""
@@ -189,6 +191,7 @@ switch ($Verb) {
                 # An arm with no baseline before it has nothing to be compared against. Minting a fresh pin
                 # here would create a proof whose "baseline" was never taken - and it would look identical
                 # to a correct one.
+                # CONTROL (preventing). Test: TheScriptRefusesAnArmWithNoBaselineBeforeIt.
                 Write-Host "CANNOT SET AN ARM: this working tree has no active baseline pin." -ForegroundColor Red
                 Write-Host ""
                 Write-Host "A mutation arm only means something next to a baseline taken at the same commit. Pin the"
@@ -198,6 +201,7 @@ switch ($Verb) {
             }
 
             if ($existing['pinnedHead'] -ne $currentHead) {
+                # CONTROL (preventing). Test: TheScriptRefusesAnArmTransitionAfterTheHeadMoves_AndLeavesThePinOnTheBaselineHead.
                 Write-Host "CANNOT SET AN ARM: the head has moved since this proof's baseline was pinned." -ForegroundColor Red
                 Write-Host ("    proof:        " + $existing['proofId'])
                 Write-Host ("    pinned head:  " + $existing['pinnedHead'] + "   (what the baseline measured)")
--- a/scripts/mutation-proof-pin.ps1
+++ b/scripts/mutation-proof-pin.ps1
@@ -1,0 +1,211 @@
+<#
+.SYNOPSIS
+    Declares this working tree to be part of a mutation proof, so every test run in it is checked against
+    the head the proof pinned.
+
+.DESCRIPTION
+    We prove a security guard is real by mutating the one primitive it depends on, running the whole suite
+    again, and reconciling the arithmetic: passed plus failed under the mutation must equal passed under the
+    restored run.
+
+    That reconciliation has one failure it cannot see. If the BASELINE is taken on a tree where a guard block
+    has already been silently deleted and left uncommitted - which is what a KILLED run leaves behind, because
+    a killed process skips its cleanup - the tree still compiles, the baseline still looks green and complete,
+    and the two runs reconcile PERFECTLY. They agree with each other rather than measuring anything. The
+    cheapest detector we own passes while measuring nothing.
+
+    This script writes the declaration that arms the guard. From the moment a pin is set until it is released,
+    EVERY test run in this working tree is checked: the head must be the pinned head, and the working tree must
+    carry exactly the declared changes - no more (a contaminated tree) and no less (an "arm" with no mutation
+    in it, which is a second baseline in disguise and reconciles perfectly against the first).
+
+    A run with no pin is never touched. A worker mid-rework is legitimately dirty and must be able to run
+    whatever it likes; only a baseline and an arm carry a meaning that depends on the tree being exact.
+
+    The pin, and the ledger of what each proof run verified, live under the per-user local application data
+    directory - OUTSIDE every working tree. That is deliberate: removing a worktree after merging is correct
+    hygiene, and it is also what destroyed the only evidence for four already-merged proofs, which can now
+    never be shown to have been taken on clean trees.
+
+.EXAMPLE
+    # Before the baseline run:
+    ./scripts/mutation-proof-pin.ps1 set -Phase baseline -Note "tenant isolation guard, issue 1901"
+
+.EXAMPLE
+    # Before the mutation arm, after applying the mutation:
+    ./scripts/mutation-proof-pin.ps1 set -Phase arm -Mutates src/CcDirector.Gateway/Api/GatewayEndpoints.cs
+
+.EXAMPLE
+    # When the proof is finished:
+    ./scripts/mutation-proof-pin.ps1 release
+#>
+
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true, Position = 0)]
+    [ValidateSet('set', 'status', 'release', 'ledger')]
+    [string] $Verb,
+
+    [ValidateSet('baseline', 'arm')]
+    [string] $Phase,
+
+    [string[]] $Mutates = @(),
+
+    [string] $Note = ''
+)
+
+$ErrorActionPreference = 'Stop'
+
+function Get-WorkingTreeRoot {
+    $root = (git rev-parse --show-toplevel 2>$null)
+    if ($LASTEXITCODE -ne 0 -or [string]::IsNullOrWhiteSpace($root)) {
+        throw "Not inside a git working tree. Run this from the tree the proof will be run in."
+    }
+    return ($root -replace '\\', '/').TrimEnd('/')
+}
+
+function Get-PinsDirectory {
+    $local = [Environment]::GetFolderPath('LocalApplicationData')
+    if ([string]::IsNullOrWhiteSpace($local)) {
+        throw "Cannot locate the per-user local application data directory, so the pin has no environment-independent home."
+    }
+    return (Join-Path (Join-Path $local 'cc-director') 'mutation-proof-pins')
+}
+
+function Get-PinPath([string] $root) {
+    # Must match MutationProofPinGuard.ComputePinFilePath EXACTLY, or the script writes a pin the guard
+    # never reads and the whole mechanism is inert while appearing to work. The C# side of this derivation
+    # is frozen by a golden value in MutationProofPinGuardTests
+    # (ThePinFileNameIsAGoldenValue_BecauseAScriptDerivesTheSameNameSeparately) - if you change either
+    # side, run that test and reconcile both, because a silent divergence here disarms the guard.
+    #
+    # SHA256::Create rather than SHA256::HashData: HashData does not exist on Windows PowerShell 5.1, which
+    # is the shell this repository's scripts are driven from.
+    $normalized = $root.ToLowerInvariant()
+    $hasher = [System.Security.Cryptography.SHA256]::Create()
+    try {
+        $sha = $hasher.ComputeHash([System.Text.Encoding]::UTF8.GetBytes($normalized))
+    }
+    finally {
+        $hasher.Dispose()
+    }
+    $digest = ([System.BitConverter]::ToString($sha) -replace '-', '').ToLowerInvariant()
+
+    $leafSource = Split-Path -Leaf $normalized
+    $leaf = -join ($leafSource.ToCharArray() | ForEach-Object {
+        if ([char]::IsLetterOrDigit($_) -or $_ -eq '-' -or $_ -eq '_') { $_ } else { '-' }
+    })
+    if ($leaf.Length -gt 40) { $leaf = $leaf.Substring(0, 40) }
+
+    return (Join-Path (Get-PinsDirectory) ("$leaf-" + $digest.Substring(0, 16) + '.pin'))
+}
+
+function Get-WorkingTreeChanges([string] $root) {
+    $status = (git -C $root --no-optional-locks status --porcelain=v1 --untracked-files=normal)
+    if ([string]::IsNullOrWhiteSpace($status)) { return @() }
+    return @($status -split "`n" | Where-Object { $_.Trim().Length -gt 0 })
+}
+
+$root = Get-WorkingTreeRoot
+$pinPath = Get-PinPath $root
+
+switch ($Verb) {
+
+    'set' {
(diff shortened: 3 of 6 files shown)


thefrederiksen:
## Review of `dff2688e563162048efbcee8dfe0cabe732f8016`

**Standing: CHANGES REQUESTED.** The core detector does fire on a syntactically valid dirty source tree, the clean-at-pin and unpinned-dirty controls are real, malformed pins fail closed, and wiring by `.Tests` project name covers all seven current test assemblies. I found two bypasses in the supported pin workflow that are load-bearing for an instrument intended to validate every later proof.

1. **The documented baseline-to-arm transition silently re-pins a moved HEAD.** Every `set`, including `set -Phase arm`, recomputes `git rev-parse HEAD` and overwrites `pinnedHead` ([script lines 119-120](https://github.com/thefrederiksen/devthrottle/blob/dff2688e563162048efbcee8dfe0cabe732f8016/scripts/mutation-proof-pin.ps1#L119-L120), [148-163](https://github.com/thefrederiksen/devthrottle/blob/dff2688e563162048efbcee8dfe0cabe732f8016/scripts/mutation-proof-pin.ps1#L148-L163)). The documented workflow sets a baseline pin, runs it, then calls `set -Phase arm`. If HEAD moved from A to B between those steps, that second command replaces A with B. The guard then compares the arm against B, finds an exact match, and admits it; the baseline and arm measured different programs even though the guard is specifically supposed to refuse that condition. The unit test for a moved head holds one in-memory pin fixed, so it does not exercise the script transition that loses the pin. An arm transition must preserve the baseline's pinned head and refuse if current HEAD differs; setting an arm without an active baseline pin should also fail rather than create a fresh proof midpoint. This needs an end-to-end script transition test: baseline at A, commit B, attempt to set/run arm, refusal naming A and B.

2. **The PowerShell writer and C# reader use different pin homes on non-Windows systems.** The script always uses `Environment.GetFolderPath('LocalApplicationData')/cc-director/mutation-proof-pins` ([lines 67-72](https://github.com/thefrederiksen/devthrottle/blob/dff2688e563162048efbcee8dfe0cabe732f8016/scripts/mutation-proof-pin.ps1#L67-L72)). The guard uses that location only on Windows; on Linux/macOS it reads `/tmp/cc-director-{user}-mutation-proof-pins` ([guard lines 485-502](https://github.com/thefrederiksen/devthrottle/blob/dff2688e563162048efbcee8dfe0cabe732f8016/src/TestInfrastructure/MutationProofPinGuard.cs#L485-L502)). On those platforms the supported script can report `PINNED` while the module initializer looks elsewhere, sees no pin, and admits a contaminated baseline or arm. The filename golden test does not cover directory parity. The writer and reader need one cross-platform location contract, proved end to end on both Windows and Unix derivations.

A secondary documentation/operational correction is also needed: an unpinned run is admitted, but the mechanism is not “inert” and does not avoid invoking git. `Check` calls `ReadTree` and writes the per-tree run log for every assembly even when no pin exists ([lines 655-664](https://github.com/thefrederiksen/devthrottle/blob/dff2688e563162048efbcee8dfe0cabe732f8016/src/TestInfrastructure/MutationProofPinGuard.cs#L655-L664)). That is how forgotten pins leave best-effort evidence, but it means the change adds two git subprocesses plus an external log write to every ordinary run in every test assembly—not only a changed test count. Keep that tradeoff if intended, but state it consistently and avoid the contradictory “does not even invoke git unless active” claim.

The current green CI establishes that the added wiring builds and its tests pass; it cannot establish either workflow bypass above because neither transition is exercised.

thefrederiksen:
Both findings confirmed and fixed at `52ec1e93`. Both were real bypasses, and I could not have found either from inside — the second in particular was invisible to me because I only ever ran on Windows.

## 1. The re-pinned head

Confirmed exactly as described, and reproduced end to end before fixing it.

The deeper problem is the one worth naming: **that the pin is fixed is the property the whole design rests on, and it was the property nothing exercised** — precisely because it is the premise. Everything else here (refusing an undeclared change, refusing a missing mutation) is worthless if the pin can shift underneath the proof. I proved the guard fires on a contaminated tree and never asked whether the pin itself could move.

Fixed with two mechanisms, because fixing only the script fixes only the instance:

**The script.** `set -Phase arm` now carries the baseline's head forward and never recomputes it. If the tree has moved off the pinned head it refuses, naming both. Setting an arm with no baseline refuses (an arm with no baseline before it is a proof whose baseline was never taken, and it looks identical to a correct one). Pinning a baseline over a live pin refuses — re-pinning requires an explicit `release`, which is loud and discards the old identity along with its numbers.

The end-to-end transition test you asked for, run against the real script:

```
set -Phase baseline            -> pinned head eeeaa246
commit                         -> head is now 7177221e
set -Phase arm -Mutates ...    -> CANNOT SET AN ARM: the head has moved since this proof's baseline was pinned.
                                      pinned head:  eeeaa246...   (what the baseline measured)
                                      current head: 7177221e...   (what an arm would measure now)
```

The pin file after that refusal still reads `pinnedHead=eeeaa246` — it was not updated — and a suite run at `7177221e` is then refused by the guard.

**The guard.** The pin now carries a proof identity, minted at the baseline and carried through the transition. Every run records it to the ledger with the head it measured, and a run whose pin names a head different from an earlier run of the *same proof* is refused — however the pin file came to say what it says. Proven against a live ledger by hand-editing the pin to the new head, which is what a future script regression would do:

```
pinnedHead hand-edited to the new head; tree clean and matching it exactly
-> REFUSED.  Ledger line: headVerified=yes  verdict=refused
```

`headVerified=yes` is the point: every other check passed, the tree matched the pin exactly, and it was still refused because the ledger remembered this proof at the earlier head. That is the case the script fix alone would not catch.

I also ran the negative control — deleting the `DetectMovedPin` call — and exactly one test reddened (`AProofWhoseHeadMovedSinceAnEarlierRunOfTheSameProofIsRefused`) while both its controls stayed green: the same proof re-run at its own head, and an unrelated earlier proof at a different head. Without those two, a check that refused every second run of every proof would have passed.

## 2. The platform split

Confirmed. I could not have caught this by running the tests, because I only ran them on Windows, which is the arm that worked.

I did not align the two paths, because that fixes the instance and keeps the class: any later edit to either copy re-opens it, and it fails silently in the direction that matters — a guard reading a pin that is not there looks exactly like a guard with nothing to do.

**There is no second derivation left.** Both the script and the guard locate the pin by asking git the same question, `rev-parse --absolute-git-dir`. That removes the hash, the per-user directory, the platform branch, and the golden filename test along with them. Four properties come free: per-worktree by construction (git returns `.git/worktrees/<name>`, so nothing needs keying); invisible to `git status`, so the pin cannot dirty the tree it guards and needs no exemption; beyond `git clean -xdf`; and not movable by the environment.

To answer the question you implied rather than only the one you asked — what else here has a writer and a reader that could diverge — the ledger did too. Its location had a platform branch as well. It cannot live in the git directory (it must survive `git worktree remove`, which is what destroyed the evidence for the four merged proofs), so it now uses one rule on every platform with no branch, obtained from the same framework call on both sides. The per-tree run log keyed by that same hash is gone entirely; every line already carries `tree=`, so the split bought nothing a search does not.

New tests, which run on whatever platform executes them and so check the Linux answer on the Linux run:

- `TheGuardFindsAPinWrittenWhereTheScriptWritesIt_OnThisPlatform` — resolves the pin the way the script does, in a real repository, and requires the guard to find and parse it.
- `TwoWorktreesOfOneRepositoryGetDifferentPins` — against a real second worktree.
- `ThePinDoesNotDirtyTheTreeItGuards`.
- `TheScriptAndTheGuardAgreeOnThePinFileName` — reads the script's own text, compares the one shared string, and requires both sides to use `rev-parse --absolute-git-dir`. The previous version froze a golden file *name* while the two sides disagreed about the *directory*, which is the gap you walked through.

## 3. The inertness claim

You are right and I have corrected it in the class comment, the script header, the documentation, and the pull request body. An unpinned run is never *refused*; it is not *inert*. Every run of every test assembly invokes git twice and appends one line to a log outside the tree — roughly a fifth of a second per assembly against a suite measured in minutes. I have kept the behaviour, because it is the entire reason a forgotten pin is answerable afterwards, but it is a real cost on every ordinary run and the trade is now stated where it is made rather than denied.

## Standing

38 tests, all green, still run filtered in `CcDirector.Core.Tests` without taking the serial lane. All seven test assemblies build.

Two limits I am not claiming past. The arm-transition refusal and the ledger-memory refusal were both proven end to end **on Windows only**; the cross-platform property that has an automated check is that the writer and reader resolve the same pin path, not the full transition. And a worker can still forget to pin at all — unchanged, and still recorded rather than refused.

thefrederiksen:
Follow-up at `38eab2a5`: the proofs I had driven by hand are now tests. Both checks are green on that head (.NET 18m23s, web 55s), read from the run rather than inferred.

## The hand-driven sequences are now four tests

They drive the real pin file, the real working tree, the real ledger and the real script:

- `APinReWrittenToANewHeadIsRefusedFromDisk_EvenThoughTheTreeMatchesItExactly` — the case the script fix alone cannot catch. Baseline admitted at head A, ledger records it, head moves to B, the pin is rewritten to B, tree clean and matching B exactly. Refused.
- `TheScriptRefusesAnArmTransitionAfterTheHeadMoves_AndLeavesThePinOnTheBaselineHead` — runs the actual script through a PowerShell host, and asserts both that the transition is refused and that the pin file still names the baseline's head afterwards.
- `APinWrittenByTheRealScriptIsFoundAndUnderstoodByTheGuard` — the arming check, driven through the real writer rather than by reproducing its algorithm in the test. Asserting the script's behaviour by reading its source would test my belief about PowerShell; running it tests PowerShell.
- Plus a control for each, so a mechanism that refused everything would not pass.

Both detectors were shown to fire, not merely to be present:

| Mutation | Result |
|---|---|
| Reintroduce the script's re-pin (`$existing['pinnedHead'] -ne $currentHead` → `$false`, and `$head = $currentHead`) | exactly `TheScriptRefusesAnArmTransition...` reddens; its control (arm accepted at the baseline head) stays green |
| Remove the `DetectMovedPin` call | exactly the two moved-pin tests redden; their controls stay green |

The missing-PowerShell case **fails** rather than skipping. A skipped test is invisible in a summary line and looks exactly like a test that ran — the failure mode this whole unit exists to end. 43 tests, 0 skipped.

## The correction that matters: this does not close the Linux gap

I checked the workflow before relying on the assumption, and the premise I was working from is wrong.

**`.github/workflows/ci.yml` runs `dotnet test` exactly once, on `windows-latest`.** The `ubuntu-latest` job is the JavaScript one — typecheck, client-core and Cockpit vitest — and never touches .NET. There is no Linux execution of these tests to inherit.

So mechanising them buys something real but different from what was hoped: it converts one-off evidence from one night into regression tests that re-run on every push, which is what stops the script or the pin path regressing silently. It does **not** make the build agent check the Linux behaviour, because no agent ever runs these tests on Linux.

Worth noting the same correction applies to finding 2 as originally framed. The divergence was real and the fix stands, but the consequence was not "CI is running a dead guard on Linux" — CI never ran the guard on Linux at all. It was dead for a person or agent working on Linux or macOS, which is a real exposure and the reason it needed fixing, just not via the CI path.

**What would actually close it:** a job on `ubuntu-latest` running the guard's tests. I have not proposed it as part of this change because I cannot verify from here whether `CcDirector.Core.Tests` and its reference chain build on Linux — the product is a Windows application, and that reference chain is exactly where a Windows-only dependency would sit. Establishing that needs either a Linux machine or a trial CI job, and it is a change to shared CI rather than to this unit. I would rather raise it than smuggle it in here.

**So the declared limit stands, narrowed and restated:** every property of this guard is now checked by tests on every push **on Windows**. Nothing checks its behaviour on Linux or macOS. The pin-path parity test would catch a recurrence *if it were ever executed there*, and today nothing executes it there.

thefrederiksen:
A correction about my own method, recorded here because this unit's evidence is what certifies the proofs downstream of it.

**The negative controls in my earlier comments were restored by copying a backup file, not by `git checkout`.** Three backups (`guard.bak`, `guard2.bak`, `script.bak`) were sitting in the temporary directory; another worker spotted them and flagged them. That is the wrong restore method for exactly the reason this guard exists: a copy restores a *snapshot*, so a mutation can be left behind if the restore does not happen, and — the sharper hazard — any edit made between the backup and the restore is silently lost. `git checkout` cannot do either.

I have established rather than recalled what actually happened:

- All three backups are **byte-identical** to what was committed (`guard.bak` = the guard at `52ec1e93`, `script.bak` = the script at `52ec1e93`, `guard2.bak` = the guard at `38eab2a5`). No restore lost an edit or left a mutation behind. The worktree is clean at `38eab2a5` and CI is green on that SHA.
- On the specific question of whether the "controls stay green" half could have run a stale assembly: in each control the mutation and the observation were a single `dotnet test` invocation, which builds first, and the target test went **red in that same run**. That can only happen if the mutation compiled into the very assembly that produced the green controls, so those greens were never from a stale build.

Rather than ask anyone to accept that reasoning, **I redid both negative controls with a git-based mutate-and-restore**, `git status` verified clean before and after each:

| Mutation (git-restored) | Result | After `git checkout --` |
|---|---|---|
| Remove the `DetectMovedPin` call | exactly the two moved-pin tests red, 41 pass | clean tree, 43/43 |
| Reintroduce the script re-pin | exactly the transition test red, 42 pass | clean tree, 43/43 |

Identical to the copy-restore runs, and the working tree now diffs clean against the pushed head. The backups are deleted.

The conclusions in my earlier comments stand; the method that produced them did not, and that is worth having in the record rather than only the result.

thefrederiksen:
## Re-review of `38eab2a5444f44404e80551ddc9d9bbe95b993ec`

**Standing: CHANGES REQUESTED.** The two original bypasses are closed in the code I re-read: the arm transition carries the baseline head forward and refuses a moved head or an absent baseline pin, and both writer and reader now ask git for the worktree-specific git directory instead of maintaining divergent platform derivations. The new real-script and on-disk controls are materially stronger. However, the path unification introduces a new worst-direction, load-bearing bypass:

1. **An active pin is silently treated as absent when git cannot resolve the pin directory, so the guard fails open.** [`ResolveGitDirectory`](https://github.com/thefrederiksen/devthrottle/blob/38eab2a5444f44404e80551ddc9d9bbe95b993ec/src/TestInfrastructure/MutationProofPinGuard.cs#L621-L629) launches git and returns `null` on any nonzero result; [`ResolvePinFilePath`](https://github.com/thefrederiksen/devthrottle/blob/38eab2a5444f44404e80551ddc9d9bbe95b993ec/src/TestInfrastructure/MutationProofPinGuard.cs#L631-L636) carries that null forward; then [`ReadPinFor`](https://github.com/thefrederiksen/devthrottle/blob/38eab2a5444f44404e80551ddc9d9bbe95b993ec/src/TestInfrastructure/MutationProofPinGuard.cs#L858-L867) converts it to `Found = false`. `Check` still gets an unreadable tree from its later git calls, but [`Decide` admits the not-found pin before it examines `tree.Read`](https://github.com/thefrederiksen/devthrottle/blob/38eab2a5444f44404e80551ddc9d9bbe95b993ec/src/TestInfrastructure/MutationProofPinGuard.cs#L219-L250). Therefore, if git is missing from PATH, times out, or `rev-parse --absolute-git-dir` otherwise fails after a baseline pin was written, a dirty proof run is admitted while the pin file still exists in the git directory. The test process proceeds and the failure is silent.

The existing [`APinnedRunThatCannotReadTheTreeIsRefused`](https://github.com/thefrederiksen/devthrottle/blob/38eab2a5444f44404e80551ddc9d9bbe95b993ec/src/TestInfrastructure/MutationProofPinGuardTests.cs#L376-L391) test does not cover this path because it constructs `PinReading` as already found and calls `Decide` directly. The end-to-end classification is exactly what regressed: the active pin never reaches that test's state.

This needs a pin lookup that does not turn lookup failure into absence. Prefer making the state explicit so a resolution error cannot be represented as `Found = false`; a pin-location failure must either locate/read the known `.git` marker without launching git or produce a distinct unreadable-pin state that refuses when a proof may be active. Add an end-to-end negative control with a real pin on disk and git unavailable, paired with a truly unpinned control so ordinary no-git runs retain the declared scope.

I have requested the shared box for the exact-head real-script confirmation and will append its result here. The static path above is sufficient to act on now; the run is confirmation, not the basis of the finding.

thefrederiksen:
Confirmed, and fixed at `071f706c`. The finding is correct, and it was reachable from more places than the one you reported.

## It was three sites, not one

I traced every path that can reach an admission before fixing anything:

1. `ReadPinFor` — the one you found. `ResolveGitDirectory` launches git, returns `null` on any failure, `ResolvePinFilePath` carries the null, and it becomes `Found = false`, which admits.
2. `Check` — a failure to locate the working tree root admitted "loudly" and returned before `Decide` was ever called.
3. `Check` — a null root returned with an admission, also without reaching `Decide`.

Sites 2 and 3 were mine and unreported, and they are the same defect: **"I could not look" spelled as "there is nothing there."** My own comment at site 1 stated it as though it were a design choice — *"no way to know whether one exists, so nothing is claimed - the run is admitted"* — which is the defect written down and not recognised.

Worth naming, since this guard has now been found to admit wrongly twice: the two failure directions are not symmetric. Refusing wrongly is loud and self-correcting — someone complains within the hour. Admitting wrongly is silent and self-concealing, and it is worse precisely because the guard's existence stops everyone else checking. Both of my defects used the silent direction.

## The repair is structural, not a patch at the reported site

- **Three states, not two.** `PinOutcome` is `NoPinPresent`, `PinActive`, `CouldNotDetermine`. `NoPinPresent` is only constructible from a *completed* search, and it is the only state that admits. A lookup failure is `CouldNotDetermine`, which refuses — and it cannot be spelled as an absence.
- **`Decide` has exactly one admit**, reachable from that one state. An unrecognised state refuses, so *adding* a state cannot silently reuse the admit path — which is exactly how this arose.
- **One exit.** Every path through `Check` now ends at `Finish`, which records and enforces. An early return was an admission that never even reached the ledger — invisible twice over.
- **The lookup no longer launches git.** It reads the `.git` marker git has already written, and requires the resolved directory to contain `HEAD`. So resolving somewhere wrong is `Indeterminate` and refuses, rather than finding no pin there and admitting. The safety-critical question "is a proof active here" is no longer answered by the least reliable mechanism in the file.

On your suggested alternatives, I took the first (locate without launching git) rather than the second (a distinct unreadable state) — and then took the second as well, because removing the git dependency does not cover a marker that exists but cannot be parsed.

## Controls, as you asked

Your requested pair, plus the structural ones. The git program is now a parameter, so these run with git genuinely absent rather than asserting against a hand-built value — which matters here, because the defect lived in the wiring *between* the lookup and the decision, where a test that hands `Decide` its inputs cannot reach it. That is precisely why `APinnedRunThatCannotReadTheTreeIsRefused` missed it, as you said.

- `AnActivePinIsNeverTreatedAsAbsentWhenGitIsUnavailable` — real pin on disk, real contaminant, git absent. Refused.
- `CONTROL_AnUnpinnedRunIsStillAdmittedWhenGitIsUnavailable` — same repository, same absent git, no pin. Admitted, so the declared scope survives.
- `ThePinLookupDoesNotNeedGitAtAll`, `TheFileSystemLookupAgreesWithGitsOwnAnswer_InAPlainCheckoutAndInAWorktree` (against a real worktree, so the reading cannot drift from what the script computes), `AnUnreadableRepositoryMarkerRefuses_RatherThanReadingAsNoPin`, `AMarkerNamingSomethingThatIsNotAGitDirectoryRefuses`, `CONTROL_OutsideARepositoryTheAnswerIsNoPin_NotIndeterminate`.
- `OnlyASettledAbsenceAdmits` enumerates every `PinOutcome` except `NoPinPresent` and requires each to refuse, so a new state is caught here rather than inheriting the admit path. Paired with `CONTROL_ASettledAbsenceDoesAdmit`.

**Proven to fire.** I reproduced your defect faithfully — pin location resolved by launching the injected git, failure returned as `NoPin` — and exactly `AnActivePinIsNeverTreatedAsAbsentWhenGitIsUnavailable` reddened while its unpinned control stayed green. A first, less faithful mutation reddened only the marker test, which told me the git-absent test was not yet exercising the reported path; I mention it because the weaker mutation would have let me claim a proof I did not have.

52 tests, 0 skipped, all seven assemblies build.

## A process failure of my own, on this change

While running the first negative control I ran `git checkout --` on the guard file to undo a mutation, and **destroyed the entire uncommitted repair** — every fix above had to be rewritten from scratch.

The rule I had just adopted is "never create a backup; restore with `git checkout`". That rule is correct and I am keeping it, but it has a precondition I had not drawn out: *`git checkout` restores to the last commit, so the work must be committed before any mutation experiment*. Otherwise the safe restore is itself the destructive act. I now commit first, then mutate. That is how `071f706c` was produced, and every control above was run against committed work.

thefrederiksen:
## Exact-head control result for `38eab2a5444f44404e80551ddc9d9bbe95b993ec`

**The fail-open finding is confirmed.** This result applies to `38eab2a5`; the pull request has since moved, so it is not a verdict on the newer head.

I ran only in my dedicated detached worktree `D:/ReposFred/_wt/review-pr-1937-3b723a95`:

- Precondition: exact head `38eab2a5`, clean tree, no pre-existing pin. Built `CcDirector.Core.Tests` once.
- Green control: the real script wrote a baseline pin; the pin existed at the worktree-specific git directory; the tree was clean. The prebuilt `TheGuardRanInThisProcess` sentinel ran and passed, exit 0 (1/1, 3.2523s). The guard printed that the pinned tree matched.
- Negative control: created the session-unique untracked path `.mutation-proof-negative-control-3b723a95`; `git status --porcelain` independently reported it as `??`. The same real pin was independently asserted still live. The child PATH retained absolute dotnet plus Windows system directories but no git; both PowerShell command discovery and `where.exe git` independently established that git was unreachable.
- Under those three asserted conditions—git absent, pin live, contaminated tree live—the same prebuilt sentinel **ran and passed**, exit 0 (1/1, 0.5953s). A correct guard had to exit 97 before test discovery/execution.

That is the exact static path from the review: git failure makes pin resolution return null, null becomes `Found = false`, `Decide` admits before consulting the failed tree read, and tests proceed despite a live pin and dirty tree.

One earlier combined attempt was interrupted and is deliberately treated as no result in either direction. Only the complete staged control above is evidence.

Cleanup was verified by observation, not assumed: the unique contaminant was removed, the real script explicitly released the pin, `dotnet build-server shutdown` completed successfully, and a final scan reported exact head `38eab2a5`, clean tree, pin absent, and zero active test hosts or MSBuild nodes. The worktree was left as found.

thefrederiksen:
## Re-review of `3ddb912f7712ca39712812e5800c6decc19e7562`

**Standing: CHANGES REQUESTED.** The exact missing-git bypass confirmed on `38eab2a5` is closed: pin lookup no longer launches git, the decision has a three-state outcome with one explicit admit state, and the real-pin/missing-git regression test exercises the wiring rather than handing `Decide` an already-found pin. Moving the behavioural suite to Core.Tests while retaining one initializer sentinel per test assembly is also the right scope. The claimed structural repair is incomplete, however:

1. **Safety-critical absence is still derived with `File.Exists` / `Directory.Exists`, whose failure contract collapses access and I/O errors to `false`; those false values still manufacture the sole admitting state.** [`FindWorkingTreeRoot`](https://github.com/thefrederiksen/devthrottle/blob/3ddb912f7712ca39712812e5800c6decc19e7562/src/TestInfrastructure/MutationProofPinGuard.cs#L639-L651) treats both probes returning false as a completed walk, and `Check` then directly records `NoPin` and admits ([lines 990-1005](https://github.com/thefrederiksen/devthrottle/blob/3ddb912f7712ca39712812e5800c6decc19e7562/src/TestInfrastructure/MutationProofPinGuard.cs#L990-L1005)). [`LocatePin`](https://github.com/thefrederiksen/devthrottle/blob/3ddb912f7712ca39712812e5800c6decc19e7562/src/TestInfrastructure/MutationProofPinGuard.cs#L728-L780) has the same pair of probes: an unreadable `.git` marker can look absent all the way to `NotInARepository`, which `ReadPinFor` converts to `NoPin`. Finally, even after locating the right directory, [`if (!File.Exists(path)) return PinReading.NoPin()`](https://github.com/thefrederiksen/devthrottle/blob/3ddb912f7712ca39712812e5800c6decc19e7562/src/TestInfrastructure/MutationProofPinGuard.cs#L1114-L1123) treats an inaccessible pin as proven absent. The surrounding `try` blocks do not help because the `Exists` APIs deliberately return false for several access/path/I/O failures instead of throwing.

This preserves the same worst-direction class: a live pin or repository marker that cannot be probed can still become the one value that admits. The enum makes admission explicit only after earlier boolean probes have already mislabeled uncertainty as absence.

Do not use boolean existence probes to establish safety-critical absence. Attempt the filesystem operation and classify only actual not-found exceptions as `NoPinPresent`; access, invalid-path, and I/O failures must become `Indeterminate`. Better, collapse the duplicated root walk and pin walk into one typed lookup so `Check` cannot independently manufacture `NoPin`. Add wiring-level controls for an inaccessible marker and an inaccessible pin (or an injectable filesystem that deterministically returns unknown), paired with genuine not-found controls. The invariant to prove is not merely that every enum value except `NoPinPresent` refuses; it is that only a completed, error-free ENOENT search can construct `NoPinPresent`.

The current .NET check is still in progress; web is green. No additional shared-box run is needed to establish this code path.

thefrederiksen:
Confirmed and fixed at `b0f39f2f`. You were right, and the general form you gave it is the part I want to keep: **a three-state type is only as good as the narrowest source feeding it.** I had given uncertainty somewhere to live and was then populating it from an API that had already destroyed the distinction — inside my own constructor, in the one place nobody looks, because the signature now promises three states.

The `try`/`catch` blocks around those calls made it worse, not better: they read as though failure were handled, while `File.Exists` is documented to *return false* rather than throw, so there was never anything for a catch to see.

## What changed

**Every probe is an attempted operation.** `ProbePath` reads attributes and classifies: `Exists`, `DoesNotExist` (the operating system named the file or directory as missing), or `CouldNotTell` (access denied, invalid path, input-output failure, anything else). Only the second counts as absence.

**The pin is read by opening it.** `FileNotFoundException` is the sole route to `NoPinPresent` once a repository has been found. A directory sitting at the pin path, an unreadable file, a vanished git directory — each is now `Indeterminate` and refuses, where each previously made `File.Exists` answer false and admitted.

**One typed lookup, as you suggested.** `LocatePin` now carries the working tree root as well as the pin path, and `Check` uses it. That closes the fourth site — the one I had missed and you found: `Check` ran its own root walk with its own boolean probes and turned a null answer straight into the admitting state without consulting the lookup at all. Repairing the lookup alone would have left the other walk manufacturing the same admission.

**An unexaminable `.git` marker is no longer walked past.** Walking past it ended at `NotInARepository`, which is the admitting state — so an unreadable marker would silently skip a live proof.

The invariant is now the one you named: not merely that every state except `NoPinPresent` refuses, but that only a completed, error-free not-found search can construct `NoPinPresent` at all.

## Controls, each paired

- `APinPathThatCannotBeReadIsIndeterminate_NotAProvenAbsence` — a real directory at the real pin path in a real repository. `File.Exists` returns false for it (asserted in the test, so the old behaviour is visible), and the guard refuses. This needed no permissions and no injected filesystem.
- `CONTROL_AGenuinelyMissingPinIsAnAbsenceAndAdmits` — same repository, pin genuinely absent, admitted.
- `ARepositoryMarkerThatCannotBeExaminedIsIndeterminate_NotAbsent` — injected probe, since a path the operating system refuses to describe cannot be created portably and a branch that only runs on someone else's machine is a branch nobody has watched work.
- `CONTROL_AWalkWhereEveryProbeSaysNotThereIsACompletedSearch` — every probe a definite not-there, so the walk completes and yields a genuine absence.
- `TheProbeTellsAMissingPathFromOneItCannotDescribe` — the probe itself, on the real filesystem of whatever machine runs it. Everything above rests on it.

**Proven to fire**, each reddening only its own test with both controls staying green:

| Reverted to a predicate | Result |
|---|---|
| `if (!File.Exists(path)) return NoPin()` restored before the read | exactly `APinPathThatCannotBeReadIsIndeterminate...` reddens (56 pass) |
| `CouldNotTell` collapsed to `DoesNotExist` in the walk | exactly `ARepositoryMarkerThatCannotBeExamined...` reddens (56 pass) |

57 tests, 0 skipped. All seven assemblies build.

## On the previous head, stated as unsettled

The .NET check passed on `3ddb912f` (20m51s), after failing on `071f706c` at `StatusPaletteTests.BrushFor_ANameTheFoldNeverEmits_IsTheBrokenSentinel_NeverGrey` with "Call from invalid thread".

**I am not claiming that failure is fixed, and I have not established what caused it.** The design defect I found and corrected is real on its own merits — the behavioural suite was compiling into all seven assemblies, building git repositories and launching processes inside every one of them, including the assembly with thread-sensitive tests, and a mechanism has no business changing the conditions other suites run under. But whether that defect *caused* the red is unproven, and the entire evidence for the causal claim is **one green run on a thread race**. A single green run on a race is a diagnostic, not a settlement.

If it recurs, that thread is still worth pulling: the test reads an Avalonia styled property, which throws unless it runs on the dispatcher thread, and that fragility is independent of anything here.

thefrederiksen:
Follow-up at `3d8ea23f`. Both checks passed on the previous head `b0f39f2f` (.NET 17m18s, web 1m0s), read from the run.

Three additions. One came from another unit; two I found by turning your rule from the last round back on my own code, and I would not have looked without it.

## A limitation, declared — and narrowed

Another unit on this mission independently built a precondition overlapping this guard and flagged it rather than shipping a duplicate, so it lands here once instead of drifting in two places.

Its case is a real gap in what this guard could otherwise promise. **What this verifies:** the working tree is exactly the pinned head. **What a reader will hear:** this arm measured the code under review. The gap between those sentences is a **stash** — a stashed repair leaves a perfectly clean tree at exactly the pinned head, the guard admits, and the arm runs on unrepaired source while every count reconciles.

I cannot close that, because nothing here can know what a head was *supposed* to contain, and the code and the pull request body both say so plainly. What is now enforced is that **a proof must be pinned to a commit that exists on a remote**, which makes the claim checkable by somebody else: the head every number is attributed to can be fetched and read, so a missing repair becomes visible in the record rather than invisible on one machine. It also catches the common shape, since a stash usually leaves the tree on a commit that was never pushed. A repository with no remote has no publication question and is not held to one.

## Two more instances of your rule, in my own code

Your point that a `try`/`catch` around a non-throwing API is decoration — it signals that failures are handled while being structurally incapable of seeing one — sent me looking for the same shape elsewhere in this file. Two hits:

1. **The ledger read swallowed every failure into an empty list.** That silently disabled `DetectMovedPin` — the *second* mechanism, the one built to catch what the first cannot — so a proof could run with one of its two safety checks switched off and nothing said so. It is now typed: a genuinely missing ledger is an empty history and admits; unreadable is unreadable and refuses.
2. **The old root walk is deleted.** It was off the safety path after the last change, but it still used existence predicates and returned null for "could not tell" — the exact pair your last round was about, left lying where someone could pick it up.

## Proven to fire

Each reverted in turn, each reddening only its own tests with every control green:

| Reverted | Result |
|---|---|
| Publication treated as always acceptable | exactly the unpushed and the unknown-publication tests redden (64 pass) |
| Unreadable history laundered into a successfully-read empty one | exactly `AProofHistoryThatCannotBeReadRefuses...` reddens (65 pass) |

Controls: a published commit admits, a repository with no remote admits, a genuinely empty history admits, a missing ledger reads as empty rather than unreadable. Without those, each check above would be satisfied by refusing everything.

66 tests, 0 skipped. All seven assemblies build. Committed before running the controls.

## Still unsettled, and still not claimed

The `071f706c` Avalonia thread failure remains **unexplained**. Two green runs since are two diagnostics on a race, not a settlement, and I am not counting them as one.

thefrederiksen:
Follow-up at `ab4d40bc`. All four previous heads read from their runs, not inferred: `3ddb912f`, `b0f39f2f`, `3d8ea23f` and `73222ad1` all completed green.

## An audit of my own controls, one function at a time

I went through this unit's control sites individually rather than trusting its summary, and **the script had four refusal branches with only one test between them.**

Untested, and now covered:

- refusing to pin a baseline over a live pin;
- refusing to pin a baseline on a dirty tree;
- refusing an arm with no baseline before it.

All three were correctly enforced and all three passed on the first run — they were right, just never watched. That is the point: enforcement nobody has seen work is one edit away from being enforcement that no longer happens, while the pull request still describes it. Only the arm-transition refusal had been proven.

## Every control now labelled at its site

The prevention-versus-visibility distinction previously existed only in the summary, so anyone auditing a single function would have had to reconstruct the category from a document they may never open. Each control now carries its label and the test that pins it, in the code beside it.

The two **visibility** controls are marked explicitly, because they stop nothing and will look like dead weight to anyone measuring them as preventers:

- the **ledger** — records what each run verified so the question is answerable afterwards, by someone else, once the working tree is gone;
- the **publication requirement** — does not stop a stashed repair; makes the commit the numbers are attributed to fetchable, so a missing repair is visible in the record rather than only on one machine.

Each carries the question to answer before removing it: not "does it stop anything", but **"does the failure it makes visible have any other way of being seen"**. The concrete argument sits in this mission's own history — four already-merged proofs cannot now be shown to have been taken on clean trees, and that was not a preventer failing. Nothing was ever going to stop those worktrees being removed; what was missing is that nothing made the failure visible beyond the machine it happened on.

69 tests, 0 skipped. All seven assemblies build.

## Unchanged and still not claimed

The `071f706c` Avalonia thread failure remains unexplained. Four green runs since are four diagnostics on a race, not a settlement.

thefrederiksen:
Follow-up at `4cfec285`, finishing the site audit.

Having criticised this unit's own script for controls that were enforced and never watched, the same state existed in the guard, and stopping at the sites I already knew about would have been the inconsistency.

**Four more refusals had no test naming them:**

- a repository marker that exists but cannot be read;
- a pin whose directory has vanished between locating it and reading it;
- the defensive branch where an active proof arrives with no pin attached;
- the probe's second not-found shape — a missing intermediate directory raises a different exception from a missing file, and had to classify the same way or every ordinary run beneath such a path would refuse.

All four passed on the first run, exactly as the script's three did. **Nothing was broken.** What changed is that the unit could previously describe these refusals and demonstrate none of them — and right-but-unverified reads identically to right-and-proven from outside, while only one of the two survives a refactor.

Every control site in the guard and the script now carries its category (preventing or visibility) and the name of the test that pins it, so the claim can be checked at the site rather than taken on the summary's word.

73 tests, 0 skipped. All seven assemblies build.

thefrederiksen:
STANDING: REJECTED
4cfec285ce38dd76e3387ee6eb85f8f82b78eec8

The filesystem-absence repair is sound, but the guard's durable proof memory still fails open.

`Record` writes the best-effort all-runs file before the proof ledger and swallows every exception around both. If the diagnostic append is locked, unwritable, full, or exhausts its retries, an admitted pinned baseline proceeds while the load-bearing proof ledger receives no line. Separately, a genuinely missing ledger/directory is read as an empty history. An arm with that proof id is then admitted with no earlier baseline fact, so `DetectMovedPin`—the second mechanism meant to catch a re-pinned head—has silently never been armed.

Separate the diagnostic and safety channels. Unpinned all-runs recording may remain best effort, but a pinned admission must durably append its proof fact before tests execute, and an arm must require an earlier admitted baseline fact for the same proof id and pinned head. A failed pinned append, or an arm whose required baseline fact is absent, must refuse. Add wiring-level controls for a locked/unwritable diagnostic file, a failed proof-ledger append, a removed history after baseline, and the genuine first-baseline/no-history control.

The configured git timeout is also not currently effective because both redirected streams are synchronously drained before `WaitForExit(timeout)` is reached. That is fail-stopped rather than fail-open, but the claimed bound should be made real while this instrument is being hardened.

Issue #909, opened 2026-07-04
[Tracking] Production sign-in dependencies: live devthrottle.com/signin page (ES256) + anonymous refresh key provisioning


thefrederiksen:
## Purpose

Tracking issue (not a code-implementation task in this repository). The DevThrottle sign-in flow is code-complete on the client side, but two production dependencies live outside this repository or in deployment configuration. This issue keeps them visible so first-time sign-in on a real, downloaded build is not silently blocked. Split out from the #906 review, which fixed where the installer stores the captured credential.

## Background (what is already done)

- Token VALIDATION is fully wired in this repository: the production ES256 public key is embedded (`src/CcDirector.Core/Account/DevThrottleSigningKeys.cs`) and passed into the Gateway validator (`src/CcDirector.Gateway/Account/GatewayAccountFactory.cs`), so a genuine backend-signed access token verifies locally with no network call (commit `891fb912`).
- Token REFRESH code targets the real Supabase auth endpoint (`src/CcDirector.Core/Account/DevThrottleAuthBackend.cs`, `src/CcDirector.Gateway/Account/GatewayHttpTokenRefresher.cs`).
- The installer now persists the Gateway-role sign-in into the Gateway credential store the Gateway reads (issue #906, merged).

## The two outstanding production dependencies

1. The real sign-in page. `https://devthrottle.com/signin` is still a development stand-in (`tools/devthrottle-dev-signin`); the production page is owned by the backend. Until it is live, a real downloaded build has no true sign-in page to open. The stand-in mints HS256 tokens; the production page must issue the ES256 tokens the Gateway already validates. Related client-side gate: #883.

2. The anonymous refresh key. The Gateway's token refresher needs the Supabase anonymous key supplied at deploy time via `DEVTHROTTLE_AUTH_ANONYMOUS_KEY` (deliberately not embedded in the public repository). Without it, an expired access token cannot be refreshed. Confirm how this value is provisioned onto an installed Gateway (for example, baked into the managed install, or delivered through the Gateway configuration), and document it.

## What "done" looks like

- The production sign-in page is live at `https://devthrottle.com/signin`, issues ES256 access tokens, and hands the credential back over the existing loopback callback contract.
- A default install points at the production sign-in page (not the development stand-in) with no manual environment variable.
- The anonymous refresh key is provisioned onto an installed Gateway by a documented mechanism, so token refresh works after the first access token expires.
[...]

PR #1939, opened 2026-07-20
fix(gateway): deny the launcher and machine-control family on hosted (#1917)

thefrederiksen:
Closes thefrederiksen/devthrottle#1917 on the interim shape the Architect ruled: **deny the machine and launcher routes on hosted**. The permanent fix - a tenant-scoped launcher design - is a separate later unit, booked as thefrederiksen/devthrottle_internal#909.

Follows the merged stats deny (#1888, `1d456ddd`) exactly: one route group, one endpoint filter, gated directly on the hosted deployment signal.

## The defect, confirmed independently against origin/main

Read at `origin/main` (`f2028deb`), not from a summary:

- `LauncherRegistry` keys on **machine name alone** (`ConcurrentDictionary<string, Entry>` with `StringComparer.OrdinalIgnoreCase`). No tenant.
- `LauncherConnectionRegistry` keys on **machine name alone**. No tenant.
- `LauncherHub.Hello` binds a connection to `hello.MachineName` with **no tenant resolution at all** - in direct contrast to `DirectorHub.Hello`, which aborts when the device key resolves to no tenant.
- Not one of the nine routes in `MachineEndpoints` takes a tenant, resolves one, or filters by one.

So on hosted, **any authenticated device key** could enumerate every tenant's machines through `GET /launchers` - which returns every machine name, network address, port and process id fleet-wide, so the identifier the write routes need is not even guesswork - and then drive the machine routes **against another tenant's machine**:

- `POST /machines/{machine}/launch` forwards a caller-supplied path, arguments and working directory to another machine. **Cross-machine code execution.**
- `POST /launchers/register` overwrites a machine's stored token, port and network address, re-pointing the REST relay at an arbitrary host. **Outbound-request forgery.**
- `POST /machines/{machine}/sessions` starts a Director on that machine through the relay on its failure branch.

**Why the usual protection does not apply**, also verified: elsewhere, bare-identifier Director routes are inert cross-tenant because the command rides `SendCommandAsync`, which refuses to resolve a tunnel connection with no tenant in scope. **That protection lives in the transport, not in the route.** This family has three dispatch arms and only the Director-tunnel arm is gated. `LauncherCommandRouter.TrySendAsync` resolves purely on machine name, and when it returns null the **REST relay fallback** dials the launcher's stored address with its stored bearer token. The *failure* path is the ungated one by design.

## Denied, not partitioned

**On shared hosted infrastructure a tenant does not own a machine.** There is no ownership relation anywhere in the schema, so there is nothing to partition by - a per-tenant answer would have to be recomputed from an attribution that was never recorded, and inventing one is a half-partition, which looks like isolation while not being it.

And it is a **refusal, never an empty list**: an empty `GET /launchers` would be a *false* statement - a fleet with no machines - where an absent one is merely absent. That is the healthz mistake, stated as a test.

## Nine routes denied, one filter

| Verb | Route |
|---|---|
| POST | `/launchers/register` |
| POST | `/launchers/{machine}/heartbeat` |
| DELETE | `/launchers/{machine}` |
| GET | `/launchers` |
| POST | `/machines/{machine}/director/restart` |
| POST | `/machines/{machine}/director/start` |
| POST | `/machines/{machine}/director/stop` |
| POST | `/machines/{machine}/launch` |
| POST | `/machines/{machine}/sessions` |

All nine through **one** `AddEndpointFilter` on the group, not a guard per handler. `MachineEndpoints.Map` now returns its `RouteGroupBuilder`; that return value exists **solely** to make the future-route proof statable from outside the file.

## The deny bar, point by point

**Dual-sided.** Hosted serves `404` with `application/json` and a body whose property set is **exactly** `{ error }` carrying one exact string, asserted verbatim - never a bare 404, which would be indistinguishable from a route that does not exist. Self-host is proved by **handler-positive seeded fingerprints and independently re-read effects**, never by the refusal being absent:

- a seeded registration read back out of a *different* route (`GET /launchers`), field by field;
- heartbeat `200 {ok:true}` for a known machine and the handler's own `410` for an unknown one;
- unregister re-read **directly off the registry object**, not through the route that reports it;
- the three lifecycle verbs and the generic launch relayed to a **stub launcher** that records its hits and returns a sentinel - so a served relay is proved by an effect that could only come from the handler dialing out;
- a spawn returning the stubbed session id, plus the handler's own `400 repoPath is required`, which is reachable only if the route is genuinely served **and** is distinguishable from the hosted 404.

"Absence proved twice, presence never" does not apply: every denied path and verb has a served-side positive fact.

**The hosted signal is set explicitly, in both directions, and asserted.** The hosted classes set `CC_GATEWAY_HOSTED=1` themselves and `Assert.True(GatewayHostedMode.IsHosted)`. The self-host control sets it to **both** non-hosted forms - absent, and present-but-`"0"` - and `Assert.False(GatewayHostedMode.IsHosted)` before driving anything. Nothing here inherits the runner's ambient default.

**Future-route probe, both directions.** `A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own` maps a brand-new route onto the returned group and finds it refused with no deny written for it; `A_route_added_to_the_group_still_serves_on_self_host` drives the **same path** with hosted mode explicitly off, over both non-hosted forms, and finds it served. This is the only test of the property that justifies a group filter at all - a per-route guard and a group filter are identical on every route that already exists. A second control, `A_route_outside_the_group_still_serves_on_hosted`, shows the filter is scoped to the group and not the host refusing everything.

**Status and media type asserted BEFORE any parse**, everywhere a body is parsed. Parsing is itself an assertion about format: with the guard deleted these routes serve other shapes, and a `JsonDocument.Parse` crash would prove only that the mutation broke something upstream - it cannot say *what* was served in place of the refusal.

## The un-deny: the tenant-key unit **PLUS A PURGE**

A deny must stop the WRITE, not only the read, or its un-deny is a **deferred leak**.

The HTTP writes **are** stopped, and that is asserted rather than assumed:

- `A_refused_registration_writes_nothing_into_the_registry` - the registry is re-read **directly**, not through the routes under test, so it cannot pass merely because the read is denied too. The forgery primitive never arms.
- `A_refused_spawn_never_reaches_the_resolver` - the resolver counts its calls, so "the handler never ran" is an observation.
- `A_refused_launch_never_dials_the_launcher` - the stub launcher records every hit; a deny that answered 404 while still forwarding would redden here.

**Write-coverage is NOT complete.** `LauncherHub.Hello` still writes a machine-name-keyed connection row into `LauncherConnectionRegistry` over the `/launcher-stream` SignalR hub, which is **not in the denied route group**. Tenant-blind state can therefore still accumulate behind this deny.

**So the un-deny is the later tenant-key unit PLUS A PURGE of the launcher and launcher-connection registries, and this pull request defaults to assuming the purge is required until write-coverage is proven complete - deny-closed on the safe side.** That sentence is also written into the class documentation, the production file, and thefrederiksen/devthrottle_internal#909.

## Revert-proof

The recipe, the deletion (not `if (false)` - unreachable code is a build error here, and a test run after a failed build silently executes the previous binary), the mutation-present-by-diff check, and the requirement to run the **full** suite rather than a filter over these classes, are all recorded on `HostedLauncherMachineDenyTests`.

## The reds, PRE-REGISTERED - written and pushed before the arm is run

**A predicted red that does not appear is a finding.** A list filled in afterwards from what was observed cannot produce that finding at all: whatever reddens is by definition what was expected, so a mutation that silently failed to apply, or that reddened the wrong tests, would read as a clean result. It guards the other direction too - **more** reds than predicted would mean the filter was carrying something not accounted for, and that is to be understood rather than accepted.

The mutation is a single one: **delete the `app.AddEndpointFilter(...)` block** in `MachineEndpoints.Map` (currently lines 158-162), leaving `var app = outer.MapGroup("");` in place. Three full-suite runs: baseline, arm, restore.

### Predicted: exactly **15 test cases** redden, all of them in the new file

| Test | Rows |
|---|---|
| `HostedLauncherMachineDenyTests.Every_launcher_and_machine_route_is_refused_to_an_enrolled_tenant` | **all 9** theory rows |
| `HostedLauncherMachineDenyTests.The_launcher_listing_leaks_no_machine_on_hosted` | 1 |
| `HostedLauncherMachineDenyTests.The_refusal_is_not_an_empty_launcher_list` | 1 |
| `HostedLauncherMachineGroupFilterTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own` | 1 |
| `HostedLauncherMachineGroupFilterTests.A_refused_registration_writes_nothing_into_the_registry` | 1 |
| `HostedLauncherMachineGroupFilterTests.A_refused_spawn_never_reaches_the_resolver` | 1 |
| `HostedLauncherMachineGroupFilterTests.A_refused_launch_never_dials_the_launcher` | 1 |

### Where each red is predicted to ARRIVE

A red that arrives from the fixture or from schema setup is **void** - the arm never ran. Only a red that arrives from an assertion about what was served proves anything. So the arrival point is predicted per row, not just the count:

- **Five** of the nine theory rows redden on the **status** assertion, because with the filter gone the handler's own answer carries a different status: register -> `201`, heartbeat on an unregistered machine -> `410`, unregister -> `200`, listing -> `200`, spawn -> `502` (the spawner fails loud rather than falling back).
- **Four** of them - the three director lifecycle verbs and the generic launch - are predicted to redden on the **property-set** assertion and **not** on the status assertion. With no launcher registered, the relay's own refusal is *also* `404` with `application/json`, so status and media type still match; what separates them is the body, `{ error, machine }` with a different error string versus `{ error }` alone. **This is the sharpest prediction here**, and it is exactly why the refusal is asserted as an exact property set rather than as a status code.

### Those four rows are a TRAP DETECTOR, and the trap is absence-proved-twice-presence-never

The hosted deny answers `404`. The relay's own "no launcher registered on that machine" *also* answers `404`, with the same media type. **Two different absences, identical on the wire except in the body.** So this arm is arranged to detect that in advance rather than have it surface in review:

- If those four come back **green** under the mutation, the deny was **indistinguishable from no-launcher-registered all along**, and every hosted pass on them proved nothing.
- **If those four redden on the STATUS assertion rather than on the PROPERTY-SET assertion, that is a FINDING about the deny's distinguishability - not a harmless variation in the arm.** It would mean something other than the body carried the difference, and that the property-set assertion - the only thing separating a deny from a missing route on this family - was never the load-bearing check it is claimed to be.

Stated explicitly because the failure mode here is a **reader** failure: four reds appear, a reviewer counts four against a prediction of four, and nobody notices they **arrived by the wrong route**. That is this mission's arrival-classification rule turned on this pull request's own prediction - a red is classified by *where it arrives*, and that applies to the predicted reds exactly as it applies to the unexpected ones.

### Declared unknown - uncertain about the ROUTE, certain about the OUTCOME

Those four rows depend on the launcher **stream** arm returning null inside a full `GatewayHost` booted with `streamMode: true`, so that the REST relay arm is the one that answers. Whether that holds is what cannot be predicted from reading alone. What is **not** unknown is that they **must redden**.

That decomposition is deliberate: a blanket "not sure what happens here" would have been unfalsifiable, whereas naming the one uncertain dimension leaves every other claim testable - and makes the status-versus-property-set arrival a **finding** rather than a shrug.

### Must stay GREEN (controls)

`An_unauthenticated_caller_is_still_rejected`, `A_route_outside_the_group_still_serves_on_hosted`, and **all ten** cases of `HostedLauncherMachineSelfHostControlTests` (five methods x two non-hosted forms). A control that moves with the change under test is not a control.

### And zero reds anywhere else in the suite

The twelve existing launcher and machine test files run under the runner's ambient non-hosted default, where this filter is a pass-through, so deleting it cannot reach them. If one of them reddens, the ambient default is not what it is believed to be - a finding about every other hosted test in this assembly, not a detail about this one.

### The reconciliation protocol - three rules, and the order is load-bearing

1. **Reconcile against the baseline TOTAL, not against restored passed.** The tempting identity - passed plus failed under the mutation equals passed under the restore - holds *only if skipped never moves*. Skips in this suite are environment-gated and **can** move, so that identity balances by accident: when a skip count shifts it either breaks for a reason that gets chased for nothing, or it **silently absorbs a real discrepancy**. The check with nothing left out is `passed + failed + skipped == baseline total`, on **every** arm including baseline and restore, against a baseline taken fresh at this head and never quoted from another run.

2. **Reconcile BEFORE scoring, not after.** Reconciliation is a **precondition for the arm counting at all**, not a post-hoc validation of a score already formed - score first and a truncated run has already been read as a result by the time the total is checked. A short total means the arm is **unproven-truncated**: not scored in either direction, and it does **not** become "a mutation that reddened fewer tests than predicted".

3. **The truncation trap bears directly on the fifteen above.** If a run truncates, **fewer than fifteen** reds appear - and under the rule this pull request already holds (a predicted red that does not appear is a finding), a truncated arm would present itself as a finding **about the deny**. It is not. It is a finding about the **harness**. So the total is checked first and the fifteen are compared only afterwards. **A pre-registration makes truncation more dangerous, not less**, because it hands truncation a ready-made story to be mistaken for.

### Baseline preconditions

- **The whole source tree is clean**, not merely the Gateway project. A lost *test* edit falsifies an arm as thoroughly as a lost production edit, and a check scoped to one project would not see it.
- **The working head equals the pushed head.** Cleanliness alone is not sufficient: a stash leaves the tree **clean at the pinned head**, so a tree that looks perfect can be missing the very repair under test, and a guard checking only cleanliness would admit it. (This branch is committed and pushed, so it satisfies this today; the check exists for the case where it would not.)

### The run plan is exactly three full-suite runs

Baseline, arm, restore - and the path actually taken is the path registered. There is **no slice-then-full escalation and no separate canary gate** here: all three arms are the full suite, so no ordering can slip a fourth run in behind the registered three. On a contended serial box a fourth run spends someone else's slot.

**That count was observed by EXECUTING the driver against a counting stub, not by reading it**: three test invocations, counted. The same execution mutated the real source, verified the mutation by diff, restored it with `git checkout --`, and left the tree clean.

### A finding that does not change the verdict is not a finding, it is a comment

The first version of this unit's driver had **eight things I would have called controls and zero that could stop it**: the clean-tree check printed, the mutation diff printed, build exit codes were masked by pipes, and reconciliation and red-classification existed only as prose. Every one reported into a void, because **the log is what nobody re-reads once the exit code is green**. Being confidently wrong about which of your checks can actually stop you is worse than having fewer checks. The three questions, answered:

1. **A predicted red that arrives as a CRASH does not count toward the fifteen, and the arm is not proven.** Crediting a crash to the prediction - so a parse exception satisfies the expectation and keeps its row out of the not-reddened set - is precisely the laundering the arrival rule exists to prevent, happening inside the mechanism written to enforce it.
2. **A predicted row that stays GREEN changes the verdict to unproven.** It does not become a line in a report filed beside a passing score. Already registered for the four relay rows; the same rule now covers all fifteen.
3. **A total that does not match stops the arm being scored at all.** Not mentioned alongside a score already formed. Reconcile first; a short total is unproven-truncated, scored in neither direction.

Those three, plus controls-must-not-move and nothing-else-may-redden, are **five preventing controls**. Each was validated by **executing** it against a synthetic arm built to trip it - a truncated arm, a crash-red arm, a stayed-green arm - confirming a distinct non-zero verdict, and against a clean arm to confirm it does not fire on the happy path. Observed exit codes: perfect `0`, truncated `2`, crash `3`, stayed-green `3`. A detector that has only ever been read is another printout.

### Why all three arms are the full solution, and not the Gateway assembly per arm

The obvious answer is the **reference graph**: of the seven test projects, only `CcDirector.Gateway.Tests` references `CcDirector.Gateway`, so nothing else can even load the mutated code. **That answer is wrong**, and it is worth recording because it is the answer anyone would give - including me, until I checked.

`CcDirector.Core.Tests` contains **ten architecture fitness functions that read production source under `src/` as text**, with no assembly reference to the Gateway at all. One of them, `NoCrossMachineLoopbackGuardTests`, carries a **path-keyed allowlist naming `src/CcDirector.Gateway/Api/MachineEndpoints.cs` explicitly** - the exact file this pull request mutates - and by its own documentation it fails on a **stale** entry, meaning a file that no longer contains the literal. **So a source edit in the Gateway can redden a project that does not reference the Gateway.** The coupling is textual, over the filesystem, and the reference graph is structurally blind to it.

For **this** mutation it is settled, and settled **for free by text** rather than by spending a box slot: the deleted block is five lines of routing code containing **zero** loopback literals, and `MachineEndpoints.cs` still contains **four** afterwards, so the allowlist entry stays honest and that guard cannot flip. Verified by applying the mutation, grepping, and restoring.

**The deciding argument is the reconciliation gate, not the coupling.** Scoping the arm narrower than the baseline **breaks the primary truncation detector**: `passed + failed + skipped` would be compared against a total from a *different population* and could never match. Preserving the gate under a scoped arm requires a **scoped baseline too - a fourth run**, on a contended serial box, to save part of one.

The per-arm saving **scales with the number of arms** while the extra full run is **fixed**, so the scoping that is correct for an eight-arm proof **inverts at three arms**. Measured against roughly 17 minutes for a full solution and 10 for the Gateway suite: three full runs is about 50 minutes, and every scoped variant that keeps the gate intact lands between 47 and 56. **There is no saving here to buy, only a detector to lose** - so the registration stands: three full-solution runs.

Self-host is completely unchanged and is the control.

### DIFF
--- a/src/CcDirector.Gateway.Tests/HostedLauncherMachineDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedLauncherMachineDenyTests.cs
@@ -185,7 +185,33 @@ namespace CcDirector.Gateway.Tests;
 /// THE RUN PLAN IS EXACTLY THREE FULL-SUITE RUNS, and the path actually taken is the path registered:
 /// baseline, arm, restore. There is no slice-then-full escalation and no separate canary gate here - all
 /// three arms are the FULL suite, so no ordering can slip a fourth run in behind the registered three. On a
-/// contended serial box a fourth run spends someone else's slot.
+/// contended serial box a fourth run spends someone else's slot. THAT COUNT WAS OBSERVED BY EXECUTING the
+/// driver against a counting stub, not by reading it: three test invocations, counted. Reading a driver has
+/// been wrong repeatedly tonight; executing it has not.
+///
+/// A FINDING THAT DOES NOT CHANGE THE VERDICT IS NOT A FINDING, IT IS A COMMENT. The first version of this
+/// unit's driver had EIGHT things that would have been described as controls and ZERO that could stop it:
+/// the clean-tree check printed, the mutation diff printed, the build exit codes were masked by pipes, and
+/// the reconciliation and red-classification existed only as prose. Every one of them reported into a void,
+/// because THE LOG IS WHAT NOBODY RE-READS ONCE THE EXIT CODE IS GREEN. Being confidently wrong about which
+/// checks can actually stop you is worse than having fewer checks. So the three questions, answered:
+///
+///   1. A PREDICTED RED THAT ARRIVES AS A CRASH DOES NOT COUNT TOWARD THE FIFTEEN, and the arm is NOT
+///      proven. Crediting a crash to the prediction - so a parse exception satisfies the expectation and
+///      keeps its row out of the not-reddened set - is precisely the laundering the arrival rule exists to
+///      prevent, happening inside the mechanism written to enforce it.
+///   2. A PREDICTED ROW THAT STAYS GREEN CHANGES THE VERDICT TO UNPROVEN. It does not become a line in a
+///      report filed beside a passing score. This is already registered for the four relay rows, where a
+///      green means the deny was indistinguishable from no-launcher-registered; the SAME rule covers all
+///      fifteen.
+///   3. A TOTAL THAT DOES NOT MATCH STOPS THE ARM BEING SCORED AT ALL. It is not mentioned alongside a score
+///      already formed. Reconcile FIRST; a short total is UNPROVEN-TRUNCATED, scored in NEITHER direction.
+///
+/// Those three, plus the controls-must-not-move check and the nothing-else-may-redden check, are five
+/// PREVENTING controls - each one validated by EXECUTING it against a synthetic arm built to trip it (a
+/// truncated arm, a crash-red arm, a stayed-green arm) and confirming it returns a distinct non-zero verdict,
+/// and against a clean arm to confirm it does not fire on the happy path. A detector that has only ever been
+/// read is another printout.
 /// </summary>
 public sealed class HostedLauncherMachineDenyTests : IAsyncLifetime
 {
--- a/src/CcDirector.Gateway.Tests/HostedLauncherMachineDenyTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedLauncherMachineDenyTests.cs
@@ -106,7 +106,7 @@ namespace CcDirector.Gateway.Tests;
 ///       - ALL NINE theory rows.
 ///   HostedLauncherMachineDenyTests.The_launcher_listing_leaks_no_machine_on_hosted
 ///   HostedLauncherMachineDenyTests.The_refusal_is_not_an_empty_launcher_list
-///   HostedLauncherMachineExclusiveGroupTests.A_route_added_to_the_group_later_is_refused_on_hosted_with_no_deny_of_its_own
+///   HostedLauncherMachineExclusiveGroupTests.A_body_bound_route_added_to_either_group_later_is_refused_on_hosted_with_no_binder_run
 ///   HostedLauncherMachineExclusiveGroupTests.A_refused_registration_writes_nothing_into_the_registry
 ///   HostedLauncherMachineExclusiveGroupTests.A_refused_spawn_never_reaches_the_resolver
 ///   HostedLauncherMachineExclusiveGroupTests.A_refused_launch_never_dials_the_launcher
@@ -461,7 +461,8 @@ internal sealed class MachineGroupProbeHost : IAsyncDisposable
     }
 
     public static async Task<MachineGroupProbeHost> StartAsync(
-        Action<HostedDenyGroup>? mapIntoGroup = null,
+        Action<HostedDenyGroup>? mapIntoMachineGroup = null,
+        Action<HostedDenyGroup>? mapIntoLauncherGroup = null,
         Action<IEndpointRouteBuilder>? mapOutsideGroup = null,
         bool withStubLauncher = false)
     {
@@ -506,11 +507,14 @@ internal sealed class MachineGroupProbeHost : IAsyncDisposable
             (directorId, req, ct) => Task.FromResult<(bool, SessionDto?, string?)>(
                 (true, new SessionDto { SessionId = SpawnedSessionId }, null)));
 
-        // The probe routes are /machines paths, so mapIntoGroup targets the MACHINE exclusive group. On hosted
-        // its handler is discarded and the group's catch-all refuses the path; off hosted the handler is mapped
-        // and served - which is exactly the future-route property both directions of the proof turn on.
-        var (_, machineGroup) = MachineEndpoints.Map(app, launchers, spawner, sendLauncherCommand: null);
-        mapIntoGroup?.Invoke(machineGroup);
+        // BOTH exclusive groups are handed back, because the family owns TWO prefixes and the future-route
+        // property has to be proved on EACH - a proof on /machines alone is silent on whether /launchers got
+        // the same coverage. On hosted a handler mapped into either group is discarded and that group's
+        // catch-all refuses the path; off hosted the handler is mapped and served. That is the future-route
+        // property both directions of the proof turn on, per prefix.
+        var (launcherGroup, machineGroup) = MachineEndpoints.Map(app, launchers, spawner, sendLauncherCommand: null);
+        mapIntoMachineGroup?.Invoke(machineGroup);
+        mapIntoLauncherGroup?.Invoke(launcherGroup);
         mapOutsideGroup?.Invoke(app);
 
         await app.StartAsync();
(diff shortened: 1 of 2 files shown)


thefrederiksen:
STANDING: REJECTED
deb154219321c3661fcb151bda9b2b56074b80e1

The hosted denial is not uniform across all request shapes: part of the request pipeline can still answer before the family-wide refusal boundary. The current valid-body coverage cannot prove the stated exact-refusal guarantee.

Security details and the required proof direction were reported privately.

thefrederiksen:
Additional census evidence for the already-ruled single-unit scope:

The launcher family has a second mapped ingress outside `MachineEndpoints`: `/launcher-stream`. Its inbound `LauncherHub.Hello(LauncherStreamHello)` invocation binds/deserializes the hello payload before the hub's machine-binding/re-claim decision, and a successful call populates `LauncherConnectionRegistry`.

The host-wide credential middleware authenticates the SignalR transport before message binding when enabled, but it does not implement the hosted-family refusal. The current diff changes only the HTTP route group, so the hosted launcher family still has a non-HTTP door.

Clearance therefore requires this one unit to close both doors:

- all nine HTTP launcher/machine routes; and
- the launcher hub ingress on hosted, before it can populate launcher connection state.

The proof must show the hosted hub path refuses without writing the connection registry, with an explicit self-host positive control. This is the same family-completeness guarantee, not a separate feature unit. Detailed security-boundary census evidence was reported privately.

PR #1940, opened 2026-07-20
hosted: refuse the shared-vault transcription key provisioner, and pin it

thefrederiksen:
## The reported defect does not hold on hosted today

State this first, because a future reader who finds only the gate will assume there was a live cross-tenant hole in production. There was not.

The transcription-key auto-provisioner writes the key vault with a set-if-absent on sign-in, and revokes-and-deletes that same entry on sign-out. The vault is one `keyvault.json` per box with no tenant dimension. On a multi-tenant Gateway those two operations would be exactly the two harms reported: the first tenant to sign in owns the key and every later tenant silently transcribes on it, spending the first tenant's credits; and any one tenant's entirely ordinary sign-out revokes and deletes the key every other tenant is using.

Neither can execute on hosted. Walked end to end at `f2028de`:

- The hosted Gateway is a **Linux container** (`Dockerfile`, `mcr.microsoft.com/dotnet/aspnet:10.0`).
- `GatewayHost` builds its DevThrottle credential service only on Windows and macOS; Linux logs an explicit null. `GatewayAccountFactory` has no Linux path, and the only non-test construction (`GatewayService`) injects no account.
- The provisioner **and** the sign-in service are both built only inside `if (Account is not null)`, so on hosted neither exists.
- Consequently the startup ensure is skipped by its own null guard, `/account/sign-in-start` and `/account/sign-in-callback` report not-available, and `POST /account/logout` returns at its `account is null` branch **before** the sign-out hook or any vault delete.

On a self-hosted Gateway, where this code does run, there is exactly one account — and there both behaviours are the deliberate safeguards their own documentation describes.

## Why change anything, then

Because the safety is an accident of which platform hosted happens to run on, and nothing anywhere asserts it. Nothing in the provisioner or the logout hook is tenant-aware or hosted-gated. A credential service on hosted is a product direction, not a hypothesis; the day it lands, both harms arm themselves and **nothing goes red**. Whoever wires it will have no way to know they were relying on the platform.

That shape — a property that holds because of how something happens to be constructed, with no assertion that it holds — is what this pull request converts into an asserted property.

## What it does

**The gate.** `TranscriptionKeyAutoProvisioner.CreateUnlessHosted` is now the only way to obtain a provisioner and returns null on hosted; the constructor is private, so there is no second construction route to bypass it. Null is the state hosted is already in for its own unrelated reason, so no new code path is introduced — every caller already handles it. That is deliberately **one** removable line rather than a guard at the top of each of the two methods: with two guards either could be individually removed while the other stayed correct, so each would need its own proof; with one construction route there is nothing to bypass.

The reasoning — including that the gate is redundant **today** and why that is not a reason to delete it — is written at the factory and at the wiring site in `GatewayHost`, not only here. A rule without its reason gets removed by whoever is next and is sure it is redundant, and they will be right about today.

The gate is not the right hosted answer forever. That is a per-tenant credential, the way hosted enrollment already partitions a shared store by prefixing with the authenticated tenant. Until that exists, refusing is the answer that cannot mis-bill anyone.

**The tests**, which are the point.

- Two tenants on one shared vault, on a provisioner explicitly told it is not hosted: the second inherits the first's key without minting at all, then the second's sign-out revokes the *first's* key id and deletes the shared entry. This asserts the defect rather than describing it, and it is the **destructibility control** for the hosted arms — without it, "the other tenant's key survived" would be satisfied by a revoke that never destroys anything.
- The same two tenants, hosted: neither can obtain a provisioner, nothing mints, and the shared vault is never written — so there is no key to inherit and none to destroy.
- The **wiring**, pinned separately against the real `CC_GATEWAY_HOSTED` signal on a real `GatewayHost`, with a credential service **present** and injected. That matters: it models the day hosted has one, and it means the test is not passing for the incidental reason hosted is safe now. The self-host arm is the control — same construction, same credential service, only the hosted flag differs.

The nine existing provisioner tests now state `isHosted` explicitly, so they assert on a stated value instead of on whatever the runner happens to leave in the environment.

## Proof

Continuous integration runs on `windows-latest`, which is the platform where the wiring test genuinely fires: with a credential service present and hosted set, the pre-change code constructs a provisioner, so that test is red without the gate.

Revert-prove: delete the hosted branch from `CreateUnlessHosted` and the three hosted assertions go red by name, while the two-tenant harm demonstration and both self-host controls stay green — those are what must not move.

### DIFF
--- a/src/CcDirector.Gateway.Tests/HostedTranscriptionKeyProvisioningGateTests.cs
+++ b/src/CcDirector.Gateway.Tests/HostedTranscriptionKeyProvisioningGateTests.cs
@@ -1,0 +1,260 @@
+using CcDirector.Core;
+using CcDirector.Core.Account;
+using CcDirector.Core.Configuration;
+using CcDirector.Gateway.Account;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// Pins that a HOSTED Gateway cannot run either of the two key-vault operations in
+/// <see cref="TranscriptionKeyAutoProvisioner"/>.
+///
+/// THE REPORTED DEFECT DOES NOT HOLD ON HOSTED TODAY, AND THAT IS THE POINT OF THIS FILE. A future reader
+/// who finds only the gate will assume there was a live cross-tenant hole in production. There was not. As
+/// deployed, the hosted Gateway is a Linux container, <c>GatewayHost</c> builds its DevThrottle credential
+/// service only on Windows and macOS, and the provisioner and the sign-in service are both built only when
+/// that credential service exists - so on hosted no provisioner is ever constructed, no sign-in flow is ever
+/// mapped, and <c>POST /account/logout</c> short-circuits at "no credential service" before it reaches the
+/// sign-out hook. Neither the write nor the delete can execute. This work was done because that safety was
+/// UNASSERTED, not because it was absent.
+///
+/// SAFE ONLY BY WIRING is exactly the problem. The property held because of which platform hosted happens
+/// to run on, and nothing anywhere said so. The day hosted gains a credential service - a product
+/// direction, not a hypothesis - both harms below arm themselves and NOTHING GOES RED:
+///
+///   - BILLING INTEGRITY. <c>EnsureAsync</c> writes the shared, tenant-less key vault with SET-IF-ABSENT, so
+///     the first tenant to sign in owns the key and every later tenant silently transcribes on it, spending
+///     the first tenant's credits. Nobody is refused; nothing errors.
+///   - AVAILABILITY. <c>RevokeMintedKeyAsync</c> revokes and deletes that same shared entry, so one tenant's
+///     entirely ordinary sign-out breaks transcription for every other tenant on the box.
+///
+/// <see cref="Two_tenants_on_one_shared_vault_is_a_real_harm_when_the_provisioner_exists"/> DEMONSTRATES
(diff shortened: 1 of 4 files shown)


thefrederiksen:
STANDING: REJECTED
a9b880f2026a3d230b439e354143da1bd187caea

The claimed sole construction boundary is still caller-selectable: a future production caller can request the self-host branch without changing or removing the gate. That means the exact extension this PR is meant to make safe can bypass the boundary through its supported signature, while the current wiring tests remain green.

Bind the production factory directly to the independent deployment signal and expose no production-callable override. Keep test control of mode outside that security signature, then retain the real-wiring hosted/self-host controls and revert attribution. Security details were reported privately.

thefrederiksen:
Superseded by #1904 (merged). On main, the TranscriptionKeyAutoProvisioner already refuses every dangerous operation on hosted at the method level: EnsureAsync returns before the vault read/mint/SetIfAbsent/Set, RevokeMintedKeyAsync returns before the vault read/remote-revoke/deletes, and GatewayHost startup seeding refuses before touching the shared vault - each reading the production hosted identity directly. #1940's only substantive delta is a construction-level no-instance invariant, which is defense-in-depth over an already-closed gap. Moreover #1940's gate is a caller-supplied nullable isHosted argument (isHosted:false creates an operational instance even when the process is hosted) and it removes #1904's method-level production-signal guards - so it is weaker, not stronger, than what is on main. Closing rather than merging. Analysis: docs/reviews/production-readiness/mission/validate-1940.md.

PR #1942, opened 2026-07-20
fix(gateway): tenant scoping for Director removal

thefrederiksen:
## What this changes

The Gateway's Director registry keys entries by `(tenant, director id)`. The removal event it raises,
`DirectorRegistry.OnDirectorRemoved`, was typed `Action<string>` and so carried only the identifier - it
dropped the tenant that the removal path already had resolved in its hand. Every subscriber inherited that
loss, and the roster cache's `Forget` had nothing to scope by.

This widens the event to carry the owner and scopes the forget to it.

- New `DirectorRemoval(TenantId Tenant, string DirectorId)` is the event payload; `OnDirectorRemoved` is
  `Action<DirectorRemoval>`. Every raise site already held the resolved key and now passes it.
- `FleetRosterCache.Forget` **requires** the tenant and clears exactly the one `(tenant, director)` entry.
- The bare-identifier overload is **deleted rather than kept and made to filter internally**, so a call that
  does not name an owner is a compile error rather than a runtime judgement. Chosen over an internal filter
  because a filter can be dropped by a later editor and a missing parameter cannot.
- The other subscribers read `removal.DirectorId` and are behaviourally unchanged.

The general lesson, which is why the type moved rather than a check being added: a signature that cannot
represent what the caller knows destroys that knowledge silently, and every consumer downstream inherits the
loss with no place to put a correction.

## Noted at their sites, not addressed here

- `FleetSessionNumberAllocator` records assignments against a bare director identifier, so it cannot scope its
  release even now that the event carries the tenant. Partitioning it is its own unit of work; the impact is a
  duplicated rail number, not data loss or disclosure. Commented at the method.
- `GET /events` is the fleet-global event feed and has no tenant of its own to filter against; it is on the
  existing list of not-yet-tenant-aware routes. Commented at the site.
- Whether the Director identifier should be namespaced or minted server-side at ingestion is a larger change
  belonging to the census remedy work. This fix is complete without it: with a composite registry key and a
  tenant-scoped forget, an identifier reused across tenants is contained within its own tenant at both the
  write and the removal.

## Proof

Three tests in `DirectorRemovalTenantScopeTests`, each asserting the guarantee **and a destructibility control
beside it** - that the same call really does clear the owner's own entry - because "the other tenant's entry
survived" is also satisfied by a forget that removes nothing at all.

`DirectorRegistry.SweepStale` was made `internal` (the project already has `InternalsVisibleTo`) so the
end-to-end test drives the real removal path through the real event and the real subscriber wiring, rather
than the tenant-ambiguous graceful `Remove` overload.

**Refusal branch watched firing:** deleting the bare-identifier overload produced `error CS7036` at the one
pre-existing call site on the first build. The compiler refusing an unscoped forget was observed, not assumed.

**Revert-proof.** Predicted reds were registered before either arm was pushed; both arms run the full suite
on continuous integration. Arm one restores the pre-fix `Forget` body; arm two fills the widened signature in
with a fixed owner instead of the removed entry's own. Results are recorded with the mission.

Detailed mechanism, call path and conditions are recorded in the private tracker.

### DIFF
--- a/src/CcDirector.Gateway.Tests/DirectorRemovalTenantScopeTests.cs
+++ b/src/CcDirector.Gateway.Tests/DirectorRemovalTenantScopeTests.cs
@@ -1,0 +1,151 @@
+using System;
+using System.IO;
+using CcDirector.Core.Tenancy;
+using CcDirector.Gateway.Contracts;
+using CcDirector.Gateway.Discovery;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// Removing one tenant's Director must not destroy another tenant's cached roster.
+///
+/// WHAT THESE PIN. The registry keys entries by (tenant, id), and an id is unique only WITHIN a tenant, so
+/// the same id may be held in two tenants at once. The registry's removal was already tenant-correct, but
+/// <c>OnDirectorRemoved</c> was typed <c>Action&lt;string&gt;</c>, which dropped the tenant the removal path
+/// had in hand; the roster cache's forget was correspondingly unscoped and cleared every partition whose id
+/// matched case-insensitively. A removal in one tenant could therefore clear a cached roster in another,
+/// dropping the last-known-good sessions the grace window existed to keep serving.
+///
+/// WHAT EACH TEST HOLDS DOWN, and why both halves are needed. Survival alone proves nothing - a forget that
+/// removed nothing at all would also leave the other tenant intact - so every test here asserts the
+/// DESTRUCTIBILITY CONTROL beside it: the owner's own entry really is cleared by the same call.
+///
+/// REVERT-PROOF. Make <see cref="FleetRosterCache.Forget"/> ignore its tenant argument and enumerate every
+/// partition again (the pre-fix body) and both survival assertions go RED while both destructibility
+/// controls stay GREEN. Separately, make <c>DirectorRegistry.RaiseDirectorRemoved</c> stamp a fixed tenant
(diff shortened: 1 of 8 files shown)


thefrederiksen:
STANDING: APPROVED
922360ba4cee5f13276046932b3c69b24a0dd201
REASON-TO-TRUST CHECK: The claimed boundary is that Director removal cannot lose its owner and roster eviction can destruct only the exact (tenant, director) cache entry. Static review confirms every removal raise site passes the resolved DirectorKey into the single event seam, the event payload carries TenantId, FleetRosterCache.Forget requires that tenant and has no bare-id overload, and its implementation performs one composite-key TryRemove.

The proof exercises that property rather than ordinary traffic. On this exact feature head, CI is green and all three focused tests pass. Closed ARM-ONLY #1943 restores a tenant-ignoring forget: exactly the two cross-tenant survival assertions fail while the destructibility controls remain green. Closed ARM-ONLY #1944 stamps the wrong owner at the event seam: exactly the event-owner assertion and the end-to-end owner's destructibility assertion fail, while direct scoped forget remains green. The arm totals reconcile against the feature run (3012 Gateway tests; 3002 passed plus 10 skipped on feature, 3000 passed plus 10 skipped plus 2 expected failures on each arm).

The feature's executable code is identical to the commit tested by both arms; the later head delta is prose-only. Current main's intervening #1934 does not touch this unit's files, the PR is mergeable, and diff-check is clean.

#1943 and #1944 are accepted solely as closed mutation evidence. They must never be merged.

PR #1943, opened 2026-07-20
ARM ONLY - DO NOT MERGE: revert-proof arm one for #1942

thefrederiksen:
Mutation arm for pull request #1942, so the full suite can show which tests detect the regression. Restores the pre-change body of `FleetRosterCache.Forget`. Predicted reds were registered with the mission before this was pushed. Close without merging once read.


thefrederiksen:
Arm read: exactly the two pre-registered reds, both genuine assertion failures, all controls green, run totals reconciled against baseline. Closing unmerged.

PR #1944, opened 2026-07-20
ARM ONLY - DO NOT MERGE: revert-proof arm two for #1942

thefrederiksen:
Second mutation arm for pull request #1942. Fills the widened removal-event signature in with a fixed owner instead of the removed entry's own - the shape of a corrected type that is then populated wrongly. Predicted reds were registered with the mission before this was pushed. Close without merging once read.


thefrederiksen:
Arm read: exactly the two pre-registered reds, both genuine assertion failures - the end-to-end test failed at its destructibility control, which is the point of having one. All controls green, run totals reconciled against baseline. Closing unmerged.

PR #1945, opened 2026-07-20
fix(gateway): require the tenant to construct the voice upload store

thefrederiksen:
Detail: devthrottle_internal#454

## What changed

`VoiceUploadStore` had two public constructors that named no partition - a no-argument one and a root-only one - and both bound the local tenant on the author's behalf. They are replaced by a single constructor that takes the root and the tenant.

The guarantee this buys: **there is no way to construct an upload store without naming the partition it belongs to.** An unscoped store is not something to guard against, enumerate, or remember to check - it cannot be built, and the attempt is a compile error at the call site. That is why this change is small: a structurally impossible state has no bypass set to prove.

Naming the local tenant resolves to exactly the path it always did. Self-host is unchanged, nothing migrates, no behaviour moves. What changed is what an author has to write down, and therefore what a reviewer can see.

Construction and re-scoping now share one partition-computing path, so the canonical-identifier rules that make a tenant id safe as a directory name apply identically however a store is obtained. This builds on the dictation partition work and expects to land after it.

## Base

Cut from `fix/1884-tenant-key-upload-store` rather than around it, so its commits are shared and it merges first.

## What proves it

**The compiler is the primary proof, and it was watched firing.** Removing the tenant-less constructors broke the build at three call sites that used target-typed `new(...)` - sites a text search for the type name does not find. Those are exactly the invisible omissions the change is aimed at, and they became build errors rather than silent objects.

**Four standing tests** (`VoiceUploadStoreRequiredTenantTests`), three of them watched reddening under a deliberate mutation, with the other three of the four staying green as controls each time:

| Property asserted | Mutation | Result |
|---|---|---|
| Every public constructor takes a tenant | re-add a no-argument constructor | RED, assertion, names the offending signature |
| An unresolved tenant is refused by the validity check | make the validity check accept everything | RED, assertion (see below) |
| A named account tenant lands inside that tenant's partition | construct against the base root instead of the partition | RED, assertion |
| Naming the local tenant leaves the root unchanged | (control - pins the no-migration claim) | green throughout |

Every red arrived in the assertion of the test under mutation. None arrived in fixture or schema setup, so no arm was void.

**A correction found by that exercise, and it is the more useful half of it.** The unresolved-tenant test first asserted only the exception type, and it stayed GREEN when the validity check was removed - an unresolved tenant has a null value, which the partition-naming rule refuses anyway because it is not a canonical identifier, so the same exception type arrived from a different guard. The test could not distinguish "this check works" from "this check does not exist". It now asserts the message, and reddens on removal. Second commit in this pull request.

Continuous integration is the run of record for the full suite on this head.

## Covered / not covered

**Covered**
- Every public construction path of `VoiceUploadStore` requires a tenant; a tenant-less one is a compile error, and a re-added convenience overload reddens a standing test at the moment it is written.
- An unresolved tenant (`default(TenantId)`, which bypasses the validating constructor of the struct) is refused at construction rather than read as local.
- Direct construction with an account tenant lands in the same partition the re-scoping path resolves - the constructor is not a second, unpartitioned way in.
- Local construction keeps today's root exactly; nothing on an existing install moves.

**Not covered**
- The voice-turn staging root is still constructed against the local tenant on the hosted Gateway. Unchanged behaviour, and that path writes no delivery record and does no lookup by upload id - but per-tenant scoping of the voice-turn path is separate work and is NOT claimed here. Named as follow-up in devthrottle_internal#454.
- This changes only how a store is OBTAINED. The behaviour of the partition once obtained is the dictation partition work's, proved there.
- Other stores in the Gateway with the same constructor shape are out of scope for this unit; nothing here surveys them.

### DIFF
--- a/src/CcDirector.Gateway.Tests/DictationDesktopLockTests.cs
+++ b/src/CcDirector.Gateway.Tests/DictationDesktopLockTests.cs
@@ -2,6 +2,7 @@ using System.Text.Json;
 using CcDirector.ControlApi;
 using CcDirector.Core.Sessions;
 using CcDirector.Gateway.Contracts;
+using CcDirector.Core.Tenancy;
 using CcDirector.Gateway.Voice;
 using Xunit;
 
@@ -30,7 +31,7 @@ public sealed class DictationDesktopLockTests
         var root = Path.Combine(Path.GetTempPath(), "cc-dictlock-xfmt-" + Guid.NewGuid().ToString("N"));
         try
         {
-            var store = new VoiceUploadStore(root);
+            var store = new VoiceUploadStore(root, TenantId.Local);
             var sid = Guid.NewGuid().ToString();
             var uploadId = Guid.NewGuid().ToString();
             store.Register(uploadId);
@@ -53,7 +54,7 @@ public sealed class DictationDesktopLockTests
         var root = Path.Combine(Path.GetTempPath(), "cc-dictlock-xfmt-" + Guid.NewGuid().ToString("N"));
         try
         {
-            var store = new VoiceUploadStore(root);
+            var store = new VoiceUploadStore(root, TenantId.Local);
             var sid = Guid.NewGuid().ToString();
             var uploadId = Guid.NewGuid().ToString();
             store.Register(uploadId);
--- a/src/CcDirector.Gateway.Tests/DictationOmittedTenantBoundaryTests.cs
+++ b/src/CcDirector.Gateway.Tests/DictationOmittedTenantBoundaryTests.cs
@@ -1,0 +1,524 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Net.Http.Json;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using CcDirector.Core;
+using CcDirector.Core.Tenancy;
+using CcDirector.Gateway.Api;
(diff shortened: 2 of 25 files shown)


thefrederiksen:
REJECT at exact head `030a3c493ec386bd302c60bb5232eae532395204`.

Two reason-to-trust blockers remain:

1. The stated out-of-scope premise is false. This endpoint itself describes the upload as resumable across retries, then uses `Register` on one request, `Exists` / `StagedBytes` / `StoreChunkAsync` on later chunk requests, and `Exists` / `AssembleAsync` / `Delete` on the complete request. Production passes the hosted voice-turn store here bound explicitly to `TenantId.Local`. Requiring a constructor argument makes that choice visible, but it does not support the claims that this path performs no cross-request lookup by upload id or that no reachable caller used the unsafe shape. Per-tenant voice-turn scoping can remain separate work, but this PR's classification and source comments must first state the live boundary honestly. The exact cross-tenant trace is recorded privately.

2. `EveryPublicConstructor_RequiresATenant` checks only that a `TenantId` parameter exists; it does not check that the parameter is required. I mutated the production signature to `TenantId tenant = default` and all four `VoiceUploadStoreRequiredTenantTests` stayed green (4/4), even though `new VoiceUploadStore(root)` then compiles. The standing test must reject an optional/defaulted tenant parameter. If “every public way to construct” is the intended durable guarantee, either cover public factory paths too or narrow the claim to constructors.

Control evidence: the unmutated four-test slice passes 4/4, and CI is green on this head. The mutation was reverted and the review worktree is clean.

PR #1951, opened 2026-07-20
fix(gateway): one shared hosted refusal boundary for every deny family

thefrederiksen:
## What this changes

Every hosted deny family adopts **one shared refusal boundary** instead of each writing its own. A family's routes are mapped through a typed handle obtainable only from this primitive, so a family maps into the guarded group **or it does not compile**. The only per-family configuration is the refusal payload.

On hosted, the family's handler is not mapped and a refusal takes its place, so the refusal answers **uniformly across request shapes** - a valid body, a malformed body, an unexpected media type, a segment that does not satisfy an inline route constraint, a verb the family never mapped, and a route added to the group later with no deny written for it.

Off hosted the primitive maps the real handler and creates no refusal at all. Self-host is unchanged and is the control, proven in **both** declared non-hosted forms, each set explicitly rather than inherited from the test runner.

Because attachment is what performs the substitution, the uniformity property is established **once, here**. An adopter owes **attachment** and its **own body-bound future-route probe**. Gate direction is *not* a per-family arm - the primitive centralises that decision, so it is proven once in this suite. That is a real loss of per-family attribution, recorded as the price of the gate existing in one place rather than six.

## How the refusal pattern is built

The pattern is **parsed by the framework's own parser** and the parameter parts are rebuilt **without their policies**. Literals, separators, parameter names, optionality, catch-alls and default values are preserved; nothing is re-serialised to text.

This replaces a hand-written scan over the pattern text. The scan was correct on the shapes it was tested against and could be silently wrong on others - and silently wrong here means a refusal that does not cover its own route while the family's author believes it does, which is manufactured coverage rather than a visible bug.

**The grammar covered is stated, one test per element.** On a part kind it does not model, the normaliser **throws at startup** rather than passing the pattern through. That arm is currently **unreachable and is documented as unreachable rather than implied to be covered** - the framework's route-part type has an internal constructor, so its three subclasses are the whole hierarchy and no test here can manufacture a fourth. What guards it instead is a canary on the assumption underneath: if a framework upgrade adds a part kind, that test reddens.

## Conflicts are caught at startup, not at request time

**Deduplication is by route SHAPE, not pattern text.** Two patterns differing only in a parameter name are different strings and the same route, so a text-keyed check maps both and produces exactly the tie it exists to prevent.

**The finalised route space is validated once everything is mapped** - the earliest moment these conflicts exist. A refusal tying with another refusal answers 500 on a denied route; a refusal tying with a live route takes an undenied route off the air.

Only **exact ties** are rejected. Refusal patterns are deliberately wide, so they overlap neighbours constantly and precedence resolves it; a validator that rejected overlap would refuse to start on every correct arrangement.

## Two failure directions, not one

Refusing too little is a leak. Refusing too much is an **outage on a route nobody denied**. The proof set carries a **neighbour probe in both directions** - a literal sibling protected by precedence, and a route on a different path shape. An alternative design that matched by path rather than route membership was measured over-refusing a sibling and was rejected for it; the neighbour probe was itself validated by confirming it fails under that rejected design.

## Scope limit on the wrong-verb property - stated rather than claimed

The wrong-verb refusal is **real in a bare probe rig and is NOT observable on the production Gateway**, because the Cockpit fallback claims everything unmatched and no wrong-verb refusal is ever produced there.

**It must not enter any proof run against the production host.** Claiming it there would be claiming something the fallback prevents anyone from seeing - a claim that cannot fail.

It does buy something concrete: a delete on a cached-voice path today falls through to the session catch-all and is proxied to a Director; after adoption it meets the refusal.

## Proof

Eight request shapes, a future route, the multi-verb duplicate path adopting families rely on, both non-hosted forms, a neighbour probe in two directions, the response media type asserted **with its parameters**, HEAD stated as status-and-headers rather than folded into a claim about bodies, and construction-validation of the refusal payload.

**The registered full-suite proof is outstanding and is not claimed by this pull request.** A previous filtered rehearsal attributed cleanly, but a filtered run cannot see whether existing coverage already caught the behaviour, nor collateral elsewhere. The advance prediction, registered before the run: the primitive is not yet adopted by any family, so **no test outside these files should move in any arm**.

### DIFF
--- a/src/CcDirector.Gateway.Tests/Tenancy/HostedRefusalPatternTests.cs
+++ b/src/CcDirector.Gateway.Tests/Tenancy/HostedRefusalPatternTests.cs
@@ -125,27 +125,6 @@ public sealed class HostedRefusalPatternTests
             partKinds);
     }
 
-    /// <summary>
-    /// The shape key is what the DUPLICATE check compares, and this is the case text comparison misses:
-    /// two patterns differing only in a parameter NAME are different strings and the same route. Mapping
-    /// both produces endpoints that tie, and the tie surfaces at request time on the denied route.
-    /// </summary>
-    [Theory]
-    [InlineData("/x/{id}", "/x/{name}", true)]              // same route, different spelling
-    [InlineData("/x/{id:int}", "/x/{name:alpha}", true)]    // policies are not part of the shape either
-    [InlineData("/x/{id}", "/x/{id?}", false)]              // optional does not compete with standard
-    [InlineData("/x/{id}", "/x/{**id}", false)]             // nor does a catch-all
-    [InlineData("/x/{id}", "/y/{id}", false)]               // different literal
-    [InlineData("/x/{id}", "/x/{id}/y", false)]             // different length
-    public void The_shape_key_sees_past_names_and_policies_but_not_past_kind_or_literals(
-        string left, string right, bool sameShape)
-    {
-        var leftKey = HostedRefusalPattern.ShapeKey(HostedRefusalPattern.WithoutPolicies(left, "probe"));
-        var rightKey = HostedRefusalPattern.ShapeKey(HostedRefusalPattern.WithoutPolicies(right, "probe"));
-
-        Assert.Equal(sameShape, leftKey == rightKey);
-    }
-
     private static System.Collections.Generic.IEnumerable<RoutePatternParameterPart> AllParameters(RoutePattern pattern)
         => pattern.PathSegments.SelectMany(s => s.Parts).OfType<RoutePatternParameterPart>();
 
@@ -158,17 +137,17 @@ public sealed class HostedRefusalPatternTests
 }
 
 /// <summary>
-/// The FINALISED route space check: the conflicts that only exist once everything has been mapped, and
-/// which the per-group duplicate check cannot see because it can only see its own group.
+/// The finalised route-space check, in the form it took AFTER review disproved the previous one.
 ///
-/// Two conflicts, in opposite directions, and only one of them is a leak:
-///   - refusal against refusal - a denied route answers 500 instead of refusing;
-///   - refusal against a LIVE route - a route nobody denied stops serving. That is an OUTAGE, and no
-///     leak-shaped test in this repository would ever notice it.
+/// The previous version tried to detect whether two patterns would TIE in the matcher, using a hand-rolled
+/// shape key. Three independent route pairs escaped it and returned HTTP 500 at request time. The reason is
+/// worth keeping: the matcher's ambiguity relation is not reachable from public API, so any check for it is
+/// a MODEL of framework semantics rather than the semantics themselves - and that was the SECOND time this
+/// primitive modelled a framework semantic and got it wrong.
 ///
-/// What is NOT a conflict is the ordinary case, and getting that wrong would refuse to start on every
-/// correct arrangement: refusal patterns are deliberately WIDE, so they overlap neighbours constantly, and
-/// route precedence resolves the overlap. Only an exact tie is rejected.
+/// So this no longer detects ambiguity. It checks the one thing that removes the conditions for it and that
+/// can be computed correctly: an EXCLUSIVE-PREFIX family claims a prefix nothing else may serve under, and
+/// that claim is verified by simple PREFIX CONTAINMENT.
 /// </summary>
 public sealed class HostedRefusalRouteSpaceTests
 {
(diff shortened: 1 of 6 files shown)


thefrederiksen:
STANDING: REJECTED
6719e7553c2b29556460aa04199816fc608d4641

Static review does not clear the shared primitive.

- Route-pattern normalization is handwritten against a richer ASP.NET route grammar. The current tests prove simple inline policies only; they do not establish preservation of optional/default/catch-all semantics, escaped delimiters, nested policy arguments, or constraints carried by the group prefix.
- Refusal deduplication is exact-text only, and the neighbour proof covers one literal-over-parameter precedence case. Structurally equivalent refusal patterns and competing nonliteral neighbours are neither prevented nor startup-validated, leaving request-time ambiguity and over-refusal unclosed.
- The specified overlap/finalization validation is absent. Payload validation is present, but it is not a substitute for validating the finalized route space.
- The exact-response proof asserts the media type without its parameters, and the unqualified “every request shape” claim needs either a documented HEAD exception or corresponding wire-level proof.

The registered full-suite proof is also still outstanding as declared, and the head is behind current `main`. Rework needs parser/model-based normalization, startup validation of the finalized route space (including protected/protected and protected/neighbour competition), bidirectional neighbour coverage beyond the single literal case, exact response assertions, and then the registered serialized full-suite arms on the rebased final head.

thefrederiksen:
STANDING: REJECTED
6719e7553c2b29556460aa04199816fc608d4641

Independent static review found route-template shapes for which the shared primitive does not provide the claimed uniform pre-binding refusal, plus an unproved duplicate-pattern path that adopting multi-verb families rely on. The self-host mode proof also does not yet cover both declared non-hosted forms.

The pull request's registered full-suite proof remains outstanding by its own declaration.

Security and mechanism details were reported privately.

thefrederiksen:
STANDING: REJECTED
`b4ba02c8f18c0231c73ac30a45ccb35df4f4e27d`

The model-based rebuild fixes the local-pattern optional/default loss from the prior head, and the explicit non-hosted forms, multi-verb row, HEAD contract, and exact media type are materially better. The shared primitive still does not establish its advertised boundary.

1. The typed handle does not make unguarded mapping structurally unexpressible. `DenyProbeHost.MapFamily(IEndpointRouteBuilder outer)` retains `outer` alongside `group` and intentionally uses it for neighbours. I changed only the `/echo` receiver from `group.MapPost` to `outer.MapPost(/family/echo, ...)`: the whole test project built with 0 errors, then five hosted assertions reddened (valid body, malformed body, wrong media type, wrong verb, and HEAD). That is a useful canary for this fixture, but it proves the opposite of the structural claim: the bypass compiles and is caught by tests. The one-proof discharge therefore cannot carry to adopters. Either make each production family’s route-mapping method accept only the handle with no unguarded builder in scope, and enforce that shape, or retain attachment/bypass arms for every independently movable production route.

2. Constraints in the group prefix remain outside normalization. `HostedDenyGroup.Map` normalizes only its local `pattern`; `_group` retains the prefix passed to `HostedRouteDeny.Group` (and any ancestor prefix). With the fixture prefix changed to `/family/{scope:int}`, `POST /family/not-an-int/echo` returned the framework 404 with no refusal media type/body. The claimed uniform pre-binding refusal therefore still fails for the prefix/parent-policy case reported on the prior head.

3. `ShapeKey` is not equivalent to the matcher’s ambiguity relation, so both per-group deduplication and final route-space validation have false negatives. I added and executed three independent hosted fixture pairs; each compiled, escaped both checks, and returned HTTP 500 at request time instead of the refusal:

   - `/optional/{id}` plus `/optional/{name?}` for a present segment. The standing test currently asserts these do not compete; the live matcher shows that they do.
   - `/Case` plus `/case`. Literal matching is case-insensitive, while the key/dictionary are ordinal.
   - `/complex/{a}-{b}` plus `/complex/{c}.{d}` at `/complex/left-right.tail`. Both complex segments match at equal precedence despite different separator text.

   This also disproves the validator’s premise that every non-identical overlap is harmlessly resolved by precedence. The final check needs to model the framework’s actual candidate/precedence rules (including endpoint order), not structural text encoded by a custom key.

Evidence: the unmodified focused set passed 47/47 before review and again after restoration. The receiver mutation built cleanly and produced the five named reds. Each route-shape/prefix probe produced the result above; all scratch edits were reverted and the worktree is clean. The registered serialized full-suite arms remain outstanding as the PR states, and `.NET` CI is still pending on this head.

thefrederiksen:
STANDING: REJECTED
`37d09e60771f0f024762f1b98b1e163da7e54768`

Removing the claimed ambiguity validator is the right correction, but the replacement does not remove the conditions that made it necessary.

1. Both modes still synthesize multiple verb-less refusals and therefore still depend on `ShapeKey` being an ambiguity relation. `ExclusiveGroup` calls `MapExclusiveCatchAll` and then returns the ordinary `HostedDenyGroup`; every later `MapGet` / `MapPost` still enters `Map` and adds the specific per-route refusal. It is not “one catch-all” on hosted. I mapped the previously measured standard/optional pair on different verbs and requested a present segment. It returned HTTP 500 in ordinary `Group` mode and again in `ExclusiveGroup` mode. The catch-all does not save the exclusive case because the two more-specific per-route candidates tie ahead of it.

   The new safety premise is also false: those two production routes would *not* already tie before the deny, because GET and POST method metadata distinguishes them. The hosted transformation discards those method constraints, creating the ambiguity. Route mirroring is therefore not identity-preserving, and the incomplete key remains load-bearing despite its corrected disclaimer.

2. Per-route mode does not carry a canary for the future-route coverage it lost. I added an unguarded body-bound `POST /family/unprobed` on `outer` beside the fixture family and ran the complete focused set: 41/41 stayed green. The existing `/future` row proves one known route whose receiver is already guarded; it does not enumerate the family or red when a newly added family route has no refusal. API documentation assigning that obligation to adopters is honest scope, but it is not the claimed standing test.

3. The exclusive-prefix validator is not yet a sound containment check for the API it accepts:

   - Exclusive prefix `/family/{scope:int}` plus live route `/family/{scope}/still-serving` produced no startup exception. Validation compares raw pattern text with `StartsWith`, even though the hosted group normalizes the parameterized prefix as a route model.
   - Two exclusive markers claiming the same `/family` prefix also produced no exception. The validator skips every other `HostedExclusivePrefixMarker`, so duplicate exclusive claims escape; their duplicate root/catch-all endpoints can tie at request time.

   Either restrict exclusive prefixes to validated literal-only patterns and reject duplicate/overlapping claims, or perform containment over the parsed pattern model. Raw-text containment cannot verify parameterized prefixes.

4. The removed claims still have consumers. The PR body still says the handle makes bypass fail compilation, that deduplication is by matcher route shape, and that final route-space validation rejects refusal/refusal and refusal/live ties. The `HostedDenyGroup` class summary repeats structural unexpressibility, and `HostedRefusalPattern.ShapeKey` still says optional parameters do not compete with standard ones. All were disproved and none describes this head. A claim deleted in one comment is not removed while the public proof artifact and adjacent API documentation still rely on it.

Evidence: unmodified focused baseline and restored run both pass 41/41; exact-head CI is green. All scratch probes were reverted and the review worktree is clean. The registered serialized full-suite mutation arms remain outstanding as declared, so evidence clearance is independently held as well.

thefrederiksen:
Superseded by #1980 (mtr/1951-refusal-validate), which carries this primitive plus the complete fail-closed route-space validation.

PR #1952, opened 2026-07-20
Bound voice-turn upload staging with a production, activity-aware age sweep

thefrederiksen:
## What this changes

Voice-turn upload staging is deleted only on the success path, so every upload that ends any other way - a size refusal, a dropped connection, an assembly that never completed, a caller that walked away - kept its recorded audio on disk for the life of the Gateway. The age sweep that would remove it (`VoiceUploadStore.SweepAbandoned`) existed and had a direct unit test, but nothing in production ever called it. On the hosted multi-tenant Gateway that is recorded speech from every account accumulating in one shared directory with no retention bound and no deletion path. Two comments (in `VoiceUploadStore` and `GatewayHost`) asserted the sweep already ran, which is how it went unnoticed.

This wires the sweep into production and makes it safe:

**1. Runs on a timer, in production.** `GatewayHost.StartAsync` starts a voice-turn upload sweep next to the existing periodic sweeps - created in `StartAsync`, disposed in `StopAsync`, logged the same way. It runs on every deployment because the audio it removes is recorded speech and no deployment should keep it forever.

**2. Four-hour maximum age, justified from purpose.** A staging directory exists only between a register and that same upload's own complete: seconds to a couple of minutes normally, tens of minutes at the pathological extreme of a phone retrying chunks over a failing link. Four hours is an order of magnitude past that, so nothing still genuinely trying is cut off, while abandoned audio has a hard ceiling in hours.

**3. Liveness is judged by an explicit activity touch, not by an incidental byte write.** Every successful operation - a register (including a resume of an existing id), an idempotent chunk retry, a real chunk write, and an assemble - refreshes the upload's last-activity signal through one `EnsureFreshStaging` touch. Without this a client that resumed an upload (re-register, or re-send an identical chunk that writes no byte) kept its stale timestamp and the sweep would delete it mid-use. The sweep judges by that explicit signal.

**4. The activity touch and the sweep coordinate through the per-upload gate, closing the resume race.** The touch and the sweep's age-check-and-delete run under the same in-process per-upload lock (keyed by the canonicalized staging directory), and the age is re-read inside that lock immediately before the delete. So a concurrent resume is serialized against the delete for that one upload: either its refresh commits first and the sweep then reads a fresh timestamp and keeps the directory, or the delete commits first and the resume's own register re-creates the staging fresh. There is no check-old-then-touched-by-resume-then-deleted-anyway window, and closing it does not depend on the window being small.

**5. Deletes only what it can positively identify - a canonical upload id - not a deny-list.** The sweep removes only a directory whose name IS a canonical 32-hex lowercase upload id (the exact form the store writes) and keeps everything else, aged or not: an unknown name, an almost-canonical one, an upper-cased alias, the per-tenant partition container, or any future sibling directory. A single hard-coded exclusion would have recursively deleted every unanticipated directory; positive validation is the safe direction for an unrecoverable delete over recorded audio, and it is also what keeps the sweep safe before and after the sibling tenant-partition change (pull request #1900) lands.

## Proof

Build is confirmed at zero errors as its own step before every test run (a build failure here would run the previous binary and report a false green). Guards are proven by removing them and watching the exact tests go red, then restoring.

- **Wiring.** A test boots a real `GatewayHost`, stages an abandoned upload on disk, and waits for the running host to remove it on its own - it fails if the timer wiring is removed (`the running Gateway never swept the abandoned staging`), which the pre-existing direct-call unit test could not detect. It also asserts a still-in-flight upload and the partition container survive.
- **Activity touch, per operation, bypassability-clean.** Three independent tests each exercise ONLY one operation on an aged upload - re-register alone, identical-chunk-retry alone, assemble alone. Removing each operation's touch in isolation reddens exactly that one test and no other: the register touch reddens only the re-register test, the chunk touch only the retry test, the assemble touch only the assemble test.
- **Positive canonical-id validation.** A test ages an unknown name, a hyphenated GUID, an upper-cased 32-hex name, and the partition container, each with a sentinel file, and asserts the sweep keeps all of them. Removing the positive validation reddens it (aged non-canonical directories deleted).
- **Full Gateway suite:** green (see the latest run recorded in the review thread).

## Not proven

Only in-process test evidence - no live hosted Gateway was exercised. Behaviour once the real partitioned store (pull request #1900) lands is covered by the canonical-id and partition-container tests using the same container name and helper as that branch, but the two have not yet been merged together.

### DIFF
--- a/src/CcDirector.Core/Storage/CcStorage.cs
+++ b/src/CcDirector.Core/Storage/CcStorage.cs
@@ -138,7 +138,9 @@ public static class CcStorage
     /// <summary>
     /// Resumable upload staging for the Gateway voice-turn front door: base/voice-turn-uploads/&lt;uploadId&gt;/.
     /// Each chunk lands here as it arrives (SHA-checked, idempotent) and the dir is deleted once the
-    /// chunks are assembled and the turn has been started. Owned by the Gateway.
+    /// chunks are assembled and the turn has been started. An upload that never reaches that point is
+    /// removed by the Gateway's voice-turn upload sweep once nothing has written to it for hours, so
+    /// abandoned recorded audio has a retention ceiling. Owned by the Gateway.
     /// </summary>
     public static string VoiceTurnUploads() => Ensure(Path.Combine(Base(), "voice-turn-uploads"));
 
@@ -146,8 +148,10 @@ public static class CcStorage
     /// Resumable upload staging for durable dictation (issue #1006): base/dictation-uploads/&lt;uploadId&gt;/.
     /// The mobile app persists the raw audio locally the instant Send is pressed and streams it here in
     /// SHA-checked chunks; once assembled the Gateway transcribes it and injects the turn into the owning
-    /// session itself, so a dead tab or a dropped connection cannot lose a recorded utterance. Abandoned
-    /// staging dirs are swept after ~1 hour. Owned by the Gateway.
+    /// session itself, so a dead tab or a dropped connection cannot lose a recorded utterance. There is
+    /// deliberately NO age sweep here (unlike the voice-turn staging above): each upload id carries a durable
+    /// delivery record, its chunks are retained while it is undelivered, and the record is retired only by
+    /// the client's acknowledgment. Owned by the Gateway.
     /// </summary>
     public static string DictationUploads() => Ensure(Path.Combine(Base(), "dictation-uploads"));
 
--- a/src/CcDirector.Gateway.Tests/VoiceTurnUploadSweepWiringTests.cs
+++ b/src/CcDirector.Gateway.Tests/VoiceTurnUploadSweepWiringTests.cs
@@ -1,0 +1,121 @@
+using System.Net;
+using CcDirector.Core.Storage;
+using CcDirector.Gateway.Voice;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// The voice-turn upload staging is bounded IN PRODUCTION, not merely bounded in principle.
+///
+/// WHY THIS TEST EXISTS AND WHY IT BOOTS A REAL GATEWAY. <see cref="VoiceUploadStore.SweepAbandoned"/> has
+/// had a direct unit test since it was written, and for that whole time nothing in production called it: the
+/// staging directory for a voice turn is deleted on the SUCCESS path only, so every refused, dropped or
+/// never-completed upload kept its recorded audio forever. A test that calls the sweep itself cannot notice
+/// that - it passes with the timer present and passes with the timer gone. So this test never calls the
+/// sweep. It stages an abandoned upload on disk, starts a real <see cref="GatewayHost"/>, and waits for the
+/// Gateway to remove that upload on its own. The only thing that can make it pass is the sweep actually
+/// running in the host, which is the fact that was missing.
+///
+/// The schedule is compressed through <see cref="GatewayHost.VoiceTurnUploadSweepScheduleForTests"/>; the
+/// AGE cut-off is production's real one, and the staged upload is aged into the past on disk, so the test
+/// exercises the deployed retention rule rather than a shortened copy of it.
+/// </summary>
+[Collection("DirectorRoot")]
+public sealed class VoiceTurnUploadSweepWiringTests : IAsyncLifetime
+{
+    private const string GatewayToken = "test-token";
+
+    private GatewayHost? _gateway;
+    private string? _originalRoot;
+
+    private readonly string _storageRoot =
+        Path.Combine(Path.GetTempPath(), "cc-sweep-storage-" + Guid.NewGuid().ToString("N"));
+    private readonly string _instancesDir =
+        Path.Combine(Path.GetTempPath(), "cc-sweep-instances-" + Guid.NewGuid().ToString("N"));
+
+    public Task InitializeAsync()
+    {
+        _originalRoot = Environment.GetEnvironmentVariable("CC_DIRECTOR_ROOT");
+        Environment.SetEnvironmentVariable("CC_DIRECTOR_ROOT", _storageRoot);
+        return Task.CompletedTask;
+    }
+
+    public async Task DisposeAsync()
+    {
+        GatewayHost.VoiceTurnUploadSweepScheduleForTests = null;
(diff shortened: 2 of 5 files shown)


thefrederiksen:
REJECTED at exact head 6df03e6a7117c6a1ba77fb8e382081860e677ee1.

The production wiring canary is sound: it passes at this head, and removing the timer registration makes it fail after the staged directory remains present. The sweep it activates is not yet safe enough to run in production, however:

1. The stated idle-age guarantee is false for the resumable/idempotent path. Registering an existing upload does not touch its directory, and StoreChunkAsync returns without a write when the same chunk is retried. I staged a chunk, aged the upload directory nine hours, then performed exactly those two successful resume operations and ran the four-hour sweep. The sweep deleted the now-active upload. That contradicts the PR's claim that a genuinely trying client cannot be cut off and that arriving chunks reset the clock. Activity needs an explicit last-activity touch/marker (including successful idempotent operations), and the sweep needs coordination with active upload operations so it cannot race a resumed request.

2. SweepAbandoned recursively deletes every aged directory under the root except the one literal tenants name, although the implementation itself states that an upload directory is a canonical 32-hex identifier. I created an aged future-partition directory with a sentinel file; the sweep deleted it recursively. Wiring this previously dormant method makes that broad deletion live on every Gateway. The deletion boundary should positively admit only canonical upload-directory names, with a regression test showing an aged unknown/reserved directory survives; one hard-coded exclusion is not a safe ownership check.

Focused evidence after restoring the exact head: VoiceUploadStoreTests plus VoiceTurnUploadSweepWiringTests pass 31/31, and the worktree is clean. Two added scratch regressions fail independently with the deletions above. Please fix both before clearance.

thefrederiksen:
## Revision addressing the two review findings (head 5f6393f8)

**Finding 1 - idle-age was false on the resume path.** Re-registering an existing upload id did not touch the staging, and an idempotent chunk returned without writing a byte, so a genuinely-active client that resumed kept its stale timestamp and the four-hour sweep would delete it mid-use. Fixed: every successful operation - register (including a resume of an existing id), an idempotent chunk, a real chunk write, and an assemble - now refreshes an explicit last-activity signal through a single `EnsureFreshStaging` touch, and the sweep judges liveness by that signal, not by whether a byte happened to land on disk.

Race prevention, stated explicitly: the activity refresh and the sweep's age-check-and-delete run under the **same per-upload gate** (`WithRecordLock`, keyed by the canonicalized staging directory, refcounted, in-process), and the age is **re-read inside that gate immediately before the delete**. So a concurrent resume is serialized against the delete for that one upload: either its refresh commits first and the sweep then reads a fresh timestamp and keeps the directory, or the delete commits first and the resume's own register re-creates the staging fresh. There is no check-old-then-touched-by-a-resume-then-deleted-anyway window - the gate closes it, and closing it does not depend on the window being small.

**Finding 2 - the sweep was a deny-list.** It deleted every aged directory except the literal name `tenants`, which would recursively delete any unanticipated directory including a future-partition directory with real data. Fixed: it now **positively admits** only a directory whose name IS a canonical 32-hex lowercase upload id (the exact form this store writes) and keeps everything else, aged or not - the safe direction for an unrecoverable delete over recorded audio.

**Red-first evidence** (both tests run against the previous head 6df03e6a, then against the fix):
- A resumed upload is not swept - stage a chunk, age the directory nine hours, perform the two successful resume operations, run a four-hour sweep, assert it survives. Previously red: `Expected: 0, Actual: 1` (the resumed upload was deleted). After the fix: green.
- Aged non-canonical directories survive - unknown name, hyphenated GUID, upper-cased 32-hex, and the partition container, each aged and carrying a sentinel file. Previously red: `Expected: 0, Actual: 3` (three of four deleted). After the fix: green.

Build was confirmed at zero errors as its own step before each test run. Full Gateway suite on the new head: **3006 passed, 0 failed, 10 skipped**.

thefrederiksen:
REJECTED at exact head 5f6393f89ee0d88ee4d2fe47e7f0b4127bc6879c.

The runtime response to both findings looks correct, and the allow-list regression is a real canary: removing IsCanonicalUploadDirName admission makes the reserved-directory test fail with four deletions. The restored focused store plus live-host suite passes 33/33.

The required live-resume regression does not yet pin either activity guard, however. SweepAbandoned_DoesNotDeleteAResumedUpload_EvenWhenItWasAgedFirst performs BOTH re-register and identical chunk retry before sweeping. I removed only the EnsureFreshStaging call from Register while leaving the idempotent-chunk touch intact: the test stayed green. I restored that and removed only the StoreChunkAsync touch while leaving re-register intact: the test also stayed green. Each operation masks the other, so either independently necessary path can regress and reintroduce deletion for a client that performs only that operation.

Please split this into independent aged-upload regressions: re-register alone must refresh activity, and an identical StoreChunkAsync retry alone must refresh activity. If AssembleAsync remains part of the documented every-successful-operation guarantee, pin that independently as well or narrow the claim.

The PR body also still describes a literal tenants-directory skip and the old skip mutation/count rather than this head's canonical-name allow-list and activity regressions. Update the merge record to match the implementation and proof. The .NET check is still pending at review time.

thefrederiksen:
PASS at exact head 1ce79282a4b0661dae4336757e82b66fd9ccc07d.

Both original blockers and the response-proof gap are cleared:

- The production timer wiring canary carries forward: the live-host test passes with the timer and reddens when timer registration is removed.
- Activity is explicit and coordinated with the sweep. The restored focused store plus live-host suite passes 35/35.
- The three activity proof units are now independent. Removing only Register's touch reddens only SweepAbandoned_ReRegisterAlone_KeepsAnAgedUpload (Expected 0, Actual 1). Removing only StoreChunkAsync's touch reddens only SweepAbandoned_IdenticalChunkRetryAlone_KeepsAnAgedUpload with the same deletion. Removing only AssembleAsync's touch reddens only SweepAbandoned_AssembleAlone_KeepsAnAgedUpload with the same deletion. No operation can mask another now.
- The destructive boundary is a positive canonical 32-lowercase-hex admission. Removing it reddens the non-canonical-directory regression with four recursive deletions; restored, unknown, hyphenated, uppercase, and partition-container directories survive.
- The PR body now matches the activity-touch, per-upload gate, allow-list implementation, and per-operation mutation proof.

The worktree is clean at the exact head. Both CI jobs are green, including the full .NET build and test job. Cleared for merge.

PR #1953, opened 2026-07-20
feat(gateway): hosted human account sign-in - the mint boundary (backend, #474)

thefrederiksen:
## What this is

The gateway-side backend for hosted human account sign-in (#474): a human on the browser cockpit or the phone \`/m\` signs into their DevThrottle account and receives a **tenant-scoped \`cc-gateway-token\`** on the hosted gateway - mirroring what \`POST /devices/enroll-hosted\` already does for a Director.

**Backend and tests only. No UI.** Ships **dormant and safe**: hosted-only, entitlement-gated, no UI to drive it yet. The UI is a thin follow-on on the existing shared sign-in screen.

## Scope: one reached entry point

The hosted human path is **\`POST /m/enroll\`** on hosted - the reached path for both the cockpit and the phone, which both return through the existing \`/device-callback\` flow. On an \`IsHosted\` branch, \`/m/enroll\` reads the account access token from the Bearer header (mirroring enroll-hosted), runs the one mint, sets the cookie, returns the key. Self-host (device key in body) is unchanged.

A hosted \`/account/sign-in-callback\` branch was initially built and has been **removed**: that flow's *start* (\`/account/sign-in-start\`) reports "not available" on hosted (its sign-in service is self-host only), so the callback is unreachable on hosted by construction - one half of a flow whose other half is disabled there. Shipping it would be a built-but-unreached auth surface. The self-host \`/account/sign-in-callback\` path is untouched as the control (empty diff against origin/main).

## The security spine - one validated mint path

The hosted entry point runs the **same** \`HostedEnrollmentEndpoint.Enroll\` - same token validator, entitlement gate, tenant mint, tenant-hashed device id. Tenant-scoping and 401-never-mint are **inherited** from that already-proven function. \`Enroll\` is **unchanged** (its tests are the no-drift proof). The \`cc-gateway-token\` cookie is set through one helper. \`GatewayHostedMode.IsHosted\` is read **directly** to select the hosted path, and the hosted dependencies are **required when hosted** - the endpoint throws at startup if hosted mode is active without them, so a hosted deployment cannot fail open to self-host through wiring drift.

## Proof (all mutations watched failing on purpose)

- **A refusal persists nothing:** every refusal case in \`HostedMobileAccountEnrollTests\` snapshots the device registry (live count and the persisted store bytes) before the call and asserts it identical after. A mutation that persists a key on the refusal branch reddens these tests; status, cookie, and tenant-mapping guards were already real.
- **The hosted decision is direct and non-optional:** a test asserts the endpoint refuses to start when hosted mode is active without its dependencies; deleting that guard reddens it.
- **The self-host control is untouched:** \`AccountSignInCallbackEndpoint.cs\` diffs empty against origin/main and its self-host tests stay green.
- \`Enroll\` unchanged; the account token is discarded after mint (fragment handback, no query/log/storage); single mint path by construction; cookie options identical to the self-host login cookie today.

Re-reviewed adversarially with all three mutations required to redden before passing.


thefrederiksen:
REJECT — I reviewed pull request 1953 at exact head ae00dba01cfc648f4cf26bbfc08b13c72b490adb.

Blocking findings:

1. The refusal proof does not actually prove that no device key was minted. HostedAccountSignInCallbackTests.cs lines 208 through 210 and HostedMobileAccountEnrollTests.cs lines 196 through 198 assert the unauthorized status, absence of the gateway cookie, and absence of a tenant mapping. They never assert that the device registry stayed unchanged. I introduced a test-only mutation that persistently created a new device key on the invalid-token refusal path while still returning unauthorized and setting no cookie or tenant mapping. Both forged-token tests remained green: two tests run, two passed. A regression can therefore mint and persist a key for a forged token while the claimed “nothing minted” proof stays green. Add an explicit unchanged-device-registry assertion to every refusal case in both endpoint suites. Cookie-on-unauthorized and tenant-mint mutations did make the tests fail, so those two guards are real; the device-key guard is the missing part.

2. The hosted versus self-host control is optional wiring, contrary to the required direct guard. AccountSignInCallbackEndpoint.cs line 102 and MobileEnrollmentEndpoint.cs line 34 accept HostedEnrollDependencies as an optional argument defaulting to null; lines 116 and 176 in the callback and line 50 in mobile select hosted behavior from that nullable argument. GatewayHost.cs line 1826 currently derives the bundle from GatewayHostedMode.IsHosted and passes it at lines 2054 and 2161, but omitting either optional argument still compiles and silently routes a hosted deployment through self-host behavior instead of the hosted mint. Make the hosted-mode decision direct and non-optional so this cannot fail through wiring drift.

Checks that passed on the clean head:

- Both hosted entry points contain exactly one call to HostedEnrollmentEndpoint.Enroll and no calls to ValidateForAuthorization, MintOrLookupBySubject, RegisterIfAbsent, or SetAccountBinding. The raw browser device identifier flows into Enroll, where the unchanged tenant hash namespaces it.
- HostedEnrollmentEndpoint.cs changes only by replacing its private bearer reader with BearerToken.Read; its validation, entitlement, tenant mint, device registration, binding, and statuses are unchanged.
- The gateway test project built with zero warnings and zero errors. Sixty-one targeted hosted enrollment, entitlement, callback, mobile, and self-host control tests passed with an explicit successful summary. The core test project also built with zero warnings and zero errors, and all eleven credential handback tests passed.
- The hosted handback keeps the account access token in the fragment-to-body handback: no query read on the hosted branch, no logging, no browser storage, and no refresh-token persistence. The server discards the account token after enrollment.
- GatewayTokenCookie and the untouched self-host login cookie currently use identical options: HttpOnly, SameSite Lax, thirty-day expiry, and IsEssential.
- A cookie set on an unauthorized response made each corresponding refusal test fail, and a tenant mint on an unauthorized response made both tests fail. The clean forged-token controls passed again after restoring the exact head.

thefrederiksen:
PASS — I re-reviewed pull request 1953 at exact head f727654a29244c664fd1029871f95a4685e7f0ac. The updated pull request body matches the descoped head.

The three required checkpoints passed adversarial review:

1. A refusal now proves the device registry is unchanged in memory and on disk. I inserted a persistent device-key creation on the forged-token refusal path. ForgedHs256Bearer_401_NoCookie_RegistryUnchanged failed at HostedMobileAccountEnrollTests.cs line 246, showing an expected snapshot of zero devices and no persisted contents against an actual snapshot of one device and persisted contents. The mutation was restored.

2. Hosted selection is read directly from GatewayHostedMode.IsHosted, and hosted dependencies are required. I deleted the map-time dependency guard. HostedMode_WithoutDependencies_RefusesToStart failed at HostedMobileAccountEnrollTests.cs line 376 because no InvalidOperationException was thrown. The mutation was restored.

3. The self-host callback control is byte-untouched: AccountSignInCallbackEndpoint.cs and CredentialHandbackPage.cs both have empty diffs against origin/main, and HostedAccountSignInCallbackTests is absent. I also mutated the existing self-host query handback condition; Get_Callback_WithQueryTokenPair_SignsIn_Transition_NoTokenEchoed failed at AccountSignInCallbackEndpointTests.cs line 171. The mutation was restored. The callback, sign-in-start, and remote-routing suites are clean and green.

The remaining security checks pass:

- MobileEnrollmentEndpoint contains one call to HostedEnrollmentEndpoint.Enroll and no independent calls to ValidateForAuthorization, MintOrLookupBySubject, RegisterIfAbsent, or SetAccountBinding. The raw device identifier flows through Enroll for tenant hashing.
- HostedEnrollmentEndpoint changes only by folding its bearer reader into BearerToken.Read. Its validation, entitlement, tenant mint, device registration, binding, and status decisions are unchanged, and its existing tests pass.
- Forged, expired, wrong-audience, and missing account tokens return unauthorized, set no cookie, create no tenant, and leave the live and persisted device registry unchanged. Entitlement refusal and failure do the same.
- The account access token is read only from the authorization header, is not logged, stored, or echoed, and is discarded after the mint. The returned credential is the tenant-scoped device key.
- GatewayTokenCookie and the untouched self-host login cookie currently use identical options: HttpOnly, SameSite Lax, thirty-day expiry, and IsEssential.
- The gateway test project built with zero warnings and zero errors. On the clean restored head, all sixty-six targeted hosted, enrollment, entitlement, mobile self-host, callback, sign-in-start, and remote-routing tests passed. The attribution scan was clean.

No blocking finding remains. I did not merge.

thefrederiksen:
Blocker fix (Codex inspection): cc-gateway-token cookie now carries Secure in hosted mode.

GatewayTokenCookie.Set (the one writer shared by the hosted mint and the self-host cookie-mirror) previously wrote HttpOnly + SameSite=Lax + essential + 30-day expiry but never Secure. Now Secure = GatewayHostedMode.IsHosted: hosted is always HTTPS behind the platform front door so the standing credential is marked Secure; self-host runs over loopback/tailnet HTTP where a Secure cookie would never be sent back, so Secure stays off and the cookie survives HTTP. HttpOnly/SameSite/expiry/essential unchanged.

Tests: hosted /m/enroll success now asserts the Set-Cookie header contains Secure (off the wire); a direct GatewayTokenCookie writer test asserts self-host stays not-Secure so it survives HTTP. Fails-on-purpose verified: dropping the Secure flag reddens the hosted-secure assertions.

Full solution green: all seven test projects pass (Core 3359, Gateway 3046, Avalonia 222, Engine 62, HostedAgent 86, Launcher 27x2, Terminal.Avalonia 21; skips are Postgres-live/OS-gated). Build 0 warnings, 0 errors. @ d7417242

PR #1954, opened 2026-07-20
fix(gateway): read the gateway's own entitlement error (diagnostic) + explicit schema guard

thefrederiksen:
## What this change is

A **behavior-preserving** change to the hosted enrollment gate's single read of the `entitlements` table: a defense-in-depth **guard** plus a **diagnostic**. It is not a fix for any live failure and it changes no SQL emitted by today's model.

1. **`GatewayDbContext` - explicit gateway schema on the entitlements mapping (Postgres only), as a guard.** On the running model this read is **already** `gateway.entitlements`, because `HasDefaultSchema("gateway")` applies the schema to every table. Naming the schema here changes no emitted SQL; what it adds is **independence**: the `entitlements` table is excluded from migrations, so its qualification would otherwise rest entirely on the model-wide default. Pinning it explicitly makes the gateway schema a property of this one read in its own right, so it stays qualified even if the model-wide default is later changed. Provider **conditional**, not a fallback: SQLite is schemaless and takes no schema argument, so the self-host model is byte-for-byte unchanged.

2. **`EntitlementRegistry` - a PostgreSQL diagnostic.** On a PostgreSQL read failure the log now also carries the `SQLSTATE` code and the server's message text (whether the `PostgresException` arrived directly or wrapped by EF). This lets the server's own error be read from the log; it is not a fix. Neither field carries PII for a schema/relation/column error, so the account subject is still never logged. Example line when the table is absent:
   `[EntitlementRegistry] LookupBySubject: PostgreSQL error SqlState=42P01 MessageText=relation "gateway.entitlements" does not exist`

## Proof

Read from EF's own SQL generation, never re-derived in another client:

- **Postgres branch (guard):** the entitlement query under the Npgsql provider generates `FROM gateway.entitlements` and the entity's schema annotation is `gateway`. Fail-on-purpose verified in development: because the explicit mapping is independent of the model-wide default, removing the default alone leaves it green; removing both the explicit schema and the default drops it to a bare `entitlements` and the assertion reddens.
- **SQLite branch (self-host control):** the same query generates the bare table with no schema, and the full self-host enrolment gate suite stays green.
- **Real Postgres, end to end:** the gateway-qualified read resolves a `gateway.entitlements` table when it exists; and when it is absent the read raises a genuine `PostgresException` whose `SqlState` and `MessageText` are logged, and the subject is not.
- Full Gateway suite green: 3010 passed, 0 failed, 12 skipped (the Postgres-gated facts, in the default configuration).


thefrederiksen:
REJECT

The implementation behavior is sound, but the new test file records the disproved incident diagnosis as fact. `src/CcDirector.Gateway.Tests/Data/EntitlementSchemaQualificationTests.cs:12` says the live outage was traced to the entitlement read resolving through `search_path`. It was not: the baseline already applies the `gateway` default schema under the PostgreSQL provider, and the running binary already contained that behavior. This permanent comment would send future incident responders back toward a cause that has been ruled out.

Required change: rewrite the class summary to describe explicit table qualification as a behavior-preserving defense-in-depth guard. It must not say or imply that this pull request fixes, or explains, the live 503 response.

The remaining evidence passed at exact head `c1af4ac44cf7fa1a232edad19205b6936178597b`:

- Entity Framework generated the exact same 157-character PostgreSQL query on `origin/main` and on this head: `FROM gateway.entitlements AS e`, with no bare table reference.
- The failure path adds `SqlState` and `MessageText` for a PostgreSQL exception. No account subject or email variable reaches any constructed log line.
- The SQLite provider emits no schema, and the self-host enrollment control remains green.
- `ExcludeFromMigrations` remains on both provider branches.
- The change is limited to the two production files and the new test file.
- The full Gateway test project passed: 3,022 total, 3,010 passed, 12 provider-gated tests skipped, zero failed. The real-PostgreSQL collation test remained skipped in the default run, so the known baseline-only expectation defect was not encountered.

No merge performed.

thefrederiksen:
PASS

Prose-only confirmation passed at exact head `59c0f1cb785b167b79805bd3d3fd45daaae9528f`.

- The reframe from the previously reviewed head changes 78 lines, all comments or summaries; no production or test logic changed.
- The test summary now says the read is already qualified by `HasDefaultSchema(gateway)` and presents the explicit table schema only as a behavior-preserving defense-in-depth guard.
- The fail-on-purpose explanation is now accurate: removing the model-wide default alone remains green because the explicit table schema is independent; removing both qualification mechanisms makes the assertion fail.
- The database-context comment explicitly says the guard changes no emitted query today.
- The registry comment describes `SqlState` and `MessageText` as the diagnostic that exposes the server's own error, while retaining the no-subject and no-email rule.
- No added line contains the rejected false-premise language: `search_path`, unqualified, outage, 503, caused-by, the-fix, or go-live.

The previously completed behavior, privacy, generated-query, SQLite, self-host, surgical-diff, and full-suite evidence stands unchanged. No merge performed.

PR #1955, opened 2026-07-21
fix(gateway): map entitlements.subject as Postgres uuid via a value converter

thefrederiksen:
## What changed

`gateway.entitlements.subject` is a Postgres `uuid` column, but `EntitlementEntity.Subject` is a CLR `string` with no value converter and no column type. Under Npgsql that types the account-subject key-lookup parameter as `text`, so Postgres rejects the entitlement read with `42883 operator does not exist: uuid = text`. That failure made every entitlement read fail and hosted device enrollment answer 503.

The `subject` column stays `uuid` by contract - the website's `read_gateway_entitlement(p_subject uuid)` depends on it and the subject is the Supabase auth identifier. The Gateway was the side out of step, so this is a Gateway-side fix.

Inside the existing Postgres-only model block (the one that sets the `gateway` default schema), `EntitlementEntity.Subject` is now mapped through a `string`<->`Guid` value converter with its store type pinned to `uuid`, so Npgsql emits a UUID parameter for the key lookup instead of a text one. The CLR property stays a `string`, transparent to the token validator and the entitlement registry, which pass and compare it as a string. Applied only under Npgsql; self-host SQLite has no `uuid` type and keeps the plain-string column unchanged - a provider conditional, not a fallback.

## Proof

New model-only tests (`EntitlementSubjectUuidMappingTests`) that open no live database and run in the ordinary suite:

- **Under the Postgres provider** the `Subject` primary-key property maps through the value converter (model type `string`, provider type `Guid`) and its resolved relational store type is `uuid` - so the generated key-lookup parameter is a UUID, not text. Red-first: with the converter removed the property has no converter and its store type resolves to `text` (the defect); restoring the converter returns it to `uuid`.
- **Under the SQLite provider** the same `Subject` key has no converter and its store type stays a plain string, proving the Postgres-only conditional does nothing off Postgres (self-host unchanged).

The full Gateway test suite is green (one unrelated prompt-endpoint test flaked under concurrent-suite load and passes in isolation; it is structurally disjoint from this change).

The necessary condition proven here is that the key parameter now types as `uuid`. The definitive arbiter remains the live hosted Gateway's own `enroll` returning 200 after deploy.

## Note on ordering

This coexists with the sibling entitlements-schema pull request #1954, which changes the `ToTable(...)` call in the same `EntitlementEntity` block to schema-qualify it on Postgres. That change lands in the entity-config block; this one lands in the separate Postgres-only model-shape block, so they are adjacent but non-overlapping. This branch should be rebased onto `origin/main` to include #1954 once #1954 merges.

### DIFF
--- a/src/CcDirector.Gateway.Tests/Data/EntitlementSchemaQualificationTests.cs
+++ b/src/CcDirector.Gateway.Tests/Data/EntitlementSchemaQualificationTests.cs
@@ -29,6 +29,12 @@ public sealed class EntitlementSchemaQualificationTests
 {
     private const string PgTestConnectionEnvVar = "CC_GATEWAY_TEST_PG_CONNECTION";
 
+    /// <summary>A throwaway but VALID canonical account subject for the query-shape probes. The subject column
+    /// is Postgres uuid and the Gateway maps it through a string&lt;-&gt;Guid value converter, so the WHERE
+    /// constant must be a real uuid - EF evaluates the converter on it when it generates the SQL literal. The
+    /// value is irrelevant to what these facts assert (the FROM clause schema), it only has to parse as a uuid.</summary>
+    private const string ProbeSubject = "35543491-85cb-468d-a0c9-560193683105";
+
     /// <summary>A Fact that skips itself when the throwaway-Postgres connection is unset, so the normal
     /// SQLite test run and CI are unaffected. Anything reading a real PostgresException needs a real server.</summary>
     private sealed class RequiresPostgresFactAttribute : FactAttribute
@@ -81,7 +87,7 @@ public sealed class EntitlementSchemaQualificationTests
         Assert.Equal("gateway", entity!.GetSchema());
         Assert.Equal("entitlements", entity.GetTableName());
 
-        var sql = ctx.Entitlements.AsNoTracking().Where(e => e.Subject == "any").ToQueryString();
+        var sql = ctx.Entitlements.AsNoTracking().Where(e => e.Subject == ProbeSubject).ToQueryString();
 
         // The relation is schema-qualified. Postgres quotes an identifier only when it must, so EF emits the
         // bare-but-qualified form gateway.entitlements here; assert on that exact qualified reference.
@@ -107,7 +113,7 @@ public sealed class EntitlementSchemaQualificationTests
         Assert.NotNull(entity);
         Assert.Null(entity!.GetSchema());
 
-        var sql = ctx.Entitlements.AsNoTracking().Where(e => e.Subject == "any").ToQueryString();
+        var sql = ctx.Entitlements.AsNoTracking().Where(e => e.Subject == ProbeSubject).ToQueryString();
         Assert.DoesNotContain("gateway", sql);
         Assert.Contains("entitlements", sql);
     }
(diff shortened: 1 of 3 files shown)


thefrederiksen (commented):
REJECT

The guarded production mapping is sound, but two existing opt-in PostgreSQL facts are now incompatible with it, and the default run hides that regression by skipping them.

1. EntitlementRead_AgainstRealPostgres_ResolvesGatewaySchemaTable still creates gateway.entitlements.subject as text and queries with a value beginning sub-live- followed by a compact identifier. The new converter requires the canonical D form, so that value throws FormatException before the intended read. Changing only the value would still leave PostgreSQL comparing a text column with a uuid parameter. This fact can no longer prove the schema-qualified read.
2. EntitlementReadFailure_OnPostgresException_LogsSqlStateAndMessageText_NeverSubject still queries with a value beginning sub-PII-marker-. The converter rejects it before PostgreSQL can return the intended undefined-table error, so the fact cannot reach or prove the PostgreSQL diagnostic and subject-nondisclosure path.

The exact head and parent are correct. The converter application is entirely inside Database.IsNpgsql(); the static converter field outside that guard is inert. The application property remains string, conversion is Guid.ParseExact with D in and ToString with D out, the PostgreSQL mapping resolves the primary key to uuid, and SQLite retains a string key with no converter. Removing the converter makes the focused PostgreSQL proof fail. Removing both schema-qualification sources makes the updated schema probe fail on the unqualified generated FROM clause, so the new canonical probe value did not weaken that proof. The default Gateway suite passed with 3,012 passed and 12 skipped, and both continuous-integration jobs are green; those results do not execute the two broken real-PostgreSQL facts.

Please update the real PostgreSQL fixture to create subject as uuid and use canonical D-form subjects in both gated facts, then run those facts against PostgreSQL. The production fix should not land while its provider-specific live proofs are unable to exercise their intended paths.

thefrederiksen:
PASS

I reviewed exact head 693a53faf410f08c07f0bc1a1f4b0238adb6a2a6 after fetching origin. Its parent and origin/main are both c0ccdaf6a3a16f2a1e053651fcbbbcf3de182a11. The rework from the previously reviewed head 4a21d23f changes exactly one file, EntitlementSchemaQualificationTests.cs. GatewayDbContext.cs is byte-unchanged, so no production or model line moved in this rework.

The release build completed first with zero warnings and zero errors.

I myself then executed EntitlementSchemaQualificationTests against a throwaway PostgreSQL 16 server with CC_GATEWAY_TEST_PG_CONNECTION set. All four facts executed and passed, with zero skipped. The passed list included both required real-server facts by name:

- EntitlementRead_AgainstRealPostgres_ResolvesGatewaySchemaTable
- EntitlementReadFailure_OnPostgresException_LogsSqlStateAndMessageText_NeverSubject

The successful-read fact created gateway.entitlements.subject as the native universally unique identifier type, seeded it with a native identifier parameter, and read the active row back through EntitlementRegistry. I also confirmed the resulting server column type and seeded row directly.

The failure fact reached the real PostgreSQL undefined-table error 42P01. The server log recorded that gateway.entitlements did not exist, proving the read reached PostgreSQL rather than failing in the converter. The fact then passed its assertions that the diagnostic contains SqlState and MessageText and that no emitted line contains the subject.

With CC_GATEWAY_TEST_PG_CONNECTION removed, the full default Gateway test project also passed: 3,012 passed, 12 skipped, zero failed. The two named real-server facts were present as skipped in that default run. The schema-qualified generated-read fact is unchanged from the prior approved head and passed in both runs, so the fixture correction did not weaken the red-first schema guard.

PR #1956, opened 2026-07-21
fix(gateway): derive cockpit URL from ONE public base URL (path per surface)

thefrederiksen:
## What

Reworks the P1 first cut to the owner's settled design (2026-07-20): there is ONE public base URL for a Gateway, and every surface is a PATH under it, derived on the Gateway exactly as localhost already works (`host:port/cockpit`). The Cockpit URL is `{base}/cockpit`; the mobile URL is `{base}/mobile`. One derivation rule, hosted and self-host alike. The client is dumb (CLAUDE.md rule 7) and opens whatever URL the Gateway hands back.

## Changes

- Replace the cockpit-specific env var `CC_GATEWAY_PUBLIC_COCKPIT_URL` with the ONE base var **`CC_GATEWAY_PUBLIC_URL`** (e.g. `https://gateway.devthrottle.com`, set on the App Service).
- `GatewayCockpitUrl` -> **`GatewayPublicUrl`**: the resolver takes a surface PATH and returns `base + path`. HOSTED (`CC_GATEWAY_HOSTED=1`) uses the configured base and **fails loud** if it is unset/blank (deploy misconfiguration, no fallback); SELF-HOST uses the tailnet front door (null when Tailscale is down) with the path appended - the base-resolution branch is byte-identical to before, only the explicit surface path is new.
- Route **every** cockpit-URL handout through `ResolveCockpit()`: `GET /cockpit` (`CockpitInfoDto.Url`), the `CockpitUrl` on `GET /gateway/about`, and the `cockpit.url` on `GET /gateway/settings` (the first cut missed this third site, which still emitted the raw front-door root).
- Add `ResolveMobile()` (`{base}/mobile`) with its own pure both-mode unit tests so the resolver is one complete thing - but wire **nothing** mobile-serving and leave `/m` exactly as-is. The `/m`->`/mobile` app re-base (Vite base, React Router basename, service-worker scope, the `/m/signin` + `/m/enroll` seam), the 301, and the first consumer of `ResolveMobile` all land together in **P3** where a phone proves it. A `DEFERRED to P3` note sits next to `ResolveMobile` so it is not mistaken for wired.

## Proof

Tests cover both modes for BOTH cockpit and mobile, each proven to fail on purpose:
- pure resolver tests redden when the surface-path append is dropped (cockpit AND mobile);
- the three endpoint tests redden when any call site is reverted to the tailnet front door (the front-door root appears where `{base}/cockpit` should be).

Full `CcDirector.Gateway.Tests` suite green: **3029 passed, 0 failed, 12 skipped** (skips are live-Postgres-gated). No mobile/web/TS changes, so no web/mobile test project is affected.


thefrederiksen:
Phase-one inspection fixes (HEAD d0368b62)

F1 - Removed the desktop Learn button. Verified the desktop already has a Documentation menu item opening https://devthrottle.com/docs, so the retired Learn page has no replacement to point at. Deleted the Learn toolbar button and all of its plumbing: ResolveLearn and LearnPath, CockpitInfoDto.LearnUrl and its populating site in GET /cockpit, BtnLearn_Click, and the Learn-destination tests. ResolveCockpit and the Cockpit URL fix are unchanged.

F2 - The endpoint tests now configure a non-production base (https://gw.test.invalid) and assert values derived from it, so a hardcoded production constant reddens. Added live-wrapper assertions that drive ResolveCockpit() through the real environment for both hosted (deterministic pin to the configured base) and self-host (proves the wrapper ignores the hosted variable and routes through the hosted gate), so the wrapper is proven wired to the pure resolver.

F3 - Extracted a pure seam SelectCockpitOpenUrl(info) used by the Cockpit button, and added a desktop test pinning that it opens the handed-back Url with no subpath appended.

F4 - Updated the CockpitInfoDto.Url documentation to say {base}/cockpit (hosted) / {frontDoor}/cockpit (self-host).

Guard mutations re-run and confirmed red:
- Hardcode ResolveCockpit() to a production constant: 5 failed (including the hosted live-wrapper and the endpoint tests).
- Reintroduce the bad desktop composition (info.Url + subpath): 2 failed.

Suites: CcDirector.Gateway.Tests 3031 passed, 0 failed, 12 skipped (live PostgreSQL); CcDirector.Avalonia.Tests 219 passed, 0 failed; cockpit web production build passed.

thefrederiksen:
## Phase-one closing pass (HEAD 3ded3eee)

Closes the two blockers and one documentation defect from the re-inspection at d0368b62.

**F1 - consumer boundary pinned.** `BtnCockpit_Click` no longer holds any cockpit-URL logic: its entire fetch->select->open decision now delegates to the testable `MainWindow.ResolveCockpitOpenUrlAsync(fetch)`, which returns `info.Url` verbatim via `SelectCockpitOpenUrl`. A new desktop test drives that method with a fake fetch and asserts the returned URL has no subpath. Re-running the inspector's exact mutation (compose `info.Url + "/learn"` at the consumer) now REDDENS: `Failed: 1, Passed: 5` with an ordinary xUnit string assertion.

**F2 - exact self-host live wrapper.** The live Cockpit resolver now takes an injectable front-door provider (`GatewayPublicUrl.ResolveCockpit(Func<string?>)` -> `Resolve(surfacePath, provider)`), so the self-host branch is driven with a known front door. The new test asserts the self-host live wrapper equals `{frontDoor}/cockpit` exactly, and null when the provider returns null. The self-host-only production-hardcode mutation now REDDENS: `Failed: 1, Passed: 19` (previously stayed 19-green). The pure overload and all prior tests are intact.

**F3 - documentation.** Removed `/learn` from the `CockpitReactApp` browser-routing comment (Learning was retired; Help -> public docs). Repository source search confirms no other comment claims `/learn` is a live React page.

**Verification**
- `CcDirector.Avalonia.Tests`: 221 passed, 0 failed.
- `CcDirector.Gateway.Tests`: 3032 passed, 0 failed, 12 skipped (live PostgreSQL proofs).
- Cockpit web build: TypeScript check and Vite production build pass.
- Both mutations redden with ordinary assertion failures (no framework errors), then reverted.

PR #1958, opened 2026-07-21
feat(client): hosted callback forwards access_token as Bearer to /m/enroll

thefrederiksen:
## What

Phase C of the hosted multi-tenant sign-in mission: make the client complete the HOSTED sign-in round trip.

The shared device-enrollment callback (`DeviceCallback`) now branches on which credential the website returned in the URL fragment:

- **Hosted gateway**: the fragment carries the account's Supabase `access_token`. The client POSTs `/m/enroll` with `Authorization: Bearer <access_token>` and body `{deviceId, name, platform}` (no `device_key`). This is the mint boundary the hosted endpoint expects (it reads the account subject from the verified token, checks the paid entitlement, resolves the tenant, and mints a tenant-scoped device key). Previously the client sent a `device_key` body with no auth header, which 401s on a hosted gateway.
- **Self-host gateway**: the fragment carries a `device_key`. Unchanged, byte-for-byte: `device_key` in the body, no auth header.

## How

- `readEnrollCredential(params)` — a pure helper keyed only on the fragment (`access_token` -> hosted, `device_key` -> self-host, access_token wins if both present). Makes the branch decision testable without a DOM and keeps the self-host path additive-only.
- `enrollDeviceHosted(accessToken, ...)` — new request that sends the Bearer header and omits `device_key`. `enrollDevice` (self-host) is unchanged. Both read the local device key back through one shared reader.
- `DeviceCallback` selects the enroll call from the credential mode; the state verification, cookie mirror, `next=` landing, and error handling are shared unchanged.

## Contract

Matches the pinned mission contract: website returns `#access_token=<token>&state=<state>` for hosted (self-host still returns `#device_key=...`); hosted `/m/enroll` expects `Authorization: Bearer <access_token>` + body `{deviceId, name, platform}`.

## Tests

Both branches, proven to fail on purpose:
- Hosted: sends `Authorization: Bearer <token>`, body omits `device_key`, token never leaks into the body.
- Self-host: sends `device_key` in the body with no `Authorization` header (unchanged).
- `readEnrollCredential`: hosted for access_token, self-host for device_key, access_token wins when both present, null for neither.

Full `client-core` vitest suite green (552 passed). All three web workspaces typecheck. Mobile + cockpit builds green. No .NET touched. Lint clean on all changed files (two pre-existing `rule definition not found` errors remain on untouched files identical to origin/main).

### DIFF
--- a/packages/client-core/src/api/enroll.test.ts
+++ b/packages/client-core/src/api/enroll.test.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from "vitest";
-import { GATEWAY_NOT_SIGNED_IN, isGatewayNotSignedIn } from "./enroll";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { GATEWAY_NOT_SIGNED_IN, enrollDevice, enrollDeviceHosted, isGatewayNotSignedIn } from "./enroll";
 import { GatewayError } from "./client";
 
 // There are TWO different sign-ins in this product and they are easy to conflate:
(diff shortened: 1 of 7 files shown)


thefrederiksen:
Addressed all three inspection findings at f98723ac.

**F1 (BLOCK, login-CSRF) — state now fails closed.** The callback enrolls only when this browser still holds the exact anti-forgery nonce it minted AND the site echoed that same nonce back. Missing saved state, missing returned state, or a mismatch all reject via the existing error path — no enroll call runs. Verification happens before either enroll call. Both branches are held to the same bar: the website echoes `state` for both the self-host (`buildMobileCallbackUrl`) and hosted (`buildHostedAccountTokenCallbackUrl`) callback shapes and `SignIn` always mints it, so a legitimate callback of either kind always carries a matching state — uniform fail-closed does not regress the self-host happy path.

**F2 (BLOCK) — runtime dispatch is now pinned.** The dispatch (state check, credential classification, and which enroll request is sent) is extracted into `runEnrollmentCallback` (enrollCallback.ts) so it is unit-tested without a DOM; `DeviceCallback` is a thin consumer that maps the outcome to a phase and the shared side effects. New tests drive the real dispatch: a hosted fragment must reach the Bearer request (body `{deviceId,name,platform}`, no `deviceKey`) and a self-host fragment the device-key body request. Mutation proof: reverting the hosted arm to the old `enrollDevice` shape reddens the hosted dispatch test; reverting the state gate to best-effort reddens the missing-state test.

**F3 (non-block) — mixed fragment fails closed.** A fragment carrying BOTH `access_token` and `device_key` is ambiguous and now returns null instead of silently picking hosted. The precedence test was updated to assert ambiguous-rejects.

Verification: client-core typecheck passing; full suite 51 files / 560 tests passing (was 50 / 552); cockpit and mobile production builds green. Scope unchanged — all edits under `packages/client-core/src` (3 modified, 2 new).

PR #2016, opened 2026-07-22
fix(gateway): isolate the events feed by request tenant

thefrederiksen:
## Summary

- resolve the authenticated request tenant before opening the `/events` stream
- deny hosted requests when no tenant is bound, instead of exposing the fleet-wide feed
- carry tenant ownership through Director arrival notifications and filter both arrivals and removals
- preserve the Local tenant behavior on self-hosted Gateways

## Reproduction

A hostile two-tenant test subscribed tenants A and B to the real `/events` route, then connected a new Director owned only by tenant B through the real tunnel. Before the fix, tenant A received tenant B's Director notification. The test failed with `Assert.ThrowsAny() Failure: No exception was thrown` because tenant A's stream returned the foreign event instead of remaining quiet.

## Verification

- 23 affected Gateway tests passed, including cross-tenant arrival and removal isolation, hosted fail-closed behavior, registry ownership, and self-hosted Local behavior
- `CcDirector.Gateway` built successfully with zero warnings
- `CcDirector.Gateway.Migrations.Postgres` built successfully with zero warnings

### DIFF
--- a/src/CcDirector.Gateway.Tests/DirectorRouteTenantScopingTests.cs
+++ b/src/CcDirector.Gateway.Tests/DirectorRouteTenantScopingTests.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 using CcDirector.Core.Tenancy;
 using CcDirector.Gateway;
 using CcDirector.Gateway.Contracts;
+using CcDirector.Gateway.Discovery;
 using Xunit;
 
 namespace CcDirector.Gateway.Tests;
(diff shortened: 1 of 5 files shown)


thefrederiksen:
Independent verification before merge (release-gate seat): built the branch in a clean detached worktree at cd5c9af8; the full DirectorRouteTenantScopingTests plus EventsFeedSelfHostedTests run is 16 of 16 green. Revert-proof executed: with the tenant filter neutered (unconditional writes restored in the /events handler), both isolation facts go red - tenant A receives tenant B's arrival and removal events - and green again with the filter restored. Self-host behavior preserved by its own suite. This closes the last open route-level cross-tenant exposure from the round-four review; merging.

PR #2024, opened 2026-07-22
fix(mac): single 'Director' app, upgrade-in-place install, working Launch button

thefrederiksen:
Makes the macOS install runnable and stops the copy pile-up, ahead of the v1.7.3 release.

## What this fixes
- **Five copies pile up.** Fresh install now purges stale bundles (legacy "CC Director.app", the new name, and Finder's numbered duplicates) from both ~/Applications and /Applications before placing one canonical bundle, so reinstalling upgrades in place.
- **App named "CC Director".** Renamed to a single app "Director" across packaging, setup engine, launcher, and updater. Bundle id, binary name, and download asset are unchanged so the auto-updater is unaffected; the old name is kept as a detection alias (issue thefrederiksen/devthrottle_internal#871) so a pre-rename host is refreshed, not orphaned.
- **Launch button vanished on hover / did nothing.** The final "Launch Director" button now uses a reusable Button.launch style with a :pointerover template override so it stays visible, and on macOS it launches the .app via /usr/bin/open (correct path + Dock registration) instead of running the inner binary at the wrong path.

## Verification
- setup-engine build clean; 369 setup-engine tests + 27 launcher tests pass; setup wizard builds clean.
- The .app launch and Dock behavior are verified on the Mac after release (cannot be exercised from a Windows CI build).

## Follow-up (not in this PR)
- Rename the dev-slot build scripts off /Applications so developer slot builds cannot collide with the user install, and add a DMG drag-to-Applications package + a Director icon.

### DIFF
--- a/docs/public/release-notes/v1.7.3.md
+++ b/docs/public/release-notes/v1.7.3.md
@@ -5,6 +5,7 @@ This release brings all of your machines together on one cockpit, tells you whic
 ### Highlights
 - See every machine on one cockpit: the Fleet Map now lays out all of your connected machines, the Directors running on each, and their sessions in a single view. Group the map by machine, by repository, or by working tree. Point several machines at the same gateway and watch them all appear together.
 - Know which gateway you are on: the desktop status box now shows the gateway host and whether it is connected, so you can tell at a glance where this Director is pointed.
+- A cleaner Mac install: the Mac app is now a single app named "Director". Reinstalling replaces the previous copy in place instead of leaving a pile of duplicates, and installing removes older copies left behind by earlier versions. The "Launch Director" button at the end of setup now works on Mac and opens the app in the Dock.
 - No more first-party telemetry: DevThrottle no longer reports any usage telemetry back to us. The reporting code is gone, and this release also cleans up any telemetry files and settings left behind by earlier versions.
 
 ### Also in this release
--- a/scripts/package-mac-app.sh
+++ b/scripts/package-mac-app.sh
@@ -18,11 +18,11 @@
 # Usage:
 #   scripts/package-mac-app.sh --binary <path-to-self-contained-binary> --out <output-dir>
 #
-# Produces:  <output-dir>/cc-director-mac-arm64.zip  (contains "CC Director.app")
+# Produces:  <output-dir>/cc-director-mac-arm64.zip  (contains "Director.app")
 #
 set -euo pipefail
 
-APP_NAME="CC Director"
+APP_NAME="Director"
 BIN_NAME="cc-director"                 # CFBundle binary name (matches AssemblyName)
 BID="com.devthrottle.ccdirector"
 ZIP_NAME="cc-director-mac-arm64.zip"
--- a/src/CcDirector.Core/Update/UpdateInstaller.cs
+++ b/src/CcDirector.Core/Update/UpdateInstaller.cs
@@ -24,7 +24,7 @@ public static class UpdateInstaller
     /// <summary>
     /// The path a new build must overwrite to "become" the installed app. On
     /// Windows this is the running cc-director.exe; on macOS it is the enclosing
-    /// "CC Director.app" bundle (not the binary inside it). Falls back to the
+    /// "Director.app" bundle (not the binary inside it). Falls back to the
     /// process path when not running from a bundle (e.g. a bare dev binary).
     /// </summary>
     public static string InstallTarget()
--- a/src/CcDirector.Core/Update/UpdaterState.cs
+++ b/src/CcDirector.Core/Update/UpdaterState.cs
@@ -34,7 +34,7 @@ public sealed class UpdaterState
 
     /// <summary>
     /// Absolute path the staged build should overwrite. On Windows the installed
-    /// cc-director.exe; on macOS the installed "CC Director.app" bundle directory.
+    /// cc-director.exe; on macOS the installed "Director.app" bundle directory.
     /// </summary>
     [JsonPropertyName("installTarget")]
     public string? InstallTarget { get; set; }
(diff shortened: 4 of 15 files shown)


thefrederiksen:
# Mac installer mission - handover to devthrottle - mac (Sorens-Mac-mini)

You are taking over the macOS installer for DevThrottle (the app whose Director binary is
cc-director; the product is DevThrottle). The Mac install experience is bad and the owner
wants it made genuinely good on real hardware. Everything below was written and verified on
Windows only - the whole point of handing this to you is that you can build, run, and SEE
the installer on a Mac, which a Windows CI build cannot.

## Ground state (read first)

- The fixes you need to TEST are on branch `mac/rename-director` (pull request thefrederiksen/devthrottle#2024). That
  pull request is NOT merged to main yet (continuous integration still running), so `main`
  does NOT have them. Check out the branch to test:
      git fetch origin
      git worktree add ../devthrottle-mac -b mac-work origin/mac/rename-director
  (or once thefrederiksen/devthrottle#2024 merges, work from origin/main). Never work in the shared checkout.
- Repo on this machine: /Users/soren/ReposFred/devthrottle. Version 1.7.3.
- A v1.7.3 release has been DRAFTED but NOT published. DO NOT publish any release. We cut the
  release only AFTER the installer is good. Hold.
- Rules of the house: trunk-based development - land small pull requests on origin/main, delete
  the branch, park on main. Work in a worktree, never the shared checkout. NEVER put Claude /
  Anthropic / AI attribution in any commit, pull request, issue, comment, or file. Plain
  English, no abbreviations. ASCII only, no emoji.

## What was already implemented on `mac/rename-director` (verify each on the Mac)

1. Renamed the Mac app bundle "CC Director.app" -> "Director.app" across packaging, the setup
   engine, the launcher, and the updater. The bundle id (com.devthrottle.ccdirector), the
   binary name (cc-director), and the download asset (cc-director-mac-arm64.zip) were kept
   STABLE so the auto-updater is unaffected. The old name is registered as a detection alias
   (InstallLayout.LegacyAliasesFor) so a pre-rename host is refreshed, not orphaned.
   Files: scripts/package-mac-app.sh (APP_NAME), tools/cc-director-setup-engine/InstallLayout.cs,
   tools/cc-director-setup-engine/MacAppPlacer.cs, src/CcDirector.Core/Update/UpdateInstaller.cs,
   src/CcDirector.Core/Update/UpdaterState.cs, src/CcDirector.Launcher/DirectorSupervisor.cs.

2. Forced cleanup on fresh install: MacAppPlacer.PurgeStaleBundles removes stale copies (the
   legacy "CC Director.app", the new "Director.app", and Finder's numbered duplicates like
   "CC Director 2.app") from BOTH ~/Applications and /Applications before placing the one
   canonical bundle, so reinstalling upgrades in place instead of stacking icons.

3. The setup wizard's final "Launch Director" button:
   - Was invisible/unclickable on hover because it set its green background inline with no
     :pointerover template override, so the base theme repainted it near-white on hover. Fixed
     with a reusable Button.launch style (tools/cc-director-setup-avalonia/Styles.axaml) applied
     in Steps/CompleteStep.axaml.
   - Did nothing when clicked on Mac because it looked for the binary at the bundle root and ran
     the raw binary (the binary is in Contents/MacOS, and running it directly gives no Dock
     registration). Now on macOS it launches the .app via /usr/bin/open, which is what gives the
     Dock icon and foreground activation. See Steps/CompleteStep.axaml.cs LaunchButton_Click.

   Verified on Windows: 369 setup-engine tests + 27 launcher tests pass; setup wizard builds
   clean. NOT verified on a Mac - that is your job.

## What to TEST on the Mac (end to end)

Build the Director and the setup wizard for osx-arm64, package the .app with
scripts/package-mac-app.sh, and run the installer for real. Then confirm:
- After install there is exactly ONE app in ~/Applications named "Director" with the correct
  icon and name (not "CC Director", no duplicates).
- Reinstalling REPLACES it in place - no second icon, no "Director 2.app".
- Pre-existing "CC Director*.app" copies (in ~/Applications and /Applications) are removed by
  the install. Confirm the cleanup does not over-delete or fail silently, and handle the case
  where a /Applications copy needs admin rights to remove (log clearly, do not crash).
- The "Launch Director" button is visible on hover, clickable, launches the app, and the app
  appears in the Dock.
- In the launched Director you can select a gateway, including the HOSTED gateway.

## What to IMPLEMENT / fix (outstanding - this is the real work)

1. ALL the installer buttons are bad - audit and fix EVERY button on EVERY wizard step
   (Welcome, Prerequisites, Skills, GatewayConnect, Install, Complete). Look for the same
   inline-background-with-no-pointerover-override bug elsewhere, plus sizing, contrast,
   disappearing on hover, dead clicks, and anything that does not look and behave like a solid
   native button. The owner's words: the buttons are "really bad". Treat this as the headline.
2. The cleanup issue - make sure the pile-up can never happen again and the purge is correct
   and safe (see the /Applications admin case above).
3. Dev-slot scripts collide with the user install: scripts/local-build/mac/make-app-bundle.sh,
   scripts/mac-rebuild.sh, scripts/mac-setup.sh still build "CC Director N.app" into
   /Applications, which the new purge would delete. Rename them to "Director" and move dev slots
   OFF /Applications (into a dev-only dir) so developer builds can never collide with the real
   install.
4. Ship a DMG (drag-to-Applications) instead of the bare cc-director-mac-arm64.zip so a manual
   install always lands in one canonical place, and give the app a proper "Director" icon
   (scripts/local-build/mac/AppIcon.svg is the current source of the Mac icon).
5. First-run polish: the Dock icon, the one-time Gatekeeper right-click-Open, Intel-vs-arm64
   messaging (install-mac.sh currently fails on non-arm64).

## Deliverables

- Small pull requests to origin/main, each with a screenshot of the installer showing the fix.
- Do NOT publish a release. When the installer is good, report back and the owner cuts v1.7.3.

PR #2126, opened 2026-07-24
Morning report endpoint: one honest JSON report per account (#2119)

thefrederiksen:
Closes thefrederiksen/devthrottle#2119. Second slice of thefrederiksen/devthrottle_internal#925 (the brain / morning report). The website's 07:00 cron reads this endpoint, renders the approved email design from it, and sends it - so this is the piece that makes tomorrow morning's first email possible.

## What this adds

`GET /gateway/reports/morning?account=&date=yyyy-MM-dd&tz=<IANA>` returning, for ONE account and ONE calendar day:

- `window` - the resolved UTC range, plus the date and zone it came from. Every number carries its coordinates.
- `stats` - `sessionsRan`, `workDelivered`, `hostedAiSpendUsd`, each individually optional.
- `attention[]` - typed rows, today `waiting-session` only.

Read-only. It writes nothing and needs no table of its own: the numbers come from the governance event ledger, the workflow-run table, and the mirrored hosted-AI spend, all of which already exist.

## The honesty rule, as code rather than as a promise

A stat whose backing store holds **no rows at all** for the account is ABSENT from the JSON. A store that holds rows but none inside the window reports a measured **zero**. Both are reachable and they mean different things - "we have never measured this" versus "we looked and it was nothing" - and an email that printed `0` for the first would state a measurement nobody made.

This is also what makes the slice shippable on its own: with no repo-state store yet (#2118), the stale-worktree and unmerged-branch rows are simply not present, rather than present and empty.

## The window

Resolved from `date` + IANA zone. The end is the start of the NEXT calendar day resolved independently in the same zone - **not** the start plus 24 hours. So the spring-forward day is 23 hours, the autumn-back day is 25, and consecutive days abut exactly across a transition. A zone whose local midnight does not exist that day (Santiago in September) starts at the first instant that does. An unparseable date or unknown zone is a 400, never a default.

## Authorization

The caller is a server with no device key, so the route is exact-match exempt from the host-wide token gate and carries its own: a bearer token from `REPORT_SERVICE_TOKEN`, compared in fixed time over digests. Every branch denies by default:

| condition | answer |
|---|---|
| `REPORT_SERVICE_TOKEN` unset or blank | 503 - refuse to serve, never serve unguarded |
| missing / malformed / wrong / near-miss credential | 401 |
| valid token, unknown account | 404 |
| valid token, an email two tenants share | 409 - never a guess about whose day to email |
| a sibling path under the route | still 401 from the host gate |

## Tenancy

The tenant is resolved ONCE from the named account (by tenant id, or by account email case-insensitively) and passed explicitly into the builder, which reads through a context stamped with that tenant. Another tenant's rows are therefore unreachable rather than merely unasked-for. Proven with two seeded tenants at the builder AND at the endpoint, including the case where both tenants run a session with the same raw id - the live roster of the other tenant never labels this tenant's row.

Self-host has one tenant and no account census, so there the report is served for the single local tenant; the service token still gates it.

## Waiting rows

Derived from the durable governance event ledger, not from a live connection - the ledger appends only on a real transition, so the last event IS the start of the current state and "waiting since" is read rather than inferred. A Director that happens to be offline at 07:00 costs a row its friendly name and repository path, never its place in the email. A session that came back, that exited, or that the owner deliberately snoozed is not reported. A session with no transition in 30 days is not reported either - the Gateway will not assert the current state of something it has not heard about in a month, and the scan cap is logged rather than silently truncating.

## Tests

63, all green.

- **Window**: ordinary day, winter offset, spring-forward (23h), autumn-back (25h), consecutive days abutting across a transition, a zone with no local midnight, a far-from-UTC zone, and seven malformed-input refusals. Revert-prove is written into the file: compute the end as `startUtc.AddDays(1)` and exactly the two transition tests go red.
- **Builder**: the absent-versus-zero pair for every stat; distinct-session counting; inclusive start / exclusive end; accepted-only delivery counting; ceil rounding across seven micro-dollar cases and on the summed total; waiting-row derivation, ordering, snooze and exit exclusions, and the lookback horizon; two-tenant isolation including a shared session id.
- **Endpoint**: the whole authorization matrix above, account resolution by id and by email casing, the exemption asserted through the middleware itself (and a sibling path proven still gated), the two-tenant negative with a raw-text assertion that the other tenant's identifiers appear nowhere, and the exact wire shape - camelCase keys, absent keys genuinely absent, money as a number, and no synthetic `$type` from the polymorphic attention list.

### DIFF
--- a/src/CcDirector.Gateway.Contracts/MorningReportDtos.cs
+++ b/src/CcDirector.Gateway.Contracts/MorningReportDtos.cs
@@ -1,0 +1,165 @@
+using System.Text.Json.Serialization;
+
+namespace CcDirector.Gateway.Contracts;
+
+/// <summary>
+/// The morning report for ONE account and ONE calendar day (issue #2119, slice 2 of #2096) - the JSON the
+/// website's 7:00 cron reads, renders into the approved email design, and sends. Three headline numbers, the
+/// needs-your-attention list, and the exact window every number was measured over.
+///
+/// THE HONESTY RULE IS STRUCTURAL, NOT EDITORIAL. Anything the Gateway has no data for is ABSENT from the
+/// JSON - never zero-filled, never estimated. A missing <see cref="MorningReportStatsDto.SessionsRan"/> means
+/// "this Gateway holds no session history for this account", which is a different statement from "no sessions
+/// ran", and the email must be able to tell them apart. Every optional member below is therefore
+/// <see cref="JsonIgnoreCondition.WhenWritingNull"/>: a null does not serialize at all.
+///
+/// The field names here are the CONTRACT the website sender is coded against (owner-relayed, 24 July 2026).
+/// They are camelCase on the wire (minimal-API web defaults). Renaming one breaks the email.
+/// </summary>
+public sealed class MorningReportDto
+{
+    /// <summary>The account this report is about, echoed back exactly as the caller asked for it, so the
+    /// sender can prove the report it is about to email belongs to the recipient it is emailing.</summary>
+    public string Account { get; set; } = "";
+
+    /// <summary>The exact coordinates every number below was measured over - the resolved UTC range, plus
+    /// the calendar day and zone it came from. Every number carries its coordinates.</summary>
+    public MorningReportWindowDto Window { get; set; } = new();
+
+    /// <summary>The three headline numbers. Each is individually optional (see the honesty rule).</summary>
+    public MorningReportStatsDto Stats { get; set; } = new();
+
+    /// <summary>
+    /// The needs-your-attention list, one typed item per row the email renders. ALWAYS PRESENT, possibly
+    /// empty: an empty list is real knowledge ("nothing is waiting on you"), unlike an absent stat.
+    /// Items are polymorphic and discriminated by their <c>type</c> string; the sender tolerates a type it
+    /// does not know (it logs and does not render it), so a new item type never breaks a sent email.
+    /// </summary>
+    public List<MorningAttentionItemDto> Attention { get; set; } = new();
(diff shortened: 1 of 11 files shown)


thefrederiksen:
## Local test evidence

Run against this branch (main + these two commits) on Windows, .NET 10.

**The change's own tests - 75 passed, 0 failed**

```
dotnet test src/CcDirector.Gateway.Tests --filter "FullyQualifiedName~MorningReport|FullyQualifiedName~TenantRegistryEmailBackfill"

Passed!  - Failed: 0, Passed: 75, Skipped: 0, Total: 75, Duration: 10 s
```

Covering: window resolution including the three daylight-saving cases and consecutive days abutting exactly across a transition; the absent-versus-zero pair for every stat; ceil rounding across seven micro-dollar cases and on a summed total; waiting-row derivation from the durable ledger, ordering, and the snooze/exit/horizon exclusions; the whole authorization matrix (503 unconfigured, 401 missing/malformed/near-miss, 404 unknown account, 409 ambiguous account); the route exemption asserted through the middleware itself with a sibling path proven still gated; two-tenant isolation at the builder AND the endpoint, including a session id both tenants share; the exact wire shape; and the email backfill (fires on null, never overwrites, never blanks, and makes the account findable by email afterwards).

**The whole gateway suite - no regression, no hang**

```
dotnet test src/CcDirector.Gateway.Tests --blame-hang-timeout 300s

Test Run Successful.
Total tests: 3881
     Passed: 3864
    Skipped: 17
 Total time: 13.1094 Minutes
```

The hang detector was armed at 300 seconds and never fired.

PR #2131, opened 2026-07-24
Director repo-state snapshots: branches and worktrees to the Gateway (#2118)

thefrederiksen:
Closes thefrederiksen/devthrottle#2118. First slice of thefrederiksen/devthrottle_internal#925 (the brain / morning report). The report endpoint (#2126, issue thefrederiksen/devthrottle#2119) is already merged and deployed; this is the git-hygiene feed it reads, plus the two report-side pieces that were falling between the two issues.

## What this adds

**1. The Director ships repo-state snapshots** for each registered repository, and the Gateway stores the newest one per (tenant, director, repository). This is the one git-hygiene feed the morning report cannot get from any Gateway-side store: the Gateway knows a repository exists, but only the machine holding the checkout can enumerate its branches and worktrees.

**2. The report's hygiene rows** - stale worktrees and unmerged branches - which thefrederiksen/devthrottle#2118 leaves to "the next slice" and thefrederiksen/devthrottle#2119 omits "until snapshots exist", so neither issue as written actually put them in the email despite their being its flagship content.

**3. A 48-hour reporting bar on waiting rows**, which the first real production call made necessary - see below.

## Director side

`RepoStateSnapshotCollector` builds each snapshot from the git services the Director already uses, so there is no second notion of "merged" anywhere in the product:

- `GitBranchService` gains `ListInventoryAsync`, returning the branches **and** the default branch every containment verdict was measured against. Carried together on purpose - a caller resolving the default branch independently could disagree with the service that computed containment, and "merged into main" measured against a different ref than the caller believes is exactly the confidently-wrong statement this email must not make.
- New `BranchInfo.MergedIntoDefault`, deliberately **narrower** than `SafeToDelete` and not a substitute for it. Safe-to-delete also accepts "origin branch gone" and "pull request merged", and refuses any checked-out branch - which is precisely the case the report wants to talk about.
- Worktree merged-ness is read back out of that one inventory rather than recomputed, so a snapshot cannot show a worktree as safe to remove whose branch it also lists as unmerged, in the same payload.

**It never guesses.** An undetectable default branch records `null`, and every merged-ness in that snapshot is `null` - not `false`, which a reader would legitimately treat as "this definitely has unmerged work". A repository that fails to enumerate is **omitted** rather than pushed empty: an empty snapshot reads downstream as a clean bill of health invented out of a failure.

`RepoStatePusher` runs two minutes after start, then every six hours. Fail-safe is the design, not a `try`/`catch` around it: a refused push, a throwing transport, and a failed enumeration all return quietly and **the next cycle still runs** (asserted, not promised). No outbox, deliberately - this pushes CURRENT state, so a lost push is superseded rather than lost, and replaying one would lay a stale picture over a fresher one.

## Gateway side

`POST /gateway/repostate`, device-authenticated, write-only. The tenant comes from the caller's authenticated key and **never** from the body. **No read route** - the only consumer is the report, which reads in-process; a public read would put every repository path and branch name of every account on an HTTP surface for no caller that exists. `repo_state` is keyed `(tenant_id, DirectorId, RepoPath)` because both natural key parts are caller-supplied; overwritten, never appended. A snapshot older than the freshness bar is **excluded** from a read, never aged into a recommendation.

## The hygiene rows

Every rule is conservative in the same direction, because a wrong answer here costs the owner work rather than time:

- **Safe to remove is the narrow claim.** Old, clean, unoccupied, AND provably merged. Unknown merged-ness goes in the not-safe row: it is honest to say "this is old" and dishonest to say "this is finished".
- **Safe and not-safe are separate rows**, never one blended count. "Six worktrees, four of them safe to remove" is a sentence a person acts on wrongly.
- A **dirty** worktree is never listed (uncommitted work exists nowhere else); nor is an **occupied** one (that is not stale work, that is someone's desk).
- An age needs an **observed timestamp**. No last-activity and no tip date means skipped: an unknown age is not a small one.
- A branch counts as unmerged only on a definite `false`, never a `null`; the default and checked-out branches are excluded; anything touched in the last day is work in progress, not a recommendation.

No repo-state store, or no snapshot fresher than 24 hours (four missed push cycles), means **no hygiene rows at all** rather than empty ones.

## The 48-hour bar, and why production taught us to add it

The first real call to the deployed endpoint returned nine `waiting-session` rows aged **69 to 129 hours**, for sessions long gone, shown as raw identifiers because no Director is connected to the hosted Gateway to name them.

Every one was a true statement *about the ledger* and a false statement about the owner's morning. The ledger records what it is told; a session that stops without an exit event leaves its wait open forever, so "waiting since" ages without bound and never resolves. After two days of silence the Gateway has no business asserting a session is waiting on a person today, so now it says nothing about it - counted and logged, never silently dropped. It is a reporting bar, not a data change: the ledger keeps every transition.

## Two defects the tests caught before they shipped

1. **`repo_state` was missing from the model's tenant-scope list**, so its global query filter was absent: one account's push overwrote another's row for the same repository path, and one account's read returned the other's rows. Both cross-tenant tests went red on the first run. Both migrations were regenerated over the corrected model; both snapshots verify clean with `has-pending-model-changes`.
2. The report builder was being constructed **before** the repo-state store, so it would have silently received `null` and quietly omitted every hygiene row forever. Construction order fixed, with a comment saying why it matters.

## Tests

44 new, and 109 green across the fold, the report, and the repo-state store and endpoint together.

- **Collector (6)**, against REAL git repositories: merged versus unmerged in both directions with ahead-counts; an undetectable default branch recording null everywhere; worktrees reported with the primary checkout excluded; a broken repository omitted rather than pushed empty; a vanished path skipped; and a no-content assertion against a repository whose files **and commit messages** carry a distinctive marker, with a positive control so it cannot pass by testing an empty payload.
- **Store (15)** and **endpoint (7)**, including both cross-tenant negatives and an authenticated key with no bound tenant denied rather than served the Local partition.
- **Hygiene fold (16)**, each written as the harm it prevents.
- **The bar (2)**, in both directions: a wait past it is dropped, a wait just inside it is still reported. A bar set wrong the other way makes every genuine multi-day wait vanish, which is the same silence dressed as tidiness.

## Honest scope note

The hosted tenant is **thin** because the owner's fleet reports to the LOCAL Gateway, not the hosted one. This feed only enriches the hosted morning report once a hosted-enrolled Director registers real repositories - so this ships the mechanism and the correctness, not hygiene content in tomorrow's email.

### DIFF
--- a/src/CcDirector.ControlApi/ControlApiHost.cs
+++ b/src/CcDirector.ControlApi/ControlApiHost.cs
@@ -1,4 +1,4 @@
-using System.Net;
+﻿using System.Net;
 using CcDirector.Core.Configuration;
 using CcDirector.Core.Fleet;
 using CcDirector.Core.Sessions;
@@ -196,6 +196,7 @@ public sealed class ControlApiHost : IAsyncDisposable
     private Core.Activity.ActivityEventOutbox? _activityOutbox;
     private Core.Activity.ActivityEventProducer? _activityProducer;
     private ActivityEventUploader? _activityUploader;
+    private RepoStatePusher? _repoStatePusher;
     private TransientErrorAutoResume? _transientErrorAutoResume;
     private TerminalSessionRecorder? _sessionRecorder;
     private Core.Storage.TurnReviewLogger? _turnReviewLogger;
@@ -751,6 +752,20 @@ public sealed class ControlApiHost : IAsyncDisposable
         // prompt sink's pattern) and deletes an outbox record only on the Gateway's acknowledgement.
         _activityUploader = new ActivityEventUploader(() => _gatewayClient, _activityOutbox);
         _activityUploader.Start();
+        // The repo-state feed (issue #2118): every six hours, ship this machine's branches and worktrees
+        // to the Gateway so the morning report can make its stale-worktree and unmerged-branch
+        // recommendations. Only when a repository registry exists - with none there is nothing to report.
+        // Late-binds the Gateway client per cycle, exactly like the uploader above.
+        if (_repositoryRegistry is not null)
+        {
+            _repoStatePusher = new RepoStatePusher(
+                // Late-bound per cycle: the Gateway client is created after this and can be replaced when the
+                // configuration changes. A null client is an install with no Gateway - nowhere to report to.
+                (req, ct) => _gatewayClient is { } c ? c.PushRepoStateAsync(req, ct) : Task.FromResult<RepoStatePushResponse?>(null),
+                _repositoryRegistry, DirectorId, Environment.MachineName,
+                liveSessions: LiveSessionRefs);
+            _repoStatePusher.Start();
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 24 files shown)


thefrederiksen:
## Local test evidence

```
dotnet test src/CcDirector.Gateway.Tests --filter "RepoHygieneFold|MorningReport|RepoState|TenantRegistryEmailBackfill"

Passed!  - Failed: 0, Passed: 123, Skipped: 0, Total: 123, Duration: 24 s
```

```
dotnet test src/CcDirector.Core.Tests --filter "RepoStateSnapshotCollectorTests"

Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6, Duration: 15 s
```

The collector tests run against **real git repositories** (bare origin + clone + real worktrees), because the whole value of this feed is that it reports what git actually says - a stubbed git would prove nothing about "is this branch merged".

Both EF migration snapshots verify clean after the tenant-scope fix:

```
dotnet ef migrations has-pending-model-changes ... (SQLite)   -> No changes have been made to the model since the last migration.
CC_GATEWAY_EF_PROVIDER=postgres dotnet ef ... (Postgres)      -> No changes have been made to the model since the last migration.
```

### Why the 48-hour bar is in this pull request rather than a separate one

It was found by the deployed thefrederiksen/devthrottle#2126, not by review: the first real production call returned nine waiting rows aged 69-129 hours for sessions that were long gone. It is a change to the same report builder this PR already touches, and shipping it separately would mean a second deploy tonight for one condition. The two tests cover both failure directions, since a bar set wrong the other way makes every genuine multi-day wait disappear.

PR #2175, opened 2026-07-26
Returning to a language your old engine speaks restores that engine

thefrederiksen:
Choosing Danish moves the account to the multilingual engine, which is right. Choosing English again left it there, which is not.

Nothing was broken by that - the multilingual engine speaks English perfectly well. It just costs **61% more per character** for a language that never needed it, and nothing ever moved the account back. A round trip silently upgraded the bill.

| Engine | Rate card | Per million characters |
|---|---|---|
| Default | 1.24 u$/char | $1.24 |
| Multilingual | 2.00 u$/char | $2.00 |

At a ~500-character narration that is +$0.23/month for light use, +$1.14 steady, +$5.70 for a heavy voice user. Small money, but nobody should pay it by accident.

## What changed

The Gateway remembers what it switched away from, and restores it the moment the remembered engine can speak the language being chosen.

Three details that make it behave rather than merely work:

- **Restores the remembered model, not "the default."** A tenant who deliberately picked an engine gets that engine back, not ours.
- **Only the first auto-switch is recorded**, so hopping Danish to French to English still returns to the English engine originally left, not the multilingual one it was already sitting on.
- **Picking a model by hand clears the memo** - a deliberate choice supersedes anything we were holding.

## Verified on a running Gateway, the whole cycle

| Step | Result |
|---|---|
| Start on the default engine, English | `switched: false` |
| Choose Danish | -> multilingual, stale voice cleared |
| Back to English | **-> default engine restored, with its voice** |
| Danish, then French | stays multilingual (the default engine cannot speak French) |
| Then English | **-> default engine restored** |

Nine new resolver tests cover the stored state, including a corrupt memo reading as "nothing remembered" so a bad value can never break a language change.

### DIFF
--- a/src/CcDirector.Gateway.Tests/TenantSettingsResolverTests.cs
+++ b/src/CcDirector.Gateway.Tests/TenantSettingsResolverTests.cs
@@ -22,6 +22,102 @@ public sealed class TenantSettingsResolverTests
     private static TenantSettingsResolver NewResolver(GatewayDbTestHarness h)
         => new(new TenantSettingsStore(h.Open()));
 
+    // ---- the spoken language, and the model it moves ------------------------------------------
+
+    [Fact]
+    public void SpokenLanguage_Unset_IsEnglish()
+    {
+        using var h = new GatewayDbTestHarness();
+        Assert.Equal("en", NewResolver(h).SpokenLanguage(TenantA));
+    }
+
+    [Fact]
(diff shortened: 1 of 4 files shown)


thefrederiksen:
Superseded by #2177, which contains this branch plus the two client-side defects found while verifying it on a phone. Consolidated so it is one review and one hosted-gateway deploy instead of two.

PR #2185, opened 2026-07-26
Settings: one tabbed page shared by the phone and the Cockpit

thefrederiksen:
The two surfaces did not offer the same settings. The Cockpit had Notifications, AI and Car Mode, with the microphone and transcription checks on a different page entirely. The phone had a single untabbed "AI settings" scroll: no notification settings, no Car Mode end phrase, and the two checks linked out to screens of their own. Same account, two different products.

The page body now lives in `client-core/settings` and both shells mount it, so there is one tab set and one copy of every card.

## The tabs, identical on both surfaces

| Tab | What is on it |
|---|---|
| Notifications | snooze lengths, display time zone, notifications on this device |
| AI | hosted provider, thinking + fast models, spoken language, speech model, voice |
| Transcription | the transcription model, how your microphones are measuring, Test microphone, Test transcription |
| Car Mode | model, end phrase, live phrase tester |

Transcription is a tab because "dictation keeps coming out wrong" is its own problem and the four things that diagnose it were in four places. The Cockpit's Transcription Health page keeps the report over time and links across; it no longer carries a second copy of the checks. The Car Mode model left the phone's AI screen for the Car Mode tab, where the desktop already had it.

## What the phone gains

Snooze lengths, display time zone, notifications, and the Car Mode end phrase with its live tester - none of which it had. Its screen is titled "Settings" and is only a frame around the shared body; `styles.css` re-tunes the shared cards for touch under a `.screen` ancestor, so the layout differs and the content cannot.

`?tab=` resolves the same on both surfaces, and `/mic-test` and `/transcription-test` redirect into the Transcription tab, so existing bookmarks, home-screen shortcuts and any stale service-worker shell still land on what they asked for.

## Also in here

Neither dev config proxied `/voice-quality` or the transcription report endpoints, so those panels could not be exercised under `npm run dev`. The Cockpit proxies the three exact `/transcription` API paths rather than the prefix - its own Transcription page is routed at `/transcription` and would otherwise answer its own URL with JSON.

## Proof

Driven in a real browser against a stubbed Gateway, at phone size and desktop size:

- all four tabs render with content on both surfaces
- both retired phone links land on the Transcription tab with it selected
- Transcription Health still renders, keeps the microphone-quality panel, and no longer contains the checks

New rendering test (`SettingsTabs.test.tsx`) mounts the real panels, not stubs, and was checked by removing `MicTestPanel` from the Transcription tab on purpose and confirming it failed.

Typecheck clean across all three workspaces, both apps build, 694 client-core tests and 183 Cockpit tests pass.


thefrederiksen:
Rebased onto main after the spoken-language revert (#2181) landed. The shared AI tab no longer has the Spoken language control, its per-language sample sentences are gone, and `ttsSample` is called with the model and voice the page is showing - matching what #2181 left on the Cockpit. The Transcription tab's wording changed with it: dictation still understands every language, DevThrottle speaks back in English.

Re-verified after the rebase: typecheck clean across all three workspaces, both apps build, 694 client-core tests and 189 Cockpit tests pass.

PR #2195, opened 2026-07-26
About page: the Gateway, Cockpit and mobile app versions (the Director leaves)

thefrederiksen:
## Why

The About page was a relic of the days we hosted the Gateway ourselves. It reported the wrong things about the wrong product:

- **"Product: Director"** - a hard-coded string, on a page whose next row is labelled "Gateway version". The Director has its own About box, its own tray window and its own Cockpit screen; its facts do not belong here.
- **Install root** - it printed `C:\Users\<user>\AppData\Local\cc-director`, spelling out the operating-system user name of the Gateway box to any enrolled device that reads the endpoint.
- **Installed components** - the installer's manifest from `installed.json` on that box.
- **Machine name and run mode** - internal detail about infrastructure a hosted caller does not own.
- **Port 7878 printed directly beneath a public HTTPS address**, where the two do not compose into anything reachable. This is what made the page look broken.

## What it shows now

Three server-side products, and how the Gateway is reached.

| | |
|---|---|
| Gateway | version, build date |
| Cockpit | the bundle this Gateway serves: commit and build time |
| Mobile app | the bundle this Gateway serves: commit and build time |
| This Gateway | deployment, public address, Cockpit URL, uptime, Gateway time, and the listen port **self-hosted only** |

### The bundle versions needed a build stamp

The Gateway could not name either bundle it serves. Each app's commit is compiled into its minified JavaScript, which the server never executes, so the old page could only report the build of the bundle the browser was *already running* - never the mobile app's, and never from a server-side call.

Both builds now emit a `build.json` (commit and build time) into their bundle, staged into `wwwroot/c` and `wwwroot/mobile` by the existing recursive copies in the release MSBuild targets. `BundleStamp` reads it **per request and deliberately does not cache**: a Cockpit-only redeploy replaces the bundle under a live Gateway, so a cached stamp would keep naming the previous build - precisely the staleness the stamp exists to expose.

The **mobile app gained a build identity it never had anywhere**. Its About row showed the content-hashed script filename scraped out of the live page (`index-CUHLjBg5.js`) - a value that changed every build and named nothing you could look up.

The Cockpit page also keeps an always-visible line naming the bundle **this tab** is running. That is a different fact from the served bundle: when the two disagree, the tab is stale. It renders even when the Gateway call fails.

### Why the port is not 443, and why it is now omitted on hosted

7878 is correct and stays. Nothing outside the container ever connects to it: App Service terminates encryption on the public 443 and forwards inward, so no customer network or firewall sees 7878. Self-hosted, Tailscale Serve maps 443 to it the same way. Binding 443 inside the container would buy nothing, and on a self-hosted Windows box it would fight whatever else wants that port.

So the port was never the defect - **printing it was**. It is now served only when self-hosted, where it is a real reachable port. On hosted it is absent from the payload, so no client can render it. The Gateway decides; the page renders what it is given.

### Removed from the payload, not merely hidden

`Product`, `MachineName`, `InstallRoot`, `InstalledComponents`, `Directors`, `Mode`, `State` are gone from the response object, not just from the page. The install root is why this matters: it was readable by any enrolled device even while the page rendered it, so deleting the row alone would not have stopped it leaving the Gateway.

Also corrects a wrong `Dockerfile` comment: it claimed the Gateway binds loopback. It binds `0.0.0.0`, which is exactly what lets the App Service front end reach it.

## Proof

- Full `CcDirector.Gateway.Tests` suite: **still running at the time of opening**; will be reported in a comment below. Continuous integration also runs it on this branch.
- Other six .NET test projects, zero failures: Core 3710, Engine 62, Avalonia 305, Launcher 27, HostedAgent 86, Terminal.Avalonia 23.
- Cockpit vitest: 199/199 including 10 new About tests.
- New `BundleStampTests`: 6/6 - stamped, no bundle staged, missing root, corrupt JSON, commit-less stamp, time-less stamp.
- Targeted endpoint tests: 24/24 across the About, Cockpit-URL and public-URL suites.
- Zero-warning builds of Gateway, Gateway.Host, Gateway.Tests and Avalonia.
- Typecheck clean across all three npm workspaces.
- **Revert-proof**: reinstating the old unconditional port reddened the new hosted test with `Expected: null, Actual: 60875`. The test fails for the right reason rather than passing by accident.
- Verified the emitted stamps carry the correct commit, and that the mobile bundle now contains a commit literal where it previously had none.
- Both rendered variants checked in a browser: self-hosted shows `Listening on port 7878`; hosted shows no port row, no filesystem paths, and `Deployment: Hosted service`.

## Note

`packages/client-core/src/api/schema.ts` is a generated OpenAPI snapshot that still lists the old About fields. Nothing type-checks against it for this endpoint, and regenerating needs a live Gateway and would sweep in every unrelated endpoint change since the last run - so it is left exactly as stale as it already was.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -146,9 +146,19 @@ ENV COCKPIT_COMMIT=${COCKPIT_COMMIT}
 
 USER gateway
 
-# The Gateway binds to loopback (127.0.0.1:7878) by design - it is reached over the tunnel, not a public
-# port. Exercise it from inside the container (docker exec ... curl 127.0.0.1:7878/...); this EXPOSE is
-# documentation of the internal port, not a public bind.
+# 7878 is the container's INTERNAL port and nothing outside the container ever connects to it. App Service
+# terminates TLS on the public 443 and forwards to this port, so every real client reaches the Gateway over
+# https on 443 at CC_GATEWAY_PUBLIC_URL and no customer network or firewall ever sees 7878. That is why the
+# port does not need to be (and must not be) 443 here: the platform owns the public port, and binding a
+# privileged one inside the container would buy nothing.
+#
+# The bind is 0.0.0.0:7878, NOT loopback (GatewayHost.ConfigureKestrel listens on IPAddress.Any in both
+# hosted and self-hosted mode) - which is exactly what lets the App Service front end reach it. An earlier
+# version of this comment claimed loopback; it was wrong. Reachable-but-authenticated, not open: every route
+# except /healthz, /login and /logout requires a credential.
+#
+# Exercise it from inside the container with: docker exec ... curl 127.0.0.1:7878/healthz
+# This EXPOSE documents the internal port; it is not a public bind.
 EXPOSE 7878
 
 ENTRYPOINT ["dotnet", "CcDirector.Gateway.Host.dll", "--port", "7878"]
--- a/apps/cockpit/src/about/AboutView.test.tsx
+++ b/apps/cockpit/src/about/AboutView.test.tsx
@@ -1,0 +1,121 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { render, screen, cleanup, waitFor } from "@testing-library/react";
+import type { AboutInfo } from "@devthrottle/client-core/about/aboutClient";
+
+// The About page renders whatever the Gateway ruled (CLAUDE.md rule 7). These tests pin the two things a
+// regression would quietly get wrong: an unstamped bundle is REPORTED as unstamped rather than rendered as
+// a blank build, and the internal listen port appears only when the Gateway hands one over - the hosted
+// service omits it because it composes with nothing a caller can reach.
+const getAboutMock = vi.fn();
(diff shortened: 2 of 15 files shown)


thefrederiksen:
Full `CcDirector.Gateway.Tests` suite, run locally on this branch: **4105 passed, 0 failed, 17 skipped** in 24m39s.

That closes the one outstanding item in the proof list above. Every test project is now green on this branch:

| Project | Result |
|---|---|
| CcDirector.Gateway.Tests | 4105 passed, 0 failed, 17 skipped |
| CcDirector.Core.Tests | 3710 passed, 0 failed, 8 skipped |
| CcDirector.Avalonia.Tests | 305 passed, 0 failed |
| CcDirector.HostedAgent.Tests | 86 passed, 0 failed |
| CcDirector.Engine.Tests | 62 passed, 0 failed |
| CcDirector.Launcher.Tests | 27 passed, 0 failed |
| CcDirector.Terminal.Avalonia.Tests | 23 passed, 0 failed |
| Cockpit (vitest) | 199 passed, 0 failed |

PR #2236, opened 2026-07-27
feat(gateway): send hosted owner email by naming the tenant (internal #986)

thefrederiksen:
Stacked on #2233 (issue #984). Review that one first; this targets its branch, so the diff here is only the new work.

## What this closes

#2233 stopped `POST /account/email` telling a signed-in hosted user they were not signed in. It could not make it **send**: the original cloud primitive resolves the recipient from an account access token, and a hosted Gateway holds none - hosted enrollment validates the account token, mints a tenant and a device key, and stores neither.

devthrottle_internal #986 adds a tenant-addressed cloud route. The Gateway names a **tenant**; the cloud resolves the recipient server-side from that tenant's account subject, through the same authority the JWT route uses.

## The safety properties, and how each is held

**The Gateway cannot address anyone.** No recipient parameter, no recipient field on the wire, no code path that could add one. The cloud refuses a recipient field with a hard 400 rather than ignoring it - braces here, belt there. A test asserts eleven recipient-shaped key names can never appear in a constructed body.

**The account subject is a cross-check only.** Compared byte-for-byte cloud-side, so a wrong tenant id becomes a refusal instead of a correctly-delivered email to the wrong person. It never resolves the recipient: a subject that selects the target is a subject that could select someone else's.

**No user account token is stored, read, or forwarded** on this path. The property that hosted enrollment keeps neither token stays true - that was a hard constraint from the cloud side and it is preserved by construction, not by discipline.

**An unconfigured Gateway fails closed** rather than calling unauthenticated. This is load-bearing: if it called anyway, the cloud's 401 would mean either "wrong secret" or "no secret", and neither end could tell which. Refusing keeps that 401 diagnostic.

**The refusal names the signed-in identity**, exactly as the #984 fold does. That was missed on the first cut and caught by #984's own test going red when this path was added - the invariant test doing precisely the job it was written for.

**Rate limits keep their 429** and the cloud's `Retry-After` passes through untouched. It is computed from when the oldest hit in the window actually expires, so a backoff of our own would only fight it. No retry on any 4xx.

## Verification

Committed tests run a real HTTP stub of the contract, so they stay hermetic and CI-safe.

Separately - and this is the part worth knowing - the real client was driven against **the cloud side's own contract server**, which runs their production decision code (`api/_lib/notify-tenant.js`) and fakes only the tenant lookup, the owner lookup and the send. 24 checks passed, including:

- both directions of the cross-tenant refusal (tenant A id with tenant B subject, and the reverse) -> 403 `subject_mismatch`
- unknown tenant -> 404 `tenant_not_found`
- tenant exists but owner unresolvable -> 404 `owner_not_resolvable`
- `simulate-lookup-not-installed` -> 503 `tenant_lookup_not_installed`, the state production is actually in today
- a wrong service token -> 401 `gateway_auth_failed`
- the recipient resolved cloud-side matching the fixture owner
- all eleven recipient-shaped field names refused with `recipient_not_accepted`

Neither side's reading of the contract was the referee - their code was. A stub encodes the contract as its author read it; this does not.

301 account tests pass; the full Gateway suite was green at 4265/0/17 on the parent branch.

## What this does NOT prove

**Not end-to-end.** The contract server fakes the lookup and the send, and the tenant-resolution migration is not applied on any environment. Two manual configuration steps remain, both needing credentials no automated session holds:

1. **Supabase SQL editor** - apply the migration creating the tenant-resolution function and its grant. Until this lands the cloud route authenticates and then answers 503 `tenant_lookup_not_installed`.
2. **Azure App Service** - add `NOTIFY_OWNER_SERVICE_TOKEN` to the hosted Gateway, matching the Vercel value. Until this lands the Gateway will not call the route at all.

Neither side misbehaves half-configured: the Gateway fails closed without its secret, and the route refuses to resolve without the function. There is no window where something is half-wired and doing something odd.

Pairs with devthrottle_internal #987.

### DIFF
--- a/src/CcDirector.Core/Account/AccountNotifyByTenantClient.cs
+++ b/src/CcDirector.Core/Account/AccountNotifyByTenantClient.cs
@@ -1,0 +1,229 @@
+using CcDirector.Core.Network;
+using System.Net.Http;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using CcDirector.Core.Utilities;
+
+namespace CcDirector.Core.Account;
+
+/// <summary>
+/// The outcome of a TENANT-ADDRESSED owner-email send (devthrottle_internal #986).
+/// </summary>
+/// <param name="Sent">True only when the cloud accepted and sent. Never inferred from a status alone.</param>
+/// <param name="ProviderId">The provider message id when the cloud returned one; may be null even on a send.</param>
+/// <param name="Error">The cloud's own human-readable message, to be surfaced VERBATIM. Null on success.</param>
+/// <param name="StatusCode">The cloud HTTP status (0 when the request never got a response).</param>
+/// <param name="ErrorCode">
+/// The cloud's machine-readable code (<c>tenant_not_found</c>, <c>owner_not_resolvable</c>,
+/// <c>subject_mismatch</c>, <c>notify_rate_limited</c>, <c>tenant_lookup_not_installed</c>, ...). Used for
+/// LOGGING and for the retry rule only - never for composing a user-facing sentence, which is what
+/// <see cref="Error"/> already is.
+/// </param>
+/// <param name="RetryAfterSeconds">
+/// The cloud's <c>Retry-After</c> in seconds on a 429. Authoritative: the caller waits at least this long and
+/// applies no backoff of its own, because the cloud computes it from when the oldest hit in the window
+/// actually expires.
+/// </param>
+public sealed record TenantNotifyResult(
+    bool Sent, string? ProviderId, string? Error, int StatusCode, string? ErrorCode, int? RetryAfterSeconds);
+
+/// <summary>
+/// The Gateway's client for the TENANT-ADDRESSED owner-email primitive
+/// (<c>POST /api/v1/account/notify-owner-by-tenant</c>, devthrottle_internal #986). It exists because the
+/// hosted Gateway holds NO account access token for its tenants and must never start holding one: hosted
+/// enrollment validates the account token, mints a tenant and a device key, and stores neither. So the
+/// existing JWT-authenticated <see cref="AccountNotifyClient"/> cannot be used on hosted at all - that gap is
+/// what made <c>POST /account/email</c> report a signed-in user as signed out (issue #984).
+///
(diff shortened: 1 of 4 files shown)


thefrederiksen:
## If the first live run does not return 200: which side to look at

Mirrored from the triage table in devthrottle_internal #987 so it is reachable from this repo too - whoever debugs a bad first run may be standing in front of the Gateway side with no access to the cloud pull request.

Every response code names exactly one side. That is by design: the codes were kept deliberately separable during the contract negotiation precisely so a bad run would not need guesswork.

| Response | Which side | What it means |
|---|---|---|
| `503 tenant_lookup_not_installed` | cloud config | The migration has not been applied. |
| `503 gateway_service_auth_not_configured` | cloud config | The cloud's secret is unset, too short, or set but not redeployed - Vercel bakes env into the build. |
| `401 gateway_auth_failed` | both ends, different values | Cannot mean "missing": this Gateway fails closed and never calls without its secret. So it means the two ends hold different values. |
| `404 tenant_not_found` | Gateway | The lookup **ran** - so resolution is live - and the id sent is not in `gateway.tenants`. |
| `403 subject_mismatch` | Gateway | The tenant id and account subject came from different places. Both are read off the same row here, so this should be impossible. |
| `404 owner_not_resolvable` | neither - data | Tenant is real, no address behind it. Do **not** fall back to `tenants."Email"`; that is a display value recorded at mint time and allowed to be null, not a delivery address. |
| `429` | neither | Honour `Retry-After` and add no second backoff - it is computed from when the oldest hit in the window expires. |
| `502` | provider | The mail provider rejected it; the reason is carried in the message. |
| `500` | cloud | The only code that is nobody's configuration: the lookup was reached and the database was unreachable or the query failed. |

A `503 gateway_service_auth_not_configured` and a `401 gateway_auth_failed` being separable is the line doing the most work here. Collapsing them would have made "your secret is wrong" and "the cloud has no secret" indistinguishable from either end - the same defect as `#984` itself, one layer down.

The Gateway's own unconfigured state is separate again: it answers 503 locally with a message naming `NOTIFY_OWNER_SERVICE_TOKEN` and never calls out, so it can never be confused with any row above.

PR #2260, opened 2026-07-29
Hold a whole skill directory, and put it where each agent looks

thefrederiksen:
## What this is

Two things, which together are what makes a central skill library actually usable by an agent on a machine: the Gateway can hold a **whole standard skill directory**, and the Director **puts it where each agent already looks**. A skill that carries Python or an executable cannot be run from the Gateway - it has to reach the disk.

### The store

- File identity becomes a relative path, validated per segment (no empty segment, no dot or double-dot segment, no backslash, no drive letter, no reserved Windows device name, at most five segments deep) rather than by forbidding the separator outright.
- The extension allow-list becomes a small deny-list, because an allow-list cannot be completed when the specification permits any file.
- Files may be binary (an explicit `utf8` or `base64` encoding discriminator, size cap applied to decoded bytes) and may be marked executable.
- The standard's frontmatter fields are stored rather than invented, so a skill authored elsewhere survives a round trip through the library unchanged.
- Matching migrations for both database providers. The new migration timestamps after `AddSkillLibrary`, the latest on main, so the chain is intact.
- The Cockpit can author the files.

### The placement

Every skill is materialized **exactly once**, into the shared `~/.agents/skills` that six of the eight agent families read natively with no configuration at all. Claude Code and Cursor, which do not read it, get **one link per skill** pointing into it - a directory junction on Windows (no elevation needed), an ordinary symlink on Linux and macOS. The agent's own skills directory is never replaced, only named entries inside it, so skills already living there are untouched. A full copy left by the earlier three-directory scheme is rebuilt as a link on the next launch.

### Freshness, in two halves

- **Network half, off the launch path.** `SkillStoreRefresh` pulls every enabled skill from the Gateway when the connection goes green and once a minute thereafter, reconciling rather than accumulating.
- **Launch half, synchronous and local.** `SkillDirectoryInstaller.InstallFor` reflects the store into the shared directory and reconciles the links, with no network call, so it cannot slow or block a session launch.

A session therefore opens with skills at most about a minute old, and a skill switched off on the Gateway is gone from disk the next time any session starts.

## Proof

Claude Code discovering a skill through a Windows junction, run unelevated against the real binary. A skill was written only to `~/.agents/skills`, carrying a token that appears **only in the body** of its `SKILL.md` and never in the description - so reciting the token proves the body was loaded, not merely that a name was listed.

```
Control, no junction:  -> "I don't have a skill named junction-proof-skill.  NO SUCH SKILL"
Junction created:      -> "Yes - the junction proof token is:  QX7-VELVET-ANVIL-4419"
```

The control is half the proof: it establishes the skill was unreachable before the junction existed, and independently confirms Claude Code does not read the shared path - which is the entire reason the link exists.

The rest is covered by test: that the entry created in the agent's own directory is a reparse point reading back the same bytes as the one real copy, that a second launch leaves one working link rather than a broken one, that a copy left by the previous scheme is rebuilt as a link, that a withdrawn skill is removed from both places while a skill we did not write is left alone, and that removing one of our links does not empty the one real copy.

Full `CcDirector.Core` suite green: 3812 passed, 0 failed.

## Owner decisions recorded on this pull request

- **Executable content: allowed.** Confirmed 29 July 2026. The store already held and marked executables, so this confirms the built behaviour rather than changing it.
- **Install path: one directory, using the agents standard.** Decided 29 July 2026. Written up in `docs/architecture/skill-placement-2026-07-29.html`, with the per-agent path table and the junction proof above.

## After merging

The materializing happens in the Director, so the Director has to be rebuilt and restarted before any of this takes effect on a machine.


thefrederiksen:
Owner decision, 29 July 2026: **skills may carry executable files.**

This answers the policy question the plan said had to be settled before the feature exists. No code change follows from it - the change as built already stores and marks executables, so this confirms the built behaviour rather than altering it.

The reasoning: the point of the feature is that a skill can carry a script, there is one tenant publishing skills today, and the safeguards are real - the Cockpit shows which files a skill carries and which are executable before anyone publishes, and the whole tree is hash-verified before anything is written to disk. A per-tenant deny-executables policy can be added later; a store that cannot hold them at all is the harder thing to undo.

The install-path question from the plan is still open.

thefrederiksen:
## Placement decided: one directory, using the agents standard

Owner decision, 29 July 2026. Skills are materialized **once**, into the shared `~/.agents/skills`. Claude Code and Cursor get one link per skill pointing into it. Full write-up added to this branch as `docs/architecture/skill-placement-2026-07-29.html`.

### The fact that changes the design

This change currently **copies** each skill into up to three per-agent directories. The recorded reason for choosing copies over links was that symlinks need elevation on Windows. That is true - and incomplete. A directory **junction** does not.

Verified on Windows 11, unelevated, 29 July 2026:

```
Elevated: False
SYMLINK:  FAILED -> Administrator privilege required for this operation.
JUNCTION: OK (readable through it)
```

So the objection that produced three copies does not stand. Three copies is three things that can drift; one copy cannot disagree with itself.

### Why the shared path rather than a DevThrottle-owned one

Six agents (Codex, Gemini, Grok, Copilot, opencode) plus pi read `~/.agents/skills` natively and need **nothing at all**. Inventing `~/.devthrottle/skills` would take seven agents that need zero configuration and give all eight something that can be misconfigured. The failure surface becomes one link instead of eight.

### Still to do on this branch

- `SkillInstallTargets` / `SkillDirectoryInstaller`: materialize once into the shared path; create a per-skill junction (Windows) or symlink (Linux, macOS) for Claude Code and Cursor. Per **skill**, never replacing the agent's own skills directory, so the owner's own skills are untouched.
- Reword the session preamble line in `SkillIndexStore.cs:189`. "Nothing is installed on this machine" stops being accurate once files are written there, and that sentence is injected into every agent in every session. The freshness half stays - it is the point.
- One test worth running before this ships: watch Claude Code actually discover a skill through a Windows junction. Reading through a junction is verified; the agent doing so is not.

PR #2272, opened 2026-07-29
Reduce unnecessary GitHub builds

thefrederiksen:
## Why

Remote pushes currently start long shared builds even when an older run is already obsolete or the change only contains documentation and agent instructions.

## Changes

- Teach repository agents to batch implementation, tests, and proof into one remote push.
- Cancel superseded pull-request runs without cancelling main-branch or release runs.
- Use a fail-closed changed-path decision to avoid unrelated product and web builds.
- Keep draft pull requests on the inexpensive decision path until they are ready.
- Provide one final continuous integration decision suitable for the main-branch ruleset.

## Local validation

- Seven changed-path classifications passed.
- The workflow configuration parsed successfully.
- Updated skill metadata passed validation.
- The committed difference passed the whitespace check.

### DIFF
--- a/.claude/skills/commit/skill.md
+++ b/.claude/skills/commit/skill.md
@@ -31,6 +31,13 @@ The four supporting rules (see the TRUNK DEVELOPMENT section of the global CLAUD
   concurrent agent uses its own worktree off origin/main.
 - Never commit directly on main. If the checkout is on main, create a branch first.
 
+## One remote build request per coherent batch
+
+Local commits are checkpoints; a push requests shared GitHub validation. Complete all known code,
+tests, proof, and applicable local checks before the first push. Do not push intermediate progress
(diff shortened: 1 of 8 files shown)


thefrederiksen:
Closing as stale, not as rejected - and deliberately WITHOUT deleting the branch.

The intent is sound and the problem is still real: unnecessary continuous-integration builds cost money and time, and this repository has felt both. But the branch is a hundred and forty-one commits behind main and has not been touched since 29 July, and continuous integration itself has changed underneath it since - the release pipeline was rewritten twice, the download mirror moved to a stable address, and the local test suite became the everyday gate rather than continuous integration. A change-classifier written against July's workflows is unlikely to classify August's correctly.

`ci/throughput-review` stays on origin. If build throughput is picked up again, start from today's workflows; the reasoning here is worth reading first.

Closed while sweeping the open list toward a release. Reopening costs nothing if it is wanted.

PR #2285, opened 2026-07-30
The launcher owns the Director's update: stop, swap, start, confirm - and roll back on both platforms

thefrederiksen:
Implements thefrederiksen/devthrottle_internal#1033.

## The principle, and what it forced

A process cannot replace its own binary and then vouch for what came back, because the only thing that could check is the process that just exited. That was not a bug inside the Director's updater, it was the shape of it: `ApplyUpdate` swapped the build, called `Process.Start`, logged `complete`, and called `FileLog.Stop()`. A relaunch that never came up was recorded as a success and then logging ended - no witness, no record, nothing on the machine that knew anything was wrong.

The launcher outlives the swap, so it can hold the question open until there is a real answer, and act on the answer.

## What the launcher now does

Its existing update loop gained a second job. When a Director update is staged and the Director holds no sessions, the launcher stops the Director, swaps the build, starts it, and keeps asking until the **new version** answers on the Director's own health endpoint. A build that never answers is rolled back to its predecessor, which is then started and confirmed too, and the bad version is pinned so it is not offered again.

No prompt. Nothing is lost and nothing is interrupted, so there is nothing to wait for.

The gate is the rule that was already right, moved to where acting on it is safe: a Director holding live work is never restarted out from under it. One case is decided differently from the strict reading of the issue - a Director that is **not running** is left alone rather than started, because updating it would mean reopening an application somebody deliberately closed, and installing a build with nothing running to witness whether it works. A closed Director picks the update up on its next start, which is a start somebody asked for.

**Certification is by version, never by liveness.** A staged update is by definition newer than what was running, so a health answer carrying the new version cannot have come from the old build - where "something answered on the port" could. This is the lesson `LauncherHealthProbe` already carries, applied here from the start.

## The macOS hole is closed, at its source

`RecoverHalfAppliedSwap` and `TryRollBackFailedUpdate` both began `if (!OperatingSystem.IsWindows()) return null;`. Removing those guards alone would have achieved nothing: `SwapMac` **deleted** the installed bundle and moved the new one in, so on a Mac there was never a backup for a rollback to have used. The backup was the missing half of the hole.

So the swap now keeps a backup on both platforms, through a new bundle-aware `DirectorBuildSwapper`, and one rule drives both recovery paths. The bundle branch is written in portable file operations, so **the bundle swap and rollback are exercised on Windows in continuous integration** - a hole that only opens on macOS cannot be proved closed by code that only runs on macOS.

## What the Director stops doing

It no longer decides to replace itself while running. The idle auto-apply, the session-ended trigger and `ShouldAutoApplyWhenIdle` are gone; it stages, shows "Restart now", and stops there.

**One deliberate exception, please push back if you disagree.** The startup apply path stays, for a start a person or the launcher already initiated. Deleting it would permanently strand any install whose launcher is missing or broken - the thing that fetches every future fix would be gone, and no later release could repair that machine. It is a different event from a running Director deciding to swap itself: no session exists yet, and somebody asked for this start. It is also hardened rather than left as it was:

- it waits for the relaunched build to report healthy before calling anything complete, and says `NOT VERIFIED` and returns non-zero when it does not (point 5 of the issue);
- the startup cleanup keeps the backup while a health check is still pending, closing a window in which a build could start, delete the only way back, and then die before proving itself;
- the build that proves healthy deletes its predecessor's backup itself, so backups do not accumulate.

Unattended updating is **preserved, not removed** - the unattended-update capability from devthrottle_internal#2047 moved from the Director to the launcher.

## Two traps that would have failed silently

**Where the staged record lives.** `UpdaterState.FilePath` resolves against the calling process's own storage home. The launcher runs at the storage root; the installed Director keeps its whole home one level in, under its instance folder. A launcher that simply asked for "the" updater state would have read an absent file and concluded, every single time, that nothing was staged - wired, logged, and never once firing. It scans every instance home, the same way instance registrations are scanned, and this is pinned by a test.

**Which backup name the launcher uses.** The freshly started Director deletes `.old` during its own startup cleanup. Had the launcher kept its rollback material there, a build that started far enough to run cleanup and then died would have destroyed the only way back *while the launcher was still waiting to find out whether it was healthy*. The launcher's backup is `.prev`, it owns it, and it removes it itself. Also pinned by a test.

## Proof

**The negative test is the deliverable and it was verified to fail.** Two separate deliberate sabotages, each reverted:

| Sabotage | Result |
|---|---|
| rollback made a no-op (`restored = true`) | 5 of 9 tests red |
| success reported without waiting for the version (`answered = newVersion`) | 6 of 9 tests red |

The health stand-in in those tests is not a boolean that says yes: it **reads the installed build off disk** and answers with the version it finds there, and only once started. So an assertion about what answered is an assertion about what was actually swapped in, and in what order.

Covered in both install shapes - single executable (Windows) and application bundle (macOS): the new build comes up; it never comes up; something answers but not as the new version; there is no backup to restore; the staged build is missing; and the launcher's backup is not the one the Director deletes.

Windows is not traded for macOS. The Windows file swap is the same sequence as the `SwapWindows` it replaces - copy beside, delete the old backup, atomic replace keeping a backup - and the auto-update that carried this machine from 1.8.0 to 1.8.6 is unchanged in mechanism.

**Full suite, as continuous integration runs it (`dotnet test cc-director.sln`): see the comment below for the run against this exact commit.** Note for anyone re-running it: the Gateway suite takes a fleet-wide per-user lock, so a concurrent run in another worktree makes this queue for many minutes before starting. That is not a hang.

## Not done here

- The rescue install (a launcher that downloads and places a Director when none is installed) - noted in the design document as still unbuilt.
- A preference for being asked before updating. The owner's ruling is that the default is to just do it; a preference can be added later.
- Surfacing "am I up to date" and an on-demand check (devthrottle_internal#1030). The launcher is now the thing that holds that answer, which is what that issue needs, but no interface for it is added here.

## Reconciliation

That issue sits **inside** thefrederiksen/devthrottle_internal#578 and completes its last unmet acceptance point; the decision is written on both issues. Four of the five capabilities it lists already shipped, which is why this was wiring rather than new machinery; what remains with 578 afterwards is only the item it marks optional, a watchdog that restarts a Director that died without being asked.

The parked mutual-update design in `docs/plans/cc-launcher-mac-tunnel-and-mutual-updates.md` agreed with this on the backup, the health check, the rollback and the version pin - all built as it described. It differs on one point, and the document has been annotated rather than left to mislead: it had the launcher call `director/restart` and let `UpdateInstaller` apply the staged build during that restart, "the launcher never fights UpdateInstaller". Handing the swap back to the Director is the exact thing this change removes.


thefrederiksen:
## Test result against this exact commit (fa1ca8c1)

All three test steps continuous integration runs (`ci.yml`), all green:

| Step | Result |
|---|---|
| `dotnet test cc-director.sln` | **9035 passed, 0 failed**, 25 skipped |
| `dotnet test tools/cc-director-setup-engine.Tests/ -c Release` | **429 passed, 0 failed** |
| `dotnet test tools/cc-director-setup.Tests/ -c Release` | **21 passed, 0 failed** |

Per project in the solution: Core 3931, Gateway 4450, Avalonia 313, HostedAgent 88, Launcher 83, Engine 63, Terminal.Avalonia 24.

Worth knowing for anyone re-running this: **the new rollback tests are in the setup-engine project, which is NOT in `cc-director.sln`** - it is the separate step above. Running only the solution would report a green that never executed them. The nine of them are named individually in Release below.

<details>
<summary>The nine apply and rollback tests, Release configuration</summary>

```
Passed DirectorUpdateApplyTests.NewBuildComesUp_Swaps_ReportsUpdated_AndRemovesTheBackup
Passed DirectorUpdateApplyTests.NewBuildNeverComesUp_RollsBackToThePreviousBuild_AndDoesNotReportSuccess
Passed DirectorUpdateApplyTests.SomethingAnswersButNotAsTheNewVersion_StillRollsBack
Passed DirectorUpdateApplyTests.NoPreviousBuildToRestore_ReportsFailed_AndSaysTheRollbackWasNotPossible
Passed DirectorUpdateApplyTests.StagedBuildIsMissing_ReportsFailed_AndPutsTheDirectorBack
Passed DirectorUpdateApplyTests.TheLauncherBackupIsNotTheOneTheDirectorsCleanupDeletes
Passed DirectorUpdateApplyTests.Bundle_NewBuildComesUp_SwapsTheWholeBundle
Passed DirectorUpdateApplyTests.Bundle_NewBuildNeverComesUp_RollsBackTheWholeBundle
Passed DirectorUpdateApplyTests.Bundle_SomethingAnswersButNotAsTheNewVersion_StillRollsBack
```
</details>

## The read path, run against this machine's live state

The tests above prove the swap and the rollback. They cannot prove that the launcher will find a real staged update on a real machine - so that was run separately, through the production code (`FindStagedUpdate`, `ReadHealthAsync`, `ShouldApply`), read-only, with no call to `RunOnceAsync`:

```
installed Director path: C:\Users\soren\AppData\Local\cc-director\app\cc-director.exe   (exists=True, running=True)
candidate state file: ...\cc-director\config\director\updater-state.json                 (exists=True)
candidate state file: ...\cc-director\instances\default\config\director\updater-state.json (exists=True)
FindStagedUpdate: version=1.8.7,
  build=...\instances\default\config\director\updates\1.8.7\cc-director-win-x64.exe,
  target=...\cc-director\app\cc-director.exe,
  from=...\instances\default\config\director\updater-state.json
ReadHealthAsync: version=1.8.6, sessions=15, pid=24952, port=7879
ShouldApply(staged, 15 sessions) = False
```

Three things are established there, none of which were assumptions:

1. **The discovery trap was real, and is closed.** Both candidate files exist on this machine. The launcher's own home holds no staged version; the instance home holds a genuine staged **1.8.7**. A launcher reading only its own home would have reported "nothing staged" while a real update sat waiting - and looked perfectly healthy doing it.
2. **The health contract is what the code parses.** The live answer is `{"status":"ok","directors":1,"sessions":15,"version":"1.8.6",...}` - so the version certification has something real to compare, and 1.8.6 is correctly not 1.8.7.
3. **The gate holds.** Fifteen sessions are running on this Director, and the decision is `False`. No session is interrupted to update.

One incidental finding while doing this, worth recording because it is a trap for anyone testing the launcher by hand: a process started from inside an agent session inherits `CC_DIRECTOR_ROOT` from the Director that spawned it, so it resolves the storage root one level deep (`...\cc-director\instances\default\app\...`) and finds nothing. The first run of the probe did exactly that. In production the launcher starts from autostart and has no such variable, which is why the second run - with it cleared - resolves the true root. It is a testing artifact, not a product defect, and nothing in the Director starts the launcher.

## The negative test was verified to fail

Neither of these is a claim about the code being careful; both were run.

| Sabotage applied to `DirectorUpdateApply` | Result |
|---|---|
| rollback made a no-op (`var restored = true;`) | **5 of 9 red**, including both the file and the bundle rollback |
| success reported without a witness (`var answered = newVersion;`) | **6 of 9 red** |

Both reverted; the file in this pull request contains neither.

PR #2289, opened 2026-07-30
A held session is listable, so the ghost card can be reaped

thefrederiksen:
Fixes thefrederiksen/devthrottle_internal#1019.

> Note on the reference: `devthrottle#1019` in this repository is a different, unrelated issue, so the link is fully qualified deliberately. A bare `#1019` would have pointed at the wrong one.

## What was actually wrong

The session was never lost. It was **held, displayed, and denied** — and the denial came from one list.

`GET /fleet/sessions` is the roster every `cc-devthrottle` verb resolves a target against *before it sends anything*. That makes it the reap floor: a session this roster omits cannot be named, and a session that cannot be named cannot be interrupted, messaged, or marked done. Two independent omissions put rows there.

**Omission one — the roster hid exactly the rows that need reaping.** The listing filtered out `ActivityState.Exited`. A crashed session sits in precisely that state (a crash was never modelled as its own state; it lives on the separate `Session.Crashed` fact) and is **deliberately kept** in the roster and on the rail by issue #959, so the user sees that work stopped. So the one row the user could *see* and most needed to clear was the one row the command line could not *name*. This needs no Gateway at all.

**Omission two — with a Gateway configured, the Director served the relayed roster only.** That relay silently drops the sessions of a Director it cannot reach while still returning 200 (documented on `GatewayClient.ListFleetSessionsAsync`). A Director whose registration was failing — which is what the report's logs show — therefore vanished from its own fleet listing, live sessions and all.

Both explain the report's four "verified absent" rows with one mechanism, and they explain why the card stayed on screen: the rail renders the local store, which was right, while every tool read a roster that had dropped the row.

**The reap machinery was always willing.** `/fleet/done` looks the session up in the local store and would have found it; `ReapPendingDeletions` explicitly reaps an `Exited` session. Only the naming was impossible. That is why "restart the Director" was the only remedy.

## What changed

1. **`GET /fleet/sessions` lists every session this Director holds** — the store is the same store the rail renders, so a row the user can see is now a row the command line can name.
2. **This Director's own sessions are unioned into the relayed roster**, with the Gateway's copy winning for any session it already knows, so the session numbers and identity stamping it hands out are never overwritten. When the relay omitted rows, that is now written to the Director log naming the count and the ids — the diagnostic the report found missing.
3. **A failed create no longer leaves debris or silence.** `SessionManager.CreateSession` reserves the worktree just before `backend.Start`, but the matching release is wired in `RaiseSessionCreated`, which runs only once the session is in the roster. A throw in between left a permanent reservation held by a session id that no longer existed anywhere, silently blocking the worktree reaper. It is released now, and the failure is written through `FileLog` rather than only the optional `_log` that several callers never supply.
4. **`session list` marks a crashed row as crashed.** Listing dead rows is what makes them reapable, and the STATUS column reads `activityState`, so a crash would have printed `Exited` — byte-identical to a session that finished on purpose, the exact misreading #959 was filed for twice. The raw crash fact is now surfaced. The client renders the fact and does not rule on it; deciding what a state *means* stays in the Gateway fold.

## Against the issue's four asks

1. *Spawn must not return an id unless the session was actually created.* Already true on the local leg, and now true end to end in the sense that matters: the id spawn returns is an id every verb can resolve. See the correction below.
2. *Roll back the UI card when the start fails, or never add it until the session is confirmed.* Already the behaviour — the rail is only ever populated from `RaiseSessionCreated`, which runs after the store insert, so a card cannot precede a confirmed session. No change needed, and I have not made one.
3. *Log the failure path.* Done — the failed create writes through `FileLog`, and an omitted roster row is now logged with its count and ids.
4. *`session done` should succeed on a ghost.* Done, at the root rather than by special-casing: the ghost is now nameable, so the reap path that always worked can be reached. `CrashedSession_isReapableEndToEnd` walks it.

## One correction to the issue

Defect 1 says spawn "returns a session id for a session that was never created". The create path does not do that: the local leg of `/fleet/spawn` re-reads the session out of the store after the executor returns and fails with a 500 if it is absent (`ControlEndpoints.CreateLocalSessionAsync`). The session in the report **was** created — that is why a card appeared at all, since the rail is only ever populated from `RaiseSessionCreated`, after the store insert.

What actually happened is worse in a way the report captures correctly: spawn returned an id that no read path would then accept. That is fixed here at the root — create and read now agree by construction, so the id spawn hands back is an id every verb can resolve. I have not invented a guard for a code path that does not exist.

**Left deliberately unfixed, and worth its own issue:** the remote (`--machine`) leg can time out at 30 seconds (`DirectorCommandRouter.DefaultCommandTimeout`) while the target Director goes on to create a perfectly live session. The caller is told it failed and never learns the id. That is the inverse leak and needs an idempotency key or a reconciliation pass, not a patch here.

## Proof

**The regression tests fail on purpose.** With the `ActivityState.Exited` filter reinstated and everything else unchanged, the three endpoint tests fail with the roster empty while the Director holds the session:

```
Failed GhostSessionReapableTests.CrashedSession_theDirectorStillHolds_isListedSoTheCliCanNameIt
  Assert.Contains() Failure: Filter not matched in collection
  Collection: []
Failed GhostSessionReapableTests.CrashedSession_isListedAsCrashed_notSilentlyDressedUpAsLive
Failed GhostSessionReapableTests.CrashedSession_isReapableEndToEnd_soTheCardCanBeCleared
Total tests: 10   Passed: 7   Failed: 3
```

The tests drive the real producer — a backend raising a genuine process exit into `Session.OnBackendProcessExited`, so nothing sets `Crashed` by hand and the test cannot pass on a fact production never emits — then the real endpoint, the real `/fleet/done`, and the real `ReapPendingDeletions`. `CrashedSession_isReapableEndToEnd` walks the whole supported route the report found impossible: list it, mark it done, sweep, and confirm the row is gone.

`UnknownSession_stillFailsLoud` is the counterpart: widening the roster must not make the Director accept an id it has never seen, and a genuinely unknown target with no Gateway is still a clear 404.

The command line change is pinned the same way — it renders a bare `Exited` without it.

**The command line tests are green:** `157 passed` in `tools/cc-devthrottle` (153 before, plus the 4 added here).

### What is NOT yet proven, stated plainly

`dotnet test cc-director.sln` ran with the fix in place and **six of the seven projects are green with zero failures** — `83, 83, 63, 313, 24, 88` passed, and `3931 passed / 8 skipped` on Core.Tests. The 8 are pre-existing deliberate skips, not failures; there is not one `Failed` line in the run.

The seventh project, `CcDirector.Gateway.Tests` — the one holding the tests above — **did not run.** Its per-user suite lock was held continuously by other worktrees (`dt-tools-1002`, then `devthrottle.wt-audit5`), so after 45 minutes it refused to start rather than run alongside a live holder, which is that lock working as designed. `vstest` reports that refusal as `Test Run Aborted`, which is why the solution run exits non-zero.

So: **these tests are proven to fail without the fix, and have not yet been observed passing with it.** Those holders are legitimately working, so I have not ended anyone else's run to get the lock. A retry is waiting for the lock to go free; I will post the result here as a comment, pass or fail, and this should not merge until that green is on the record.

### DIFF
--- a/src/CcDirector.ControlApi/ControlEndpoints.cs
+++ b/src/CcDirector.ControlApi/ControlEndpoints.cs
@@ -371,17 +371,43 @@ internal static class ControlEndpoints
             return (answer, outcome);
         }
 
-        // GET /fleet/sessions - the fleet directory. With a Gateway, relay its aggregated list;
-        // standalone, serve this Director's own sessions (the no-Gateway acceptance criterion).
+        // GET /fleet/sessions - the fleet directory, and the roster EVERY cc-devthrottle verb resolves a
+        // target against before it sends anything. That makes it the reap floor: a session this roster
+        // omits cannot be named, so it cannot be interrupted, messaged or marked done. Issue #1019 - a card
+        // that no tool could remove, with restarting the Director as the only remedy - was two independent
+        // omissions from this one list.
         app.MapGet("/fleet/sessions", async (CancellationToken ct) =>
         {
+            // Omission one: EVERY session this Director is still holding belongs on the roster, including
+            // crashed and exited-pending-reap rows. This store is the same store the desktop rail renders,
+            // and a crashed session is deliberately KEPT in it (issue #959) so the user sees that work
+            // stopped - in ActivityState.Exited, because a crash was never modelled as its own state. The
+            // filter that used to sit here dropped exactly that state, so the one row the user could SEE
+            // and most needed to clear was the one row the CLI could not NAME. The reaper on the other
+            // side was always willing: /fleet/done finds a session in this store and ReapPendingDeletions
+            // removes an Exited one. Only the naming was impossible.
+            var own = sessionManager.ListSessions()
+                .Select(s => MapWithIdentity(s, turnSummaryCache))
+                .ToList();
+
             var gw = gatewayClientProvider?.Invoke();
             if (gw is { IsEnabled: true })
             {
                 try
                 {
                     var fleet = await gw.ListFleetSessionsAsync(ct);
-                    return Results.Json(fleet);
+                    // Omission two: the relayed list silently DROPS the sessions of a Director the Gateway
+                    // cannot reach while still returning 200 (see GatewayClient.ListFleetSessionsAsync).
+                    // A Director whose registration is failing therefore vanishes from its OWN fleet
+                    // listing, which is how three live local sessions came to be denied by the CLI and the
+                    // Control API at once while the rail still showed them. A Director is first-hand
+                    // authority on its own machine, so its rows go back in. The Gateway's copy WINS for any
+                    // session it already knows: the session numbers and identity stamping it hands out are
+                    // never overwritten here.
+                    var (roster, restored) = UnionOwnSessions(fleet, own);
+                    if (restored.Count > 0)
+                        FileLog.Write($"[ControlEndpoints] /fleet/sessions: the Gateway roster omitted {restored.Count} of this Director's own session(s); served from the local store: {string.Join(", ", restored)}");
+                    return Results.Json(roster);
                 }
                 catch (Exception ex)
                 {
@@ -391,11 +417,7 @@ internal static class ControlEndpoints
                 }
             }
 
-            var local = sessionManager.ListSessions()
-                .Where(s => s.ActivityState != ActivityState.Exited)
-                .Select(s => MapWithIdentity(s, turnSummaryCache))
-                .ToList();
-            return Results.Json(local);
+            return Results.Json(own);
         });
 
         // GET /fleet/repositories + /fleet/worktrees - the repository/worktree directory (#510 phase C).
@@ -1446,6 +1468,42 @@ internal static class ControlEndpoints
         => gatewayEnabled ? HoldRoute.Gateway
          : sessionIsLocal ? HoldRoute.LocalMirror
          : HoldRoute.NotFound;
+
+    /// <summary>
+    /// Fold this Director's OWN sessions into the fleet roster relayed from the Gateway (issue #1019), and
+    /// report which of them the relay had left out.
+    ///
+    /// The relay is authoritative for the fleet and STAYS authoritative for every session it knows: a row
+    /// already in <paramref name="fleet"/> is returned untouched, so the session numbers and identity
+    /// stamping the Gateway hands out are never overwritten by a local copy. This adds only the rows the
+    /// relay omitted - which it does silently, while still returning 200, for any Director it cannot reach.
+    /// A Director is first-hand authority on its own machine and must never deny a session it is itself
+    /// holding, because this roster is what the CLI resolves an id against before it can reap anything.
+    ///
+    /// A row with no session id cannot be addressed by any caller, so it is never restored on that basis.
+    /// </summary>
+    internal static (List<SessionDto> Roster, List<string> Restored) UnionOwnSessions(
+        IReadOnlyList<SessionDto> fleet, IReadOnlyList<SessionDto> own)
+    {
+        var roster = new List<SessionDto>(fleet);
+        var known = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        foreach (var s in fleet)
+        {
+            if (!string.IsNullOrEmpty(s.SessionId))
+                known.Add(s.SessionId);
+        }
+
+        var restored = new List<string>();
+        foreach (var s in own)
+        {
+            if (string.IsNullOrEmpty(s.SessionId) || !known.Add(s.SessionId))
+                continue;
+            roster.Add(s);
+            restored.Add(s.SessionId);
+        }
+
+        return (roster, restored);
+    }
 
     /// <summary>Folder name of a repo path, for display fallback. Empty path -> "Unknown Project".</summary>
     // Gateway Cleanup Phase 0 (Worker R2): widened to internal so CatalogReadExecutor's claude-sessions core
--- a/src/CcDirector.Core/Sessions/SessionManager.cs
+++ b/src/CcDirector.Core/Sessions/SessionManager.cs
@@ -752,7 +752,22 @@ public sealed class SessionManager : IDisposable
         catch (Exception ex)
         {
             session.MarkFailed();
+            // Issue #1019, defect 3: a create that fails must SAY SO somewhere a person can find it later.
+            // _log is optional - anything that constructs a SessionManager without one (the Control API's
+            // own host among them) failed completely silently, which is why the original report could not
+            // tell a failed spawn from a UI glitch after the fact. FileLog always lands in the Director log.
+            FileLog.Write($"[SessionManager] CreateSession FAILED: session={id}, repo={repoPath}, agent={agent.Kind}, backend={backendType}: {ex.Message}");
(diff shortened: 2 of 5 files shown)


thefrederiksen:
## The relay leg, named separately — and the half of it this pull request does NOT fix

Omission two deserves its own name, because it is the worse of the two and this change only closes part of it.

`GET /sessions` on the Gateway **silently drops the sessions of a Director it cannot reach and still returns 200.** So a caller receiving a roster cannot distinguish:

* "that Director has no sessions", from
* "I could not reach that Director, and its sessions are missing from what you are holding".

Absent reads identical to empty. A 200 with a short list is indistinguishable from a 200 with a complete one, and nothing in the response says which you got.

This is already known in the code, which is what makes it worth naming rather than discovering twice. `GatewayClient` carries a second method for exactly this reason:

> `ListFleetSessionsWithReachabilityAsync` — "asks for the envelope (`GET /sessions?envelope=true`), which also carries per-Director reachability (Online / Wobbly / Offline). **The plain list silently DROPS an unreachable Director's sessions while still returning 200**, so a caller that must not act on a partial roster — the destructive worktree reaper — needs the reachability to tell whether the fleet view is complete."

So the hazard was understood, an envelope was built for it, and the one caller judged to need it was wired to the envelope. `/fleet/sessions` was left on the plain list.

**What this pull request fixes:** the *local* half. This Director's own rows can no longer be dropped from its own roster, because they are unioned back in from the store, and when the relay omits them that is written to the Director log with the count and the ids. That is the half that produced the reported ghost, since the reported Director was the one failing to register.

**What it does NOT fix, and I am not going to pretend otherwise:** the *cross-Director* half. If Director B is unreachable, B's sessions are still silently absent from the roster this endpoint returns, still under a 200, and a caller still cannot tell. A session on B is therefore still unnameable from here — the same defect as the one fixed, one machine over.

Fixing that properly means `/fleet/sessions` moving to the envelope and reporting per-Director reachability to its callers, which changes the response shape and every reader of it (`cc-devthrottle`, `cc-status`, `cc-history`). That is a deliberate, wider change and does not belong in a release-gate fix. It should be its own issue, and I would rather it be filed than folded in here quietly.

Worth stating why the local half was still the right cut: a Director is first-hand authority on its own machine and can answer completely without asking anyone. For another machine it genuinely cannot, and there the honest answer is reachability, not a guess.

thefrederiksen:
The cross-machine half described in the comment above is now filed separately as thefrederiksen/devthrottle_internal#1051 (labelled post-release). This pull request stays scoped to the local half.

thefrederiksen:
## The pair is complete: red without the fix, green with it

Both halves are now on the record, from two different observers.

### Green with the fix — CI, clean checkout, Release

[Run 30515756303](https://github.com/thefrederiksen/devthrottle/actions/runs/30515756303) — `Build & Test (.NET)` **pass** (41m27s), `Build & Test (web)` **pass**. The step is `dotnet test cc-director.sln -c Release --no-build`, so `CcDirector.Gateway.Tests` is included.

All ten tests added here ran and passed, by name:

```
Passed GhostSessionReapableTests.CrashedSession_theDirectorStillHolds_isListedSoTheCliCanNameIt
Passed GhostSessionReapableTests.CrashedSession_isListedAsCrashed_notSilentlyDressedUpAsLive
Passed GhostSessionReapableTests.CrashedSession_isReapableEndToEnd_soTheCardCanBeCleared
Passed GhostSessionReapableTests.UnknownSession_stillFailsLoud_soListingHeldRowsDidNotWeakenTheGuard
Passed UnionOwnSessionsTests.OwnSessionsTheRelayOmitted_areRestored_andReported
Passed UnionOwnSessionsTests.SessionsTheRelayAlreadyKnows_keepTheGatewaysCopy_notTheLocalOne
Passed UnionOwnSessionsTests.IdMatching_isCaseInsensitive_soCasingNeverDuplicatesARow
Passed UnionOwnSessionsTests.ARowWithNoSessionId_isNeverRestored_becauseNoCallerCouldAddressIt
Passed UnionOwnSessionsTests.AnEmptyRelayRoster_stillYieldsOurOwnSessions
Passed UnionOwnSessionsTests.NoOwnSessions_leavesTheRelayedRosterExactlyAsItCame
```

Every test run in the job reported `Test Run Successful`, and there is **not one test failure anywhere in the log**. Totals across the run: `63, 83, 83, 313, 24, 88, 3939 (8 skipped), 4477 (17 skipped), 21, 429`. The skips are pre-existing and deliberate.

Two things I checked rather than assumed, because a green tick is not by itself evidence that the relevant tests ran:

* **The tests actually executed.** Grepping the log for the six lines matching `Failed <word>` returns only MSBuild's `Failed to load prune package data from PrunePackageData folder` — unrelated noise, no test failures. And the ten names above are present with `Passed`, which they could not be if the assembly had been skipped.
* **The suite lock played no part.** `gateway-test-lock` appears **zero** times in the CI log — a fresh runner has no competing run, so `CcDirector.Gateway.Tests` acquired immediately. That is the project that could not run locally.

I have deliberately not claimed which per-project total belongs to `Gateway.Tests`: the job runs projects in parallel, so the summary lines are interleaved and adjacency does not prove the mapping. The named passes above are the evidence, and they do not depend on that mapping.

### Red without the fix — local, observed earlier

CI cannot give this half, because it only ever sees the fixed tree. Reinstating the `ActivityState.Exited` filter and changing nothing else:

```
Failed GhostSessionReapableTests.CrashedSession_theDirectorStillHolds_isListedSoTheCliCanNameIt
  Assert.Contains() Failure: Filter not matched in collection
  Collection: []
Failed GhostSessionReapableTests.CrashedSession_isListedAsCrashed_notSilentlyDressedUpAsLive
Failed GhostSessionReapableTests.CrashedSession_isReapableEndToEnd_soTheCardCanBeCleared
Total tests: 10   Passed: 7   Failed: 3
```

`Collection: []` is the defect itself: the Director is holding the session, the rail is showing it, and the Director's own roster returns an empty array.

The seven that passed in both directions are the ones that should — the six pure fold tests and `UnknownSession_stillFailsLoud`, none of which depend on the filter. A test that passes either way pins nothing, so it is worth being explicit that only three of these are load-bearing.

The command line tests are pinned the same way: `157 passed` in `tools/cc-devthrottle`, and the crash-marker test renders a bare `Exited` without the change.

### Scope note

CI green is the observation that was missing, not a merge gate — the standing rule for this repository is that CI does not gate merges. This says the fix is proven, not that it should land; that call is the release manager's.

thefrederiksen:
Correction to my evidence comment above, caught while running the same check on #1051.

I wrote that `gateway-test-lock` appears **zero** times in that CI log. What I actually measured was narrower: `grep -c "REFUSING TO RUN\|gateway-test-lock] WAITING"`, which returned 0. The bare string does appear — on a healthy runner the lock logs `[gateway-test-lock] Acquired the per-user Gateway test lock. Starting the run.`, which is the line you *want* to see and which I had not looked for.

The conclusion stands and is unchanged: no waiting, no refusal, so the lock caused no interference and `Gateway.Tests` ran. But I stated a broader piece of evidence than the one I checked, which is the same shape of error as reading a narrow grep as a wide absence. Recording it here rather than leaving a slightly-too-strong claim sitting in the proof comment.

PR #2293, opened 2026-07-30
THROWAWAY revert proof for #2292 - DO NOT MERGE

thefrederiksen:
**THROWAWAY - DO NOT MERGE. DO NOT REVIEW. This branch exists to be observed failing, then deleted.**

Revert proof for pull request #2292 (thefrederiksen/devthrottle_internal#819 and
thefrederiksen/devthrottle_internal#824).

This branch is `origin/main`'s **production code** with only the **new tests** from #2292 applied.
Nothing else. `git diff origin/main --stat` is two test files.

Why it exists: a green *with* a fix proves the tests pass. It does not prove they can **fail**, and a
test that has never been observed red is a test whose passing means nothing. For #824 in particular, a
guard that refuses a hold on an exited session is easy to write so that it never fires and still goes
green.

### Expected RED

`DirectorCommandRouterTimeoutTests`
- `TrySendAsync_Timeout_WithoutAMachineName_KeepsTheSameMessageShape` - main still says "The command was
  not carried out."
- `TrySendAsync_Timeout_DoesNotClaimTheCommandWasSkipped`

`SnoozeEndToEndTests`
- `Holding_a_session_that_has_already_exited_is_refused_and_records_nothing` - main returns 200 and
  records an armed snooze entry.
- `Holding_a_crashed_session_is_refused_so_the_crash_stays_visible` - main returns 200 and the row flips
  to grey "Snoozed".

### Expected GREEN even here

`Unsnoozing_an_exited_session_is_still_allowed_because_it_is_the_repair` - it asserts behaviour main
already has, and the fix must not change it. If this one goes red, the guard is over-reaching and
blocking the repair path.

### DIFF
--- a/src/CcDirector.Gateway.Tests/DirectorCommandRouterTimeoutTests.cs
+++ b/src/CcDirector.Gateway.Tests/DirectorCommandRouterTimeoutTests.cs
@@ -106,7 +106,31 @@ public sealed class DirectorCommandRouterTimeoutTests
 
         // Assert - the degraded message is the SAME sentence minus the machine clause, never vaguer and never
         // a second message style. The machine name is presentation only.
-        Assert.Equal("The Director did not answer within 30 seconds. The command was not carried out.", result!.Error);
+        Assert.Equal(
+            "The Director did not answer within 30 seconds. It is not known whether the command was carried out.",
+            result!.Error);
+    }
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


thefrederiksen:
Throwaway revert proof, its job done. It carried origin/main's production code with ONLY the new tests applied, and its FAILURE is the evidence for 2292 rather than a problem: the four tests predicted red went red, and the one predicted to stay green did. Closing and deleting the branch as designed.

PR #2299, opened 2026-07-30
THROWAWAY fault-injection proof for #2298 - DO NOT MERGE

thefrederiksen:
**THROWAWAY - DO NOT MERGE, DO NOT REVIEW. This branch exists to be observed FAILING, then deleted.**

Revert proof for **#2298** (thefrederiksen/devthrottle_internal#1070 and #1071).

This is `origin/main` + pull request #2298's **tests**, with four faults injected into #2298's
**production** code. The tests are **byte-identical** to #2298 - `git diff fix/gateway-cards-1070-1071 HEAD
-- src/CcDirector.Avalonia.Tests tools/cc-director-setup-engine.Tests` is empty.

## Why fault injection and not a plain revert

A blanket `git checkout origin/main -- <production files>` does not work here, and calling it a revert would
misdescribe what ran. The old cards were `Border`s; restoring them changes the generated field types and the
**test project stops compiling**. A build error is not the same evidence as a failing test - it proves
nothing about whether the test can detect the defect. So the seam that hands the cards to the tests is typed
`Control` rather than `RadioButton`, and the faults are applied surgically so **every test still runs**.

| Fault | What it puts back | Issue |
| --- | --- | --- |
| 1 | `ShowGatewayFailure` hides the choice view, as the old exclusive failure view did - the cards vanish | #1070 |
| 2 | The step's entry no longer calls `ClearGatewayFailure`, so a stale banner survives leaving and re-entering | #1070 |
| 3 | All three engine cancel messages name a control again: `Click "Sign in to DevThrottle" to try again` | #1070 |
| 4 | The three cards go back to `Border`s carrying `Focusable` and an `AutomationProperties.Name` | #1071 |

## Predicted RED - 7 tests

`CcDirector.Avalonia.Tests.GatewayStepOptionsTests`
- `EveryOption_IsARealControlWithAnAccessibleName_NotDecoratedText` (F4)
- `EveryOption_IsOneOfASingleChoiceGroup_SoSelectionIsExposedAndExclusive` (F4)
- `ChoosingAnyOption_MovesTheWizardsChoice_SoTheControlIsWiredNotJustPresent` (F4)
- `ACancelledSignIn_StillLetsTheUserDeclineTheGatewayAndContinue` (F4)
- `ACancelledSignIn_LeavesEveryOptionReachable` (F1)
- `ACancelledSignIn_DoesNotStickAcrossLeavingAndReenteringTheStep` (F1 + F2)

`CcDirector.Setup.Engine.Tests.GatewayAccountEnrollRunnerTests`
- `EverySignInCancelledMessage_StatesTheFact_AndNamesNoButton` (F3)

## Predicted GREEN - 2 tests, and both matter

- **`Control_TheStepsOwnButtonsAreVisibleAndFocusable_SoAbsenceMeansSomething`** - THE POSITIVE CONTROL. It
  must stay green, because it is what proves the reds above are about the cards rather than about a harness
  that stopped seeing the screen at all. A card that cannot be driven and a card that is not there look
  identical from a test. **If this one goes red the run is void, not damning.**
- **`TheFailureMessage_NamesNoControlThatIsNotOnTheScreen`** - it guards the *banner's* own guidance text,
  which fault 3 does not touch; the engine string is guarded by the engine test instead. Predicted green so
  the limit of that test is on the record rather than discovered later.

Everything else must stay green: the other **313** Avalonia tests and the other **430** setup-engine tests.
A wide red means a broken harness, not four injected faults.

### DIFF
--- a/src/CcDirector.Avalonia.Tests/GatewayStepOptionsTests.cs
+++ b/src/CcDirector.Avalonia.Tests/GatewayStepOptionsTests.cs
@@ -1,0 +1,238 @@
+using Avalonia.Automation;
+using Avalonia.Controls;
+using Avalonia.Controls.Primitives;
+using Avalonia.Headless.XUnit;
+using Avalonia.LogicalTree;
+using CcDirector.Core.Configuration;
+using CcDirector.Core.Onboarding;
+using Xunit;
+
+namespace CcDirector.Avalonia.Tests;
+
+/// <summary>
+/// The gateway step's three options - Hosted, Self-hosted, Not now - must stay REACHABLE. All of them,
+/// by mouse and by keyboard, before a sign-in attempt and after one that failed or was cancelled.
+///
+/// That is not a styling preference. The step gives "Not now" equal prominence deliberately, because it
+/// is the no-account, local-only choice and the product installs and runs without an account; the source
+/// comment says a first screen reading as sign-up-to-continue "would change what the product is". Two
+/// separate mechanisms were quietly taking that promise back:
+///
+///   * #1070 - a cancelled sign-in swapped the whole choice view for a failure view, so all three cards
+///     vanished, and re-entering the step never reset the sub-view, so Back-then-forward brought the
+///     user back to the cards-less screen. No path declined the gateway and continued.
+///   * #1071 - the cards were Borders with handlers hung on them. A Border gets no automation peer, so
+///     they exposed no name, no pattern, and never entered the tab order: Tab cycled Back and Sign in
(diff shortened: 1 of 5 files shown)


thefrederiksen:
Purpose served. 6 of 7 predictions confirmed - Avalonia 315 passed / 6 failed of 321, the 6 exactly as named, both predicted greens held including the positive control, no collateral reds. The 7th (the engine message test) was never run: ci.yml runs the installer-tests step AFTER the solution tests, so it was skipped. Recorded as unverified, not confirmed. Result linked in #2298. Deleting.

PR #2300, opened 2026-07-30
Stop cc-image printing a Unicode ellipsis, and run the guard that caught it

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1077.

## It was a true positive

`tools/test_shipped_tools_contract.py` has been failing on `origin/main`:

```
FAILED tools/test_shipped_tools_contract.py::test_installs_ascii_truncation[cc-image]
AssertionError: cc-image: renders Rich tables but does not install the ASCII truncation patch
```

I checked whether that was a real defect or a test wanting to be satisfied. It is real. Driving
the **installed** cc-image at a narrow console and reading the raw bytes:

```
non-ASCII bytes emitted by the shipped cc-image: ['0x85']
0x85 is cp1252 for U+2026 (horizontal ellipsis): True

    Image Info:
     shot.png
+-----------------+
| Prope? | Value  |
| Dimen? | 1920 x |
```

That is shipped output breaking the ASCII-only house rule. Read back as UTF-8 it decodes to the
replacement character, which is what a consumer of the stream actually sees. With the patch, the
same command through the repo source:

```
BEFORE (installed, no patch): non-ASCII bytes -> ['0x85']
AFTER  (repo source, patched): non-ASCII bytes -> none

| Pro... | Value  |
```

cc-image was the only shipped table-rendering tool without the patch that cc-html, cc-gmail,
cc-outlook, cc-devthrottle and cc-comm-queue all install at import. The patch here is copied from
cc-html escape for escape - including writing the ellipsis as `"\u2026"` rather than the literal
character, since the shipped source must itself be ASCII. I got that wrong on the first attempt and
the same contract file would have caught me.

## The half that matters more

**Nothing ran this file.** CI ran the .NET solution, the installer projects and the JavaScript
workspaces, and no Python at all. A guard written specifically to catch what has silently broken a
clean install before - a pyproject name that no longer matches the registry, a missing
console-script entry point, non-ASCII in shipped source - sat red where nobody could see it.

A guard nobody runs cannot fail, and a red nobody sees is indistinguishable from green. Fixing
cc-image alone would have closed the instance and left the mechanism, and the next dropped patch
would go unnoticed in exactly the same way. This adds a `Tool contracts (Python)` job running the
contract guard and cc-devthrottle's own tests, which were equally unrun.

Windows on purpose: that is where these tools run and where I verified both suites. Dependencies
are named explicitly rather than installed from cc-devthrottle's pyproject, because that package
depends on the local `cc-shared` distribution which is not on PyPI - I tried the pyproject route
first and it fails to resolve. A new dependency will fail this job loudly with a
`ModuleNotFoundError` rather than quietly skipping anything.

## Proof the new job can actually fail

A CI step that cannot go red is the thing this pull request is about, so I removed the patch again
and ran the exact command the step runs:

```
patch REMOVED  -> pytest exit code: 1  (non-zero = CI step fails)
patch RESTORED -> pytest exit code: 0  (zero = CI step passes)
```

## Tests

| suite | result |
|---|---|
| `tools/test_shipped_tools_contract.py` | 36 passed (was 1 failed, 35 passed) |
| `tools/cc-devthrottle/tests/` | 157 passed |
| `tools/cc-image/tests/` | 58 passed |

Both new-job suites were also run in a clean virtual environment holding only pytest, typer, rich
and requests, to confirm they do not depend on anything this machine happens to have: 36 and 157
passed, in under two seconds combined.

I did not run `dotnet test cc-director.sln` for this change - it touches one Python file and the
workflow, no .NET - and CI runs the whole solution on this pull request anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,11 +106,17 @@ jobs:
   # on main with nobody to see it. A guard nobody runs cannot fail, and a red nobody sees is
   # indistinguishable from green. Found while cutting cc-playwright from the toolbelt (#1002).
   #
-  # Windows on purpose: these tools are installed and run on Windows, and that is where these
-  # suites were verified. The dependencies are named explicitly rather than installed from
-  # cc-devthrottle's pyproject, because that package depends on the local cc-shared distribution
-  # which is not on PyPI. If the tool gains a dependency this job fails loudly with a
-  # ModuleNotFoundError rather than quietly skipping anything.
+  # Windows on purpose: these tools are installed and run on Windows, and that is where this suite
+  # was verified. The dependencies are named explicitly rather than installed from cc-devthrottle's
+  # pyproject, because that package depends on the local cc-shared distribution which is not on
+  # PyPI. If a dependency is added this job fails loudly with a ModuleNotFoundError rather than
+  # quietly skipping anything.
+  #
+  # cc-devthrottle's OWN tests are deliberately not run here yet. Adding them was the first thing
+  # this job did, and it immediately found why they had never survived an unattended run: five of
+  # them assert plain substrings against Rich-STYLED output, so they pass on a plain console and
+  # fail wherever colour is on - which is what a CI runner is. That is a real defect in those tests
+  # and it gets its own change rather than being silenced here with a colour flag.
   tool-contracts:
     name: Tool contracts (Python)
     runs-on: windows-latest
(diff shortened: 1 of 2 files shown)


thefrederiksen:
Pushed a correction, and the reason is worth recording because the job earned its keep on its first run.

The `Tool contracts (Python)` job went red immediately - not on the contract guard, which passed on the runner and is the suite that was actually red on main, but on the cc-devthrottle step I had added alongside it. The cause is not the runner being wrong. Five of those tests assert plain substrings against Rich-STYLED output (`"--subject" in result.output`, `"900 directories" in ...`), so they pass on a plain console and fail wherever colour is on - which is what a CI runner is. Reproduced locally with `FORCE_COLOR=1`: 5 failed, 152 passed, against 157 passed without it.

I could have made that green with `NO_COLOR=1` on the job. That is the same move this whole change is against - making a red disappear rather than reading it - so instead the cc-devthrottle step is out of this pull request, the workflow carries a comment saying why, and the colour fragility is filed as thefrederiksen/devthrottle_internal#1082 with the five test names and the verification recipe (require green both with and without `FORCE_COLOR=1`).

So this pull request now ships exactly what internal#1077 was about: cc-image stops emitting the Unicode ellipsis, and the guard that caught it runs in CI where a future regression will be seen. Step 5, `Run the shipped-tool contract guard`, passed on the runner.

PR #2302, opened 2026-07-30
Say when the fleet list is not the whole fleet

thefrederiksen:
Fixes thefrederiksen/devthrottle_internal#1051. Follows thefrederiksen/devthrottle#2289 (#1019), which fixed the local half of the same defect.

> `devthrottle#1051` in this repository is an unrelated issue, so the link is fully qualified deliberately.

## The defect

`GET /sessions` on the Gateway **drops an unreachable Director's sessions and still answers 200.** A caller holding that response cannot tell:

* "that Director has no sessions", from
* "I could not reach that Director, and its sessions are missing from what you are holding".

**Absent reads identical to empty.** A 200 with a short list looks exactly like a 200 with a whole one, and nothing in the response says which arrived.

That roster is what every `cc-devthrottle` verb resolves a target against, so this is #1019's consequence one machine over: a session on an unreachable machine cannot be **named**, and a session that cannot be named cannot be interrupted, messaged, or marked done.

Three sentences were flatly false in that state:

| Said | When the roster was short |
|---|---|
| `No session matches '<id>'` | It may exist perfectly well on the machine we could not read |
| `No sessions are running in the fleet` | A claim about the whole fleet, from a partial list |
| `(no sessions running)` (cc-status) | The same claim again |

Each reads as a fact about the world and was only ever a fact about the bytes that arrived — and the two readings call for opposite next steps: give up, or go and look at machine B.

## What changed

`GET /fleet/sessions?envelope=true` answers with the roster **and a finished completeness verdict**; the three tools print it. The bare array stays the default shape, because a Director that has not restarted yet still serves it.

**Only `offline` counts as incomplete**, and that distinction carries the design. An offline Director has had its sessions dropped, so rows really are missing. A `wobbly` one is inside the grace window with its last-known-good sessions still served — the roster is whole, merely part-stale. Reporting wobbly as incomplete would put a caveat on the most frequently run command in the tool for a case where nothing is missing, which teaches the reader to skip it, and then it goes unread on the day offline finally happens.

## Three things worth a reviewer's attention

**1. The verdict is folded in `Gateway.Contracts`, beside `SessionOrdering` — not in each tool.** Deciding what a reachability state *means* is ruling, not rendering. Three tools each deciding for themselves would drift, and a client that rules for itself renders something plausible the moment it meets a state it did not expect. The shared fetch and wording also moved into `cc_shared/director.py`, so the "No session matches" line has one definition instead of three copies.

**2. Unknown is a third state and is never collapsed into complete.** A Director that predates this change serves the bare array and cannot vouch either way. Reading that silence as a guarantee of completeness would rebuild this very defect one layer up, so the tools say they *cannot confirm* rather than implying all is well. This is a real case, not a hypothetical: a running Director is long-lived and the command line is invoked fresh, so a newer tool will meet an older Director.

**3. The read path deliberately does NOT reuse `ListFleetSessionsWithReachabilityAsync`, and this is the near-miss worth reading.** My first cut did reuse it. That method **fails closed** — it throws when the Gateway cannot vouch for completeness. That posture is correct for its own caller, the worktree reaper, which deletes directories. It is catastrophic for a listing: against a Gateway that ignores the envelope query (the hosted Gateway is deployed separately from the desktop app, so version skew is ordinary) it would have turned `session list`, every target resolve, `cc-status` and `cc-history` into a 502.

So there are now two methods with two explicit postures, and the strict one is untouched — weakening it would have quietly disarmed the reaper's guard, which is a worse outcome than the bug being fixed here.

## Proof

**Fail-on-purpose, both halves.** These were run as mutations of the shipped code, not as a description of what the tests would do:

*The wobbly/offline distinction* — making `wobbly` count as incomplete (the plausible-but-wrong implementation) is caught by exactly one test, the one that exists for it:
```
Failed RosterCompletenessFoldTests.AWobblyDirector_isNOTreportedIncomplete_becauseItsSessionsAreStillServed
Total tests: 13   Passed: 12   Failed: 1
```

*The caveat itself* — making `roster_caveat` always return empty (the pre-fix behaviour, where a partial roster looks whole) is caught by seven:
```
FAILED test_an_incomplete_roster_reports_the_directors_own_reason
FAILED test_an_incomplete_roster_with_no_reason_still_warns
FAILED test_an_unknown_verdict_is_not_treated_as_complete
FAILED test_failed_resolve_on_an_incomplete_roster_says_the_list_may_be_short
FAILED test_empty_and_incomplete_does_not_claim_the_fleet_is_empty
FAILED test_a_listed_roster_that_is_incomplete_is_qualified_after_the_table
FAILED test_json_output_stays_a_bare_array_and_warns_on_stderr
7 failed, 166 passed
```

**Green:** 13 fold tests and `173 passed` across `tools/cc-devthrottle` (157 before, plus 16 added).

The fold tests live in `Core.Tests` rather than `Gateway.Tests` on purpose: it is a pure fold with no host, so it does not belong behind that suite's machine-wide serialisation lock, and it can be run and re-run while other worktrees hold that lock.

`--json` output keeps its bare-array shape — agents and pipes parse it — and the warning goes to **stderr**, so a partial roster is still announced without corrupting the parse. There is a test asserting stdout still parses as JSON.

### Not yet proven

`CcDirector.Gateway.Tests` has not run locally: another session holds its per-user suite lock, and that lock never breaks for a live holder. `FleetBodyShapeToleranceTests` (the array-vs-envelope shape tolerance, item 3 above) therefore has CI as its first observer. I will post the CI result here pass or fail, and this should not merge before it lands.

### DIFF
--- a/src/CcDirector.ControlApi/ControlEndpoints.cs
+++ b/src/CcDirector.ControlApi/ControlEndpoints.cs
@@ -405,7 +405,14 @@ internal static class ControlEndpoints
         // omits cannot be named, so it cannot be interrupted, messaged or marked done. Issue #1019 - a card
         // that no tool could remove, with restarting the Director as the only remedy - was two independent
         // omissions from this one list.
-        app.MapGet("/fleet/sessions", async (CancellationToken ct) =>
+        //
+        // Issue #1051 - the other end of the same defect. The relayed roster DROPS an unreachable Director's
+        // sessions and still answers 200, so a caller cannot tell "that Director has no sessions" from "I
+        // could not reach it": absent reads identical to empty. `?envelope=true` answers with the roster AND
+        // a finished completeness verdict, so a caller that must not mistake a partial list for a whole one
+        // can say so. The bare array stays the default, because a Director that has not restarted yet still
+        // serves the old shape and the tools must keep working against it.
+        app.MapGet("/fleet/sessions", async (bool? envelope, CancellationToken ct) =>
         {
             // Omission one: EVERY session this Director is still holding belongs on the roster, including
             // crashed and exited-pending-reap rows. This store is the same store the desktop rail renders,
@@ -424,7 +431,6 @@ internal static class ControlEndpoints
             {
                 try
                 {
-                    var fleet = await gw.ListFleetSessionsAsync(ct);
                     // Omission two: the relayed list silently DROPS the sessions of a Director the Gateway
                     // cannot reach while still returning 200 (see GatewayClient.ListFleetSessionsAsync).
                     // A Director whose registration is failing therefore vanishes from its OWN fleet
@@ -433,10 +439,47 @@ internal static class ControlEndpoints
                     // authority on its own machine, so its rows go back in. The Gateway's copy WINS for any
                     // session it already knows: the session numbers and identity stamping it hands out are
                     // never overwritten here.
+                    //
+                    // Read posture, NOT the reaper's. The strict ListFleetSessionsWithReachabilityAsync throws
+                    // when the Gateway cannot vouch for completeness, which is correct for something that
+                    // deletes directories and wrong here: it would take `session list`, every target resolve,
+                    // cc-status and cc-history down against a version-skewed Gateway. This one returns null
+                    // reachability for "the Gateway did not say" and still serves the rows.
+                    var (fleet, reachability) = await gw.ReadFleetSessionsWithOptionalReachabilityAsync(ct);
                     var (roster, restored) = UnionOwnSessions(fleet, own);
                     if (restored.Count > 0)
                         FileLog.Write($"[ControlEndpoints] /fleet/sessions: the Gateway roster omitted {restored.Count} of this Director's own session(s); served from the local store: {string.Join(", ", restored)}");
-                    return Results.Json(roster);
+
+                    if (envelope != true)
+                        return Results.Json(roster);
+
+                    // The verdict is FOLDED HERE, not in the client. A caller must never have to decide what
+                    // "offline" means for completeness - that is the Gateway-owns-ruling rule, and the reason
+                    // the desktop rail renders pushed display state verbatim. The tools print these strings.
+                    //
+                    // Null reachability means the Gateway would not say, so rosterComplete is reported as NULL
+                    // - explicitly UNKNOWN, never true. Coalescing unknown to complete here would rebuild the
+                    // exact defect this issue is about, one layer up: absent reading identical to empty.
+                    bool? complete = null;
+                    string? incompleteReason = null;
+                    if (reachability is not null)
+                    {
+                        var folded = RosterCompleteness.Fold(reachability);
+                        complete = folded.Complete;
+                        incompleteReason = folded.Reason;
+                    }
+                    else
+                    {
+                        FileLog.Write("[ControlEndpoints] /fleet/sessions?envelope=true: the Gateway supplied no reachability; reporting completeness as UNKNOWN");
+                    }
+
+                    return Results.Json(new
+                    {
+                        sessions = roster,
+                        directors = reachability,
+                        rosterComplete = complete,
+                        rosterIncompleteReason = incompleteReason,
+                    });
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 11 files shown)


thefrederiksen:
## End-to-end proof of the compatibility path, against a real old Director

The version-tolerance case in `FleetBodyShapeToleranceTests` is pinned with hand-written bodies, which proves the parser but not that a real Director behaves the way I assumed it would. Every Director running on this machine right now is on the **pre-change build**, so it is exactly the "older Gateway/Director that ignores the envelope query" case — a better witness than a stub.

Driving the new `director.get_fleet()` against a live one on `http://127.0.0.1:7879`:

```
sessions returned: 15
complete       : None
reason         : None
caveat         : 'This Director cannot confirm the roster is complete, so a session may be missing from it.'
```

Three things that had to hold, and did:

* **The rows still came back** — all 15. This is the path my first cut would have broken: the old build answers with the bare array, and the strict `ListFleetSessionsWithReachabilityAsync` throws when the `directors` field is absent, so `session list`, every target resolve, `cc-status` and `cc-history` would have been a 502 against precisely this Director.
* **`complete` is `None`, not `True`.** The Director did not vouch, and its silence is reported as unknown rather than as a guarantee. Collapsing that to `True` would rebuild this issue's own defect one layer up.
* **The caveat renders the honest sentence** rather than either claiming completeness or crying incompleteness.

Worth being precise about what this does and does not show: it proves the **unknown** path end to end against a real Director. It does not exercise the offline path, because nothing in this fleet is currently offline — that one rests on the fold's unit tests and CI.

thefrederiksen:
## Green. The pair is complete.

[CI run 30538536251](https://github.com/thefrederiksen/devthrottle/actions/runs/30538536251) — `Build & Test (.NET)` **pass** (55m12s), `Build & Test (web)` **pass**.

**All sixteen tests added here ran and passed, by name:**

```
Passed FleetBodyShapeToleranceTests.TheEnvelopeShape_yieldsRowsAndReachability
Passed FleetBodyShapeToleranceTests.ThePlainArrayShape_fromAnOlderGateway_yieldsRowsAndUNKNOWNreachability
Passed FleetBodyShapeToleranceTests.AnEnvelopeWithoutReachability_isUNKNOWN_notComplete
Passed FleetBodyShapeToleranceTests.AnEmptyReachabilityArray_isPreserved_asAnAuthoritativeNothingUnreachable
Passed FleetBodyShapeToleranceTests.AnEnvelopeWithNoSessionList_throws_becauseThatIsMalformedNotDegraded
Passed FleetBodyShapeToleranceTests.AnEmptyBody_throws_ratherThanReadingAsAnEmptyFleet
Passed RosterCompletenessFoldTests.AllOnline_isComplete_withNothingToSay
Passed RosterCompletenessFoldTests.AnOfflineDirector_makesTheRosterIncomplete_andIsNamed
Passed RosterCompletenessFoldTests.AWobblyDirector_isNOTreportedIncomplete_becauseItsSessionsAreStillServed
Passed RosterCompletenessFoldTests.SeveralOfflineDirectors_areAllNamed_andCountedPlurally
Passed RosterCompletenessFoldTests.OneOfflineDirector_readsSingular
Passed RosterCompletenessFoldTests.ADirectorNeverReached_saysSo_ratherThanClaimingAnAge
Passed RosterCompletenessFoldTests.AnOfflineDirectorWithNoMachineName_fallsBackToItsId_soItIsStillIdentifiable
Passed RosterCompletenessFoldTests.NoReachabilityReported_isTreatedAsComplete
Passed RosterCompletenessFoldTests.StateMatching_isCaseInsensitive
Passed RosterCompletenessFoldTests.Age_readsInTheCoarsestUsefulUnit
```

`FleetBodyShapeToleranceTests` is the one that needed CI as its observer — it pins the array-vs-envelope tolerance behind the near-miss described above, and it could not run locally.

**Whole solution:** every test run reported `Test Run Successful`, and there is **not one** `Failed` test line in the log. Totals: `83, 313, 24, 88, 3997 (8 skipped), 4568 (17 skipped), 21, 430`. The skips are pre-existing and deliberate.

**Checked rather than assumed**, because a green tick is not evidence that the relevant tests ran:

* The sixteen names above are present with `Passed` — they could not be if the assemblies had been skipped.
* The only log lines matching `Failed <word>` are MSBuild's `Failed to load prune package data from PrunePackageData folder`. Unrelated noise, no test failures.
* **The Gateway suite lock acquired immediately on the runner** — `[gateway-test-lock] Acquired the per-user Gateway test lock. Starting the run.` at 11:30:06, with no `WAITING` and no `REFUSING TO RUN`. That is the project that could not run locally, so this is the line that matters.

**Local, in parallel:** 4560 passed / 0 failed across the six non-Gateway projects, and `173 passed` in `tools/cc-devthrottle`. That includes `AuthoritativeRosterCompletenessTests` — the worktree reaper's own completeness tests — still passing, so the guard this change deliberately left byte-identical is verified working, not merely untouched on disk.

**Red without the fix**, from the two mutations of shipped code recorded earlier: making `wobbly` count as incomplete is caught by exactly the one test that exists for it (12 passed, 1 failed); making the caveat always empty is caught by seven.

### Still not claimed

The **offline** path has no end-to-end witness. The live smoke test above proved the *unknown* path against a real pre-change Director, but nothing in this fleet is currently offline, so `offline` rests on the fold's unit tests. That is a gap in the evidence, not a known defect, and it is the one I would attack first if this misbehaves in the field.

CI green is the observation, not a merge gate — the standing rule is that CI does not gate merges here. This says the fix is proven; whether it lands is the owner's and the release manager's call.

PR #2303, opened 2026-07-30
Assert what the user reads, not how Rich painted it

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1082.

**Stacked on #2300**, which adds the `Tool contracts (Python)` job this restores a step to. Review or merge that one first; until it lands, the diff here shows its commits too.

## What was wrong

Five cc-devthrottle tests asserted plain substrings against Rich-**styled** output - `"--subject"` in a help panel, `"900 directories"` in a results table, an error message quoted back. Rich threads ANSI escape sequences through that text, so an `in` check misses a substring that is plainly on the screen:

```
assert '--subject' in '\x1b[1m ... \x1b[1;36m--subject\x1b[0m ...'   ->  False
```

Whether styling is on is a property of the **environment**, not of the code under test. Captured to a buffer on a developer's console Rich usually renders plain and these passed; on a runner colour is on and they fail. They had never run anywhere but a console, so nobody had seen it - the same hole as the guard that found them (#1077): a suite nobody runs tells you nothing about whether it works.

## The fix, and the one I did not take

The assertions now compare **text**, through a `plain` fixture in a new `tests/conftest.py`.

I deliberately did **not** fix this by forcing colour off in the job. That would make these five pass while leaving the next test written the same way just as fragile, and it depends on knowing which variable a given runner uses to switch colour on - which I have not established, and did not want to guess at. Stripping the escapes is correct whatever the reason colour is on.

Two of the changed assertions are negative (`"--to" not in ...`). Those get **stronger**: a styled occurrence could previously have slipped past the check.

## The CI step comes back, running the harder condition

`FORCE_COLOR: "1"` is set on the restored step on purpose. Running these the safe way would re-hide exactly what the job caught, so CI now exercises the condition that failed rather than the one that passes.

## Proof

Green **both** ways, which had never been true before:

| condition | result |
|---|---|
| `FORCE_COLOR=1` (what CI now runs) | 157 passed |
| no colour | 157 passed |
| `tools/test_shipped_tools_contract.py` | 36 passed |

Run in a clean virtual environment holding only pytest, typer, rich and requests, so this does not lean on anything a particular machine happens to have.

And the assertions still discriminate - stripping escapes did not make them pass on anything. Changing one expectation to `"901 directories"` fails as it should:

```
with a deliberately WRONG expectation: 1 failed, 10 passed
restored:                              11 passed
```

No .NET or web code is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,11 +112,11 @@ jobs:
   # PyPI. If a dependency is added this job fails loudly with a ModuleNotFoundError rather than
   # quietly skipping anything.
   #
-  # cc-devthrottle's OWN tests are deliberately not run here yet. Adding them was the first thing
-  # this job did, and it immediately found why they had never survived an unattended run: five of
-  # them assert plain substrings against Rich-STYLED output, so they pass on a plain console and
-  # fail wherever colour is on - which is what a CI runner is. That is a real defect in those tests
-  # and it gets its own change rather than being silenced here with a colour flag.
+  # cc-devthrottle's own tests run here too. Adding them was the first thing this job did, and it
+  # immediately found why they had never survived an unattended run: five asserted plain substrings
+  # against Rich-STYLED output, so they passed on a plain console and failed wherever colour is on -
+  # which is what a runner is. Those assertions now compare TEXT rather than rendering (#1082), and
+  # the suite is required to be green both with colour and without.
   tool-contracts:
     name: Tool contracts (Python)
     runs-on: windows-latest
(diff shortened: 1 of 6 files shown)


thefrederiksen:
#2300 has merged, so this is rebased onto it and now contains only its own commit - `tools/cc-image/src/cli.py` has left the diff, and what remains is the five assertions, the new `tests/conftest.py`, and the restored CI step.

Main moved twice while this was open (#2300 and #2302), so I re-ran both suites against the NEW base rather than relying on the earlier green:

| suite | result on the rebased branch |
|---|---|
| `tools/test_shipped_tools_contract.py` | 36 passed |
| `tools/cc-devthrottle/tests/` with `FORCE_COLOR=1` | 173 passed |

That count is up from 157 because #2302 added sixteen tests to this suite while this pull request was open. They pass under forced colour too, so nothing new arrived carrying the same fragility.

PR #2304, opened 2026-07-30
THROWAWAY fault-3-only proof for #2298 - DO NOT MERGE

thefrederiksen:
**THROWAWAY - DO NOT MERGE, DO NOT REVIEW. Exists to be observed failing, then deleted.**

Completes the revert proof for **#2298**. The combined fault-injection branch (**#2299**) could not prove
one of its seven predictions, and the reason is a CI-ordering fact worth knowing:

> `Run tests` (the solution) failed, and `Run installer tests` runs **after** it - so that step was
> **skipped** and the setup-engine test never executed.

Six of seven predictions were confirmed on #2299. The seventh was **unverified** - which is not the same as
confirmed, and not the same as wrong. This branch gets it verified.

## What this is

#2298 in full, with **one** fault injected: all three engine cancel messages name a control again
(`Click "Sign in to DevThrottle" to try again`). `git diff` against the fix branch is 3 lines in
`GatewayAccountEnrollRunner.cs`. Nothing else differs, so the solution tests pass, the installer step
**runs**, and the engine test is reached.

## Predicted RED - exactly one test

`CcDirector.Setup.Engine.Tests.GatewayAccountEnrollRunnerTests`
- `EverySignInCancelledMessage_StatesTheFact_AndNamesNoButton`, on
  `Assert.DoesNotContain("Sign in to DevThrottle", ...)`

The job therefore fails at **`Run installer tests`**, not at `Run tests`.

## Predicted GREEN

- The whole `Run tests` step, including all **321** Avalonia tests. **If that step fails here, the
  isolation is wrong and this run proves nothing** - that is this branch's positive control.
- The other **430** setup-engine tests.

Already observed locally on #2299's branch: that test fails on exactly that assertion, with the substring
found at position 30. This is the same observation on CI, in a clean environment, at a step that can
actually be reached.

### DIFF
--- a/src/CcDirector.Avalonia.Tests/GatewayStepOptionsTests.cs
+++ b/src/CcDirector.Avalonia.Tests/GatewayStepOptionsTests.cs
@@ -1,0 +1,238 @@
+using Avalonia.Automation;
+using Avalonia.Controls;
+using Avalonia.Controls.Primitives;
+using Avalonia.Headless.XUnit;
+using Avalonia.LogicalTree;
+using CcDirector.Core.Configuration;
+using CcDirector.Core.Onboarding;
+using Xunit;
+
+namespace CcDirector.Avalonia.Tests;
+
+/// <summary>
+/// The gateway step's three options - Hosted, Self-hosted, Not now - must stay REACHABLE. All of them,
+/// by mouse and by keyboard, before a sign-in attempt and after one that failed or was cancelled.
+///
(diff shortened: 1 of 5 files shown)


thefrederiksen:
Closing without waiting for its run. The fault-3 claim is already established: the engine test fails on exactly Assert.DoesNotContain("Sign in to DevThrottle"), observed directly. A further full CI run to re-confirm one string assertion is not worth another red draft on the repository - #2298 is merged and green. Deleting.

PR #2313, opened 2026-07-30
The product stops feeling slow, and stops lying about what it is doing

thefrederiksen:
Closes #1111, #1105, #1107. Substantially addresses #1100 (remainder filed as #1125).

Four issues given to one seat because they are one subject and would have collided on the same user-interface thread. The release manager's hypothesis was that #1111 - per-session work on the paint thread every second - might be the shared cause, and that the late-answering buttons could be the same thread being busy. **It is not, and the honest answer matters**, so it is the first thing below.

---

## Was it one cause? No - and here is the measurement

The dictation poll really was burning the paint thread, and it really did get worse with every session. Measured with a harness added in this pull request (`ResponsivenessCostMeasurement`), against a 28-marker store shaped like the one on SOREN_NORTH:

| sessions | old: ask per session | new: read once per tick |
|---|---|---|
| 1 | 2.3 ms | 0.8 ms |
| 10 | 25.4 ms | 0.7 ms |
| 19 | 41.7 ms | 0.7 ms |
| 27 | 58.1 ms | 1.0 ms |

**The positive control the release manager asked for is the left-hand column**: the cost climbs with the session count, so the defect is observable on this machine before anything is claimed about removing it. At two dozen sessions the Director spent roughly a tenth of every second doing synchronous file input/output on the thread that paints, for ever.

But that does not explain the Cockpit button. Eight seconds of nothing is not thread contention - it is an eight-second `HttpClient` timeout with no state change before the await and no re-entrancy guard. **Those buttons are a separate defect that happens to live nearby.** Fixing the thread makes the whole application feel fresher; it would not have moved the Cockpit button by a single millisecond. So this is fixed once for the thread, and once for the buttons, rather than eleven times.

---

## #1111 - the Director stops getting slower with every session

**The dictation poll**, three things wrong, all three fixed:

- Read once per tick, not once per tick *per session*. Each session used to re-enumerate the store and re-read every marker for itself, so the work was sessions x markers to answer a question with one store-wide answer.
- Moved off the dispatcher onto the thread pool; only the result is posted back. A tick is skipped while the previous read is still out rather than queued behind it, so a slow disk cannot build a backlog of answers that are already stale.
- A settled marker is read once and skipped thereafter, keyed on its last-write time.

That third point deserves its correctness note. **Only `Delivered` and `Abandoned` are memoized, deliberately not "anything that is not `Pending`"** - `DictationDeliveryRecord` documents a `FAILED` back to `PENDING` transition, so settling every non-pending marker would pin a retried dictation unlocked for the life of the process. There is a test for exactly that transition, and it holds the file's last-write stamp still so the terminal-state rule is what it actually proves (written the obvious way it passed with the rule removed - see the second commit).

The store itself is never cleaned up: `SweepAbandoned` is wired to the voice-turn root, not this one, which is why that machine held 28 markers, every one terminal, the oldest three weeks old. The Director cannot fix that by deleting them - they are the Gateway's deduplication guard - so bounding the store is filed as **#1126**.

**Git status** was probed once per session instead of once per repository: 23 sessions over 6 trees meant 23 `git status` runs against trees that live agents are writing to. Grouped, that is **6**. Counted rather than timed, because the number that matters is how many times the monitor asks a question it already has the answer to.

One correction to the issue: there *is* already a shared ten-second cache in `GitStatusProvider`, so item 2 slightly overstates the problem - but it cannot save these, and the reasons are worth writing down. It is keyed on the raw path so two spellings of one directory never share an entry; it is populated on completion so probes issued together all miss together; and its ten seconds are shorter than the fifteen-second poll, so it is cold at the top of every cycle. Deduplicating in the monitor is what actually collapses the work.

**Path normalization** (item 3) is handled where it bites: the grouping key is canonical (`RepoPathKey`), so the fix is not defeated by one directory being stored as both `D:/x` and `D:\x` - which was 8 spellings for 6 repositories on that machine. It is a comparison key and never a display string, so `RepoPath` keeps the casing the user chose.

---

## #1105 and #1107 - a click is answered at once, once, and out loud

`BusyAction.RunAsync` is the shared helper both issues asked for, modelled on `ConnectionsView.OpenConnection` - which #1107 correctly identified as the best existing example - rather than on a new pattern. Feedback before the await; a real guard tracked in an explicit in-flight set rather than by reading `IsEnabled`; always restored in a `finally` including on the failure path; and failures surfaced to the **screen**.

That last rule is its own bug: three of the nine sites reported failure to `FileLog` and nothing else, so a failed Record left the interface saying IDLE next to a button that appeared not to exist.

**The Cockpit decision, recorded in #1105 as the issue required.** The issue's option 1 - "open the browser anyway, we know the Cockpit URL" - **is not available**: the Gateway owns that URL and hands it back from `GET /cockpit`, and composing one from the base address is the dumb-client violation that once pointed the Learn button at a route that did not exist. The probe *is* how we learn where to send the user, so it stays. What was wrong was the silence around it: the button now shows a working state before the call, the timeout is 4 seconds rather than 8, and the failure offers **Retry** with the address it tried instead of a bare OK.

Eleven sites converted, each with its own defect - the recorder failing silently and orphaning a temp directory per click, the destructive branch delete running inert and re-entrant, Compact's label that was never a guard (three clicks bought three compactions, which costs real money), the gateway scan where last-finisher-wins, the commit that re-fired `git commit`, and the rest. `SendPrompt` was safe *by accident* and now has an explicit guard.

---

## #1100 - the gauge stops inventing the context window

The gauge showed `ctx 184k / 200k (92%)` in red for a session with roughly 800,000 tokens of headroom. The used count was honest; the denominator was derived from the letters "opus" in a model name.

The rule, applied to every driver: **a driver may only report a context window its agent told it, and may never derive one.** Both derivation tables are **deleted** rather than left unused, because code that stays gets called again. Claude and pi now report used tokens and no denominator; Codex is unchanged and is the reference, because it reads `model_context_window` from its own session file. A new `WindowSource` field records provenance - the difference between a measurement and an assumption, which the screen previously could not express.

**A Claude session now shows `ctx 184k` with no percentage, no bar and no colour.** That is a deliberate, visible product change, and it is the state #1100 specifies: "a missing number is a minor annoyance, a confident wrong number gets acted on." Restoring the number honestly - via Claude Code's status-line payload and pi's extension call - is **#1125**, which carries the two caveats #1100 raised about the status line.

The old tests asserted that the table returned its own contents and were green throughout the entire life of the bug. They are replaced by tests that pin the rule: no window derived from a model name, from a launch argument, or by self-correcting against the observed size.

---

## Proof

Every guard here was verified by breaking it on purpose, after the fix was committed:

| fault injected | what went red |
|---|---|
| memo settles `Failed` markers too | `LockedSessionIds_AFailedMarkerThatGoesBackToPending_IsStillSeen` |
| git grouping keyed on the raw `RepoPath` | `RefreshOnce_TreatsTheSameDirectorySpelledTwoWaysAsOneRepository` |
| `BusyAction` re-entrancy guard removed | `FiveClicksProduceOneRunAndOneFailure` |
| the 200,000 guess restored | 11 cases across `ReadContextUsage_NeverDerives*` |

The first of those is the interesting one: written the obvious way, that test passed **with the rule removed**, because a state transition moves the file's last-write time and the second guard re-read the marker anyway. It was named after one guard and exercising another. Pinning the stamp isolates the rule, and only then did removing it turn the test red.

Two defects in this branch's own work were caught reviewing it against the surrounding code and fixed in the last commit: the branch-delete summary was written just before a fire-and-forget reload that hides the status line, so the outcome would have flashed and vanished; and the Cockpit retry called itself instead of looping.


thefrederiksen:
## Rebased onto main, and one claim in the description above has changed

`da0228fce` (#2312, "read the dictation marker store once a tick, not once per session") landed on main while this branch was in flight. **That is the same fix as the first item here**, and not by coincidence: I built it by adopting uncommitted work I found in an abandoned `dt-1111-dictation` worktree this morning, and #2312 is that same work, finished and merged by whoever owned it. My reading that it was orphaned was wrong.

So the honest accounting after the rebase:

- **Not mine any more:** reading the marker store once per tick instead of once per session. That shipped in #2312. The three conflicting files resolved to my side because mine is a strict superset of the same code, not because I disagreed with it.
- **Still only here:** moving that read OFF the paint thread (#2312 left it on the dispatcher), the settled-marker memo, the git probe grouping, `RepoPathKey`, the measurement harness, the eleven buttons, and the context-window rule.

The measurement table in the description is unchanged and still correct - the "old" column is the pre-#2312 shape either way - but one number in the code comment was stale and is corrected in this rebase: it quoted 3.9ms/119.9ms from my first harness run, where every later run and everything I have reported says 2.3ms at one session and 58ms at twenty-seven. Same shape, different machine load; the comment now matches what I actually report.

Re-verified after the rebase rather than assuming it survived: solution builds clean, 202 affected Core tests pass, full Avalonia suite 330 pass.

PR #2319, opened 2026-07-30
WIP: remove SQLite from the hosted Gateway (statistics store to Postgres)

thefrederiksen:
**Draft. Lands nothing.** Opened so the CI pull_request trigger fires on every push to this branch.

CI runs only on push to main and on pull requests based on main, so no commit on this mission had ever been built or tested by CI. Local full-suite runs had been stopped in favour of CI on the commit - correct in principle, and false for these branches, which left no whole-suite signal at all. This restores the one we assumed we had, and it makes the green belong to a commit.

Work in progress against the incident of 2026-07-30 (docs/incidents/gateway-roster-500-2026-07-30.md). Progress and evidence status are tracked in docs/step2-proof-ledger.md - a row is closed by evidence, never by an honest account of failing to obtain it. Do not merge; the ledger is the gate.


thefrederiksen:
Superseded. The mission's code shipped to main through separate pull requests, and this branch was 79 commits behind with only four files of net difference - two of them older than the copies already on main.

The two files that were genuinely unmerged have been carried across in #2371, now merged: the proof ledger (`docs/step2-proof-ledger.md`) and the merge-ancestry guard (`scripts/check-mission-branch-ancestry.ps1`). Nothing else on this branch was unique.

The `nosqlite-stats` branch is left on the remote as the record of the commit history; its worktree has been removed.

PR #2320, opened 2026-07-30
WIP: a statistics failure can no longer take down the session roster

thefrederiksen:
**Draft. Lands nothing.** Opened so the CI pull_request trigger fires on every push to this branch, and so the final commit has a CI result before it is considered for merge.

CI runs only on push to main and on pull requests based on main, so no commit on this branch had been built or tested by CI.

On 2026-07-30 GET /sessions wrote to a SQLite statistics database on a shared network mount on every read. A slot swap put two containers on that mount, the indexes corrupted, the write threw, and the roster handler answered HTTP 500 to every client for 32 minutes. This makes the roster read pure: statistics are observed at the push ingress behind a bounded single-consumer queue that never blocks the caller, drops are counted rather than silent, and a stalled write cannot convoy the ingress. The snooze prune - an Entity Framework write that also sat on the read path - moves to the ingress too.

Outstanding before this is ready: two endpoint-level tests proving GET /sessions answers 200 with the full roster while the statistics store and the snooze store are failing, each watched red with the pre-change wiring restored.


thefrederiksen:
Abandoned. Its third independent review rejected it on a finding that is fatal to the design rather than to the details: the snooze prune it moves onto a queue holds the same lock the roster's own reads take, so a stalled prune convoys GET /sessions anyway - the branch does not deliver the isolation it exists for. The roster also still performs a per-session directory create on the shared Azure Files mount, so the comment claiming no store write remains was false.

What was good in it - the bounded single-consumer queue, the health counters, the accepted-push gate, the statistics page degrading to a named 503, and about a dozen tests that assert real things - is being salvaged into a smaller piece rather than resurrected here.

The branch pure-roster-read stays for salvage. Review at codex-review-step1-round3.md on that branch. The underlying incident is docs/incidents/gateway-roster-500-2026-07-30.md, and the statistics store has since been moved off SQLite entirely and deployed (#2335, live as 1.9.1).

PR #2336, opened 2026-07-31
Allowlist the statistics design-time factory, so main builds green again

thefrederiksen:
Main has been red on `NoCrossMachineLoopbackGuardTests.No_new_production_file_hardcodes_cross_machine_loopback` since the no-SQLite merge (#2335) landed, which means the Core suite fails for anyone who runs it.

The merge added `GatewayStatsDbContextPostgresDesignTimeFactory.cs`, and it carries a `Host=localhost` connection string. That is the same case as `GatewayDbContextDesignTimeFactory`, which is already on the allowlist for the same reason: `dotnet ef migrations add` builds the model and writes source without ever opening a connection, so the string is a throwaway design value. The running Gateway derives its statistics connection through `StatsConnectionSelection` and never reads it.

I checked every production file against the allowlist rather than trusting the single name the failure printed. That factory is the only offender, and the companion `Allowlist_has_no_stale_entries` check finds nothing stale.

Verified on a worktree cut fresh from origin/main: both guard tests pass.


thefrederiksen:
Closing as superseded, not rejected - the change this makes is already on main.

This pull request adds one line: the allowlist entry for `GatewayStatsDbContextPostgresDesignTimeFactory.cs` in `NoCrossMachineLoopbackGuardTests`. That exact entry is present on main today, so the whole diff against main is a single line that duplicates a key already in the dictionary. That also explains the red build here rather than any defect in the reasoning: adding a key twice does not compile.

The branch was ninety-four commits behind main, which is how the duplication went unnoticed - the fix landed by another route while this sat.

Nothing is lost by closing it. The design-time factory is allowlisted, the guard passes on main, and the reasoning that justified the entry is in the comment that shipped with it.

PR #2340, opened 2026-07-31
Add a memory diagnostics library: read it, rule on it, spot the leak

thefrederiksen:
A Director left running for a day reached 9.4 gigabytes and nothing in the product could see it coming. This adds the measurement and the ruling. **No wiring: no sampler, no panel, no caller** - the wiring is a separate change so this one can be judged on its own.

## Three layers, following the existing Facts/Fold shape

| Type | Job |
|---|---|
| `MachineMemoryProbe` -> `MachineMemoryReading` | The machine's true totals, from one `GetPerformanceInfo` call |
| `ManagedHeapProbe` -> `ManagedHeapReading` | This process's own heap by generation, Large Object Heap called out |
| `MemoryPressureRule`, `LargeObjectHeapLeakRule` | The two rules, stated once each |
| `MemoryPressureFold` -> `MemoryPressureStatus` | The finished verdict: headline, detail, colour, advice, offered actions |
| `MemoryPressureTracker` | Bounded history, debouncing and trend |

The reading is deliberately **not** a sum over processes. On the machine this came from, process working sets added to 46 GB while 53 GB was in use; the missing 7 GB was kernel pool, which belongs to no process and cannot be discovered by walking a process list. A report built by summing processes silently loses it and then reads as wrong.

## Four decisions, each from a measurement rather than a preference

**Pressure is keyed on commit, not physical memory.** The measured machine sat at 84 percent physical and felt fine while its commit peak reached 106.28 GB against a 105.78 GB limit. Commit exhaustion is what makes allocations fail; physical pressure only makes things slow. A physical-based monitor reports "fine" right up to the failure. Two tests pin this: high physical with commit headroom is not critical, and commit near its limit is critical even with 25 GB of physical free.

**Large Object Heap growth only counts if it survived a generation 2 collection.** That heap is swept only on such a collection, so growth without one is uncollected garbage rather than retention. Without this clause the detector would fire on healthy processes constantly. It returns `Undetermined` rather than guessing.

**A level must hold for three readings before it is announced.** Build activity moved the measured totals by 8 to 10 GB within minutes, so a single sample crossing a threshold says nothing, and an alert that cries wolf is worse than none. Recovery is announced immediately, so a machine that is fine stops shouting at once.

**The reclaim advice promises no byte figure**, with a test enforcing it. Which build nodes are genuinely free is not knowable by inspection - 52 of 81 apparent orphans turned out to be serving live builds - so a number there would be invented.

Thresholds are fractions of the machine's own limits rather than absolute byte counts, so they do not become a stale measurement of one machine.

## Verification

Twenty tests, using the real numbers from the incident so they answer a specific question: would this have caught it. The rules are pure functions, so they are tested with no machine, no timer and no leak.

The probes were also run against a live machine, reading 63.78 GB total, a 105.78 GB commit limit and a 106.28 GB peak, and correctly reporting that the machine had already exhausted commit since boot. That live run is what caught a defect in the fold itself: at level Normal it correctly withheld the reclaim action while the advice still told the reader to use it - advice pointing at a button that was not there. Advice and action now come from the same decision, with a regression test. It only appeared against a real machine, not a fixture.

## Note on the .NET check

The full `CcDirector.Core.Tests` suite was run locally: **4094 passed, 1 failed**. The failure is `NoCrossMachineLoopbackGuardTests` flagging `src/CcDirector.Gateway.Migrations.Postgres/GatewayStatsDbContextPostgresDesignTimeFactory.cs`.

That failure is **pre-existing on main and unrelated to this change**. This branch adds six files, all under `Memory/`; the flagged file is already on main unchanged, and it is the exact file pull request thefrederiksen/devthrottle#2336 exists to allowlist. Expect the .NET check here to be red for that reason until main is green again.

Recommend not merging this until main is green, so it is not merged past a red that could hide a real regression.

## Scope

Library only. Nothing calls it yet, so it changes no behaviour. Related: issue thefrederiksen/devthrottle_internal#1286, and `docs/incidents/2026-07-30-director-memory-investigation.md`.

### DIFF
--- a/src/CcDirector.Core.Tests/Memory/MemoryPressureFoldTests.cs
+++ b/src/CcDirector.Core.Tests/Memory/MemoryPressureFoldTests.cs
@@ -1,0 +1,342 @@
+using CcDirector.Core.Memory;
+using Xunit;
+
+namespace CcDirector.Core.Tests.Memory;
+
+/// <summary>
+/// The memory pressure rule, the leak rule, and the fold.
+///
+/// The numbers in these tests are the real ones measured on SOREN_NORTH on 2026-07-30, when a
+/// Director that had been up for a day held 7.53 GB of Large Object Heap out of a 9.74 GB heap
+/// and the machine's commit peak passed its own limit. Using the real shape means these tests
+/// answer a specific question: would this library have caught it?
+/// </summary>
+public class MemoryPressureFoldTests
+{
+    private const long GB = 1024L * 1024 * 1024;
+
+    private static MachineMemoryReading Machine(
+        long physicalTotal, long physicalAvailable,
+        long commitTotal, long commitLimit, long commitPeak = 0)
+        => new(
+            PhysicalTotalBytes: physicalTotal,
+            PhysicalAvailableBytes: physicalAvailable,
+            CommitTotalBytes: commitTotal,
+            CommitLimitBytes: commitLimit,
+            CommitPeakBytes: commitPeak == 0 ? commitTotal : commitPeak,
+            KernelPagedBytes: 0,
+            KernelNonPagedBytes: 0,
+            ProcessCount: 0, ThreadCount: 0, HandleCount: 0);
+
+    private static ManagedHeapReading Heap(
+        long loh, long total, int gen2Collections, DateTimeOffset at, long fragmentation = 0)
+        => new(
+            TotalHeapBytes: total,
+            Gen0Bytes: 0, Gen1Bytes: 0,
+            Gen2Bytes: Math.Max(0, total - loh),
+            LargeObjectHeapBytes: loh,
+            PinnedObjectHeapBytes: 0,
+            LargeObjectHeapFragmentationBytes: fragmentation,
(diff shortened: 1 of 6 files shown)


thefrederiksen:
Closing as stale, not as rejected - and deliberately WITHOUT deleting the branch, so nothing is lost.

The work is real: a memory diagnostics library, about 1,100 lines across six files, including a pressure tracker and a pressure fold. What has gone stale is its footing. The branch is ninety-three commits behind main, its build is red, and it has not been touched since 31 July - during which main took the tenant boundary, the local trust boundary, prompt retention, the hosted statistics rewiring and two releases. Reviving it is a rebase-and-re-verify job, which is real work and should be a decision rather than something that drifts along in the open list.

`feat/memory-diagnostics` stays on origin. If the diagnostics are wanted, branch fresh from today's main and bring the ideas across, rather than merging a fortnight-old view of the Gateway.

Closed while sweeping the open list toward a release. Reopening it costs nothing if it is wanted.

PR #2343, opened 2026-07-31
Prompt history gets a retention window, export, and delete (CR-3b)

thefrederiksen:
Closes the product half of CR-3b (devthrottle_internal issue 1180, label release-must-have): the Gateway's prompt log held every prompt and agent reply forever, with no delete route and no export, and its own doc comment said so. Prompt text is customer content - proprietary code, file paths, error output, credentials pasted for debugging - so the unbounded model was itself the release-blocking finding.

What changes:

- **Retention is bounded.** New PromptLogRetentionSweep enforces a window on the same footing as the other bounded stores (activity 30 days, session history 90, dictation transcripts 30, voice-turn audio 24 hours). The default is 30 days in ONE named constant, clearly marked PROVISIONAL: the owner's 30-versus-90 decision is still open, and 30 ships as the conservative interim per the release coordinator's ruling today. A self-host operator may choose their own window via CC_GATEWAY_PROMPT_RETENTION_DAYS; the hosted Gateway always runs the published default, and a malformed override throws at startup rather than silently running a different window than the operator believes.
- **The purge walks partitions on disk, not the tenant census** - so a partition orphaned by a deleted tenant still ages out instead of holding that customer's prompt text forever. Wired in GatewayHost on the exact ActivityRetentionSweep timer pattern (overlap guard, startup delay, disposal).
- **Export and delete.** GET /prompts/export downloads the requesting account's entire prompt history as one JSON file; DELETE /prompts erases it immediately. The Gateway holds the single copy (the Director keeps none) and makes no backups of it, so the delete IS the erasure. Both routes are tenant-scoped in PromptEndpoints.cs exactly like the existing verbs - per the release plan, nothing touches GatewayEndpoints.cs or ControlApiHost.cs.
- **The hosted first-run disclosure.** The wizard's hosted-gateway card now states, at the moment of choosing hosted: your agents run on your machines; fleet metadata and conversation history are stored on the hosted gateway so remote history and supervision work.

The customer-facing data map this code enforces is devthrottle_internal pull request 1181 (docs/architecture/data-map.md), the single source the /privacy page renders - contract agreed with the website mission, recorder policy confirmed with the local-trust-boundary mission (recorder off by default, purge on session removal, local disk only).

Tests: PromptLogRetentionTests covers the purge (aged day removed, cutoff day kept, orphaned partition aged out, foreign files untouched), the erasure (one tenant's history gone, nobody else's touched), the export read, the window-resolution rules (hosted ignores the override, malformed override refused loudly), and the sweep end to end. PromptEndpointsTests adds export and delete over a real HTTP pipeline, proving deleted rows are gone from both the ranged read and the export. Test-run results are being posted as a comment once the machine-wide test lock frees (the statistics release-gate run has it at the moment).

### DIFF
--- a/src/CcDirector.Avalonia/FirstRunWizardDialog.axaml
+++ b/src/CcDirector.Avalonia/FirstRunWizardDialog.axaml
@@ -587,7 +587,7 @@
 
                         <RadioButton x:Name="GatewayHostedCard" GroupName="GatewayChoice"
                                      Classes="gatewayCard emphasized" IsChecked="True"
-                                     AutomationProperties.Name="Hosted gateway, recommended. We run it for you - sign in and this machine is enrolled."
+                                     AutomationProperties.Name="Hosted gateway, recommended. We run it for you - sign in and this machine is enrolled. Your agents run on your machines; fleet metadata and conversation history are stored on the hosted gateway so remote history and supervision work."
                                      IsCheckedChanged="GatewayCard_IsCheckedChanged">
                             <StackPanel Spacing="6">
                                 <StackPanel Orientation="Horizontal" Spacing="10">
@@ -601,6 +601,12 @@
                                 </StackPanel>
                                 <TextBlock TextWrapping="Wrap" FontSize="13.5" Foreground="#5A616B"
                                            Text="We run it for you. Sign in and this machine is enrolled - phone access, voice, and the morning report work immediately." />
+                                <!-- The hosted data disclosure (CR-3b): the one honest sentence about what
+                                     leaves this machine, stated at the moment of choosing hosted - not a
+                                     broad "runs on your machines" that hides the history push. The data map
+                                     the /privacy page renders says the same thing in full. -->
+                                <TextBlock TextWrapping="Wrap" FontSize="12.5" Foreground="#5A616B"
+                                           Text="Your agents run on your machines. Fleet metadata and conversation history are stored on the hosted gateway so remote history and supervision work." />
                                 <!-- State the plan, never assert what this user already owns: on a
                                      brand-new free account "included with your subscription" is false. -->
                                 <TextBlock FontSize="12.5" Foreground="#8A909A"
--- a/src/CcDirector.Gateway.Tests/ClientErrorEndpointsTests.cs
+++ b/src/CcDirector.Gateway.Tests/ClientErrorEndpointsTests.cs
@@ -45,39 +45,50 @@ public sealed class ClientErrorEndpointsTests
 
     private const string Sensitive = "the password is hunter2 and the prompt said fix login";
 
-    [Fact]
-    public void DurableLine_reduces_every_client_controlled_field_to_structure()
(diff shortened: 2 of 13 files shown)


thefrederiksen:
Local proof, run on this branch (commit 2090c44df):

- Gateway prompt suites (GatewayPromptLogTests + PromptEndpointsTests + PromptLogRetentionTests, filter FullyQualifiedName~Prompts): **43 passed, 0 failed** (16 seconds, run under the per-user Gateway suite lock, no concurrent run).
- Avalonia suite (covers the first-run wizard step model and views): **345 passed, 0 failed**.
- CcDirector.Gateway and CcDirector.Avalonia both build with 0 warnings, 0 errors.

What the tests prove against the exit rows:
- A daily file aged past the window is deleted by the sweep and a young one kept, in the Local partition, an account partition, and a partition orphaned by a deleted tenant; the file on the cutoff day itself is kept, so nothing ever lives less than the window (PromptLogRetentionTests).
- DELETE /prompts removes one account's entire history and the rows are then shown GONE from both the ranged read and the export, while another account's and the local partition's files are untouched (PromptEndpointsTests, real HTTP pipeline; PromptLogRetentionTests for the partition isolation).
- GET /prompts/export downloads the whole history as one attachment regardless of age.
- The window resolution rules: unset means the default everywhere, self-host may set its own number, hosted IGNORES the override, and a malformed value refuses loudly at startup.

The continuous-integration .NET job on this pull request is red with the pre-existing GatewayStatsDbContext pending-model-changes failures - the known red main (devthrottle_internal 1170, the statistics mission's area). No failure names any file this pull request touches. It re-greens when that fix lands on main.

Flag for the owner (also marked in the code and the data map): the 30-day retention default is the release coordinator's conservative interim; the 30-versus-90 decision is still the owner's to settle, and it lands in exactly two places - PromptLogRetentionSweep.DefaultRetention and the data map.

thefrederiksen:
Updated proof after the review rounds, now rebased onto green main (past the hosted-image marker fix, pull request 2341):

- Combined filtered run over EVERY area this branch touches (Prompts, Dictionary suggestion stores, ClientError, Launcher, AiModels): **189 passed, 0 failed** (1 minute 5 seconds, under the suite lock). The 16 failures seen before the rebase were the known hosted-image marker leak this branch inherited from pre-fix main - the leaked marker file was observed directly in the test output directory, and the same tests pass untouched on the rebased branch.
- New commit on this branch from the independent review: customer content and raw tenant ids driven out of every Gateway log line within this mission's reach (LauncherHub and ClientErrorEndpoints tenant ids now hashed; dictionary-suggestion terms and the Car Mode end phrase now logged by length only). The remaining raw-tenant log sites cross other missions' file ownership and are tracked with an enforcement mechanism in issue 2347.
- The customer-facing data map was corrected twice under adversarial review and /privacy on the live website now renders it exactly (devthrottle_internal pull requests 1181 pending its recorder merge-gate, and 1187 merged and verified live).

PR #2344, opened 2026-07-31
Gateway load-test harness: Stage 0 instrumentation, the rig, and the Stage 1/2 drivers (#1173)

thefrederiksen:
Mission 05 of the 31 July release (devthrottle_internal issue #1173, read-model epic #1159): the hosted Gateway had never been load-tested. This adds the instrument and the harness, measurement only - the fixes belong to the read-model epic. The first baseline produced with it is devthrottle_internal pull request #1184.

## Product changes (measurement only, no behavior change)

- **New: src/CcDirector.Gateway/Diagnostics/LoadTestMetrics.cs** - lock-free counters and fixed-bucket histograms served at GET /diag/loadmetrics (auth-gated; ?reset=true starts a fresh window). Measured: snooze-gate lock wait, snooze database read count (the N+1 counter), fold duration, roster latency, sweep duration and OVERLAP count, DirectorHub connection count and push in-flight/duration, device-credential lookups, and live process numbers. Deliberately no logging in the hot recording paths - these run millions of times per minute under load.
- **Instrumentation touch points**: the three SnoozeRegistry hot reads (wait to enter the gate, read count), StampFleetRolesAndFold (duration), the display-sweep timer (duration + overlap count - counted, never prevented), the access-log middleware (roster latency), DirectorHub (connection and push counters; the two push handler bodies moved to Core methods so a try/finally can time them), DeviceRegistry.ResolveCredential (lookup counter).
- **One real fix: SeedEntitlementForTest now works on Postgres.** The test seam had SQLite-only SQL (INSERT OR REPLACE into an unqualified table) and the load rig is its first Postgres caller - it was a hard syntax error there. Provider conditional on the same IsNpgsql() signal the model shape uses; the Postgres arm creates and upserts the website-owned gateway.entitlements shape (uuid subject); the SQLite arm is byte-identical to before.

## The harness (tools/loadtest, outside the solution, per the tools/harnesses precedent)

- **LoadRig**: boots the REAL GatewayHost hosted-by-environment against a throwaway Postgres, seeds synthetic tenants and device keys through the real registries (GUID subjects - the production shape), writes key files, serves as the target. Storage isolated via CC_DIRECTOR_ROOT.
- **DirectorSim**: N SignalR connections to /director-stream sending the real Hello / PushSnapshot / PushDelta contract with strictly increasing per-connection sequences (the PushedSessionStore acceptance rule); hold mode doubles as the Stage 1 background fleet; scrapes /diag/loadmetrics into a JSONL.
- **stage1-roster.js**: k6 ramping viewers polling GET /sessions at the real 2-second cadence, with a STAGES override for knee-finding and repeat-for-comparison runs.
- **scripts/**: throwaway Postgres up/down (teardown REMOVES the container, and with it every synthetic tenant), and a Stage 1 runner that resets the metrics window and scrapes it through the run.
- **tools/loadtest/README.md**: the one-page how-to-run, including the honesty notes (rig versus production differences).

## The one hard safety rule, enforced in code

Production-looking hosts (azurewebsites.net, anything containing devthrottle) are REFUSED with no override; loopback is free; any other host must be named explicitly in LOADTEST_ALLOW_HOST. The guard is compiled into both consoles (Shared/LoadTargetGuard.cs) and mirrored in the k6 script. The first baseline ran entirely against a local rig with synthetic tenants, torn down after.

## Proof

- Stage 0 sanity on the rig: roster serves the pushed sessions with the Gateway fold stamped; the metrics endpoint returned exactly 3 database reads per session per fold (96 reads = 4 folds x 8 sessions x 3) - the N+1, confirmed by the new counter.
- The full baseline (knee run, standard profile, wedge thread-stack capture, Stage 2) is in devthrottle_internal pull request #1184.
- CcDirector.Gateway builds with zero warnings under TreatWarningsAsErrors; the Gateway test suite run is reported on this pull request before merge.

### DIFF
--- a/src/CcDirector.Gateway/Api/GatewayEndpoints.cs
+++ b/src/CcDirector.Gateway/Api/GatewayEndpoints.cs
@@ -527,6 +527,13 @@ internal static class GatewayEndpoints
         // time, not server work - keeping the reported latency honest.
         app.MapGet("/diag/ping", () => Results.Json(new { t = DateTime.UtcNow }));
 
+        // GET /diag/loadmetrics: the Stage 0 load-test instrumentation window (issue #1173) - lock wait,
+        // fold duration, sweep overlap, hub connection and push-ingress pressure, roster latency, and the
+        // standard process numbers. Behind the same auth gate as every other route. `?reset=true` starts a
+        // fresh window after the read, so each load step scrapes its own numbers.
+        app.MapGet("/diag/loadmetrics", (bool? reset) =>
+            Results.Json(Diagnostics.LoadTestMetrics.Snapshot(reset == true)));
+
         // Result logging: the phone/Cockpit POSTs its completed speed-test result here; the Gateway stamps
         // what IT saw about the connection, writes one greppable log line, and keeps it in a small ring so
         // an agent can read the recent history at GET /diag/results with no phone. This is the "log all of
@@ -3495,6 +3502,11 @@ internal static class GatewayEndpoints
         if (roleUniverse is null) throw new ArgumentNullException(nameof(roleUniverse));
         if (toStamp is null) throw new ArgumentNullException(nameof(toStamp));
 
+        // Load-test Stage 0 (issue #1173): time the whole fold pass. This method is the shared hot path -
+        // the roster, the display sweep, and every accepted hub push all run it - so its duration under
+        // load is one of the numbers the load-test baseline exists to capture.
+        var foldStart = Stopwatch.GetTimestamp();
+
         var all = toStamp;
 
         // Snooze Length mission: an EXPIRED snooze must read as "needs you" again. The registry is the
@@ -3602,6 +3614,8 @@ internal static class GatewayEndpoints
             // pushes this down to the desktop, and the single-session read all emit the same deadline.
             s.SnoozeUntil = snoozeRegistry?.SnoozeUntilFor(s.SessionId);
         }
+
+        Diagnostics.LoadTestMetrics.FoldDurationMs.RecordSince(foldStart);
     }
 
     // Gateway Cleanup CUT RESTORATION (SB-4a): map a tunnel command's null-or-failed result to the faithful
--- a/src/CcDirector.Gateway/Diagnostics/LoadTestMetrics.cs
+++ b/src/CcDirector.Gateway/Diagnostics/LoadTestMetrics.cs
@@ -1,0 +1,259 @@
+using System.Diagnostics;
+
+namespace CcDirector.Gateway.Diagnostics;
+
+/// <summary>
+/// Stage 0 of the Gateway load-test plan (devthrottle_internal issue #1173, mission 05): the numbers that
+/// did not exist and without which a load test is blind on exactly the thing it expects to break. The
+/// read-model review says the hot cost is database work and N+1 reads held under process-wide locks, the
+/// five-second display sweep has no overlap guard, and cross-tenant harm comes from shared locks. This
+/// class measures those claims so the load-test harness can confirm or refute them with numbers:
+///
+///   * how long callers WAIT to enter the snooze registry's process-wide gate (the shared-lock claim);
+///   * how many database reads the snooze registry performs (the N+1 claim - divide by roster requests);
+///   * how long one <c>StampFleetRolesAndFold</c> pass takes (the fold cost);
+///   * whether two display sweeps were ever in flight at once (the missing-overlap-guard claim);
+///   * how many Director stream connections are held and how many pushes are in flight (the socket and
+///     ingress-pressure numbers);
+///   * server-side <c>GET /sessions</c> latency, beside what the load driver measures from outside.
+///
+/// MEASUREMENT ONLY. Nothing here changes behavior, adds a guard, caches a read, or fixes anything the
+/// numbers reveal - the fixes belong to the read-model epic (#1159). Everything is a lock-free counter or
+/// a fixed-bucket histogram (a handful of interlocked adds per observation), cheap enough to stay on
+/// permanently. Deliberately NO FileLog entry/exit logging in this class: these methods run millions of
+/// times per minute under load, and a log line per observation would make the instrument the bottleneck
+/// it exists to find.
+///
+/// Read it at <c>GET /diag/loadmetrics</c>; pass <c>?reset=true</c> to also start a fresh window, so each
+/// load step reads its own numbers rather than a mix with the previous step's.
+/// </summary>
+public static class LoadTestMetrics
+{
+    /// <summary>Time callers spent waiting to ENTER the snooze registry's process-wide gate on the hot
+    /// read path (HoldStateFor / IsExpired / SnoozeUntilFor). Near zero uncontended; the read-model
+    /// review predicts this is what grows first under multi-tenant load.</summary>
+    public static readonly DurationHistogram SnoozeLockWaitMs = new();
+
+    /// <summary>Duration of one whole <c>StampFleetRolesAndFold</c> pass (roles + hold + color/label/triage
+    /// over the stamped set, including its per-session snooze reads).</summary>
+    public static readonly DurationHistogram FoldDurationMs = new();
+
+    /// <summary>Server-side duration of <c>GET /sessions</c>, recorded by the access-log middleware. The
+    /// outside-the-process twin of the load driver's own latency measurement.</summary>
+    public static readonly DurationHistogram RosterDurationMs = new();
+
+    /// <summary>Duration of one display-state sweep tick (all tenants), measured around the whole
+    /// <c>ForEachTenant(Sweep)</c> pass in the five-second timer.</summary>
+    public static readonly DurationHistogram SweepDurationMs = new();
+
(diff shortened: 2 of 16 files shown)


thefrederiksen (commented):
REVIEW FAIL — changes required before approval.

Verified passes:
- The Gateway touch points are additive measurement instrumentation; the existing request/fold/sweep/hub/device behavior is preserved. The PostgreSQL SeedEntitlementForTest arm uses the mapped website-owned shape (gateway.entitlements, uuid subject, timestamptz/boolean/text columns) and a valid ON CONFLICT upsert, while the SQLite arm is unchanged.
- The run is a synthetic local target, not production: the report identifies the local hosted-mode LoadRig, 20 synthetic tenants/100 synthetic Directors/800 synthetic sessions and throwaway Postgres; the committed thread capture is from LoadRig, and the harness seeds only loadtest identities into its guarded throwaway database.
- Required local builds all succeeded with 0 warnings and 0 errors: CcDirector.Gateway Debug with RunWorkspaceTypecheck=false, DirectorSim Debug, and LoadRig Debug.

Blocking defects:
1. DirectorSim does not enforce the strict per-connection send order it claims. Program.cs:233-236 sends reconnect Hello/PushSnapshot from the Reconnected callback while the main pump at 172/250-261 may concurrently send PushDelta calls for the same SimDirector. Interlocked.Increment makes sequence values unique, but it does not serialize assignment plus InvokeAsync; N+1 can be written before N, causing PushedSessionStore to drop the later-arriving N as stale and contaminating the throughput measurement. Put reconnect snapshot and every delta behind one per-SimDirector async gate, including sequence allocation and DTO mutation.
2. The production guard is bypassable with a DNS trailing dot. stage1-roster.js:27-34 and scripts/run-stage1.ps1:29-34 do not trim a terminal '.', so https://name.azurewebsites.net. is the same production host but does not match EndsWith('azurewebsites.net'). The PowerShell runner sends the metrics-reset request before k6's staging check; with LOADTEST_ALLOW_HOST set to the dotted form, k6 permits the full run. Normalize the host exactly as Shared/LoadTargetGuard.cs already does before every request.
3. PR 1184's report is not fully supported by its committed artifacts. There is no Stage 0 raw artifact and no 100-events/s Stage 2 artifact, so the 96-read floor and the all-tunnels-dropped claim cannot be independently verified; there is also no 10-minute health timeline or Postgres telemetry supporting permanent non-recovery / “Postgres never close to limits.” One committed artifact directly contradicts the report: stage1-knee-run-loadmetrics.jsonl reaches sweepInFlight=36 at line 58, while the report twice says the maximum was 16 (the final 91 overlaps / 98 ticks is supported). Commit the missing evidence and correct the maximum before calling the baseline verified.

GitHub does not allow the pull request owner account to submit a REQUEST_CHANGES state, so this fail verdict is recorded as a review comment.

thefrederiksen (commented):
RE-REVIEW FAIL — the sequence-gate fix, direct k6 trailing-dot fix, Stage 0 artifact, and local synthetic-target provenance are verified. All three required Debug builds again pass with 0 warnings and 0 errors. Approval is still blocked by these defects:

1. The production guard is not closed across the Stage 1 tool. scripts/run-stage1.ps1:31-32 still does not trim a terminal DNS dot, and line 45 sends /diag/loadmetrics?reset=true before k6 evaluates its now-correct guard. https://name.azurewebsites.net. therefore reaches production once from the runner. Normalize exactly as Shared/LoadTargetGuard before the first request.
2. DirectorSim's rate/duration accounting is unsound under the measured backlog. Program.cs:110 divides completions by the configured REPORT_SECONDS even though each metrics scrape extends the interval; the fresh 100-eps artifact spans 108.31 s from first to last sample and gains 495 completions (4.57/s), while its last recorded achievedPerSecond says 12.9. The 10-eps artifact similarly records 10.9 while its sample-to-sample aggregate is 7.67/s. Also, the inner catch-up loop at 166 does not check runUntil, so once the 2,000-send semaphore fills it can continue scheduling/draining past DURATION_SECONDS. The committed 100-eps timeline is at 545 completions near 120 s, not the report's 1,054 “after 120 s.” Measure actual elapsed time, enforce the deadline inside the inner loop/wait, record scheduled/completed/backlog at the boundary, then rerun Stage 2 and commit the final summary.
3. PR 1184 remains internally contradictory: the corrected narrative says the sweep maximum is 36, but the resource table at line 107 still says “up to 16 deep.” The claimed 10-minute /healthz non-recovery and “Postgres never close to limits” still have no committed health timeline or Postgres telemetry.

GitHub does not permit the owner account to submit REQUEST_CHANGES, so this fail verdict is recorded as a review comment.

thefrederiksen (commented):
SECOND RE-REVIEW FAIL — the trailing-dot fixes, measured-interval rate calculation, corrected 36-sweep table, Stage 0 evidence, and all three required 0-warning builds are verified. Approval remains blocked:

1. DirectorSim still cannot enforce DURATION_SECONDS under backlog. The inner catch-up loop at Program.cs:166 has no runUntil condition, and its WaitAsync uses stopRequested.Token; that token is not cancelled until line 201, after both loops return. Once the 2,000-send semaphore fills, the inner loop can spend minutes catching up to an old `due` value past the deadline, exactly the prior defect. Use a deadline-linked cancellation token in the inner loop/wait.
2. The claimed WINDOW END totals are printed before cancellation. Program.cs prints deltasOk/deltasFailed, then calls stopRequested.Cancel; queued/in-flight sends therefore have not yet reported their cancellation failures, and their continuations are never tracked/awaited. Cancel at the boundary, await every scheduled send/continuation, then print one final in-window scheduled/completed/failed/backlog record.
3. The Stage 1 runner still does not apply the full shared guard before its bearer-keyed reset: it refuses production-shaped names but does not require LOADTEST_ALLOW_HOST for any other non-loopback host. A custom production domain or arbitrary remote host receives the reset before k6 refuses it. Mirror all three Shared/LoadTargetGuard rules before line 47.
4. PR 1184 references `2026-07-31-artifacts/rig-boot.log`, but that file does not exist in commit eca8d02 or the PR file list. The new operator note itself cites that missing file. Commit it or remove the unsupported references. The report also still says the fold “caps push completions at ~9/s however much is offered” even though its corrected 100-eps result is ~5/s; restate this as single-digit throughput that falls under queue pressure.

GitHub does not permit the owner account to submit REQUEST_CHANGES, so this fail verdict is recorded as a review comment.

thefrederiksen (commented):
REVIEW FAIL — round-three fixes are close, and the corrected internal baseline/rig-boot artifact now closes the prior evidence and provenance findings. Three correctness gaps remain:

1. DirectorSim/Program.cs:189-191: the continuation calls inFlight.Release() before incrementing deltasOk/deltasFailed. The drain exits when CurrentCount == maxInFlight, so the main thread can observe a fully released semaphore and print WINDOW END before the final counter update(s). The totals are therefore not guaranteed settled. Update the counter before releasing, or track and await the send/continuation tasks.
2. DirectorSim/Program.cs:217: windowSeconds is captured after the drain loop. A configured 100-second measurement can therefore be reported as up to 115 seconds even though sends were canceled at the 100-second deadline. Capture measurement elapsed at cancellation/before drain and report drain time separately.
3. 
un-stage1.ps1:38-39: the stated loopback rule does not accept IPv6 loopback. On this runtime, ([Uri]'http://[::1]:7891').Host normalizes to [0000:0000:0000:0000:0000:0000:0000:0001], which is absent from $localHosts, so the local target is refused unless misclassified through LOADTEST_ALLOW_HOST. Normalize/parse IP literals with a loopback check and keep the PowerShell, C#, and k6 guards aligned.

Verification at daf38d2b0: Gateway, DirectorSim, and LoadRig Debug builds all succeed with 0 warnings and 0 errors. Internal commit 5d54a2 contains 
ig-boot-log.txt, uses the corrected filename, records the local hosted/Postgres target and 20 seeded tenants, and uses the 8.5/s and 5/s window figures.

thefrederiksen (commented):
REVIEW FAIL — the counter-before-release change now makes the drain gate reliable, and the pre-drain windowEndUtc correctly excludes drain time. One blocking guard defect remains.

On the documented Windows PowerShell runtime (5.1.26100.8972), ([Uri]'http://[::1]:7891').Host is [0000:0000:0000:0000:0000:0000:0000:0001]. After .Trim('[', ']'), $hostName is still the expanded

thefrederiksen (commented):
REVIEW PASS — verified at 625402c2f.

The cancellation deadline, bounded drain, counter-before-release ordering, and pre-drain window timestamp now make Stage 2 totals settled and window-true. The PowerShell and shared C# guards parse IP literals and use loopback semantics; the k6 guard accepts the exercised compressed/expanded IPv6 loopback forms. Independent checks confirmed IPv6 loopback proceeds past the guard, trailing-dot production is refused without override, unnamed non-local targets are refused, and an explicitly named staging target proceeds.

Gateway, DirectorSim, and LoadRig Debug builds each succeed with 0 warnings and 0 errors. The internal baseline remains at 5d54a2 with the corrected rig-boot artifact and window-rate evidence.

GitHub rejected a formal approval because the authenticated account owns this pull request, so this passing verdict is recorded as a review comment.

PR #2346, opened 2026-07-31
Local testing becomes the everyday gate: concurrent-safe tests, four minutes of sleep removed

thefrederiksen:
# Local testing becomes the everyday gate (devthrottle_internal#1156, steps 1-2 + quick wins)

Makes the local test suite safe to run concurrently across worktrees and recovers ~4 minutes
of pure sleeping, so agents stop waiting fifty minutes on GitHub CI for changes the local
suite already proves.

## What changed

1. **Deterministic dead ports.** Three tests obtained a "free" port by opening a listener,
   reading the number, and closing it - leaving the port unowned, so any process could grab
   it between the close and the test's use. `DeadPortReservation` replaces the pattern with a
   socket bound but never listening: simultaneously reserved (nobody else can bind it) and
   dead (connections are refused). Both properties were measured on Windows before the type
   was written. An architecture test bans the old pattern.

2. **Live-proof isolation.** `CC_GATEWAY_TEST_PG_CONNECTION` now names a TEMPLATE database;
   each run derives its own database with a unique suffix, so two runs can share one Postgres
   server without `EnsureDeleted()` racing each other's schema. Proof artefacts write to a
   per-run subdirectory; fixed temp names are made unique.

3. **Three sleeping tests fixed - coverage up, not down:**
   - `DirectorCommandRouterTimeoutTests` 2m03s -> 1s. It slept through real 30s production
     timeouts to read a string. `DescribeTimeout` is now internal and asserted directly, plus
     a NEW seam test that runs a real 200ms timeout end to end. The timeout is injected via a
     new `GatewayHost` constructor parameter (`directorLaunchTimeout`) - never a static test
     hook, because a process-global switch is shared by every host in the process and is
     exactly what blocks parallel runs.
   - `WingmanTtsStatusPassthroughTests` 1m03s -> 7s (injected options, same principle).
   - `HostedTenantMachineControlTests` 1m46s -> 51s.

4. **The local gate itself:** `scripts/test-local.ps1` builds once and starts all seven test
   projects together, so the six unlocked projects finish while the Gateway suite queues on
   its per-user lock. The repo instruction file (rule 5a) and the commit skill stop telling
   agents to `gh pr checks --watch` the fifty-minute .NET job.

## Evidence

- Branch is rebased onto `d7953e4ac` (the first green main after #1170); solution builds with
  zero warnings on the rebase.
- All 4,691 tests outside the Gateway suite passed locally on this change merged with the
  main-green fix, before the rebase made that merge the branch itself.
- The local Gateway suite run on this exact tree is queued on the machine-wide test lock
  (three sessions contending today - the very problem this issue exists to fix); its TRX
  verdict will be posted on this PR when it lands. CI's full run on this branch is the
  same suite.
- The three fixed tests' timings above are measured, not estimated.

## Explicitly NOT in this PR

- The `DirectorRegistry` teardown drain (parked on `fix/1156-step3-registry-teardown`,
  needs its own soak - see #1156 step 3).
- The qualification harness and any change to `GatewayTestSuiteLock` (steps 4-5). The
  machine-wide lock stays exactly as it is until the concurrency soak proves it removable.

### DIFF
--- a/.claude/skills/commit/skill.md
+++ b/.claude/skills/commit/skill.md
@@ -131,12 +131,41 @@ STEP 7: Open the pull request, merge on green, and park back on main
 The job is not done until the work is MERGED to origin/main. Once the user approved the
 commit (Step 4), drive it home:
 
-1. Open the pull request: gh pr create --fill (or with a title/body matching recent PRs).
-2. Wait for checks to go green: gh pr checks <number> --watch.
-   - If checks fail, fix the failure and push a NEW commit (never amend, never --no-verify),
-     then re-watch. Report the failure to the user if it is not quick and mechanical.
-3. Merge with squash once green: gh pr merge <number> --squash --delete-branch.
+1. RUN THE TESTS LOCALLY FIRST. This is the gate - not GitHub (issue #1156):
+
+       .\scripts\test-local.ps1
+
+   It builds once and runs every test project, starting them together so the six that do not
+   serialize finish while the Gateway suite queues for its machine-wide lock. Use -Fast to skip
+   the Gateway suite when the change provably does not touch it, and say so in the pull request.
+   Do NOT hand-roll a dotnet test invocation - the script is the one place the whole fleet gets
+   faster when the suite improves.
+
+   If it is not green, fix it before opening a pull request. A red local run is a red change.
+
+2. Open the pull request: gh pr create --fill (or with a title/body matching recent PRs).
+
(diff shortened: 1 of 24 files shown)


thefrederiksen:
Local Gateway suite evidence on this exact tree (4571c8f4e, rebased on the green d7953e4ac): TRX run outcome Completed, 4,839 total, 4,799 executed, 4,799 passed, 0 failed, 22m22s. The 40 not-executed are the Postgres live-proof tests, which correctly skip without a template database configured. Judged from the TRX ResultSummary, not the console summary. The CI run on this PR is the same suite; its totals should match.

PR #2350, opened 2026-07-31
chore(release): make cutting a release need only the version number

thefrederiksen:
Cutting v1.9.2 exposed that `scripts/new-release.ps1` **cannot complete against this repo's ruleset**, and left the tag ahead of main.

## The bug

The script ended with:

```
git push origin main
git push origin $tagName
```

main is protected by *Protect main branch (... require pull request)*, so the first push is **rejected** and the second **succeeds**. The result is a tag pointing at a commit that is not on main - v1.9.2 released correctly, but main kept reporting 1.9.1 with no release notes, and needed a hand-made pull request (#2349) to catch up. Every future release would have repeated it.

## The fix

Branch -> commit -> pull request -> **merge** -> **then** tag. The tag is created only after re-reading `Directory.Build.props` from the merged main to prove the bump actually landed. A tag cannot be un-pushed, so it is created last, once there is something correct to point it at.

## Only the version is a decision

```
Current version: 1.9.2
New version [1.9.3]:          <- Enter
```

The next patch is offered and Enter takes it. The release notes may be **uncommitted** - the script commits them with the bump - so the whole procedure is: write the notes, run the script.

## Fails early, by name

Missing or signed-out `gh`; missing or under-200-character release notes; a tag that already exists locally **or on the remote**; a `.csproj` declaring its own `<Version>`; unrelated working-tree changes; not being on main; main behind the remote. Each says what to do about it.

Also fixes the closing message, which pointed at a nonexistent `example-org` URL, and now names the release page and the public download address.

## Docs

`docs/architecture/VERSIONING.md` rewritten to match - the two-step procedure, why the tag is created last, and what the workflow does after it: draft-then-publish, the download mirror, and the hash check that fails the release if the public address is not serving that build.

Validated: parses clean under PowerShell 5.1, ASCII only, no 7.x-only syntax.

### DIFF
--- a/docs/architecture/VERSIONING.md
+++ b/docs/architecture/VERSIONING.md
@@ -59,18 +59,64 @@ release and its exact commit.
   `CcDirector.Core.AppVersion` (or the inline equivalent in projects that do
   not reference Core).
 
-## How a release flows
+## How to cut a release
 
-1. Run `scripts/new-release.ps1`. It bumps **one file**
-   (`Directory.Build.props`), commits `release: vX.Y.Z`, tags `vX.Y.Z`, pushes.
-2. CI (`release.yml`) builds everything from the tagged commit. Every .NET
-   binary is stamped `X.Y.Z` automatically. A guard step fails the release if
-   the tag and `Directory.Build.props` disagree.
-3. The release manifest lists every asset with its version (= the release
-   version) and its SHA-256.
-4. Installed machines see `X.Y.Z > installed` and update every asset in their
+**Two things to do. Write the notes, run the script.**
+
+1. **Write `docs/public/release-notes/vX.Y.Z.md`.** This file *is* the release
+   page - the workflow publishes it verbatim and **fails if it is missing or
+   under 200 non-whitespace characters**. It will not generate a substitute,
+   because a page of internal pull request titles looks like release notes and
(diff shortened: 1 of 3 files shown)


thefrederiksen:
CI caught a real break from this refactor, now fixed.

`ReleaseWorkflowContractTests.NewReleaseScript_RefusesToTagWhenTheWrittenNotesAreMissing` pins the release script **by its source text**. Moving the guards into a `Fail` helper removed the literal `exit 1` those assertions grep for, so three of them failed. The guards themselves were never broken - the test was pinning a spelling.

Fixed on both sides:

- **Script:** builds `$notesPath` from the literal `docs\public\release-notes\.md` again. The forward-slash form stays because it is matched against `git status` output, but the guard's path is written out, which reads plainer anyway.
- **Test:** accepts `exit 1` **or** a `Fail` call as the terminator, and adds an assertion the old version could not make - **that `Fail` itself exits**. A helper that printed and returned would have satisfied the old regex by proximity while carrying on to push the tag, which is the exact defect these guards exist to stop.

Revert-proved: a variant whose `Fail` does not exit is rejected by the new assertion.

`dotnet test --filter ReleaseWorkflowContractTests` - **5 passed, 0 failed**.

PR #2351, opened 2026-07-31
ci(release): mirror only the current build, under names that never change

thefrederiksen:
Removes the per-version copies from the download mirror added in #2348.

## Why they were wrong

The mirror wrote every asset twice - `latest/` and `v<version>/`. I justified the second copy on the grounds that winget manifests and Store submissions pin a URL **and** a hash per version, so a moving URL would break every published manifest on the next release.

That part is true. The conclusion was not: **GitHub already serves an immutable URL per version, permanently, for free.** Anything needing a pinned address uses that. The blob copy duplicated it at ~400 MB per release and dragged in a retention question nobody had answered.

## What it does now

One copy of each file, one name, overwritten every release. That is the whole point of the mirror - an address that never moves - and per-version folders worked against it.

`release-manifest.json` sits beside the installers and carries the current version, size and SHA-256 of every asset, so anything that needs to know what is downloadable reads that rather than inferring it from a folder name. Same pattern mindzie uses.

## Effect

- Storage halves, and stops growing without bound.
- No retention policy to decide.
- Nothing lost: every previous version remains on GitHub Releases, which is and stays the archive.

The `v1.9.2/` folder already uploaded is being removed by hand.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -900,9 +900,10 @@ jobs:
             echo "\`$GITHUB_REF_NAME\` left draft with every asset attached, so there is no window in which it is latest but incomplete."
           } >> "$GITHUB_STEP_SUMMARY"
       # Mirror the files a PERSON downloads to a stable public address
-      # (devthrottle_internal #1186). GitHub stays the archive - this is not a
-      # second copy of history, only ever the latest build plus a short tail of
-      # versions.
+      # (devthrottle_internal #1186). GitHub stays the archive: this holds ONLY
+      # the current build, under names that never change. Per-version copies were
+      # removed deliberately - GitHub already serves an immutable URL per version
+      # for free, so a second copy bought nothing and doubled the storage.
       #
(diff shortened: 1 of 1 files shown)


thefrederiksen:
Pushed a follow-up: the mirror now writes to **`devthrottle.blob.core.windows.net/download/`** with no `latest/` prefix.

```
https://devthrottle.blob.core.windows.net/download/devthrottle-setup-win-x64.exe
```

Two things, both cheap now and permanent later, done before anything public points at the address:

- The account was `stdevthrottledl`, following the `st` prefix used by the gateway's storage account. That convention is for infrastructure nobody reads - this hostname is on the download link people see.
- `latest/` was a leftover from the per-version copies this PR removes. With one copy of each file it distinguishes nothing and implies other versions sit beside it.

The new account is created, the six files are copied across and **hash-verified against the manifest**, cache and content-type headers are set, the repo secret is repointed, and the old account is deleted.

PR #2356, opened 2026-08-01
Three review findings: the safe recorder state becomes the free one, the cross-language acceptance stops passing when it never ran, and the access log says which device asked

thefrederiksen:
Two findings from the independent reviews run over the release-candidate code, both small and both of the same family: a state that reports success without having earned it.

**The recorder default.** `TerminalSessionRecorder`'s constructor defaulted `captureEnabled` to `true` while the shipped product default is `false` and /privacy tells customers recording is off unless they turn it on. The single production call site passes `SessionRecordingConfig.IsEnabled()` explicitly, so nothing records today that should not - this is not a live defect. But a future call site that forgets the flag would silently record, which is the opposite of the posture we publish. The default is now `false`: forgetting it lands on the safe state. Purge-on-removal and the startup orphan sweep run regardless of capture, so `TerminalSessionRecorderPolicyTests` passes unchanged (8/8 locally).

**The acceptance that could not fail.** `CleanInstallCliAuthenticationTests` returned green when Python was absent. That test is the acceptance row for the entire cross-language credential change - the only one that drives the real Python command line against a real host - so on any machine without Python it reported success for a proof that never executed. It now fails when `CI` is set, because continuous integration always has Python and its absence there would mean the acceptance had quietly stopped running; on a developer machine without Python it still passes, but prints that it skipped instead of vanishing into a green tick.

Note for this pull request's own run: the .NET job has no `setup-python` step of its own, so this run is also the check of whether the Windows runner provides Python. If the .NET job goes red here on that assertion, the assertion is right and the environment is the thing to fix - it means the acceptance row has not been running in CI at all.

### DIFF
--- a/src/CcDirector.Core/Wingman/TerminalSessionRecorder.cs
+++ b/src/CcDirector.Core/Wingman/TerminalSessionRecorder.cs
@@ -47,7 +47,11 @@ public sealed class TerminalSessionRecorder : IDisposable
     private bool _started;
     private bool _disposed;
 
-    public TerminalSessionRecorder(SessionManager sessionManager, string? root = null, long maxBytesPerSession = 8L * 1024 * 1024, bool captureEnabled = true)
+    // captureEnabled defaults to FALSE so that the safe state is the free one: the shipped product
+    // records nothing unless the operator turns it on, and a future call site that forgets the flag
+    // inherits that posture instead of silently recording. Production passes
+    // SessionRecordingConfig.IsEnabled() explicitly; this default is the backstop, not the policy.
+    public TerminalSessionRecorder(SessionManager sessionManager, string? root = null, long maxBytesPerSession = 8L * 1024 * 1024, bool captureEnabled = false)
     {
         _sessionManager = sessionManager;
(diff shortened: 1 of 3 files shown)


thefrederiksen:
Added a third finding to this branch, from the same review pass and the same family as the other two.

**The access log could not attribute a file read.** The Gateway already logged every request with its full path, including the file path on a session file read, so the log could show that a private key had been read. The only identity on the line was the remote address, which attributes nothing when the caller is a phone on a mobile network or arriving through a tunnel. Authenticated requests now carry the device id, the device type and the account.

This matters because of devthrottle#2357: until session roots are established by the operator rather than by the request, a device credential can read files broadly on its own account's machines. That gap is a design change and is deliberately not being rushed - but it should at least be attributable while it is open, so a lost phone can be identified and revoked rather than merely suspected.

No key material is logged: the identity object holds neither the raw credential nor its stored hash, so a credential still never reaches the log. The account identifier is already disclosed on /privacy as something service logs may contain. Unauthenticated requests are unchanged.

PR #2367, opened 2026-08-01
The Director checks that the command line on PATH can actually drive it

thefrederiksen:
## The bug

An agent reports "cannot connect to DevThrottle" while the Director's own chip says Connected. Both are true: the Director is healthy, and the `cc-devthrottle` that PATH resolves belongs to an older install and cannot authenticate against it. The cost is not downtime - there is none - it is that the agent blames the network, and the owner spends the morning there.

Nothing caught it because `cc-devthrottle`'s only health check was `actions --json`, which prints a static local list and never contacts the Director. It passes on a machine where every Director-facing verb returns 401. A check answerable without the evidence is not a check.

## What this does

The Director now asks one question of itself on each tools pass: **can a session I spawn actually drive me?** It resolves `cc-devthrottle` the way `CreateProcess` would, runs it against its own Control API address, and reads the exit code.

Two rules the code holds to:

- **The verdict is functional, never structural.** "Is the resolved path under my install directory" looks like the same question and is not - a development build matches no install root and would be reported broken on a healthy machine. The path comparison only writes the explanation.
- **The Director judges; the tool is never asked to judge itself.** The case this catches is a tool too OLD to be correct, and such a tool is also too old to know about any self-report added today.

Surfaced in three places that now agree: a **Sessions** row on the home page (which previously said "All systems go - 8 of 8 tools passing" over this exact fault), the **rail badge** - beside the connection pill, never on it, because that pill was the one thing telling the truth throughout - and a **banner on Settings > Tools** naming both paths, with a non-destructive repair.

## Verification

Verified against the machine's real fault on slot 7, with screenshots: **`docs/qa/setup-health/QA-REPORT.md`**.

Full `CcDirector.Core.Tests`: `Failed: 0, Passed: 4137, Total: 4145`. 26 new tests, shown to fail on purpose twice (forcing the check to always report Working turned 2 of 8 red; forcing the Sessions row to null turned 6 of 7 red).

## Known gaps, deliberate

- **The repair button is not yet exercised.** Its logic is unit-tested and the PATH rewriting is pinned, but no run has proven the badge CLEARS after a real repair. Deliberate - the test machine was kept in its faulty state as the only reproduction.
- **`8 PASS 0 FAIL` still appears above the banner.** Making the smoke check require a Director round trip would report a false red wherever `ToolTestRunner` runs without an endpoint (Control API, setup wizard). A false red is worse than a weak green, so the Sessions row carries the real verdict instead.
- **An unreproduced test host crash.** One abort at ~test 1777, then three clean runs at the same commit. Recorded so the next person to hit it knows it predates their change.

Review to follow after merge, as agreed.


thefrederiksen:
Local suite on the rebased base (origin/main had moved 25 commits; rebase was clean, no conflicts):

| project | result |
|---|---|
| CcDirector.Core.Tests | 4171 passed, 0 failed |
| CcDirector.Avalonia.Tests | 353 passed, 0 failed |
| CcDirector.Launcher.Tests | 110 passed, 0 failed |
| CcDirector.HostedAgent.Tests | 88 passed, 0 failed |
| CcDirector.Engine.Tests | 63 passed, 0 failed |
| CcDirector.Terminal.Avalonia.Tests | 24 passed, 0 failed |

4809 tests, zero failures. Build clean, zero warnings.

Run with `-Fast`, so the **Gateway suite was skipped**. This change touches `CcDirector.Core/Setup` (new), `CcDirector.Core/Home`, and two Avalonia views - no Gateway code, no Gateway contract. Flagging it here rather than leaving it implied.

Merging on local green rather than waiting for the .NET job, as agreed; review to follow after merge.

PR #2379, opened 2026-08-02
The prompt delete erases the copy the Gateway derived from the prompts

thefrederiksen:
## What was wrong

`POST /prompts` copies each session's first prompt - the first 200 characters of it - into the
`session_history` table, where the History page serves it for ninety days. `DELETE /prompts` deleted
the prompt files and left that copy behind, while the endpoint's own documentation said the delete
WAS the erasure. A member who exercised their delete still had their own words on the screen, and
the sentence in the code is the more expensive half of the defect: an engineer reading it had no
reason to look.

## What this does

`DELETE /prompts` now erases what the Gateway derived from the prompts, in the same request, scoped
to the requesting account:

- **`FirstPromptLine` on every row** - the first 200 characters of the member's own prompt.
- **`SummaryText` and the five JSON lists beside it, on every row including a SEALED one.** Those come
  out of the same summariser reading the same prompt log, so they are the same material at one remove;
  erasing only the prompt line would erase the quote and keep the paraphrase. Sealed rows were briefly
  exempt - see below for why that was wrong and how the check that "verified" it missed.
- **Three summary metadata fields are RESET** on those same rows (`SummaryKind`, `SummaryIsPartial`,
  `SummaryAttempts`). A row still claiming a summary exists with nothing behind it is a smaller lie of the same kind.
  Reset also makes the row eligible for the background sweep again, which then finds an empty prompt
  log and settles it honestly as "none" - so the erasure is self-healing rather than something later
  work can quietly undo.
- **`session_history_rollups` rows are deleted outright**, tenant-scoped. That table caches a written
  paragraph per repository per day, derived from those same session summaries, and its own class
  comment calls a past day effectively permanent. Erasing the columns and leaving it would keep
  serving the erased words as prose for up to ninety days. The staleness hash is not a defence:
  eventually-recomputed is a promise about the future, not an erasure now.

The endpoint **reports what it erased** rather than asserting that an erasure happened, and the
counts describe rows that actually carried something - so a second delete honestly reports nothing
to do.

## Three details that are each a way for this to be present in the code and absent in the product

- **The store parameter on `PromptEndpoints.Map` is REQUIRED, not defaulted.** A forgotten store
  would leave the delete erasing files, reporting success, and quietly keeping the derived copy -
  which is exactly the defect being closed. Same reasoning as the tenant boundary above it.
- **The erasure runs inside the request tenant's ambient scope**, because the store's statements are
  filtered by the AMBIENT tenant. Run outside it, the erasure would reach whatever tenant happened to
  be current - silently, and it would be someone else's data.
- **The derived copy is erased BEFORE the files are deleted.** If the pair fails half way, the
  recoverable state is a copy whose source still exists, not an orphaned copy with nothing left to
  prove what it was.

## An in-flight writer cannot undo the erasure

**Found by the independent inspection, and it was a correctness defect rather than a proof gap.** A
delete could report success and be silently undone seconds later, with the member's own words back on
the History page and no error anywhere.

Three writers compute from material read BEFORE a delete and commit AFTER it: the summariser (a model
call in the middle), the roll-up pass, and the prompt ingest. **The metadata reset made the first one
worse rather than better** - it moves the summary kind to null, which is exactly the state in which
`StoreGeneratedSummary` stops refusing a write. The erasure re-armed the writer that undoes it.

The fix is a per-tenant **erasure watermark** (`prompt_erasure_watermarks`, with a migration for each
provider). Every prompt-derived write compares the material it was computed from against the stamp and
refuses anything older. The comparison is on the MATERIAL, not the moment of writing - a write is safe
only if what it is made of came into existence after the delete.

That distinction is the whole of the ingest case. The previous comment argued any racing material must
have been sent DURING the member's own delete; that was **false**, because `ConversationIngestor`
deliberately retries records it previously failed to deliver, so a push arriving today can carry prompts
from weeks ago. `SetFirstPrompt` is therefore judged on the PROMPT'S timestamp, the only comparison that
tells a late delivery from a new prompt.

**Durable rather than in-memory**, because that retry can cross a restart: a watermark a restart forgets
would fail open and silently. **Stamped after the clears**, so it cannot refuse a writer whose material
was read between the stamp and the clear - which is resurrecting nothing.

The required test **holds a real summarisation open across a real delete** (the model call blocks on a
completion source until the erasure has run) and asserts the words do not come back. Each writer has its
own fact AND its own control, because a guard that refused everything would satisfy the first half of
each: a summarisation started after the delete still writes, a roll-up read after it still caches, a
prompt sent after it still lands. One further fact proves the watermark survives a restart - an
in-memory watermark passes every other fact and fails that one.

**Every sequential fact on this branch stayed green while this hole was open.** A test that cannot
express the timing cannot fail on it.

## What the delete does NOT reach, said plainly

The inspection also proved the "single copy" claim false, and it was false before this work: a Director
keeps a first-prompt snippet and per-turn summaries in its own session-history files, and an
expected-first-prompt in `sessions.json` and that file's backup - all on the member's own machine, all
surviving restarts. The test-rig notes record a real machine where 7 session files carried a snippet and
36 carried turn summaries.

Extending the delete across the Directors needs a Director-side command and a story for offline
machines, so it is a feature rather than a correction: **issue thefrederiksen/devthrottle_internal#1278**. What this change does is stop
claiming otherwise - in this code, and in `/privacy` and the data map - and keep the two things apart,
because a copy on our servers and a file on the member's own disk are materially different to a member.

Also still open and named rather than left to be found: the prompt LOG can accept a retried old record,
so it can be repopulated even though the derived fields cannot. Same issue.

## The seal exemption was REVERSED, and the reason is worth more than the change

Sealed summaries were exempt for two rounds on the reasoning that a session's own farewell is not prompt
material. The verification offered for that established that `SummaryKind` reliably tracks **which writer
wrote the row** - which is true, and is the wrong question. What the exemption needed was that the
**content** is not prompt-derived, and nothing establishes that: the seal route accepts caller-supplied
prose and all five lists with no material time and no provenance at all. Arriving through the seal route
is an *operation*, not a provenance, and a seal composed from the member's own prompts is accepted exactly
like any other.

So the exemption was retaining content that may be the member's prompts - permanently, because every later
delete would exempt it again. Sealed rows are now erased with the rest, the seal route is watermarked so a
pre-delete seal cannot land after a delete, and the three landing surfaces no longer claim the closing
summary is not prompt-derived.

## The erasure holds across processes, and old material is refused at the door

The first watermark was the right idea and did not hold. Two ways, both found by inspection:

**The check and the write were separate operations.** Each writer read the watermark, then read its row,
then wrote - held together by a lock on one store *instance*. The hosted Gateway runs two containers at
once during a slot swap (`FileLog` says so in its own comment), and across two processes that lock means
nothing. All three writers are now **single conditional statements** with the watermark comparison in the
`WHERE` clause. The stamp is a conditional monotonic update for the same reason: read-compare-save is only
monotonic within one process, and two racing Gateways could *lower* a member's erasure line. The erasure
now stamps **first** and clears after - a writer blocked by a slightly early stamp is correct; one admitted
by a late stamp is the bug.

The roll-up is an upsert, and an `INSERT` cannot carry a `WHERE` in one portable statement, so it does a
conditional update and, on the insert path, a conditional delete immediately after. That leaves a cached
paragraph visible for the width of one statement if an erasure lands in that instant - stated at the
method rather than hidden.

**"Judged on the material" was false for the summary path.** The summariser stamps the clock just before
*reading* the log, so a summary built entirely after a delete out of retried pre-delete records carries an
honestly recent timestamp. No race and no clock skew needed. So the material is now refused **at the
door**: `GatewayPromptLog.Append` consults the watermark and drops records older than the erasure, which is
the only version of this that does not need every downstream reader to be perfect forever. The Director's
timestamp is clamped to the Gateway's receipt time, because a clock running ahead would otherwise stamp old
prompts into the future and walk them past the erasure.

**Proved against interleaving, not sequence.** The new facts use two independent store instances over one
database - separate locks, the way two processes have - and drive the interleave deliberately. The retry
fact appends an old batch after the delete and runs the **real** summariser: nothing is stored, the model
is never asked, the row settles as "none". Every one has a control, because a guard that refused
everything would satisfy the first half of each.

## The tenant predicate reaches the generated SQL - proved, not assumed

The erasure runs as bulk statements that never load a row, so its tenant scoping comes entirely from
the global query filter. If that predicate did not survive translation to the provider the hosted
Gateway runs on, one member's delete would erase **every account's** history in a single statement.

`TheErasureCarriesTheTenantPredicateOnNpgsqlTests` drives the **product's own statements** against an
Npgsql-configured context and captures the command text Entity Framework hands the provider. Two
interceptors suppress the connection open and the execution, so the SQL is generated, captured and
discarded - no server, no rig, no hour of the machine-wide lock. All four statements come out as:

```
UPDATE gateway.session_history AS s
SET "FirstPromptLine" = NULL
WHERE s.tenant_id = @__ef_filter__ActiveTenant_0 AND s."FirstPromptLine" IS NOT NULL

DELETE FROM gateway.session_history_rollups AS s
WHERE s.tenant_id = @__ef_filter__ActiveTenant_0
```

The assertion checks the predicate is **inside the WHERE clause** and that the **bound value** is this
tenant - text alone would be satisfied by a predicate comparing the column to anything at all. The
generated SQL is written to the test output, so it lands in the TRX and can be read rather than
trusted. It has a control: the same DELETE with the filter deliberately ignored comes out with no
tenant predicate and no bound tenant.

The same capture also shows the null semantics the seal exemption depends on:
`(s."SummaryKind" <> 'sealed' OR s."SummaryKind" IS NULL)` - so a row that never got a kind is
included rather than silently spared.

**This proves TRANSLATION, not EXECUTION.** It does not show PostgreSQL accepts or performs these
statements, and it does not stand in for a run against a real server.

## Proof

Sixteen guards, each revert-proved individually - the guard removed, the test watched failing with the
reported symptom, the guard restored:

| Fault injected | Test that caught it | Symptom |
|---|---|---|
| Endpoint stops calling the erasure | endpoint acceptance | erased count 0, expected 1 |
| Rollup delete ignores the tenant filter | tenant isolation | the other account's paragraph gone |
| Session update ignores the tenant filter | tenant isolation | the other account's prompt line null |
| `FirstPromptLine` no longer cleared | endpoint acceptance | the member's own words still in the column |
| Rollup delete removed | both acceptance facts | rollup row still present |
| The three metadata resets removed | store fact | row still claims a summary |
| The "carried something" predicate removed | second-delete fact | reports work it did not do |
| The seal exemption removed | seal fact, count fact | the session's own farewell erased |
| The always-clear prompt-line update removed | seal fact, endpoint acceptance | prompt line kept on a sealed row |
| The tenant filter dropped from the rollup delete | the Npgsql SQL fact | "no WHERE clause in the generated statement" |
| The summary watermark guard removed | the held-summarisation fact | **the member's erased summary back in the column** |
| The watermark predicate taken out of the summary write | cross-process interleave fact | a pre-delete summary committed after the delete, from another process |
| The door check removed from `Append` | the retry fact | 12 erased records stored again |
| The seal exemption restored | the sealed-erasure fact | a seal survives the delete |
| The ingest watermark guard removed | retried-prompt fact, restart fact | an erased prompt copied back after the delete |
| The roll-up watermark guard removed | the roll-up race fact | the deleted paragraph recreated |

The store fact asserts the raw COLUMNS, not the folded record, because the fold hides the prompt line
behind a description that falls back to the repository name - a DTO-level assertion would pass over a
row still holding the prompt. It seeds all ten fields first: an erasure test over empty columns proves
only that null stayed null. The endpoint fact drives the whole thing over a real HTTP pipeline against
a real database, because the store test cannot prove the endpoint calls the erasure, in the right
scope, at all.

## Not proven, stated as a gap

**A record first seen after the erasure and dated after it is accepted**, because nothing distinguishes it
from a prompt sent a second ago. That is the limit of what a receiving service can know about material it
never saw the first time; it has its own fact so the boundary is visible, and the customer-facing wording
says the service refuses material it can *tell* is older rather than claiming resurrection is impossible.
Closing it means the Director honouring the delete - issue thefrederiksen/devthrottle_internal#1278.

**The delete does not reach Director-local files**, which is now stated in the code and both documents
rather than contradicted by them. There are THREE such stores, not two: the session-history snippet and
turn summaries, the expected-first-prompt in `sessions.json` and its backup, and - found in round two -
the Director's operational log, which keeps **the first 60 characters of every prompt** in dated files
with **no retention at all**. Issue thefrederiksen/devthrottle_internal#1278 tracks reaching them with the delete; issue thefrederiksen/devthrottle_internal#1277 tracks that
missing retention, which is a finding in its own right and older than this work.

The erase-before-delete ordering is reasoned, not tested: making the file delete fail on demand is
not worth the fragility on Windows.

**The erasure is EXECUTED on SQLite only.** The tenant predicate is now proved to reach the generated
PostgreSQL statements (above), which was the failure worth closing because it is cross-tenant. What
remains unproven is that a real PostgreSQL server accepts and performs these statements - that needs
the throwaway-container rig pointed at the main database rather than the statistics one, which the
mission's handoff ruled out of scope for this item. Named rather than left for someone to assume.

## Lands with the wording

The customer-facing wording in `devthrottle_internal` (`/privacy` and the data map) states the old
behaviour and goes false with this commit. It is corrected in that repository's own pull request, and
**neither lands without the other** - a code change that makes the published wording false is the same
defect as the one being fixed, pointing the other way.

## Gate

`scripts\test-local.ps1`, judged by the TRX outcome and the executed count against the recorded
baseline, never the console line.


thefrederiksen:
Pairs with thefrederiksen/devthrottle_internal#1217 - the wording half. Neither lands without the other: this change makes the published /privacy and data-map sentences false, and that is the same defect as the one being fixed, pointing the other way.

thefrederiksen:
## Independent review complete - CHANGES REQUIRED. Do not merge.

An independent review by a session that did not write this change returned **changes required**. Five defects: four concerning the completeness of the deletion under concurrency, and one that would cause records to be duplicated indefinitely on a mixed retry batch.

**The detail is recorded privately** and is deliberately not reproduced here: it describes exploitable timing in a data-deletion path, and this repository is public.

Also noted: the paired wording change must not merge either, since several of its published sentences are not made true by this code as it stands.

The local branch gate is green - 10,163 executed, 10,163 passed, 0 failed - so this is not a test failure. The defects are ones no test run would have caught.

Maintainers: see the review recorded on the paired internal pull request.

thefrederiksen:
## Independent cross-family review (Codex, gpt-5.6-sol) - VERDICT: BLOCK

Run as the merge gate reviewer for the v2.0.2 release. Reported verbatim; nothing was changed.

1. **`SessionHistorySummarizer.cs:143-149` and `SessionHistoryStore.cs:355-360` - the roll-up watermark is judged by READ time.** Another Gateway can read pre-erasure summaries after the watermark is stamped but before the rows are cleared; its later timestamp passes the guard, and the resulting paragraph is inserted and served after the deletion. This is the same defect the watermark was added to close, one level up.
2. **`SessionHistoryStore.cs:456-494` and `SessionHistorySummarizer.cs:97-100` - repeat counts are not honest after normal background processing.** The first deletion resets `SummaryKind`; the sweep restores it to "none"; the next deletion counts that non-null metadata as another erased row. Recreated roll-ups do the same. The separate count and update also drift under concurrent deletes.
3. **`GatewayPromptLog.cs:218-245, 322-335` - the check, the append and the deletion are synchronised only per instance.** Across overlapping Gateway processes an append can pass its watermark check before the stamp, pause while the deletion completes, then recreate a prompt file after the endpoint has already reported success.
4. **`GatewayPromptSink.cs:49-58` and `ConversationIngestor.cs:231-239` - a mixed batch duplicates without bound.** A batch holding refused pre-erasure records and accepted post-erasure records is written partially, but the sink treats the whole batch as failed. Every retry appends the accepted records again and marks none complete.
5. **`SessionHistoryStore.cs:288-303` and `GatewayPromptLog.cs:156-181` - the store comment is false.** It says prompt material time is clamped to receipt time. The caller passes raw `TsUtc`, so a Director with a fast clock can date old material after the watermark and restore the prompt line and the summaries derived from it.
6. **`SessionHistoryEntry.cs:32-35`, `SessionStateStore.cs:99`, `Session.cs:2436` - Director-local copies survive.** First-prompt snippets, turn summaries, persisted expected prompts and operational log excerpts are untouched. The pull request discloses this, but the operation is not a product-wide erasure.
7. **Rebase check - nothing merged since the branch tip conflicts in behaviour.** Session keys still cannot call the prompt delete route, the new session-key model is independent, the three migrations are consistent across SQLite and PostgreSQL and match their branch snapshots, and the current-main snapshot additions auto-merge.

Findings 1 through 5 are defects. Finding 6 is a scope limit that is already disclosed. Finding 7 is a clean result.

Held out of v2.0.2 on this verdict. This branch is six days old - rebase it when the fixes land.

PR #2389, opened 2026-08-02
Gateway deploy: stop a refused connection from taking the live site down, and gate what may ship

thefrederiksen:
Closes the failure behind the 38.5-second production outage on 2 August (#2383), and closes the routes by which something unchecked could reach production.

Full investigation: `docs/reviews/gateway-deploy-outage-2026-08-02.md` in `devthrottle_internal`.

## What actually happened

The swap was not at fault - it measured `0.0s`, as every healthy deploy does. App Service starts a brand-new container once a swap completes. That container's Gateway got **one refused PostgreSQL connection**, treated it as terminal, and never bound its port. It did not exit either, so the platform waited out its full 230-second container start limit, found nothing listening, and **stopped the site** - which also killed the healthy container that had been serving throughout. The same image on the same worker opened the same database in 1.6 seconds four minutes later.

## Changes

**The open is retried for a bounded window.** One refusal is no longer proof the database is gone. The refusal is contention and it is structural: a swap briefly runs four Gateway containers, each with two Npgsql pools, against a session-mode pooler in front of a server with `max_connections=60`.

This is not a fallback and does not weaken the fail-loud contract - it never reverts to SQLite, never starts without a database, and still throws at the end of the window. Every attempt builds a fresh provider and disposes the failed one, so the retry cannot leak pooled connections into the contention it is waiting out. Ninety seconds sits well inside the platform's limit, so a genuinely dead database still fails the container rather than hanging it.

**Both pools are bounded.** Npgsql defaults to 100 per pool, two pools per container, against a server allowing sixty in total. The statistics store is the second pool and roughly doubled the demand when it landed - which is why post-swap boots had been blowing their deadlines on *every* deploy for days before one finally crossed from slow into refused. An explicit size in the connection string still wins.

**A failure records what the server said.** `SqlState` and `MessageText` are PostgreSQL's own code and text and cannot contain the connection string, which is why only the exception type was ever logged. That one line is why the root cause of 2 August is permanently unknowable. The redaction contract is unchanged for every other exception type and is now pinned by a test.

## Gating what may ship

`workflow_dispatch` runs against whatever ref the caller picks, and nothing verified the shipping commit's tests - so an unreviewed branch or a red commit was as deployable as green `main`. The run now refuses, in seconds and before it touches Azure:

- any ref that is not `refs/heads/main`;
- any commit with a check that failed, or that has not finished.

Self-exclusion is by **run id, not job name** - a stale name here would mean a deploy that can never start. The gate logic was dry-run against real check data on `origin/main` (passes) plus four negative controls: exclusion works, a failing check refuses, an unfinished check refuses, and `skipped`/`neutral` are treated as passes.

**Budget 30s to 5s.** A healthy deploy measures `0.0s`, so this was never a cost curve with slack in it - thirty would have passed a twenty-nine-second outage as green.

**The site-state comment is corrected.** It claimed to be "the earlier and truer signal". It reads the *desired* state in Resource Manager, and on 2 August every sample read `Running` straight through the 21 seconds the platform had the site stopped. The external poll is what caught it.

**The deploy skill said a minute or two of 502s was normal** and told the reader not to report it. That trained people to wave through exactly this failure. There is no cold start on the user path - the point of the warmed swap is that the user is never moved onto a cold instance. The skill now also states that this workflow is the only supported way to update the live Gateway.

## Still open, deliberately not in here

The port bind sits **behind** the database open, so any slow or erroring boot dependency can still drive the platform to stop the site. Binding first and opening the database behind a readiness flag (with a separate `/readyz` for the swap warm-up gate) removes the whole class and is the single most valuable remaining change - but it restructures the Gateway's boot sequence and deserves its own review rather than riding along with a hotfix.

### DIFF
--- a/.claude/skills/deploy-hosted-gateway/SKILL.md
+++ b/.claude/skills/deploy-hosted-gateway/SKILL.md
@@ -16,12 +16,21 @@ version numbers, release notes, tags, or the mailing list.
 
 ## What you must know first
 
+- **This workflow is the ONLY way to update the live Gateway.** Do not pin an
+  image, restart the site, change app settings, or touch the staging slot by hand
+  with `az`, and do not do it in the Azure Portal. Those paths skip the warmed
+  hand-off, the outage measurement and the refusals below, and a hand-driven
+  change is how you get an unmeasured outage on live sessions. If this workflow
+  cannot do what is needed, that is a gap to fix in the workflow, not to work
+  around. The emergency swap-back is also a workflow
+  (`rollback-hosted-gateway.yml`) - use it rather than swapping by hand.
 - **A person authorizes the go-live.** Starting this deploy pushes new code to
   the live service. Get an explicit go from the human before you start it. Do not
   start it on your own initiative.
-- **It deploys whatever commit you start it against.** By default that is the
-  current tip of `main`. There is no separate "staging" - main goes straight to
-  the live service when someone runs this.
+- **It will only ship green main.** The run refuses, in seconds and before it
+  touches anything, if it was started against any ref other than `refs/heads/main`,
+  or if the commit being shipped has a check that failed or has not finished. If it
+  refuses for a pending check, wait for continuous integration and start it again.
 - **It does not set up any infrastructure.** The Azure resource group, container
   registry, database connection, and storage are already provisioned and persist
   across deploys. This deploy only: rebuild the image, point the live service at
@@ -68,23 +77,21 @@ curl -s -m 20 -o /dev/null -w "HTTP %{http_code}\n" https://devthrottle-gw.azure
 
 A `200` means the live Gateway is up.
 
-**Expected gotcha - a brief blip right after restart.** For a minute or two after
-the restart, this check can return `502` or no response (`000`) while the fresh
-container boots and runs its database migrations. This is normal cold-start
-behavior, NOT a failed deploy. Do not panic and do not report failure. Poll every
-15 seconds until it returns `200`:
+**A blip here is NOT normal, and must not be waved through.** This used to say a
+minute or two of `502`/`000` was expected cold-start behaviour and told you not to
+report it. That was wrong, and it trained people to accept the exact failure that
+took the live service down for 38.5 seconds on 2 August 2026 (issue #2383).
 
-```
-for i in $(seq 1 12); do
-  code=$(curl -s -m 20 -o /dev/null -w "%{http_code}" https://devthrottle-gw.azurewebsites.net/healthz)
-  echo "attempt $i: HTTP $code"
-  [ "$code" = "200" ] && { echo RECOVERED; break; }
-  sleep 15
-done
-```
+The deploy is a **warmed slot swap**: the old instance keeps serving until the new
+one is proven healthy, so a healthy deploy has **no user-visible gap at all**.
+Three consecutive deploys measured `Longest unavailable stretch: 0.0s` over a
+ten-minute watch. There is no cold start on the user path, because the user is
+never moved onto a cold instance.
 
-If it is still not `200` after a few minutes, then it is a real problem - see
-Troubleshooting.
+So if `/healthz` does not answer `200` immediately after the run goes green,
+something went wrong with the hand-off. Say so; do not wait it out. The run itself
+now fails if the external outage exceeds five seconds, so a green run already
+means production stayed up.
 
 ### 5. Report plainly
 
--- a/.github/workflows/deploy-hosted-gateway.yml
+++ b/.github/workflows/deploy-hosted-gateway.yml
@@ -41,6 +41,7 @@ on:
 permissions:
   id-token: write   # required for the Azure OIDC login
   contents: read
+  checks: read      # required to read the shipping commit's check runs (the green-only gate)
 
 env:
   AZURE_RG: rg-devthrottle-hosted-gateway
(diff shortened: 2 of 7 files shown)


thefrederiksen:
Review of head `1fc8be3ff`: **FAIL — merge blocking.**

### ONE — PASS

Configured PostgreSQL never falls back to SQLite and never starts without a database. When `CC_GATEWAY_DB_CONNECTION` is present, execution enters the PostgreSQL branch and its unbounded attempt loop; the only successful exit is after open/migration succeeds, and the deadline path throws. The SQLite path is reachable only when the environment variable is unset (`GatewayDatabase.cs:151-176`, `GatewayDatabase.cs:243-261`).

### TWO — PASS

The retry itself does not retain a failed provider. Every attempt starts with a fresh local provider (`GatewayDatabase.cs:181`), publishes it only after open/migration succeeds (`GatewayDatabase.cs:225-229`), and disposes it before either sleeping or taking the final throw (`GatewayDatabase.cs:231-258`). That includes the terminal throw path because disposal happens at `GatewayDatabase.cs:233`, before the deadline branch.

### THREE — FAIL

The new credential-bearing pool helper runs before the redacting retry `try/catch` (`GatewayDatabase.cs:166`; the catch does not begin until `GatewayDatabase.cs:182`). A malformed connection string can therefore escape directly with parser text. Reproduction with an unquoted credential fragment such as `Password=alpha;ultra-secret=beta` produced an exception message beginning `Couldn't set ultra-secret...`.

That message is then written verbatim by startup at `GatewayService.cs:133-136`. This is a real credential-fragment log path, not merely missing coverage. The added redaction test only passes a fabricated exception directly to `DescribeFailure` (`GatewayDatabaseOpenResilienceTests.cs:83-95`); it never exercises `WithBoundedPool`, the constructor boundary, or the production log call.

### FOUR — FAIL

The 90-second value is not a hard bound. The deadline is created at `GatewayDatabase.cs:174`, but it is checked only after a database operation returns by throwing (`GatewayDatabase.cs:242`). `GetPendingMigrations` and `Migrate` remain blocking calls with no deadline cancellation at `GatewayDatabase.cs:212` and `GatewayDatabase.cs:220`, so one attempt can overrun the advertised window.

More importantly, even when the new terminal throw is reached, the container still does not fail. `GatewayService.cs:133-137` catches every startup exception, logs it, diagnoses it, and does not rethrow. `GatewayWorker.cs:45-51` then continues to `Task.Delay(Timeout.Infinite)`. The process therefore remains alive with no port bound until the platform's 230-second start limit—the same failure shape this pull request is intended to remove.

### Required mutation proof

Clean-head baseline: build passed; all 6 added test cases passed.

| Arm | Verified one-line production mutation | Focused result | Judgment |
|---|---|---:|---|
| Main pool wiring | Reverted `UseNpgsql(boundedConn, ...)` to the original connection at `GatewayDatabase.cs:185` | 6/6 green | The pool helper tests do not pin the main production call site. |
| Statistics pool wiring | Reverted `UseNpgsql(bounded, ...)` to `choice.ConnectionString` at `GatewayStatsStore.cs:467` | 6/6 green | The pool helper tests do not pin the second production call site. |
| Failure logging wiring | Reverted `DescribeFailure(ex)` to `ex.GetType().Name` at `GatewayDatabase.cs:241` | 6/6 green | Both failure-description tests bypass the real logging call site. |
| Control | Reinstated the old always-true Npgsql `ContainsKey` guard | 1 red, 5 green | The cap assertion failed at `GatewayDatabaseOpenResilienceTests.cs:34`: expected 10, actual 100. This proves the mutation harness detects a real broken guard. |

The three production-call-site greens are therefore findings, not a bad mutation setup. Each mutation was diff-verified before execution, restored with checkout, and followed by a clean rebuild. The final worktree is clean.

thefrederiksen:
## Rebuilt at `86c7b2b99` after the review. Both findings accepted; the previous claim is withdrawn.

**The fail-fast claim in the original body was false and I am withdrawing it.** `GatewayService.cs:134`
catches every startup exception, logs it and does not rethrow; `GatewayWorker` then awaited
`Task.Delay(Timeout.Infinite)`. The throw at the end of my retry window reached nobody - the container
stayed alive with no listener, and the retry only changed how long it took to arrive at the same silent
hang that stopped the site on 2 August.

### Finding 1 - the fix is process termination, not retry

`GatewayWorker` now consults `MustTerminate` after `StartAsync` and calls `Environment.Exit` with a
non-zero code. `Environment.Exit` rather than a thrown exception deliberately: a faulted background task
inside a live process looks identical from outside to the hang it replaces, which is the whole failure.
Ending the process is already this host's established way of stopping.

**Long database outage, as asked:** the container exits, the platform restarts it, it retries for the
bounded window and exits again. That is a restart loop and it is the correct behaviour - the service
recovers by itself the moment the database answers rather than sitting dead. It is bounded, not tight:
each container spends the full 90-second retry window before giving up, so restarts run at roughly one
per window-plus-boot. Worth noting the swap's own warm-up gate normally catches a dead database before
production is touched at all, since staging is warmed and must serve before the swap is allowed.

### Finding 2 - the redaction boundary now covers the real leak path

`WithBoundedPool` ran outside the redacting try/catch, and the caller logs `ex.Message` verbatim.
Measured on the real parser rather than assumed: `"SUPERSECRET=x"` throws `Couldn't set supersecret
(Parameter 'supersecret')`. The **value** position does not echo; the **keyword** position does. A
case-sensitive check for the leak passes while the secret sits in the log - that nearly hid it here. The
parse now sits inside a boundary that throws a redacted message and deliberately does not keep the
original as an inner exception.

### The tests were decoration and are replaced

Confirmed - reverting the two `UseNpgsql` lines and the `DescribeFailure` call left all six green.
Replaced with IL guards plus a redaction test that drives the **real constructor** with an unparseable
connection string rather than calling the helper.

Every guard was mutation-verified one production LINE at a time, with `git diff` confirming the mutation
was actually in the file before each build:

| Mutation | Guard | Result |
| --- | --- | --- |
| remove `Environment.Exit(StartFailureExitCode)` | ExecuteAsync calls Environment.Exit | RED |
| `if (MustTerminate(...))` to `if (false)` | ExecuteAsync calls MustTerminate | RED (build) |
| pass raw connection string, main store | GatewayDatabase calls WithBoundedPool | RED |
| `DescribeFailure(ex)` to `ex.GetType().Name` | GatewayDatabase calls DescribeFailure | RED |
| pass raw connection string, statistics store | GatewayStatsStore calls WithBoundedPool | RED |

**One of these was decoration on the first attempt and mutation testing caught it**: the exit guard
scanned the whole type, and the constructor's `Environment.Exit(0)` kept it green when the failure-path
exit was deleted. It is now scoped to the `ExecuteAsync` state machine, which the constructor's lambda
is not part of. That is `86c7b2b99`.

### What these guards do NOT catch, stated rather than implied

They prove the bounding helper and the redacting describer are **called** on those paths. They cannot
prove which value reaches `UseNpgsql`, so a mutation that keeps the call but passes the raw string stays
green. That is a dataflow property and needs an integration test against a real server, where
`PostgresProviderProofTests` lives. Claiming otherwise is what made the first version decoration.

thefrederiksen:
## The end-to-end test's RED state has now been observed

The point raised was that nobody had ever seen it fail, and a test whose failing state has never been
observed is not yet evidence - it may pass because the path works, or pass because it asserts something
the product cannot violate, and from outside those greens look identical.

I could not run the xunit test locally (the Gateway suite serialises on the machine-wide lock, held all
evening), so I drove the **same production path** from a standalone driver holding a direct reference to
the built `CcDirector.Gateway.dll`: set `CC_GATEWAY_DB_CONNECTION` to an unparseable value, call
`GatewayService.StartAsync`, read `service.State`.

**Unmutated:** `StartAsync` returns after 3.6s with `State = Failed ("Gateway FAILED - see logs")`.

**Mutated, one production line at a time, each confirmed present by `git diff` before building:**

| Mutation | Result |
| --- | --- |
| `GatewayService.cs` - remove `SetState(GatewayServiceState.Failed, status)` from `DiagnoseStartFailureAsync` | **RED** - `State = Starting` |
| `GatewayService.cs` - remove the `await DiagnoseStartFailureAsync()` call from the catch | **RED** - `State = Starting` |

In both red cases the driver reports exactly the regression the test exists to catch:

```
service.State = Starting   ("Gateway starting...")
FAIL  the failure did NOT reach Failed; MustTerminate would return false and the process would hang
```

That is the 2 August behaviour reproduced on demand: the failure does not reach the state the
termination decision reads, so the process stays alive with no listener.

**What this establishes and what it does not.** It establishes that the property is *violable* - the
product can break it, in at least two ways, so the assertion is not vacuous. It does **not** substitute
for seeing the xunit test itself go red; that test's own failing state has still not been observed, and
the reviewer breaking the path against the real test remains the thing that settles it. I am reporting
the driver result because it answers the "can this even fail?" half now rather than after the queue.

The driver is throwaway scaffolding and is not part of this pull request.

thefrederiksen:
## Round-two review at `5674d66db`

**Verdict: one remaining documentation finding. The process-termination implementation, redaction boundary, and new end-to-end test pass.**

### Finding: the restart-rate description is still too narrow

`GatewayWorker.cs:88-90` says the slow restart case is a database that is reachable but refusing connections, and `GatewayWorker.cs:97-98` reduces the diagnosis to “a fast loop means a misconfiguration, a slow one means the database.” That is not the boundary implemented by the code.

`GatewayDatabase.cs:203-280` retries every exception after the connection string parses and the main PostgreSQL open/migrate path begins. That includes an unavailable endpoint, authentication rejection, migration or schema failure, and provider failures—not only a reachable database refusing a connection. Conversely, blank configuration fails at `GatewayDatabase.cs:152-155`, parser rejection fails at `GatewayDatabase.cs:180-193`, and non-database startup failures occur outside that loop, so those paths are fast.

The runtime controls made the distinction concrete:

- A valid connection string targeting an unavailable endpoint (`127.0.0.1:1`) entered the retry loop, made 12 attempts, and the real hosted process exited with code 1 after 96.214 seconds.
- A malformed connection string exited the real process with code 1 after 1.358 seconds.
- A hosted-contract failure outside database startup exited with code 2 after 0.636 seconds.

The termination behavior is correct. The remaining defect is the sentence describing which failures get which rate. The accurate split is “post-parse main database open/migrate failures” versus “pre-loop configuration failures and other startup failures.”

### Termination: verified through the real process

The source path is now complete: `GatewayWorker.ExecuteAsync` awaits `GatewayService.StartAsync`, reads the resulting state through `MustTerminate`, and calls `Environment.Exit(StartFailureExitCode)` at `GatewayWorker.cs:71-105`.

I also ran the unmutated executable rather than treating a faulted test task as process evidence:

- The malformed-string path traversed the real `StartAsync` catch, `DiagnoseStartFailureAsync`, and the worker termination branch. The operating-system process exit code was 1, after 1.358 seconds.
- The hosted valid-but-unavailable database path retried for the configured window, then traversed the same state/worker path and exited with code 1 after 96.214 seconds.

The logs contain the terminal state and worker termination message; the process did not remain alive on an infinite delay.

### The new end-to-end test produced a real red

The new test at `GatewayStartFailureTerminationTests.cs:191-210` passed on the clean head. It drives the production sequence:

`GatewayService.StartAsync` → `GatewayHost` construction → `GatewayDatabase` parser failure → the swallowing catch at `GatewayService.cs:133-137` → `DiagnoseStartFailureAsync` → the state consumed by `GatewayWorker.MustTerminate`.

Because this test had never previously been observed failing, I broke the exact production transition it is meant to guard:

```diff
- SetState(GatewayServiceState.Failed, status);
+ SetState(GatewayServiceState.Starting, status);
```

That was the single line at `GatewayService.cs:307`. I confirmed the mutation with `git diff`, rebuilt, and ran only the new test. It went **RED** at `GatewayStartFailureTerminationTests.cs:208`: expected `Failed`, actual `Starting`. I then restored the production file with checkout, rebuilt, and reran the focused set cleanly.

That red establishes that the test guards the swallowed-exception-to-terminal-state path rather than asserting a property the product cannot violate. The separate executable runs establish the remaining link from that state to an actual non-zero process exit.

### Redaction: verified at the real logging boundary

The malformed connection string used a poisoned keyword—the position the provider parser echoes. The exception crossing `GatewayService.StartAsync` was the redacted wrapper from `GatewayDatabase.cs:180-193`, and the real process log contained neither the poisoned fragment nor the original parser exception text. The same credential scan was clean after the 96-second hosted run.

This covers the actual boundary that matters: `GatewayService.cs:135` writes `ex.Message` to disk verbatim, so the exception reaching that catch must already be safe.

### IL limitations and mutation controls

The revised limitation at `GatewayStartFailureTerminationTests.cs:95-113` is accurate: a call's presence in IL is not reachability, dataflow, or ordering. Scoping `CallsInMethod` at `:115-129` closes the current unrelated-method hole; the current production types contain one `GatewayDatabase` constructor and one `BuildProvider` method.

I confirmed both the strength and the limits with one-line, diff-verified production mutations:

| Mutation | Result |
| --- | --- |
| `GatewayService.cs:307`, terminal state `Failed` → `Starting` | **RED** in the new end-to-end test |
| `GatewayWorker.cs:74`, add an impossible condition while retaining both call instructions | **GREEN**, 10/10 focused tests; confirms presence is not reachability |
| `GatewayDatabase.cs:212`, pass the raw string to `UseNpgsql` | **GREEN**, 10/10; confirms the documented dataflow gap |
| `GatewayStatsStore.cs:467`, pass the raw string to `UseNpgsql` | **GREEN**, 10/10; confirms the documented dataflow gap |
| `GatewayDatabase.cs:268`, replace `DescribeFailure(ex)` with the exception type name | **RED** at the constructor-scoped IL guard (`GatewayStartFailureTerminationTests.cs:139`) |

Every mutation was applied one at a time, shown by diff before build/test, restored with checkout, and followed by a rebuild. Final verification at the exact head: build succeeded, 10/10 focused termination/redaction tests passed, and the worktree was clean.

PR #2399, opened 2026-08-02
feat(missions): a mission can adopt a session that already exists

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1768.

## What was missing

A Mission could only be joined in the instant a session was spawned. `session spawn --mission <id>`
attached at birth, and nothing could attach afterwards - the Gateway had `POST /missions`,
`GET /missions` and `GET /missions/{mid}` and no way to change a session's mission, and the command
line had `mission create` and `mission list` and no `mission attach`.

That made the feature miss the case it exists for. A mission's shape is DISCOVERED as it runs -
that is what makes it a mission rather than a task - so attach-at-birth grouped only the work
somebody had already planned, which is the work that least needs grouping. The release push that
found this began as one coordinating seat and grew over a day into about a dozen: an architect and a
manager, three review seats, three more for a different question, an investigation seat, four
independent gate reviewers commissioned one pull request at a time, and a session fixing a defect one
of them found. One body of work. None of it foreseeable at spawn. None of it representable.

## What this adds

- **`POST /sessions/{sid}/mission` on the Gateway** - attach a session that already exists to a
  Mission, or detach it on a null mission id. Resolves the mission inside the caller's own tenant,
  sends the resolved NAME down with the id, and routes the attachment to the owning Director over the
  tunnel.
- **`POST /fleet/mission` on a Director** - the loopback route the command line uses. A local target
  is attached directly (resolving the mission name against the Gateway first, as the local leg of
  `/fleet/spawn` already does); a target on another machine is relayed through the Gateway.
- **`cc-devthrottle mission attach <session> <mission>` and `mission detach <session>`**, with
  `--with-children`. Sessions are addressed the same way every other session verb addresses them
  (number, id prefix, name); a mission can be named by id, id prefix, or name.
- **Inheritance at spawn**: a session spawned with a controlling session inherits that session's
  mission by default.

The Director already had an `attach-mission` verb; it was reachable from nothing. This wires it up and
teaches it the Gateway path (stamp the id and name the Gateway resolved, instead of consulting the
Director's own single-tenant store, which is a different set - the failure #1548 fixed on the spawn
path).

## The rules, settled and written down

These are in `docs/new_architecture/mission-as-first-class-unit-of-work.md`, which is the document that
owns the attachment rules. Each was decided rather than left implied, because somebody hits every one.

1. **Attaching is a MOVE, not a one-way door.** A session that already carries a mission is re-pointed.
   The first classification of a session is always a guess about work still taking shape; one-way would
   make every wrong guess permanent until the session was killed.
2. **The move is REPORTED** - the command names the mission the session LEFT. A move that reports only
   its destination hides that anything was displaced.
3. **Detaching is supported.** No mission is the ordinary state of a session. A session that had no
   mission is told exactly that rather than told it was detached from nothing.
4. **A refused attach changes nothing** - a mistyped id leaves an existing attachment intact.
5. **Attaching a controlling session does NOT bring its children by default; `--with-children` does.**
   A controller routinely commissions work that is not part of its own mission, and a silent bulk
   re-parent cannot be undone in one step. With the flag the walk is TRANSITIVE: stopping at the first
   level would attach the Manager and leave the Workers behind, which reads as success while producing
   exactly the split view the feature exists to end. Every session it moves is NAMED, not counted.
6. **A spawned session inherits its controlling session's mission by default.** The fleet already
   records the relationship, so this costs nothing and would have grouped the release push above for
   free. `--mission <id>` wins (stated intent beats a default), `--mission none` opts out (spelled the
   way `--controlled-by none` is), and it is never silent: the spawn names the mission, names the
   session it came from, and says how to undo it.

## The workflow seat moves with the mission

Found by independent review of the first cut, and it was the finding that mattered. A Mission is not only
a record - it is also a RUN of the built-in `mission` workflow, and a mission-scoped spawn seats the
session on that run and records it in the run's participant ledger. The seat pins the CONDUCT the agent
follows. Moving only the mission link would have left a session **displayed under one mission while
governed by the one it left**, still counted in that mission's participant ledger - in exactly the case
this feature was built for.

The rule, and its one exception:

- A seat that **is a run of the mission being left** belongs to that mission and follows it: moved on a
  move, **cleared** on a detach.
- A seat the caller **chose independently** (`--workflow-run`) is **preserved**. It was never the
  mission's to take.
- A session with **no seat** has nothing to preserve and simply gains the destination's seat.
- A destination mission whose workflow the owner **switched off** seats nobody. A move must not re-enable
  something the owner deliberately turned off.

**Detach clears the seat rather than refusing.** Refusing while a seat exists was considered and
rejected: every mission-spawned session is seated, so refusal would make detach impossible for precisely
the sessions that need it. Refusal is only right where no coherent action exists, and here one does -
leave the run. Leaving is **recorded** (`LeftUtc`), not erased.

**Decided in one place.** Whether a run belongs to a mission is a fact about the Gateway's run store, so
the Gateway decides and hands the Director a finished answer, applied in the SAME verb as the mission -
two writes that must agree would disagree on a slow network. It is also why `POST /fleet/mission` relays
through the Gateway even for a session the Director hosts itself: a second implementation of this rule
would drift, and the drift would be invisible until somebody found a session governed by a mission it had
left.

**The limit, stated at the point of use.** Moving the seat corrects the RECORD - what the fleet shows,
what governs the session, who the run lists. It cannot reach into a running agent's context and replace
conduct injected at birth. The command line says so every time a seat moves, and names the command to
re-read it.

## The tenant boundary

Missions are tenant-scoped and that was hard won (devthrottle_internal thefrederiksen/devthrottle#1039 fixed a live leak where
`GET /missions` served every account's list to every account; a mission name is free text a person
typed). Attach is the first WRITE that takes a mission id from a caller and applies it, so it is exactly
the shape that would put the leak back.

The route follows `GET /missions/{mid}` line for line: resolve the caller's tenant server-side from the
authenticated device key, refuse with 403 when no tenant is bound (never fall back to a shared
partition), resolve the mission inside that tenant. An id belonging to another account is answered
identically to an id that does not exist. The mission is resolved BEFORE the session is located, so the
refusal is a property of the tenant gate alone and cannot be confused with a Director being offline.

**A refusal must not say WHICH kind of refusal it is.** This is a property of the route, not merely a
detail of a test. If "no mission has that id" and "that mission is not yours" were distinguishable - by
status code, or by wording - then attaching a session to guessed identifiers would enumerate which
missions exist in other people's accounts, one request at a time, without ever reading one. Both answer
400 with the same sentence and only the echoed id differs, and
`An_unknown_mission_id_and_another_tenants_mission_id_are_answered_alike` compares the two answers to
each other rather than to a fixed string, so a later edit cannot reword one without the other.

It is a literal route rather than an entry in the `/sessions/{sid}/{**rest}` catch-all deliberately: the
catch-all passes the raw request body straight through to the owning Director, which for this verb would
mean shipping a caller-supplied mission id with no tenant resolution at all.

`MissionAttachRouteTenantScopingTests` holds it to that, in the shape
`MissionStoreTenantPartitionTests` and `MissionRouteTenantScopingTests` established: two separately
enrolled accounts over the real routes, the real auth middleware and the real tunnel; every refusal
paired with a permitted request in the same Gateway; every refusal asserted at two levels (the status
code AND the fact that no `attach-mission` command reached the Director); and a self-test that points
the same recorder at an attach that genuinely happens and requires a positive, so a recorder that
records nothing cannot certify the file.

Revert-prove: drop the `ResolveReadTenant` gate, or resolve the mission with `TenantId.Local`, and
`Another_tenants_mission_cannot_be_attached_to_your_own_session` goes red - the attach to another
account's mission stops being a 400 and reaches the Director carrying that account's mission name.

## Which tests carry positive controls

Listed explicitly so a reviewer can check they are able to FAIL rather than infer it. Every case below
pairs the claim with something that must come out the other way in the same fixture.

**`MissionAttachRouteTenantScopingTests`** (the tenant partition)

| Case | The control |
|---|---|
| `The_attach_detector_itself_detects_an_attach_when_there_is_one` | **The self-test.** Every refusal below rests on "no `attach-mission` reached that Director". This points the same recorder at an attach that genuinely happens and requires a positive - a recorder that records nothing cannot certify the file. |
| `Another_tenants_mission_cannot_be_attached_to_your_own_session` | The same session attaching to a mission of its OWN succeeds (200, command recorded) in the same Gateway. |
| `Another_tenants_session_cannot_be_attached_to_your_own_mission` | B's own key reaches that same session (200, command recorded on B's Director). |
| `A_device_key_with_no_bound_tenant_attaches_nothing` | The identical attach AND detach with a BOUND key both land, two commands recorded. |
| `An_unknown_mission_id_and_another_tenants_mission_id_are_answered_alike` | Comparative: the two refusals are asserted equal to each other in status and wording, so neither can drift into an existence oracle. |
| `Detaching_carries_no_mission_and_needs_none` | Asserts the command ARRIVES with a null mission, so detach cannot be a no-op that reports success. |

**`MissionAttachSeatMoveTests`** (the seat, over a real run store)

| Case | The control |
|---|---|
| `Moving_a_session_updates_the_participant_ledger_on_both_runs` | Asserts the session **is** an active participant of run A *before* the move - "the session was never on run A, so nothing below means anything". |
| `Moving_a_mission_scoped_session_moves_its_workflow_seat` | The fixture fails loud if a created mission has no run, so the premise cannot silently vanish and leave the case asserting about nothing. |
| `A_seat_the_caller_chose_independently_is_preserved` | Asserted in BOTH directions: still active on the run it chose, and NOT enrolled in the destination mission's run. |
| `Detaching_clears_the_missions_seat_with_it` | The session is joined to run A first, so "no longer active" is a change and not a starting state. |
| `An_unseated_session_simply_gains_the_destination_missions_seat` | A positive claim in its own right. |

**`MissionCommandTests`** (the Director core)

`AttachMission_OnAnAlreadyAttachedSession_MovesItToTheNewMission`,
`AttachMission_WhenTheGatewaySaysMoveTheSeat_MovesMissionAndSeatTogether` and
`AttachMission_Detach_ClearsTheMissionsSeatWithIt` each assert the starting state first, so the change is
a change. `AttachMission_UnknownMission_LeavesAnExistingAttachmentIntact` attaches to a real mission
first. `AttachMission_WithMissionNamePresent_StampsDirectly_WithoutStoreLookup` proves "no local lookup"
by succeeding against an EMPTY store - a lookup would have rejected it.
`AttachMission_WithoutASeatDecision_TouchesTheSeatAtAll` is the compatibility control: a payload silent
about the seat must not clear one.

**Command line** - `test_no_warning_when_the_seat_did_not_move` is the control for the stale-conduct
warning (it must be tied to the seat actually moving, or it becomes noise ignored on the one occasion it
matters). `test_without_the_flag_only_the_named_session_moves` is the control for `--with-children`;
`test_a_spawn_with_no_controller_inherits_nothing` and
`test_standalone_inherits_nothing_because_it_has_no_controller` are the controls for inheritance.

## Deliberately not in this change

- **No user interface.** The issue asks for an endpoint and a command; the Cockpit and phone still show
  a mission but cannot change one. That is worth doing and is a separate piece of work.
- **`mission list` is unchanged.** A count of attached sessions would make an attach checkable without
  reading JSON, but `mission list` talks only to the Gateway today and adding it would give a
  Gateway-only command a dependency on a running Director.

### DIFF
--- a/docs/new_architecture/mission-as-first-class-unit-of-work.md
+++ b/docs/new_architecture/mission-as-first-class-unit-of-work.md
@@ -50,6 +50,124 @@ The Mission object supplies the missing horizontal binding:
 Without a Mission object, "Architect beside Manager" has nothing to hang on and collapses back into
 a chain.
 
+## Attaching a session that ALREADY EXISTS (issue #2387)
+
+A Mission could originally only be joined in the instant a session was spawned
+(`session spawn --mission <id>`). That made the feature miss the case it exists for. A mission's
+shape is DISCOVERED as it runs - that is what makes it a mission rather than a task - so
+attach-at-birth grouped only the work somebody had already planned, which is the work that least
+needs grouping. The case that found it: a release push that began as one coordinating seat and grew
+over a day into about a dozen - an architect and a manager, three review seats, three more for a
+different question, an investigation seat, four independent gate reviewers commissioned one pull
+request at a time, and a session fixing a defect one of them found. All one body of work. None of it
+foreseeable at spawn. None of it representable.
+
+The verb: `cc-devthrottle mission attach <session> <mission>` and `mission detach <session>`, over
+`POST /sessions/{sid}/mission` at the Gateway and `POST /fleet/mission` on a Director.
+
+### The rules, settled
+
+Each of these was decided rather than left implied, because somebody hits every one of them and an
+implied answer gets re-litigated at the worst moment.
+
+1. **Attaching is a MOVE, not a one-way door.** A session that already carries a mission is
+   re-pointed by the same command. Since the first classification of a session is always a guess
+   about work still taking shape, a one-way attach would make every wrong guess permanent until the
+   session was killed.
+2. **The move is REPORTED.** The command names the mission the session LEFT. A move that reports
+   only its destination hides that anything was displaced, which is how a session goes missing from
+   the pod somebody is looking at.
+3. **Detaching is supported.** No mission is the ORDINARY state of a session, so returning to it
+   must not require inventing a mission to park the session in. A session that had no mission is
+   told exactly that, rather than being told it was detached from nothing.
+4. **A refused attach changes nothing.** A mistyped mission id leaves an existing attachment intact
+   rather than clearing it - otherwise the failure looks like nothing happened until somebody goes
+   looking for the pod.
+5. **Attaching a controlling session does NOT bring its children by default; `--with-children`
+   does.** A controller routinely commissions work that is not part of its own mission - a reviewer
+   for one pull request, an investigation seat for something else - and a silent bulk re-parent
+   cannot be undone in one step. With the flag, the walk is TRANSITIVE (children, their children,
+   and so on): stopping at the first level would attach the Manager and leave the Workers behind,
+   which reads as success while producing exactly the split view this feature exists to end. Every
+   session it moves is NAMED, not counted.
+6. **A spawned session INHERITS its controlling session's mission by default.** The fleet already
+   records the controlling relationship, so this costs nothing and would have grouped the release
+   push above for free - every one of those sessions was spawned by a seat already on the mission.
+   An explicit `--mission` wins (stated intent beats a default), `--mission none` is the opt-out
+   (spelled the way `--controlled-by none` is), and the inheritance is never silent: the spawn names
+   the mission, names the session it came from, and says how to undo it.
+
+### The workflow seat moves with the mission
+
+**A Mission is not only a record. It is also a RUN of the built-in `mission` workflow**, and a
+mission-scoped spawn seats the session on that run and records it in the run's participant ledger. The
+seat - `WorkflowRunId` plus the workflow id and its PINNED version - is what decides the conduct the
+agent was told to follow. So the mission link and the seat are two halves of one fact, and an attach
+that moved only the link would leave a session **displayed under one mission while governed by the one
+it left**, taking its conduct from a mission it is no longer in and still counted in that mission's
+participant ledger. That is worse than an inconsistent label, and it would happen in exactly the case
+this feature was built for. (Found by independent review of the first cut, which moved only the link.)
+
+**The rule, and its one exception:**
+
+- A seat that **is a run of the mission being left** belongs to that mission and **follows it**. On a
+  move it becomes the destination mission's run; on a detach it is **cleared**.
+- A seat the caller **chose independently** - spawned with an explicit `--workflow-run` that is not this
+  mission's run - is **preserved untouched**. It was never the mission's to take.
+- A session with **no seat** has nothing to preserve, so it simply gains the destination mission's seat.
+- A destination mission with **no run of its own** (an UNGOVERNED mission, created while the owner had the
+  mission workflow switched off) seats nobody, so the session moves unseated rather than keeping the old
+  seat. A move must not smuggle a seat past a switch the owner deliberately turned off.
+
+**Detach clears the seat rather than refusing.** Refusing to detach while a seat exists was considered
+and rejected: every mission-spawned session is seated, so refusal would make detach impossible for
+precisely the sessions that most need it. A session that has left a mission cannot still be governed by
+that mission's run, and the coherent action - leave the run - exists. Refusal is the right answer only
+where no coherent action does.
+
+**Leaving is recorded, not erased.** The ledger marks the participant as having left (`LeftUtc`). That
+the session *was* in that run is true and stays true.
+
+**The decision is the Gateway's and is made in one place.** Whether a run belongs to a mission is a fact
+about the run store, which only the Gateway holds; a Director asked to decide it would have to guess. The
+Gateway sends the Director a finished answer - move the seat or do not, and which run to sit on - and the
+Director applies it in the *same verb* as the mission, so the two can never land apart. This is also why
+the Director's own `POST /fleet/mission` relays through the Gateway even for a session it hosts itself: a
+second implementation of this rule would drift, and the drift would be invisible until somebody found a
+session governed by a mission it had left.
+
+**The limit, stated because it is real.** Moving the seat corrects the RECORD - what the fleet shows, what
+governs the session, who the run lists. It cannot reach into a running agent's context and replace the
+conduct text that was injected at birth. Only telling the session to fetch its conduct again does that,
+and the command line says so every time a seat moves.
+
+### The tenant boundary, which is the constraint that governs the design
+
+Missions are TENANT-SCOPED and that was hard won (devthrottle_internal issue #1039 fixed a live leak
+where `GET /missions` served every account's list to every account; a mission NAME is free text a
+person typed - customer names, project names, people's names). Attach is the first WRITE that takes
+a mission id from a caller and applies it, so it is exactly the shape that would put the leak back.
+
+`POST /sessions/{sid}/mission` therefore follows `GET /missions/{mid}` line for line: resolve the
+CALLER's own tenant server-side from the authenticated device key, REFUSE when no tenant is bound
+(never fall back to a shared partition), and resolve the mission INSIDE that tenant. The mission is
+resolved BEFORE the session is located, so the refusal is a property of the tenant gate alone and cannot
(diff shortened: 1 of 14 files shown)


thefrederiksen:
## Evidence

### The local gate - all seven projects

Judged by the TRX `ResultSummary/@outcome` and count, not the console line.

| Project | Outcome | Total |
|---|---|---|
| `CcDirector.Gateway.Tests` | Completed | 5190 (5135 passed, 55 skipped, 0 failed) |
| `CcDirector.Core.Tests` | Completed | 4196 |
| `CcDirector.Avalonia.Tests` | Completed | 353 |
| `CcDirector.Engine.Tests` | Completed | 63 |
| `CcDirector.HostedAgent.Tests` | Completed | 88 |
| `CcDirector.Launcher.Tests` | Completed | 110 |
| `CcDirector.Terminal.Avalonia.Tests` | Completed | 24 |

Command line: 213 passed (`tools/cc-devthrottle`), and the shipped-tool contract guard: 36 passed.

### What I could NOT establish, said plainly

**I do not have a measured baseline total for the Gateway suite at this branch's base commit
(`aa8a8401e`).** Gateway totals recorded on this machine today range from 5153 to 5229, but every one of
them is from a different worktree on a different branch, so none is a baseline for this base - quoting one
would be a stale measurement presented as a control. `dotnet test --list-tests` would settle it, but
discovery loads the assembly and therefore takes the same machine-wide suite lock, which was held by
another run.

**What I did instead, which answers the question a baseline delta only approximates.** The TRX names every
result, so rather than inferring from a total I checked all nineteen new cases **by name** in the run's own
TRX. All nineteen are present and all nineteen are `Passed`:

```
The_attach_detector_itself_detects_an_attach_when_there_is_one
Another_tenants_mission_cannot_be_attached_to_your_own_session
Another_tenants_session_cannot_be_attached_to_your_own_mission
A_device_key_with_no_bound_tenant_attaches_nothing
Detaching_carries_no_mission_and_needs_none
An_unknown_mission_id_and_another_tenants_mission_id_are_answered_alike
Moving_a_mission_scoped_session_moves_its_workflow_seat
Moving_a_session_updates_the_participant_ledger_on_both_runs
A_seat_the_caller_chose_independently_is_preserved
Detaching_clears_the_missions_seat_with_it
An_unseated_session_simply_gains_the_destination_missions_seat
AttachMission_WithMissionNamePresent_StampsDirectly_WithoutStoreLookup
AttachMission_OnAnAlreadyAttachedSession_MovesItToTheNewMission
AttachMission_UnknownMission_LeavesAnExistingAttachmentIntact
AttachMission_WhenTheGatewaySaysMoveTheSeat_MovesMissionAndSeatTogether
AttachMission_WhenTheGatewaySaysLeaveTheSeat_MovesTheMissionOnly
AttachMission_Detach_ClearsTheMissionsSeatWithIt
AttachMission_Detach_LeavesAnIndependentlyChosenSeatAlone
AttachMission_WithoutASeatDecision_TouchesTheSeatAtAll
```

`outcome=Completed` is discriminating here rather than decorative: four Gateway runs in today's own data
on this machine recorded `Failed` with truncated totals of 0, 2482, 4355 and 4634, which is exactly the
crashed-host shape the outcome field exists to expose.

### Mutation proof - the seat fix can fail, with the reported symptom

Injected into the **committed** tree, so restoring is `git checkout` of one file and cannot take the fix
with it. One line in the attach route forces `seatIsTheMissions = false`, reproducing the pre-review
behaviour exactly: the mission moves, the seat does not.

Result: **4 failed, 1 passed** in `MissionAttachSeatMoveTests`.

Failing, which is the defect verbatim:

- `Moving_a_mission_scoped_session_moves_its_workflow_seat`
- `Moving_a_session_updates_the_participant_ledger_on_both_runs`
- `Detaching_clears_the_missions_seat_with_it`
- `An_unseated_session_simply_gains_the_destination_missions_seat`

**Still passing, and that is the point:** `A_seat_the_caller_chose_independently_is_preserved`. The
mutation forces the preserve behaviour, which is what that case expects, so it stays green. The mutation
is therefore targeted at the rule under test rather than breaking the suite wholesale - a mutation that
reddened everything would prove only that the code runs.

PR #2402, opened 2026-08-03
fix(director): stop the FIFO window and the main window sharing one session's terminal

thefrederiksen:
Found while reviewing the panes / tear-off design (devthrottle_internal#1225), which had cited `FifoWindow` as proof that a second window can host a terminal. It does - but it double-attaches.

## The bug

`BtnFifo_Click` opened FIFO **without detaching the main window's `TerminalControl`**. FIFO then attaches sessions to its own control, and its queue is needs-you sessions - routinely including the one already on screen. Two controls then polled one session and both drove `Session.Resize` against the real ConPTY, so the agent's output reflowed to whichever resized last.

Survivable only because FIFO is modal and full-screen, hiding the damage, plus a re-attach afterwards that repaired the geometry FIFO had changed. The comment on that re-attach described the symptom without naming the cause.

## The fix

Detach before showing FIFO; re-attach in a `finally` so a throw inside FIFO cannot leave the main window with a blank terminal.

Then make the invariant real instead of a convention: `Attach()` takes ownership of a session and detaches whichever control held it before. Last attach wins, the evicted control is genuinely detached, and every eviction is logged so a caller that should have detached first shows up in the log rather than as unexplained reflow. Release is identity-checked, so a control evicted earlier cannot clear the new owner's claim when its own window closes later.

## Proof

7 new tests, and **verified by mutation** rather than just green:

- removing the eviction -> `SecondAttach_EvictsTheFirstControl` fails
- dropping the identity check on release -> `EvictedControlDetaching_DoesNotClearTheNewOwner` fails

Suites run locally: Terminal.Avalonia 31/31, Avalonia 353/353.

`CcDirector.Core` grants internals to the terminal test project so a `Session` can be constructed - compile-time visibility only, no runtime change.

### DIFF
--- a/src/CcDirector.Avalonia/MainWindow.axaml.cs
+++ b/src/CcDirector.Avalonia/MainWindow.axaml.cs
@@ -4398,14 +4398,26 @@ public partial class MainWindow : Window
                 return;
             }
             FileLog.Write("[MainWindow] BtnFifo_Click: opening FIFO window");
-            await new FifoWindow(sm).ShowDialog(this);
 
-            // The FIFO window is full-screen, so attaching a session there resized that
-            // session's PTY to full-screen dimensions. Re-attach the main window's active
-            // session so it re-sends ITS dimensions and redraws cleanly, instead of leaving
-            // the session rendering at the FIFO window's size.
-            if (_activeSession is not null)
-                TerminalHost.Attach(_activeSession.Session);
+            // Hand the terminal over before FIFO opens. FIFO attaches sessions to its OWN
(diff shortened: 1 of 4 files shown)


thefrederiksen:
Superseded: the FIFO takeover is being removed entirely rather than repaired. It sat behind the alpha flag (off by default), and the owner's call is to delete it. The ownership-registry idea from this PR is not lost - it belongs with the panes work in devthrottle_internal#1225, where it is scoped as step 1's ownership rule.

PR #2410, opened 2026-08-03
WIP - wizard git notice, Source Control view, and window sizing (pieces 3 and 4 of a split; do not merge)

thefrederiksen:
## SPLIT DONE. This pull request is now pieces three and four, and it is NOT ready.

The owner approved splitting by RISK rather than by subject. Two pieces have left this pull request
and are open on their own:

| Piece | Where | State |
|---|---|---|
| 1. The clean-machine reword | **#2421** | Ready. Zero dependencies, never faulted in three rounds. |
| 2. The git launch guards | **#2422** | Ready. The part that fixes the real defect; not one finding across three rounds. |
| 3. `GitPresenceDetector` + the Code step notice + the Source Control view | **here** | Open findings - see below. |
| 4. The wizard window sizing | **here** | Blocked on devthrottle_internal#1241. |

**Do not merge this one.** It carries the work every review round has faulted. Its history and the
three review threads below are kept deliberately: the reasoning is worth more than the diff, and
whoever picks pieces three and four up should inherit the findings rather than rediscover them.

### What is still open on piece three

1. **The macOS probe runs Apple's shim instead of the path it just proved safe.** Having established
   that a real git exists behind `/usr/bin/git`, the code goes on to probe `/usr/bin/git` anyway. The
   fix does the hard half and throws it away.
2. **The developer-tools paths are a proxy, not the fact.** `DEVELOPER_DIR`, `xcode-select`, or an
   Xcode installed elsewhere all break the inference. "One of two hard-coded files exists" is not the
   same statement as "this shim will dispatch without prompting", and the comment asserts the latter.
3. **`RequiresGitFact` is circular.** Its availability check calls the very detector the gated test
   asserts on, so a regression from `Present` to `Undetermined` SKIPS instead of failing, and the
   blanket catch turns a detector exception into a skip too. A false green built inside work whose
   subject is false greens.
4. **`CallerCancellation_Propagates` no longer reaches the catch it names** - the pre-cancellation
   check added in the same round short-circuits it. It passes and proves nothing.
5. **`ArrivingAtTheCodeStep_StartsTheGitProbe`** passes if the `Task.Run` that runs the detector is
   deleted, so long as the flag assignment stays.
6. **The attachment generation does not order two refreshes of the SAME attachment**, though the
   comment claims it does. The counter would have to move per refresh, not per attachment.
7. **`AFailedRead_TakesTheStaleBranchBarDownWithIt`** proves a synchronous assignment, not the
   separate-timer race it describes - an in-flight sync read can still put the bar back.
8. **`AGoodReadAfterABadOne_ClearsTheProblemLine`** still shells `git init` under a plain fact, so it
   fails rather than skips on a supported git-less machine.
9. **Nothing exercises the attachment-generation guard at all.**
10. **The PATH walk is outside the probe timeout.** Detection runs off the user interface thread and
    nothing on the step waits for it, so the cost is the sentence rather than the step - but an
    unresponsive PATH entry blocks unbounded.

### What is still open on piece four

- **devthrottle_internal#1241 blocks it.** The two-stage fitter writes its clamped size into
  `Width`/`Height`, so the second stage reads an already-clamped value as the preference and a window
  clamped for the wrong screen can never grow back. That is pre-existing and shipped in v1.9.6, and I
  filed it separately - but the reviewer's counter is accepted: deferring existing debt is
  legitimate, adding a second caller of a known-defective mechanism is not, and this pull request
  puts `FirstRunWizardDialog` onto it.
- **The height assertion is `> 640`**, which 641 satisfies. It pins the direction, not sufficiency.
  A faithful content-fit assertion needs real font metrics the headless Avalonia platform does not
  provide; the only evidence of sufficiency is the screenshots in the history below.

### What IS established here, and should not be re-derived

- The wizard was a **fixed 900x640 at every resolution** - the markup declared it and nothing
  assigned `Width` or `Height` anywhere. devthrottle_internal#1046's open question is answered: the
  clipping is not a small-screen defect, and its title is wrong to scope it to 1024x768.
- `UpdateSourceControlTabVisibility: hasGit=False` in the #1048 thread is **not** git detection. It
  tests `Directory.Exists(repo/.git)` - whether the FOLDER is a checkout. A checkout copied onto a
  machine with no git still shows the Source Control tab.
- On the running application the wizard measured **916x899** where it was 916x679, with all five
  Welcome items and all seven Done receipt rows visible, including the morning-report line.
- 26 guards were revert-proven across four passes, each mutation confirmed present by diff first.

---

<details><summary>The original description, kept for the record</summary>

Closes the two first-run wizard issues from the clean-machine walk of v1.8.7.

> ## PROPOSED: split this into four, by RISK rather than by subject
>
> **Recommendation, pending the owner's decision - nothing here is settled.** Three independent
> review rounds have blocked this pull request. Recording the proposed split and how it was drawn,
> because the reasoning is worth more than the outcome and should not die with the session.
>
> This change has outgrown the two issues it was opened for. Part three's instruction - exercise the
> git paths nobody had exercised - was right and found a real defect, but the work it produced has a
> defect surface of its own, and that is what every round has faulted.
>
> | Piece | Ship when | Why |
> |---|---|---|
> | The clean-machine reword (`CodeNoneAckText`) | **Immediately** | The only piece with genuinely zero dependencies. No round has ever faulted it. |
> | The three launch guards - `GitCommandRunner`, `GitWriteService`, `ProcessRunner` | **Own pull request, own review** | They fix the actual defect the issue asked about, and they have earned it (below). |
> | `GitPresenceDetector`, the Code step notice, and their tests | **Separate work** | Where every finding across three rounds actually lives. |
> | The window sizing | **Blocked on devthrottle_internal#1241** | See below - this pull request adds a second caller of a known-defective mechanism. |
>
> **Two corrections that moved this line, both drawn from what the rounds said rather than from how
> the change is organised:**
>
> 1. **The Code step notice is not "just copy" and cannot ship alone.** The git-is-recommended
>    sentence needs `GitPresenceDetector`, so it inherits the macOS regression and the circular skip
>    attribute. Only the clean-machine reword is dependency-free. Shipping the notice as copy would
>    have shipped both open findings with it.
> 2. **The launch guards are the most settled code here, not the risky half.** Across three rounds
>    not one finding has landed on `GitCommandRunner`, `GitWriteService`, or `ProcessRunner`
>    themselves. (`GitLaunchFailure`, the shared message helper they call, was faulted once in round
>    one over reading Windows error numbers on macOS; that is fixed and has been clean since.) They
>    are also the part that repairs a real fault on a real machine - `Process.Start` throwing rather
>    than returning false, so the exception escapes into callers written against a result object.
>    Holding them back would defer the only thing here that fixes something, on the strength of
>    findings belonging to a different component: judging a part by the reputation of the whole,
>    which is a sibling of the fault this whole pull request keeps catching.
>
> **On the window sizing.** devthrottle_internal#1241 was filed as a separate issue because the
> two-stage fitter defect is pre-existing and shipped in v1.9.6. The reviewer's counter is better
> than that argument and is accepted: deferring existing debt is legitimate, but ADDING a second
> caller of a known-defective mechanism is not, and putting `FirstRunWizardDialog` onto that fitter
> does exactly that. thefrederiksen/devthrottle_internal#738 therefore blocks this half rather than sitting beside it.

## The ruling this implements

The owner settled the design: **detect and tell, never force, never install, and everything must
work without git.** Another version control system, or none at all, is a legitimate choice. What is
not acceptable is that a user finds out from the command line that the tool the product just told
them to use is not on their machine.

## Part one - tell them (#1048)

The Code step now says, in one sentence and only when git is definitely absent, that we could not
find it, that nothing here needs it, and that it is highly recommended - with a link to where to get
it. It is a sentence on the screen they are already looking at: not a warning, not a modal, not a
gate. The primary button is untouched.

Detection is deliberately **three-state**. Resolving a name on PATH is not enough, because a file
existing is not a working install, so it resolves the name the way the operating system would and
then **runs** it and requires it to identify itself. Anything that leaves the question open - a
launch that fails, a timeout, a non-zero exit, an executable that answers without printing a git
version - is `Undetermined`, and `Undetermined` says **nothing at all**. Folding it into "absent"
would tell someone who has git that they have not got it, which is a claim about their computer that
we have not established.

## Part two - stop advising the impossible (#1048)

"When you clone or create your first repository" was advice a machine with no git cannot act on.
Making a folder always works, needs nothing installed, and is exactly what the clean-machine walk
did: a plain folder with two files in it, no git anywhere. Cloning is still named as the other way
in; it is just no longer the only route offered.

## Part three - make it actually work without git

The issue flagged the Director's Source Control button and said outright that it was not exercised.
It is now, and it was not fine.

**`Process.Start` does not return false when git is missing - it throws `Win32Exception`.** So the
missing-git branch had never run, and the exception left three places by a route none of their
callers expects, every one of which is written against a result object:

| Where | What it did with no git | Now |
|---|---|---|
| `GitCommandRunner` | Threw into the worktree inventory, the branch / diff / history services and the repository monitor | Returns the failed result its callers already handle, so the Worktrees page reports the reason it already knows how to render |
| `GitWriteService` | Threw out of stage, unstage, discard and commit - desktop buttons, Control API git verbs, and through them the Cockpit and the phone | Same fix, so all three surfaces get a sentence instead of an exception |
| `ProcessRunner` | Documented a `Started=false` result it could never produce | Produces it, which is what makes the two read providers honest without a catch-all |
| `GitChangesView` | **Returned silently** and left "No changes detected" on screen - a statement about the repository, when nothing had been read | Says what happened, and clears itself when a read succeeds |
| `GitStatusProvider` / `GitSyncStatusProvider` | Reported a fixed "Failed to start git process" - also shown when git ran perfectly well and exited non-zero | Carry the real reason, and tell a failed launch apart from a non-zero exit |

Paths checked and **left alone** because they were already correct: `CodeFolderScout` (reads `.git`
directories on disk, never runs git - which is why the wizard's folder sweep works on a clean
machine), `WingmanService.GitSnapshotAsync` (catches and reports `git_failed`),
`GitHubUrls.ResolveRepoNameCached` (catches, caches the miss, groups by folder name),
`WorktreeInventoryService` and `WorktreesView` (already render "Could not read worktrees: ...").

**One correction to the issue thread, because it corrects a shared mental model and not just code.**
`UpdateSourceControlTabVisibility: hasGit=False` does **not** show that the product detects the
absence of git. It tests `Directory.Exists(repo/.git)` - whether the FOLDER is a checkout, not
whether git is INSTALLED. Anyone reading that thread would inherit the belief that the product
already knows when git is missing and acts on it. It does not, and that is precisely why the Source
Control tab is offered on a machine that cannot run a single git command.

The sentence shown for a failed launch lives in one place and does not over-claim: "not installed"
is said only for the two operating system codes that mean the file is not there. A permission
refusal keeps its own words rather than sending someone off to reinstall software already on the
machine.

## Issue thefrederiksen/devthrottle#1046 - the wizard fits the display

That issue asked one open question: is the window fixed, or does it grow with the desktop. **It was
fixed.** The markup declared `900x640` and nothing anywhere assigned the window's `Width` or
`Height`, so the 916x679 frame photographed at 1024x768 is the size it would be on any monitor. The
clipping is a defect at every resolution, and the issue's title is wrong to scope it to small
screens.

It now uses the same `WindowFit` fold the main window got in abca6682a, lifted out of `MainWindow`
so there is one copy rather than two: clamp to the work area before the window is shown, re-fit once
the frame can be measured, centre. The old fixed size did not even fit the smallest supported
display - 640 of client plus a 39 pixel frame is 679, against 672 of work area - which nothing had
noticed. Every scrolling area in the wizard also keeps its scrollbar instead of fading it out, so a
list too tall for the display reads as truncated rather than as a list that ends there.

## Evidence

**Revert-proof: 19 guards across three passes, every one mutated in the real production line and
confirmed red.** Each mutation was verified present by diff before its tests ran, because a mutation
that never applied gives a green that reads exactly like a finding. The later passes are described
in the comments below; two of them caught tests of mine that could not fail, including the macOS
guard, whose branch is never taken on Windows and which therefore had no test able to catch its own
deletion. The first twelve:

| Mutation | Turned red |
|---|---|
| `GitCommandRunner`: remove the missing-git guard | yes (2 tests) |
| `GitWriteService`: remove the missing-git guard | yes (2) |
| `ProcessRunner`: remove the failed-launch guard | yes (2) |
| Detector: fold `Undetermined` into `NotFound` | yes |
| `ShouldAdviseInstallingGit` becomes "anything but Present" | yes (4) |
| `GitLaunchFailure`: claim "not installed" for every code | yes |
| `GitStatusProvider`: restore the fixed launch message | yes |
| `GitSyncStatusProvider`: restore the fixed launch message | yes |
| Wizard: show the git notice unconditionally | yes (2) |
| Wizard: restore the clone-only acknowledgement | yes |
| Wizard: restore the fixed 640 height | yes (2) |
| `GitChangesView`: restore the silent early return | yes (2) |

**Run on the real application**, not only in tests. A Director on a throwaway instance home, so the
first-run wizard opens for real:

```
[FirstRunWizardDialog] ApplyFit (pre-show): workArea=1920x1032 ... desired=900x860, chosen=900x860
[FirstRunWizardDialog] ApplyFit (opened):   ... frame=16x39, desired=900x860, chosen=900x860 at 502,66
[MainWindow] ApplyFit (opened): ... desired=1400x900, chosen=1400x900 at 252,46
```

The wizard measured **916x899** where it used to be 916x679, and the main window is unchanged by the
extraction. Both screens the issue named were photographed whole:

- **Welcome** shows all five items, including "A browser for your agents" and its explanatory line.
- **Done** shows all seven receipt rows, including "Morning report - Every morning at 7:00 Eastern",
  the consent-adjacent row that was below the fold.

## What is NOT proven, and what is still exposed

Read this section before relying on anything above.

- **A REMAINING EXPOSURE, not a limitation of the testing.** Everything here is scoped to the
  DETECTOR and to the surfaces listed in the table. The rest of the product still launches bare
  `git` whenever a repository is open - the repository monitor, the branch and history services,
  the worktree inventory. On macOS those launches resolve to `/usr/bin/git`, so on a Mac without the
  Command Line Tools **they can still raise Apple's install dialog**. That is pre-existing behaviour
  and is unchanged by this pull request; the detector is now the one place that refuses to do it.
  Whoever picks up the macOS story next should start here rather than rediscovering it.
- **This has not been run on a Mac.** What is proven is that the detector does not launch
  `/usr/bin/git` on macOS. That launching it prompts is Apple's documented shim behaviour, not
  something observed here.
- **The git notice was never photographed on a running Director.** The walk to reach the Code step
  with git removed from the process PATH was abandoned for time. Its behaviour is covered by tests
  at both levels - the three-state detector, and the wizard panel following the verdict for all
  three states - but nobody has looked at the rendered sentence.
- **The wizard sizing tests do not prove the wizard CALLS the fitter.** They bind the size declared
  in the markup to what `WindowFit` does with it; the headless display is roomier than anything
  under test, so the wiring cannot be exercised there. The wiring proof is the running application:
  the `ApplyFit` log lines and the two screenshots above.
- **`height > 640` would be satisfied by 641.** The threshold pins the direction, not the
  sufficiency. Sufficiency is the photographed Welcome and Done screens.
- **The fitted size was measured on one display** (1920x1080). Behaviour at 1024x768 and at the
  1280x720 minimum is computed by the tested fold, not observed.
- **Several tests need real git** and would fail on a machine without it. Deliberate - they are
  integration tests, and a red on a git-less machine is informative rather than misleading.

## Unrelated defect found on the way

`scripts/agent-session-isolation.ps1` cannot launch a Director any more. It waits for
`Kestrel listening on http://0.0.0.0:<port>` in the log; the Director now binds loopback only and
writes `http://127.0.0.1:<port>`, so the launch always times out after 120 seconds and stops the
process it just started. Not touched here.

Closes devthrottle_internal#1048
Closes devthrottle_internal#1046

</details>

### DIFF
--- a/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
+++ b/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
@@ -71,6 +71,29 @@ public class GitChangesViewFailedReadTests
     }
 
     /// <summary>
+    /// The branch bar is filled by a SEPARATE read on its own timer, so a failed changes read used
+    /// to leave it standing: last known branch, ahead and behind counts, presented as current, right
+    /// beside a line saying nothing could be read. Stale numbers next to an admission of ignorance
+    /// are worse than no numbers.
+    /// </summary>
+    [AvaloniaFact]
+    public async Task AFailedRead_TakesTheStaleBranchBarDownWithIt()
+    {
+        var gone = Path.Combine(Path.GetTempPath(), "cc-director-changes-view-tests", Guid.NewGuid().ToString("N"));
+
+        var view = new GitChangesView();
+        view.Attach(gone);
+        // Stand in for an earlier successful sync read, which is what puts the bar up.
+        view.BranchBar.IsVisible = true;
+
+        await view.RefreshAsync();
+
+        Assert.True(view.ProblemText.IsVisible);
+        Assert.False(view.BranchBar.IsVisible, "the branch bar stayed up showing stale counts beside a failed read");
+        view.Detach();
+    }
+
+    /// <summary>
     /// A successful read clears the accusation. A machine that has been fixed must stop being told
     /// it is broken - and a problem line that never goes away is its own defect.
     /// </summary>
--- a/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
+++ b/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
@@ -43,6 +43,30 @@ public class GitChangesViewFailedReadTests
     }
 
     /// <summary>
+    /// A folder that is gone - deleted, or on a drive that has been unplugged - is the other way this
+    /// page used to go quiet. It returned before reading anything and left the previous repository's
+    /// changes, or the "No changes detected" empty state, describing a folder that is not there.
+    /// </summary>
+    [AvaloniaFact]
+    public async Task AFolderThatIsNoLongerOnDisk_SaysSo()
+    {
+        // A path that is simply not there - the state a deleted folder or an unplugged drive leaves
+        // behind. Nothing is created and nothing has to be cleaned up, so this test cannot fail on
+        // Windows holding a handle open rather than on the behaviour under test.
+        var gone = Path.Combine(Path.GetTempPath(), "cc-director-changes-view-tests", Guid.NewGuid().ToString("N"));
+
+        var view = new GitChangesView();
+        view.Attach(gone);
+
+        await view.RefreshAsync();
+
+        Assert.True(view.ProblemText.IsVisible);
+        Assert.Contains("no longer on disk", view.ProblemText.Text ?? "");
+        Assert.False(view.EmptyText.IsVisible);
+        view.Detach();
+    }
+
+    /// <summary>
     /// A successful read clears the accusation. A machine that has been fixed must stop being told
     /// it is broken - and a problem line that never goes away is its own defect.
     /// </summary>
--- a/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
+++ b/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
@@ -1,0 +1,101 @@
+using Avalonia.Headless.XUnit;
+using CcDirector.Avalonia.Controls;
+using Xunit;
+
+namespace CcDirector.Avalonia.Tests;
+
+/// <summary>
+/// The Source Control tab's Changes page says when it could NOT read, instead of leaving "No changes
+/// detected" on the screen (devthrottle_internal issue #1048).
+///
+/// "No changes detected" is a statement about the repository. When the read fails nothing has been
+/// established about the repository at all, so the page was asserting something it had not observed -
+/// and the two states were indistinguishable to the user.
+///
+/// The failure injected here is a folder that is not a git checkout, because that fails the read on
+/// a machine that HAS git and so can be reproduced anywhere. It is the same branch a machine with no
+/// git takes: that git-is-absent reaches this branch as a failed read is proved separately, in
+/// CcDirector.Core.Tests GitAbsentTests.
+/// </summary>
+public class GitChangesViewFailedReadTests
+{
+    [AvaloniaFact]
+    public async Task AFailedRead_SaysSo_RatherThanClaimingThereAreNoChanges()
+    {
+        var dir = Path.Combine(Path.GetTempPath(), "cc-director-changes-view-tests", Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(dir);
+        try
+        {
+            var view = new GitChangesView();
+            view.Attach(dir);
+
+            await view.RefreshAsync();
+
+            Assert.True(view.ProblemText.IsVisible);
+            Assert.Contains("could not be read", view.ProblemText.Text ?? "");
+            Assert.False(view.EmptyText.IsVisible);
+            view.Detach();
+        }
+        finally
+        {
+            TryDelete(dir);
+        }
+    }
+
+    /// <summary>
+    /// A successful read clears the accusation. A machine that has been fixed must stop being told
+    /// it is broken - and a problem line that never goes away is its own defect.
+    /// </summary>
+    [AvaloniaFact]
+    public async Task AGoodReadAfterABadOne_ClearsTheProblemLine()
+    {
+        var repo = Path.Combine(Path.GetTempPath(), "cc-director-changes-view-tests", Guid.NewGuid().ToString("N"));
+        var notARepo = Path.Combine(Path.GetTempPath(), "cc-director-changes-view-tests", Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(repo);
+        Directory.CreateDirectory(notARepo);
+        try
+        {
+            await RunGitAsync(repo, "init");
+
+            var view = new GitChangesView();
+
+            view.Attach(notARepo);
+            await view.RefreshAsync();
+            Assert.True(view.ProblemText.IsVisible);
+
+            view.Attach(repo);
+            await view.RefreshAsync();
+
+            Assert.False(view.ProblemText.IsVisible);
+            Assert.True(view.EmptyText.IsVisible);
+            view.Detach();
+        }
+        finally
+        {
+            TryDelete(repo);
+            TryDelete(notARepo);
+        }
+    }
+
+    private static async Task RunGitAsync(string workingDirectory, params string[] args)
+    {
+        var psi = new System.Diagnostics.ProcessStartInfo
+        {
+            FileName = "git",
+            WorkingDirectory = workingDirectory,
+            RedirectStandardOutput = true,
+            RedirectStandardError = true,
+            UseShellExecute = false,
+            CreateNoWindow = true,
+        };
+        foreach (var a in args) psi.ArgumentList.Add(a);
+        using var proc = System.Diagnostics.Process.Start(psi)!;
+        await proc.WaitForExitAsync();
+    }
+
+    private static void TryDelete(string dir)
+    {
+        try { if (Directory.Exists(dir)) Directory.Delete(dir, recursive: true); }
+        catch { /* a git object store on Windows can hold a handle a moment longer; the temp folder is disposable */ }
+    }
+}
--- a/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
+++ b/src/CcDirector.Avalonia.Tests/GitChangesViewFailedReadTests.cs
@@ -61,7 +61,11 @@ public class GitChangesViewFailedReadTests
         await view.RefreshAsync();
 
         Assert.True(view.ProblemText.IsVisible);
-        Assert.Contains("no longer on disk", view.ProblemText.Text ?? "");
+        Assert.Contains("cannot be reached", view.ProblemText.Text ?? "");
+        // And it stops short of naming a cause. Directory.Exists is false for a folder that was
+        // deleted AND for one that is merely unreachable or unreadable, so "no longer on disk" -
+        // which is what this said first - asserted something nothing here established.
+        Assert.DoesNotContain("no longer on disk", view.ProblemText.Text ?? "");
         Assert.False(view.EmptyText.IsVisible);
         view.Detach();
     }
--- a/src/CcDirector.Avalonia.Tests/WizardGitNoticeTests.cs
+++ b/src/CcDirector.Avalonia.Tests/WizardGitNoticeTests.cs
@@ -1,0 +1,84 @@
+using Avalonia.Headless.XUnit;
+using CcDirector.Core.Configuration;
+using CcDirector.Core.Git;
+using Xunit;
+
+namespace CcDirector.Avalonia.Tests;
+
+/// <summary>
+/// The setup wizard's Code step tells the user when git is missing, and says nothing when it cannot
+/// tell (devthrottle_internal issue #1048).
+///
+/// The owner's ruling is the whole specification: detect and tell, never force, never install, and
+/// everything works without it. So the only thing this screen does about git is show one sentence,
+/// and it shows it only for a DEFINITE absence.
+/// </summary>
+public class WizardGitNoticeTests
+{
+    private static FirstRunWizardDialog Wizard() => new(new AgentOptions());
+
+    private static GitPresence With(GitAvailability availability)
+        => new(availability, availability == GitAvailability.Present ? "C:\\Program Files\\Git\\cmd\\git.exe" : null,
+            availability == GitAvailability.Present ? "git version 2.45.1" : null, "test");
+
+    [AvaloniaFact]
+    public void NoticeIsHiddenBeforeAnythingHasBeenDetected()
+    {
+        var wizard = Wizard();
+
+        // A screen that accuses the machine before it has looked is worse than one that never does.
+        Assert.False(wizard.CodeNoGitPanel.IsVisible);
+        Assert.Null(wizard.DetectedGitPresence);
+    }
+
+    [AvaloniaFact]
+    public void GitMissing_ShowsTheRecommendation()
+    {
+        var wizard = Wizard();
+
+        wizard.ApplyGitPresence(With(GitAvailability.NotFound));
+
+        Assert.True(wizard.CodeNoGitPanel.IsVisible);
+    }
+
+    [AvaloniaFact]
+    public void GitPresent_SaysNothing()
+    {
+        var wizard = Wizard();
+
+        wizard.ApplyGitPresence(With(GitAvailability.Present));
+
+        Assert.False(wizard.CodeNoGitPanel.IsVisible);
+    }
+
+    /// <summary>
+    /// The state the whole three-way split exists for. A probe that could not reach a verdict must
+    /// leave the screen silent - telling someone with git installed that they have not got it is the
+    /// failure this avoids, and it is the one a two-state detector cannot avoid.
+    /// </summary>
+    [AvaloniaFact]
+    public void GitUndetermined_SaysNothing()
+    {
+        var wizard = Wizard();
+
+        wizard.ApplyGitPresence(With(GitAvailability.Undetermined));
+
+        Assert.False(wizard.CodeNoGitPanel.IsVisible);
+    }
+
+    /// <summary>
+    /// Part two of issue #1048: the clean-machine acknowledgement must hold on a machine that cannot
+    /// clone. "Clone your first repository" was advice the reader could not act on; making a folder
+    /// always works, and is exactly what the clean-machine walk did.
+    /// </summary>
+    [AvaloniaFact]
+    public void TheCleanMachineAcknowledgementDoesNotRequireGit()
+    {
+        var wizard = Wizard();
+
+        var text = wizard.CodeNoneAckText.Text ?? "";
+
+        Assert.Contains("Make a folder", text);
+        Assert.DoesNotContain("When you clone", text);
+    }
+}
--- a/src/CcDirector.Avalonia.Tests/WizardGitNoticeTests.cs
+++ b/src/CcDirector.Avalonia.Tests/WizardGitNoticeTests.cs
@@ -1,6 +1,7 @@
 using Avalonia.Headless.XUnit;
 using CcDirector.Core.Configuration;
 using CcDirector.Core.Git;
+using CcDirector.Core.Onboarding;
 using Xunit;
 
 namespace CcDirector.Avalonia.Tests;
@@ -64,6 +65,22 @@ public class WizardGitNoticeTests
         wizard.ApplyGitPresence(With(GitAvailability.Undetermined));
 
         Assert.False(wizard.CodeNoGitPanel.IsVisible);
+    }
+
+    /// <summary>
+    /// THE WIRING. Every other test here drives ApplyGitPresence directly, so deleting the detection
+    /// call from ShowStep would leave all of them green - the reviewer's point, and it was right.
+    /// This one proves that arriving at the Code step is what starts the probe.
+    /// </summary>
+    [AvaloniaFact]
+    public void ArrivingAtTheCodeStep_StartsTheGitProbe()
+    {
+        var wizard = Wizard();
+        Assert.False(wizard.GitProbeStarted);
+
+        wizard.ShowStepForTests(WizardStep.Code);
+
+        Assert.True(wizard.GitProbeStarted, "reaching the Code step did not start git detection");
     }
 
     /// <summary>
--- a/src/CcDirector.Avalonia.Tests/WizardWindowFitTests.cs
+++ b/src/CcDirector.Avalonia.Tests/WizardWindowFitTests.cs
@@ -41,6 +41,31 @@ public class WizardWindowFitTests
     /// </summary>
     private const double TheOldFixedHeight = 640;
 
+    /// <summary>
+    /// THE WIRING. Every other test here calls WindowFit.Fit itself, so deleting the two production
+    /// calls in the wizard would leave them all green - the reviewer's point, and it was right. The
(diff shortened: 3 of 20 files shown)


thefrederiksen:
## Independent review ran and BLOCKED this. Six real findings, all fixed. Two of my own tests could not fail.

An independent reviewer read the whole diff plus the sibling files the reasoning depends on, and
blocked the merge with ten findings. Six were real. Recording all ten, including the four I did not
change, so the next reader is not re-deriving them.

### Fixed

1. **macOS: the detector would have offered to install software.** `/usr/bin/git` is Apple's Command
   Line Tools shim. A Mac without those tools still HAS that file, `ExecutableResolver` finds it, and
   RUNNING it puts up Apple's "install the developer tools?" dialog. So the one class whose entire
   remit is to say a sentence and change nothing would have been the thing prompting an install -
   the exact opposite of the ruling. It is now the one executable the detector refuses to launch.
   The consequence is taken deliberately and stated in the code: on a Mac whose git lives there we
   reach no verdict and say nothing. A Mac with git from Homebrew resolves elsewhere and is probed
   normally.
2. **Windows error numbers were read as Windows error numbers on every platform.** Code 2 means "no
   such file" on Windows and POSIX alike and is still read everywhere. Code 3 is
   `ERROR_PATH_NOT_FOUND` on Windows but `ESRCH`, "no such process", on POSIX - reading it as a
   missing install would have told a Mac user to reinstall git that is sitting on their disk.
3. **A folder that is no longer on disk was still a silent state.** `RefreshAsync` returned before
   reading anything when the directory was gone, leaving the previous repository's changes or the
   "No changes detected" empty state describing a folder that is not there. Same defect as the one
   this pull request already fixed, one branch earlier.
4. **A refresh was not tied to the attachment that started it.** A slow FAILED read of folder A could
   land after folder B had loaded cleanly and wipe B's results with A's problem line. The path is
   pinned for the whole call and a stale result is dropped.
5. **The version banner was a loose substring match.** A program whose warning text happened to
   contain "git version" was accepted as a working git, and the line stored as its version was the
   warning. It must now be what the program LEADS with.
6. **Caller cancellation was swallowed** by the general catch and returned as a verdict, so a caller
   who asked to stop got a confident-looking answer about the machine instead.

Four claims were also cut back to what the mechanism delivers. The detector checks the PATH - the
same lookup every git call in the product makes, so it is exactly the condition under which
DevThrottle cannot run git, and it is NOT a search of the disk. The comments, the test name and the
sentence on screen now say that.

### The revert-proof pass caught two tests of my own that could not fail

Worth recording, because both would have read as coverage:

- The cancellation test supplied its own probe, so it never reached the catch inside the real
  version probe that has to let cancellation through. The exception it observed came straight back
  out of its own fake. Removing the real guard left it green. It drives the real detector now.
- The code-3 rule was **unfalsifiable on Windows**: the correct rule and "code 3 means missing
  everywhere" produce identical output on this machine, so no test running here could fail however
  wrong the rule became. The platform is passed in and both arms are asserted.

All four new guards now mutate red, each mutation confirmed present by diff first. Sixteen guards
proven across the two passes.

### Accepted, not fixed

- **The wizard sizing tests do not prove the wizard CALLS the fitter.** True. They bind the size
  declared in the markup to what `WindowFit` does with it; the headless display is roomier than
  anything under test, so the wiring cannot be exercised there. The wiring proof is the running
  application - the `[FirstRunWizardDialog] ApplyFit` log lines and the two screenshots in the
  description.
- **A preferred height of 641 would satisfy `height > 640`.** Also true. The threshold pins the
  direction, not the sufficiency; sufficiency is the photographed Welcome and Done screens.
- **Several tests need real git and would fail on a machine without it.** Deliberate - they are
  integration tests, and a red on a git-less machine would be informative rather than misleading.
- **The wizard tests call `ApplyGitPresence` directly**, so deleting the `ShowStep` detection call
  would leave them green. Acknowledged; that call is covered only by the running application.

### Still not proven, unchanged from the description

The git notice has never been photographed on a running Director with git removed from its PATH.
Nobody has looked at the rendered sentence.

thefrederiksen:
## The macOS finding is CLOSED, not noted - and the guard can now fail

A detector with a side effect on the user's machine is not a detector, so this one needed more than
a fix: it needed a test that can catch the fix being removed. On Windows the macOS branch is never
taken, so the guard as first written had **no test on this machine that could fail if it were
deleted** - coverage that cannot catch its own removal, protecting the one thing standing between our
detection and an installer dialog on somebody's screen.

**How it is closed.** `ExecutableResolver` still finds `/usr/bin/git` on a Mac, because that file
exists whether or not the Command Line Tools do. The detector now recognises that exact path on
macOS and **returns without calling the probe at all** - no process is started, so nothing can
prompt. The verdict is `Undetermined`, which the wizard renders as nothing. So on a Mac with no
developer tools the probe launches nothing and the user sees nothing from it.

**The assertion is the absence of the launch, not the verdict.** A test that only checked the
verdict would pass while the dialog still appeared:

```
Assert.False(probeWasCalled, "the macOS developer-tools stub was executed; that can open Apple's install dialog");
```

**Three mutations, each confirmed present by diff, each red:**

| Mutation | Caught by |
|---|---|
| Delete the guard - the detector runs Apple's shim again | `OnMacOs_TheDeveloperToolsStubIsNeverLaunched` |
| Widen it to every macOS path - every Mac falls silent, including Homebrew installs | `OnMacOs_AGitThatIsNotTheStubIsStillProbed` |
| Drop the platform - a non-Mac git at that path is refused too | `OffMacOs_TheSamePathIsProbedNormally` |

Nineteen guards proven across three passes.

**One limit, stated rather than implied.** This has not been run on a Mac. What is proven is that the
probe does not launch that path on macOS; that launching it prompts is Apple's documented shim
behaviour, not something observed here. And it is scoped to the DETECTOR - the rest of the product
still runs bare `git` when a repository is open, which is pre-existing and unchanged by this pull
request.

## Correcting the record for the next reader

Repeating this from the description because it corrects a shared mental model, not just code:

**`UpdateSourceControlTabVisibility: hasGit=False` in the issue thread was never git detection.** It
tests `Directory.Exists(repo/.git)` - whether the FOLDER is a checkout, not whether git is
INSTALLED. Anyone reading that thread would inherit the belief that the product already knows when
git is missing and acts on it. It does not, and that is precisely why the Source Control tab is
offered on a machine that cannot run a single git command.

thefrederiksen:
## Re-check: the reviewer BLOCKS again, and disputes all four of my accepted-with-reasons

Posting the verdict in full rather than summarising it in my favour. **Do not merge on my earlier
"all ten answered" - that was the author's account, and the reviewer does not agree with it.**

### Verdict on the original ten

| # | Finding | Reviewer verdict |
|---|---|---|
| 1 | macOS stub launched | FIXED |
| 2 | Missing directory left a false state | FIXED |
| 3 | Stale-attach race | **PARTIALLY FIXED** |
| 4 | Windows error numbers on every platform | FIXED |
| 5 | Machine-wide absence overclaim | **PARTIALLY FIXED** |
| 6 | Timeout excludes the PATH walk | **NOT FIXED** |
| 7 | Cancellation swallowed | FIXED |
| 8 | Loose version banner | FIXED |
| 9 | Wizard wiring / real-git tests | **NOT FIXED** |
| 10 | Fitter wiring / height threshold | **NOT FIXED** |

Six fixed, two partial, two not fixed.

### The four disputes, in the reviewer's terms

1. **Fitter wiring.** "Logs and screenshots demonstrate one manual run, not a regression constraint.
   The automated suite remains unable to detect removal of the production fitter calls."
2. **Height threshold.** "'Direction' is not the requirement. The requirement is sufficient visible
   content or correctly functioning scrolling. A one-pixel increase passes while preserving
   essentially the entire defect."
3. **Tests requiring real git.** "Git-less machines are explicitly supported. A suite that becomes
   red solely because an optional program is absent reports a false product regression." It wants
   them classified and conditionally skipped, or run where git is a declared prerequisite.
4. **`ShowStep` wiring.** "The author has confirmed the exact coverage hole: deleting the detection
   call leaves every wizard test green. Coverage 'only by the running application' is not an
   acceptable regression test."

I think the reviewer is right on all four. Number three is the sharpest: this pull request exists
because git-less machines are supported, and I wrote tests that go red on one. That is the same
category of mistake as the defect being fixed.

### New defects it found in the fixes themselves

My triage, marked as my judgement rather than the reviewer's:

**Real, and I would fix before merge:**
- **Every stock Mac is now permanently `Undetermined`.** The guard cannot tell working Apple git at
  `/usr/bin/git` from the prompting stub, so on the commonest Mac configuration the wizard can
  neither confirm git nor advise a Mac that genuinely lacks it. I stated this as an accepted
  consequence; the reviewer is right that "accepted" undersells losing the whole feature on macOS.
- **`ShowProblem` leaves the branch bar visible** - stale state beside a problem line, the exact
  defect class this pull request is about.
- **"No longer on disk" is asserted for any `Directory.Exists == false`**, including a path that is
  merely unavailable or unreadable. Same overclaim, one layer down.
- **A pre-cancelled detector call still returns a verdict** on the early-return paths; there is no
  cancellation check before resolution.
- **The refresh guard is path equality, not attachment identity.** A to B to A lets a stale result
  land, same-path refreshes can finish out of order, and `OrdinalIgnoreCase` conflates distinct
  paths on case-sensitive filesystems. It needs a generation token.

**Real but pre-existing, inherited by the extraction, not introduced here:**
- **The two-stage fitter treats its own clamped value as the preference.** `FitBeforeShow` writes the
  clamped size into `Width`/`Height`, and `FitOnOpened` then reads that as the desired size - so a
  window clamped for a smaller preliminary screen cannot recover on the real one. This is in
  `abca6682a` and affects the main window too. Worth its own issue rather than being folded in here.

**Marginal, recording without acting:**
- A symlinked PATH directory could route around the macOS guard, which compares path strings rather
  than canonical identity.
- POSIX `ENOENT` can also mean a missing interpreter for an existing script, which `GitLaunchFailure`
  would still describe as git not being installed.
- The test file's own summary still says "when git is not on the machine at all" when it injects a
  nonexistent command name. Cheap wording fix.

### Where this leaves the merge

Not ready. The reviewer's bar for the four disputes is real work: a generation token for the refresh,
a conditional-skip classification for the git-dependent tests, a wiring test for the two production
call sites, and a content-sufficiency assertion rather than a threshold.

thefrederiksen:
## Third review round: BLOCK again. This time the findings are mostly against my tests.

Posting the verdict unedited. **Still do not merge.** Seven items improved, but the reviewer found a
new crop, and the pattern in them is the thing worth naming: I keep writing tests whose names claim
more than their mechanism delivers, and I have now been caught on exactly that three rounds running.

### Fixed outright

- Pre-cancellation before the early returns.
- The "cannot be reached" wording.
- The Code step probe wiring exists (narrowly - see below).

### Where it says I did not finish

- **The macOS probe still runs the shim, not the executable it just proved safe.** Having established
  that a real git exists behind `/usr/bin/git`, the code goes on to probe `/usr/bin/git` anyway. It
  should probe the path it found. Correct, and embarrassing - the fix does the hard half and then
  throws it away.
- **The developer-tools paths are a proxy, not the fact.** `DEVELOPER_DIR`, `xcode-select`, or an
  Xcode installed elsewhere all break the inference. "One of two hard-coded files exists" is not the
  same statement as "this shim will dispatch without prompting", and my comment asserts the latter.
- **The attachment generation does not order two refreshes of the SAME attachment.** Both capture the
  same number, so the older can still land last. My comment claims that case is handled. It is not -
  the counter would have to move per refresh, not per attachment.
- **`AGoodReadAfterABadOne_ClearsTheProblemLine` still shells `git init` under a plain fact.** The
  skip attribute I added for exactly this reason is not on the one test that most needs it.
- **The fit-on-opened wiring is untested.** The new test constructs the wizard but never opens it, so
  deleting `OnOpened` leaves it green. It proves the pre-show call only.
- **The two-stage fitter defect is newly applied to the wizard by this pull request.** I filed the
  shipped MainWindow case as devthrottle_internal#1241 and argued the split. The reviewer's counter is
  better than my argument: deferring existing debt is legitimate, adding a second caller of a
  known-defective mechanism is not. I concede this one.

### Tests of mine that cannot fail, or prove something else

This is the part I want on the record rather than buried:

- **`RequiresGitFact` is circular.** Its availability check calls the very detector that
  `GitPresenceDetector_OnThisMachine_ReachesADefiniteVerdict` asserts on. If the detector regresses
  from `Present` to `Undetermined`, the test SKIPS instead of failing - and the blanket `catch`
  turns a detector exception into a skip too. I built a skip gate out of the thing under test. That
  is a false green of my own manufacture, added while fixing false greens.
- **`CallerCancellation_Propagates` no longer reaches the catch it names.** My own pre-cancellation
  check, added the same round, short-circuits it before the probe. The test still passes and now
  proves nothing about the catch in `RunVersionAsync`.
- **`ArrivingAtTheCodeStep_StartsTheGitProbe`** passes if the `Task.Run` that actually runs the
  detector is deleted, so long as the flag assignment stays.
- **`AFailedRead_TakesTheStaleBranchBarDownWithIt`** proves a synchronous assignment, not the
  separate-timer race the comment describes - an in-flight sync read can still put the bar back.
- **`GitStatusProvider_WhenGitCannotBeLaunched_ReportsWhy`** asserts a prefix, so
  "git could not be started: nonsense" would pass.
- **Nothing exercises the attachment-generation guard at all.**

### Where this leaves it

Three rounds, three blocks, and the remaining work is no longer small. My reading, for whoever
decides: the pull request has outgrown the two issues it was opened for. The wizard copy and the
window sizing - what #1048 parts one and two and thefrederiksen/devthrottle#1046 actually asked for - are close. The git
plumbing hardening, which came from part three's instruction to exercise the paths, has turned into
a substantial piece of work with a defect surface of its own, and it is what every round has found
fault in.

I am not making that call unilaterally, and I am not merging.

thefrederiksen:
Closing this. The findings on pieces three and four are now recorded on thefrederiksen/devthrottle_internal#1771 so they survive without a draft pull request holding them open - a pull request is the wrong place to store review findings, and this one has been aging as a draft since 3 August.

The branch `fix/wizard-git-optional-and-sizing` is KEPT, and the three review threads stay here for the record. Whoever picks the work up should start from thefrederiksen/devthrottle_internal#1771 and read these threads.

Pieces one and two shipped as thefrederiksen/devthrottle#2421 and thefrederiksen/devthrottle#2422.

PR #2436, opened 2026-08-03
Give every merge to main its own CI run, so none is evicted before it starts

thefrederiksen:
## What was wrong

Continuous integration on main was believed to keep a verdict for every commit, because #2325 turned off `cancel-in-progress` for pushes. It does not. That flag protects only the run that is already **executing**. A concurrency group holds at most **one** run in the pending slot, and the arrival of a third push discards whichever run was waiting there.

Measured on main for 3 August: **24 runs, 16 cancelled**, every one of them with `jobs: []` - no job ever started. Each cancellation timestamp matches the next run's creation time to the second (the 10:52 run was killed at 10:57:27; the next run was created at 10:57:26). The .NET job takes about fifty minutes and merges land every two to five, so nearly everything queued was evicted before it could run.

## Why it mattered

Two architecture guards were red on main that morning - `NoCrossMachineLoopbackGuardTests` and `GatewayOnlyTranscriptionGuardTests`, both misreading the new `CcDirector.Gateway.UnitTests` project as production code. The commit that repaired them (#2424) had **its own run evicted**, so GitHub reported neither the break nor the fix.

Continuous integration is also the backstop for the two suites parked out of the fast local gate. A backstop that is evicted before it starts is not a backstop.

## The change

The concurrency group is keyed on the commit for a push, so main leaves the shared queue entirely and no later merge can take an earlier one's slot. Pull requests are untouched: one run per branch, newest supersedes.

```yaml
group: ci-${{ github.workflow }}-${{ github.event_name == 'push' && github.sha || github.ref }}
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

## Cost, accepted deliberately

One full run per merge instead of one per burst. Runs may still **wait** on the account-wide runner limit, but waiting yields a late verdict where eviction yields none at all.

## Verification

The workflow parses and the concurrency expression resolves as intended (push -> the commit, pull request -> the ref). The push half cannot be exercised until this is on main; the first merge after it lands is the proof, and I will check that its run survives.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,21 +6,40 @@ on:
   pull_request:
     branches: [main]
 
-# One run per workflow per branch, and supersede an older run ONLY on a pull request.
+# A pull request keeps ONE run per branch and supersedes the older one. A push to main gets a group
+# of its OWN, one per commit, so it never shares a queue with any other commit.
 #
 # Why the asymmetry, because getting it the other way round is the expensive mistake. On a pull request
 # the head commit is the only one anyone acts on, so an older run for a commit that has already been
 # pushed past is spending a runner on an answer nobody will read - cancelling it is right. On a push to
-# main, EVERY commit must keep its own verification: cancelling there would leave main commits
-# permanently unverified, and a green that belongs to a specific commit is the whole point of having CI
-# rather than a memory of a run.
+# main, EVERY commit must keep its own verification: a green that belongs to a specific commit is the
+# whole point of having continuous integration rather than a memory of a run.
 #
-# Added after a draft pull request queued six simultaneous runs on one branch, none of which superseded
-# any other, so the answer everybody was waiting for kept being pushed further out by the very commits
-# that were waiting on it. When a resource serialises, parallelism converts into latency rather than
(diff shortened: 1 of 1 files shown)


thefrederiksen:
Independent review raised one thing worth recording, which does not block the change but bounds the claim.

Excess runs beyond the account-wide concurrent job limit enter the hosted-runner queue. They are NOT in a shared workflow concurrency group, so no later merge has a pending slot to evict them from - they run and report once capacity frees. That is the property this change was after, and it holds.

The ceiling is throughput rather than any documented queue limit. If fifty-minute .NET jobs keep arriving faster than free slots retire them, the backlog and the time to a verdict grow without bound, and the queue can starve newer pull-request jobs and other repositories sharing the account. Each commit still gets its verdict eventually once a burst stops, but during a sustained burst the time to the LATEST verdict can be worse than the eviction behaviour this replaces.

So the trade is stated precisely: eviction gave a fast answer about the newest commit and no answer about any other; queuing gives every commit an answer, at the cost of latency under sustained merge pressure. That is the right way round for a backstop. If verdict latency does become the problem, the fix is to make the fifty-minute job shorter, not to start discarding runs again.

PR #2438, opened 2026-08-03
The local gate runs the installer tests, which nothing local ran at all

thefrederiksen:
## The gap

478 tests covering the installer ran in **no local invocation at all**. `tools/cc-director-setup.Tests` and `tools/cc-director-setup-engine.Tests` are not in `cc-director.sln`, so the single solution build never produced them and the run list never named them.

Continuous integration ran them as a separate step, which hid the gap for exactly as long as that job was being waited on. The moment local became the gate (#2437), the installer - the first screen a new user ever sees - had nothing behind it, and a release could ship it untested. Found by review, not by a failure.

## Why the default run, not `-Parked`

Parked means *too slow for the two-minute budget*. These are not slow, and calling them parked would have meant this coverage only ever ran at release time. Measured on this machine, Release configuration:

| Project | Tests | Test time | Own build |
|---|---|---|---|
| `cc-director-setup.Tests` | 25 | 1s | ~3s |
| `cc-director-setup-engine.Tests` | 453 | 6s | ~3s |

A full **cold** default run with both included stayed inside budget - the slowest suite is still `Gateway.UnitTests` at 56 seconds, and the two new ones finished in 5 and 8. They start alongside everything else, so the marginal wall-clock cost is close to zero.

## Why they get their own list

A solution build cannot reach them, so they are built individually before the run loop starts - the loop uses `--no-build` and would otherwise run against a stale or absent assembly. A build failure there is fatal, for the same reason a solution build failure is: **tests that never ran must not be reported as tests that passed.**

`-Gateway` is untouched. It means "only that one suite" and still means exactly that.

## What this buys

The release gate collapses from three commands to one: `.\scripts\test-local.ps1 -Parked -Configuration Release`. A gate that depends on remembering two extra commands is one that will eventually be run without them, and a release is the single place there is no fixing it forward. The documents that spell out the three commands can be simplified once this lands.

## Verification, including what it did NOT prove

- Cold full default run: both projects ran, 25 and 453 tests, `outcome=Completed` in the TRX for each.
- Inside budget, with the two extra builds measured rather than assumed.

**The run was not all green, and the failure is pre-existing.** `Gateway.UnitTests` failed with `InvalidOperationException: The collection has been marked as complete with regards to additions`, thrown from `FileLogWriter.Enqueue`. I ran the same default gate on **unmodified `origin/main`** and it failed the same suite with the same exception - 2 failures there, 1 here - and each run named *different* tests:

- `AuthMiddlewareTests.Mobile_shell_and_enroll_mounts_are_public_without_a_credential`
- `GatewayInputStatsAggregatorTests.GrowingSnapshot_AddsOnlyTheIncrease`
- `DeviceCredentialImportTests.Reimport_IsIdempotent_FindsMarker_AndImportsNothingNew`

Different tests, one exception, thrown from process-wide logging state. So this is not three bad tests; it is one shutdown race in `FileLogWriter` hitting whichever test happens to be running when the writer is completed. It is unrelated to this change - the added projects are separate processes and cannot touch that collection - but it means **the local gate is currently red on main**, which matters now that it is the only gate. Filed separately rather than fixed here.

### DIFF
--- a/scripts/test-local.ps1
+++ b/scripts/test-local.ps1
@@ -7,9 +7,15 @@
     THE DEFAULT RUN IS ABOUT TWO MINUTES, AND THAT IS THE POINT. Local is the gate for ordinary changes
     (issue #1156), so the gate has to be cheap enough that nobody is tempted to skip it.
 
-    WHAT THE DEFAULT RUNS: every suite that finishes inside the two-minute budget, roughly 3400 tests,
-    PLUS the two installer test projects (478 tests, about seven seconds). They start together and the
-    wall clock is the slowest of them, not the sum.
+    WHAT THE DEFAULT RUNS: every suite that finishes inside the two-minute budget, PLUS the two installer
+    test projects. They start together and the wall clock is the slowest of them, not the sum.
+
+    COUNTS, MEASURED 2026-08-03 FROM THE TRX FILES OF A COLD RUN - not estimated, and not copied forward:
+      Gateway.UnitTests 2777    Avalonia 348    Launcher 110    HostedAgent 88    Core.UnitTests 82
+      Engine 63                 Terminal 24
+      installer: setup.Tests 25, setup-engine.Tests 453
+    3492 in the default suites, 3970 including the installer. Re-measure before changing these numbers;
+    the per-project comments below were carried forward for months after they stopped being true.
 
     The installer projects live outside cc-director.sln and are built separately here. They are in the
     default run because they are fast and because the thing they cover - the first screen a new user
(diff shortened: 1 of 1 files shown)


thefrederiksen:
Merged as 7be297665 on main. The squash merge itself succeeded but its API response failed with a GraphQL error (ref E22E:35B9DC:10F9897F:3B85BC58:6A7093EE), so this pull request was left showing OPEN with no merge recorded.

Verified before closing, rather than assumed from the merge command's output: origin/main carries the squash commit titled for this pull request, and `git show origin/main:scripts/test-local.ps1` contains the new installerProjects wiring. Closing manually and deleting the branch.

PR #2449, opened 2026-08-04
Show which model each session is running, on every surface that names a session

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1340.

## What this shows, and where

The model a session is actually running was already on the wire and was visible to a human in exactly one place: the Cockpit's History page, for sessions that had already ended. While a session was alive, nothing on any screen said which model was doing the work.

It now appears in six places:

- **Director session rail** - folded into the agent badge: `Claude Code | fable-5`
- **Director selected-session header** - the same badge beside the session name
- **Cockpit Fleet Map card** - a chip on the meta row, beside the agent chip
- **Cockpit Fleet Map** - a new **By model** pivot, so the fleet lays out by model the way it lays out by agent
- **Cockpit session roster row**, and **the open session's header**
- **`cc-devthrottle session list`** - a MODEL column

## The design decision worth reading

The issue said a client could read `driverCapabilities` and work out the two absences for itself, and that no new field was needed. This does not do that.

The interesting part is not the model, it is the two absences, and they mean opposite things:

- the session **can** report its model and has not finished a turn - it is coming;
- the agent **cannot** report one at all (Gemini records it nowhere readable, Cursor was never live-verified) - it is never coming.

Five surfaces each deciding that for themselves is five chances to render the hopeful wording for the hopeless case, which is the same defect shape as the Voice screen's "Generate narration now" button that could never succeed. So `ModelDisplayFold` rules it once and every surface renders the finished words.

The Gateway stamps the verdict onto `SessionDto.ModelDisplay` beside `EffectiveColor`, for the browser clients. The **desktop** calls the same pure function over its own locally-mapped session. That is not the rail ruling for itself: both inputs - the recorded model and the driver's capability list - are facts the Director produces firsthand, so the answer is identical and the badge still reads correctly with no Gateway attached. Contrast the session colour, which reads Gateway-only inputs and is deliberately never folded on the desktop.

## Two smaller calls

**The rail folds the model INTO the agent badge** rather than adding a second one (owner's choice between three mocked options). They are one identity, and the rail already carries a role glyph, a not-delivered alarm and a winding-down marker.

**`Session.OnCurrentModelChanged` is new.** The model is re-read at turn-end, and a mid-session `/model` switch raises none of the events the rail already listens to. Without its own signal the badge would keep its old value until something unrelated repainted the row - which reads as deliberate, and is therefore believed. Same class as the role-stamp defect.

**The command line gets one new column, not two.** An AGENT column beside it was built and removed: Rich lays that table out at 80 columns when stdout is a pipe, which is how every agent reads it, and the eighth column pushed STATUS far enough to elide `Exited (crashed)` - the fact issue #1019 exists to keep readable. A test now pins that budget.

## Tests

- `ModelDisplayFoldTests` - 11 cases, including that the two absences never share a string, and that a shortened badge always keeps the full id in the tooltip
- `SessionRailModelBadgeTests` - 4 desktop cases: the pre-first-turn wording, the shortened badge with the full id in the tooltip, a mid-session switch reaching the rail, and a failed read leaving the last known model standing
- `fleetMapFormat.test.ts` - +10: when the chip appears, and the model pivot's lane keys (both absences keep separate lanes; an unstamped session gets a third)
- `model.test.ts` - 5 cases on the shared reader, all of them that it composes nothing
- `test_session_list_model.py` - 6 cases, including the 80-column budget

## Gate

- `.\scripts\test-local.ps1`: **green**, 9 projects, 4,014 tests.
- Web: typecheck clean across all three workspaces; client-core 877 and cockpit 267 green. The two new test files were run individually to prove they execute rather than inferring it from a total.
- Python: green.
- `-Parked`: `Core.Tests` green (4,202). `Gateway.Tests` reported **one** failure - `HostedMissionPartitionWiringTests.An_unattributed_mission_is_quarantined_on_hosted_and_stays_on_disk`, a `POST /missions` answering 500. See below.

## About that one Gateway failure

I could not clear it, and I am not claiming it is unrelated on the strength of a hunch.

What is known: this change touches the `/sessions` fold and adds a DTO property, and goes nowhere near mission creation. Of the Gateway.Tests results kept on this machine, nine runs across six unrelated worktrees have failed this suite, each with a different set of tests and no repeats - including two other mission-family failures in worktrees whose changes had nothing to do with missions. That is the known instability of this host-bound suite.

What is NOT known: a re-run of that one class on this branch never executed. It waited 632 seconds for the suite's machine-wide lock, held by another worktree, and was killed before it ran a single test. So the evidence is a pattern, not a verdict on this diff. Anyone re-running that class when the lock is free will settle it in seconds.

### DIFF
--- a/apps/cockpit/src/fleet/FleetMapView.tsx
+++ b/apps/cockpit/src/fleet/FleetMapView.tsx
@@ -24,6 +24,8 @@ import {
   directorsByMachine,
   groupByDirector,
   machineKeyOf,
+  modelChip,
+  modelKeyOf,
   shortDir,
   type DirectorGroup as FormatDirectorGroup,
 } from "./fleetMapFormat";
@@ -57,7 +59,7 @@ const ReachabilityContext = createContext<DirectorReachability[]>([]);
 // fleet page reads that one store, this map and the Sessions roster always agree on the fleet at the
 // same moment, and a hidden tab makes no roster requests at all.
 
-type Pivot = "machine" | "director" | "repo" | "worktree" | "agent" | "list" | "mission";
+type Pivot = "machine" | "director" | "repo" | "worktree" | "agent" | "model" | "list" | "mission";
 
 const PIVOTS: ReadonlyArray<{ key: Pivot; label: string; kindLabel: string }> = [
   { key: "machine", label: "By machine", kindLabel: "Machine" },
@@ -73,6 +75,13 @@ const PIVOTS: ReadonlyArray<{ key: Pivot; label: string; kindLabel: string }> = 
   { key: "repo", label: "By repository", kindLabel: "Repository" },
   { key: "worktree", label: "By working tree", kindLabel: "Working tree" },
   { key: "agent", label: "By agent", kindLabel: "Agent" },
+  // "By model" (issue devthrottle_internal#1340): the fleet laid out by the model each session is actually
+  // running, the same way it can be laid out by agent. It sits beside "By agent" because it answers the
+  // question one step down - not "which tool", but "which brain in that tool" - and that is the one the
+  // fleet's cost and quality actually turn on. The lane title is the model id exactly as the agent's own
+  // records spell it; the two absences ("no model yet", "model not reported") keep separate lanes rather
+  // than pooling, because pooling them in the layout would undo the distinction the fold exists to make.
+  { key: "model", label: "By model", kindLabel: "Model" },
   // The flat "Fleet list" pivot (issue #1212) absorbs the retired Fleet page: every session once, as
   // the same node card, no lane grouping (machines are already their own pivot).
   { key: "list", label: "Fleet list", kindLabel: "Session" },
@@ -84,7 +93,7 @@ const PIVOTS: ReadonlyArray<{ key: Pivot; label: string; kindLabel: string }> = 
 
 // The pivots that lay the fleet out on the node canvas (root -> lanes). "list" is a flat grid and
 // "mission" is its own board; neither uses the canvas or the title search.
-const CANVAS_PIVOTS: ReadonlySet<Pivot> = new Set<Pivot>(["machine", "director", "repo", "worktree", "agent"]);
+const CANVAS_PIVOTS: ReadonlySet<Pivot> = new Set<Pivot>(["machine", "director", "repo", "worktree", "agent", "model"]);
 
 // The grouping is sticky per browser: the map opens on whatever the user last chose (rather than a
 // "smart" default that guesses from the fleet shape), so returning to the map lands them exactly where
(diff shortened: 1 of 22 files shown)


thefrederiksen:
## The Gateway failure is settled - it was luck, not this change

The parked run reported one failure (`HostedMissionPartitionWiringTests.An_unattributed_mission_is_quarantined_on_hosted_and_stays_on_disk`, a `POST /missions` answering 500). The pull request body above says I could not clear it. I can now.

Against this exact commit (`737e0a34`, clean working tree), with the suite's machine-wide lock free:

- that class alone: **3 passed, 0 failed**
- the **whole Gateway suite**: **2,456 passed, 0 failed, 47 skipped, 2,503 total** (41 minutes)

So the parked coverage gap is closed on both suites - `Core.Tests` 4,202 passed, `Gateway.Tests` 2,503 with nothing failing - and the earlier red does not reproduce.

Recorded rather than deleted, because the pattern is the useful part: nine Gateway.Tests runs on this machine, across six unrelated worktrees, have each failed a different set of tests with no repeats. A single red from this suite is not evidence about the change under it; it has to be judged against a re-run.

thefrederiksen:
## Cross-family review (Codex, gpt-5.6-sol) - three findings, one fixed here

Codex could not run a single shell command on this machine (`codex-windows-sandbox-setup.exe` is missing, so every exec dies before it starts), so it was given the diff and the sibling files inline. It reviewed the fold, the invalidation path and the absent states.

**Finding 1 - accepted and fixed in f5e9034.** The "no model yet" tooltip read "this session has not completed a turn". That names a cause nobody observed: a null model is equally consistent with a read that could not be taken - torn records, the agent restarting, a restored session - and `SessionRecordsWatcher` keeps the last known value rather than clearing on those. It now states the fact and the mechanism only. The test asserts the old claim is ABSENT as well as the new words present.

**Finding 2 - the desktop folds locally.** Codex confirmed `ControlEndpoints.Map` populates both inputs the Gateway fold reads (`CurrentModel`, `DriverCapabilities`), so identical snapshots give identical answers, and called the split client-side ruling anyway.

Keeping it, deliberately. The desktop reads no model stamp at all, so there is nothing for it to contradict; it renders its own fresher copy of a fact this Director produced firsthand. The alternative is a badge that goes blank whenever the tunnel is down - on the one surface that can still tell the truth with no Gateway to ask, which is the same "no Gateway, no fold" floor the rail already honours for colour. What the two surfaces can differ by is STALENESS (the Cockpit's stamp is a poll old), not verdict.

**Finding 3 - the handler is never unsubscribed.** True, and pre-existing: `SessionViewModel` subscribes to **19** session events, unsubscribes from none, and is not `IDisposable`. Mine is the 20th of an established pattern. Fixing one of twenty would not fix the leak and would hide it; the lifetime of that whole class wants its own issue, which I am filing rather than smuggling into a display change.

Codex also confirmed the invalidation path is complete (all four fold readers raised) and the threading correct (`SetCurrentModel` runs under `Task.Run`, the handler posts to the UI thread).

Gate re-run after the fix: local gate green, 9 projects, 4,014 tests; client-core model tests green.

PR #2490, opened 2026-08-07
workflow run accepts the short run id it prints, and a web app answer is never parsed as JSON

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1790

## What was wrong

Running 'cc-devthrottle workflow run' with a short prefix of a run id crashed with an unhandled JSONDecodeError traceback. Two defects stacked up:

1. The Gateway's run route only matches a full id ('/gateway/workflow-runs/{id:guid}'), so a short prefix matched no endpoint, fell through to the web app fallback, and came back HTTP 200 with the app shell as text/html - never a 404.
2. The workflow client parsed any 2xx body with a bare resp.json(), so that HTML became a raw traceback instead of a sentence.

The short form is the tool's own advertised input: 'workflow runs' prints run ids truncated to eight characters and then says 'Details with: cc-devthrottle workflow run <run id>'.

## The fix, root cause in the shared spot

The same unguarded resp.json() on success bodies existed in five clients (workflow, mission, schedule, email, diagnostics), so the guard now lives once in cc_shared.gateway.parse_json_body: an empty body is an empty object, a non-JSON content type is refused as the web app fallthrough in one clear sentence, and a body labelled JSON that cannot be parsed is refused too. It raises the caller's own error class because the email and diagnostics clients keep their own GatewayError types, and raising the shared class inside them would fly past their handlers. The skill client already carried its own content-type guard with skill-specific wording pinned by its tests, and is unchanged.

On top of that, 'workflow run' now resolves a short prefix against the run list the same way session commands resolve a short session id. An unknown prefix gets 'No workflow run matches ...', an ambiguous one names the candidate ids, both exit 1 in one sentence with no traceback. The workflow instructions route also refuses a text/html answer instead of printing the web app page into an agent's context as conduct.

## What was verified

- The tool's own Python test suites, which the default local gate does not run, were run directly: 354 passed (tools/cc-devthrottle/tests plus tools/cc_shared/tests), including eleven new regression tests for the prefix resolution and the fallthrough guard.
- Live against the real Gateway from this worktree: 'workflow run 3990e87c' (the exact command from the issue) now prints the run and exits 0; 'workflow run deadbeef' prints 'No workflow run matches' and exits 1. Before this change the short prefix was the recorded traceback.
- The repository gate (scripts/test-local.ps1) ran; every suite passed except CcDirector.Gateway.UnitTests, where all 8 failures are one test class (HostedSchemaRefusesAnUnownedRowTests) failing with 'Npgsql.NpgsqlException: Failed to connect to 127.0.0.1:55432'. That port is the disposable Docker Postgres those tests need, and Docker's engine is returning errors on this machine right now, so that class cannot pass here on any commit. This change touches only Python files under tools/, so the C# under test is bit-identical to the parent commit e50fa8265 - the red is the machine, not the change.
- The gate also flagged the parked CcDirector.Core.Tests and CcDirector.Gateway.Tests as a coverage gap. They are C# suites and do not execute the Python tool this change modifies; the tool's own suites above are the coverage that matters here, and they were run.

### DIFF
--- a/tools/cc-devthrottle/src/diag_ops.py
+++ b/tools/cc-devthrottle/src/diag_ops.py
@@ -29,6 +29,7 @@ _tools_dir = str(Path(__file__).resolve().parent.parent.parent)
 if _tools_dir not in sys.path:
     sys.path.insert(0, _tools_dir)
 
+from cc_shared import gateway  # noqa: E402
 from cc_shared.config import CCDirectorConfig  # noqa: E402
 
 console = Console()
@@ -102,7 +103,10 @@ class DiagClient:
                 f"Gateway at {self.base_url} did not respond within {TIMEOUT_SECONDS}s."
             ) from exc
         if 200 <= resp.status_code < 300:
-            return resp.json() if resp.content else {}
+            # The shared guard, not a bare resp.json(): a request no endpoint matches falls
+            # through to the Gateway's web app and answers HTTP 200 with text/html (issue #2486).
+            # This module's own GatewayError travels along so the handlers here still catch it.
+            return gateway.parse_json_body(resp, self.base_url, GatewayError)
         raise GatewayError(_gateway_message(resp))
 
     def network(self) -> Dict[str, Any]:
--- a/tools/cc-devthrottle/src/email_ops.py
+++ b/tools/cc-devthrottle/src/email_ops.py
@@ -28,6 +28,7 @@ _tools_dir = str(Path(__file__).resolve().parent.parent.parent)
 if _tools_dir not in sys.path:
     sys.path.insert(0, _tools_dir)
 
+from cc_shared import gateway  # noqa: E402
 from cc_shared.config import CCDirectorConfig  # noqa: E402
 
 LOOPBACK_DEFAULT = "http://127.0.0.1:7878"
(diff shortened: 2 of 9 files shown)


thefrederiksen:
Review round (from the reviewing session): resolution originally matched against the default runs page of 200 newest rows while runs are retained forever, so a matching run beyond the page was invisible and a prefix could silently select the visible one.

Fixed in commit 966e945c1. The Gateway offers no server-side prefix query and no offset paging, only a limit clamped to the store's ceiling of 1000 (WorkflowRunStore.MaxListLimit, in place since the runs endpoint was born). Resolution now requests exactly that ceiling, so the answer proves its own completeness: fewer rows back means the whole history was seen; exactly 1000 back means possible truncation, in which case a lone match or a miss is refused with a clear sentence pointing at the full id rather than picked. Two or more matches keep the specific ambiguity error regardless of the boundary.

Verified: four new boundary regression tests (lone match refused on a possibly-truncated page, a miss there not called unknown, ambiguity preserved, and the requested page size provably riding the query string); both Python suites re-run, 359 passed; live short-prefix resolution against the real Gateway re-verified.

PR #2491, opened 2026-08-07
Arm the dictation moved-on guard from both Speak flows

thefrederiksen:
Closes 2478.

## What was wrong

The Gateway's session-moved-on guard for dictation delivery (GatewayDictationEndpoint, the resumed-clip check) only engages when the request carries a baselineBufferBytes above zero. Both taught client flows - the mobile SessionControls and the Cockpit SessionComposer - omitted the field when handing a recording-stage Send to the durable background pipeline, so it defaulted to zero and the guard never armed. The recovery behavior the feature was built for (dropping a stale resumed clip instead of injecting it into a session that has moved on, with the words offered back) was unreachable from the shipped Speak flows. The Voice screen flow already sent the value; only the two Speak flows did not.

## The root-cause decision: fix the clients, not a Gateway default

Only the client knows the session's terminal-byte position at record time. A Gateway-side default would have to be stamped when the upload first reaches the Gateway - and for exactly the case the guard exists for (a clip whose first upload attempt never arrived, resumed later from the durable on-device copy), that first contact happens at resume time, so the stamped baseline would equal the already-moved-on position and the guard could never fire. Passing the real record-time baseline from the clients arms the guard in every case, changes nothing for callers that already send a value (the Voice screen), and touches no Gateway code. No new fallback logic: a failed roster read yields the pipeline's documented unknown state (field omitted, guard skipped for safety) and the dictation always still delivers.

## What changed

- packages/client-core/src/dictation/baseline.ts (new): the one shared snapshot both flows use - reads the roster (the same listSessions reading useVoiceMode has always used), returns the selected session's totalBufferBytes, or unknown when the roster cannot answer.
- apps/mobile/src/components/SessionControls.tsx and apps/cockpit/src/sessions/SessionComposer.tsx: the Speak press snapshots the baseline (alongside the existing caret snapshot) and the recording-stage Send passes it to backgroundTranscribeAndSend.
- Regression coverage proving the guard now arms from each flow:
  - apps/cockpit/src/sessions/composerSpeakSend.test.tsx now asserts the pipeline receives the snapshotted baseline (4321, above zero) for the Cockpit flow.
  - apps/mobile/src/components/sessionControlsSpeakSend.test.tsx (new) pins the same for the mobile flow, plus the roster-failure case (unknown baseline, clip still delivers). This is the mobile workspace's first test file, so the workspace gains the cockpit's exact test setup (vitest, jsdom, testing-library, and a vitest-only config so tests never load the progressive-web-app build plugin).
  - packages/client-core/src/dictation/baseline.test.ts (new): the snapshot helper and hook, including the string form of the 64-bit integer, the unknown states, and the two-presses token guard.
- .github/workflows/ci.yml: the web job runs the mobile tests, per its own add-a-line-when-a-workspace-gains-tests rule.

The chain to the Gateway is already pinned by existing tests: backgroundSend threads the stored baseline onto the upload (backgroundSend.test.ts), and the Gateway guard's behavior for a positive baseline is covered by the existing Gateway dictation suites.

## What was verified

- Web workspaces (the default gate runs no web tests): client-core 890 passed (83 files, includes the 8 new baseline tests), cockpit 267 passed (31 files, includes the extended composer test), mobile 2 passed (the new file). Workspace typecheck green; eslint clean on every changed file.
- Local gate scripts/test-local.ps1: every suite passed except CcDirector.Gateway.UnitTests, where exactly 8 tests failed - all of them HostedSchemaRefusesAnUnownedRowTests, failing with "Npgsql.NpgsqlException: Failed to connect to 127.0.0.1:55432". That is the Postgres-backed hosted-schema suite, red on this machine for environmental Docker reasons on unchanged code (this change contains no C# at all); judged against the parent rather than chased. 2973 of 2983 Gateway unit tests passed, and all eight other projects were fully green (153, 364, 63, 88, 113, 24, 25, 456).
- The gate flagged the parked suites (CcDirector.Core.Tests, CcDirector.Gateway.Tests) as a coverage gap; not run, because this change touches no product C# - it is entirely web workspace code plus one continuous-integration step, and the web tests were run directly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,9 @@ jobs:
       - name: Run Cockpit tests
         run: npm test --workspace @devthrottle/cockpit
 
+      - name: Run mobile tests
+        run: npm test --workspace @devthrottle/mobile
+
       # The shells must still BUILD, not just typecheck: the production build runs the real Vite
       # pipeline (bundling, the PWA service worker) and fails on things tsc never sees.
       - name: Build the mobile PWA
--- a/apps/cockpit/src/sessions/SessionComposer.tsx
+++ b/apps/cockpit/src/sessions/SessionComposer.tsx
@@ -322,7 +322,9 @@ export function SessionComposer({ sessionId, value, onChange, onQueued, focusHan
         // Insert the dictated words at the snapshotted caret inside the typed text: the Gateway
         // submits before + dictation + after. The caret splits the typed text into the two halves.
         composeParts: { before: composerText.slice(0, caret), after: composerText.slice(caret) },
-        // The terminal-byte position snapshotted at Speak press, for the moved-on guard (issue #2478).
+        // The terminal-byte position snapshot the Speak press started, for the moved-on guard
+        // (issue #2478). A promise: the pipeline awaits it before persisting, so a quick Send
+        // cannot outrun the roster read.
         baselineBufferBytes: baseline.read(),
         // Restore the typed text (ahead of anything typed since - valueRef reads the box as it is at
         // failure time, this callback's own `value` is a stale snapshot by then) so Send never
--- a/apps/cockpit/src/sessions/SessionComposer.tsx
+++ b/apps/cockpit/src/sessions/SessionComposer.tsx
@@ -11,6 +11,7 @@ import { describeAndReport } from "@devthrottle/client-core/errors/reportClientE
 import { DictationDialog } from "@devthrottle/client-core/dictation/DictationDialog";
 import { DictationStatusStrip } from "@devthrottle/client-core/dictation/DictationStatusStrip";
 import { backgroundTranscribeAndSend, type CapturedUtterance } from "@devthrottle/client-core/dictation/backgroundSend";
+import { useDictationBaseline } from "@devthrottle/client-core/dictation/baseline";
 import { insertAt, joinText } from "@devthrottle/client-core/dictation/transcript";
 
 // The composer (issue #972, completed in issue #1210) - the React port of the Blazor Cockpit composer.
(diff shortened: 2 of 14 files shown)


thefrederiksen:
Review defect addressed in 70951bb33. The snapshot is now handed to the pipeline as a promise and awaited before the durable record is persisted, so a quick Send waits for the roster answer (bounded by the read's own timeout) instead of racing it. Unknown and zero are now distinct end to end: a genuinely unknowable baseline (roster failed even after one retry) is persisted as unknown and omitted from the wire request - never written as zero - while a genuine zero reading passes through as valid. The dropped-clip fresh-id retry clears its deliberately unguarded baseline to unknown instead of zero for the same reason. New tests cover the quick-Send race with a roster that answers only after Send (mobile flow end to end, plus the pipeline holding until the snapshot resolves), the unknown-never-zero persistence, the genuine-zero reading, the snapshot retry, and a rejecting foreign promise costing the guard but never the words. All web suites green after the change: client-core 897, cockpit 267, mobile 3; typecheck and lint clean.

thefrederiksen:
Round two addressed in 1fa08380a. The retry is gone: the press-time roster read is the only valid reading - a call made after the press can capture bytes produced during or after the recording, over-stating the baseline and masking the very movement the guard detects - so a failed press-time read is final (unknown, field omitted on the wire, guard skipped) and the words are never refused. And durability no longer waits on the snapshot: the clip is saved immediately with the baseline unknown, then the pipeline waits for the ORIGINAL press-time promise and enriches the pending record before the first upload when it resolves; a plain-number baseline (the Voice screen) still rides the first write unchanged. Tests pin both: exactly one roster call on failure (helper unit test and mobile flow end to end), and save-at-once with unknown, upload holding for the press-time answer, same record enriched - with unknown taking exactly one durable write and no roster call of the pipeline's own. All web suites green: client-core 897, cockpit 267, mobile 3; typecheck and lint clean.

thefrederiksen:
Round three addressed in fae04dab6. The upload id is now reserved from the kick machinery for the whole save-and-enrich window, using the same in-flight set every drive already honors: a kick (app load, online, foreground) that lists the freshly saved unknown record inside the window no-ops on exactly this id, so there is no unguarded upload and no race against the enrichment write or the original drive. The reservation is released with no await before the original drive synchronously re-takes the in-flight entry, so nothing can slip into the gap - and it lives only in memory, deliberately, so a crash or reload drops it and the resume path still drives the durable unknown record. Tests pin both sides: a kick landing between the first save and promise resolution is proven a no-op with exactly one upload happening (from the original flow, enriched, not resumed), and a fresh process resumes the durable unknown record unguarded. All web suites green: client-core 899, cockpit 267, mobile 3; typecheck and lint clean.

PR #2497, opened 2026-08-07
Delete the dead speech-to-text bias path and record the ruling at the deletion site

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1785

## The ruling this carries out

Quoting the owner ruling on the issue, 2026-08-07:

> the bias path stays dead - and gets deleted. Never send vocabulary or any steering hints to the speech-to-text provider. The dictation architecture is deliberately two strict stages: transcribe the raw audio as faithfully as possible (audio only, no hints), then a separate code-driven correction pass replaces the listed words on the finished transcript. Priming the transcriber makes it steer toward suggested words, changing wording and sentence structure - corrupting the record of what was said. Meaning preservation is paramount; this was learned from real problems before. The unused prompt-building code is a rejected approach, not a missing feature: the fix for this issue is to DELETE the dead path and leave a comment recording this ruling at the deletion site, so it cannot be helpfully wired up later. The docs already describe the shipped cleanup-only truth and need no change.

## Verified before deleting anything, at origin/main

- `DictionaryLoader.BuildSttPrompt` had **no callers in any shipping code**. The only references anywhere were its own two tests and a historical problem document.
- The live speech-to-text request - the one batch POST to `/audio/transcriptions` in `BatchTranscriptionPipeline` - sends `model`, `response_format` and `language`, and **no prompt field**. Nothing was reaching the transcriber, so removing the method changes no behaviour.
- The vocabulary itself is **live and untouched**: `CleanupOrchestrator`, `FuzzyDictionaryMatcher`, `TranscriptEditEngine` and `MistranscriptionMiner` all read it. That is the correction pass working on the finished transcript, which is exactly the stage the ruling keeps.

## What changed

- Deleted `DictionaryLoader.BuildSttPrompt` and its two tests.
- Left a comment at each deletion site recording the ruling in substance - transcribe raw with audio only, substitute the listed words afterwards, priming corrupts the record of what was said, this is a rejected approach and not a missing feature.
- Corrected two stale comments that stated the rejected approach **as the design**, which is what would invite someone to rebuild it:
  - the `DictationDictionary` summary claimed the vocabulary was "Packed into the OpenAI prompt parameter as a soft decode-time bias";
  - the 2026-05-25 problem document `voice-dictionary-not-applied-on-mobile.md`, whose suggested fix told a future reader to recompute that prompt per transcription. Its record is left intact; a dated note at the top marks the bias-prompt parts as rejected and points at this issue. The class it describes, `OpenAiRecordingTranscriber`, no longer exists either.

Behaviour is unchanged. This is a deletion plus comments.

**One deliberate non-change:** `docs/features/dictation/phase0/run_phase0.py` still defines an `STT_PROMPT`. It is a dated phase-0 experiment script under `docs`, on no build path and called by nothing, and the ruling says the docs need no change - so the record of an experiment that already ran is left as it is rather than edited after the fact.

## Tests

- `.\scripts\test-local.ps1` - all nine default suites green, 4265 tests, every TRX `outcome=Completed`.
- That run then flagged a coverage gap: `CcDirector.Core.Tests` is parked and holds the test file edited here. So the parked suites were run too - `.\scripts\test-local.ps1 -Parked -Filter "FullyQualifiedName~Dictation"`: `CcDirector.Core.Tests` 159 passed, `CcDirector.Gateway.Tests` 29 passed, both `outcome=Completed`. Checked in the TRX that 15 `DictionaryLoaderTests` actually executed and that neither `BuildSttPrompt` test remains.
- **What was not run:** the two parked suites in full, unfiltered. `Core.Tests` needs 11 to 33 minutes and the machine is busy with fleet work. The filter covers every test touching the changed code, and the full build of both projects passed with zero warnings, so a compile-level break would have shown. Stating it rather than implying the whole parked gate ran.

### DIFF
--- a/apps/cockpit/src/dictionary/DictionaryView.tsx
+++ b/apps/cockpit/src/dictionary/DictionaryView.tsx
@@ -23,8 +23,10 @@ import { gatewayErrorMessage } from "@devthrottle/client-core/api/client";
 import { ConfirmDialog } from "../components";
 
 // The dictation Dictionary editor (issue #977, epic #967) - the React port of the Blazor Cockpit
-// Dictionary.razor (#183). The human edits the vocabulary chips biased into speech-to-text and the
-// common-mistranscriptions term-to-variant map fed to the cleanup pass, with a dirty-state Save that
+// Dictionary.razor (#183). The human edits the vocabulary chips and the common-mistranscriptions
+// term-to-variant map. BOTH are fed to the cleanup pass and NEITHER is ever sent to the
+// speech-to-text provider: the transcriber gets audio only, and the listed words are substituted
+// afterwards on the finished transcript (issue 2481). Edited with a dirty-state Save that
 // PUTs the whole glossary and re-renders the returned dictionary. It reads and writes same-origin
 // through the Gateway front door (client-core) - never a Director address.
 //
@@ -567,7 +569,7 @@ export function DictionaryView() {
             {/* ---- Vocabulary ---- */}
             <div className="dc-section">
               <h2>Vocabulary</h2>
-              <p className="dc-hint">Terms biased into speech-to-text.</p>
+              <p className="dc-hint">Terms corrected to this spelling after transcription.</p>
               <div className="dc-chips">
                 {dict.vocabulary.map((term, i) => (
                   <span className="dc-chip" key={`${term}/${i}`}>
--- a/docs/architecture/transcription-quality-loop.md
+++ b/docs/architecture/transcription-quality-loop.md
@@ -101,9 +101,17 @@ Findings (full report saved separately). Our approach is the standard one; we ar
     Metaphone. Add per-term precomputed IPA rather than a runtime dependency if possible.
   - Borrow Microsoft.PhoneticMatching's `EnHybridDistance` design (phonetic blended with edit distance)
     as the scoring template, swapping its English engine for the multilingual one.
-- **Orthogonal wins:** use the transcription provider's own biasing channel (Whisper `initial_prompt`,
-  Deepgram keywords, AssemblyAI word_boost, Speechmatics `sounds_like`) so fewer errors reach cleanup;
-  and consider an optional per-term `sounds_like`/pronunciation field (every serious product offers it).
+- **RULED OUT - the provider's own biasing channel.** This bullet used to propose exactly that
+  (Whisper `initial_prompt`, Deepgram keywords, AssemblyAI word_boost, Speechmatics `sounds_like`,
+  plus a per-term pronunciation field) so fewer errors would reach cleanup. The owner ruled against
+  it on 2026-08-07 (issue 2481) and the dead code for it has been deleted. **Do not propose it
+  again.** Nothing - no vocabulary, no keyword list, no pronunciation hint - is ever sent to the
+  speech-to-text provider; it gets audio only. Priming the transcriber makes it steer toward the
+  suggested words, changing wording and sentence structure rather than just spelling, and that
+  corrupts the record of what was actually said. Meaning preservation outranks the error rate: a
+  faithful transcript with a wrong spelling beats a fluent one with altered meaning, and the wrong
+  spelling is what the cleanup pass exists to fix. This constrains the whole section - every recall
+  improvement has to come from the cleanup matcher, which is why that is where the work goes.
 
 ## 6. Continuous improvement + always-deploy-best - PROCESS
 
--- a/docs/architecture/transcription-quality-loop.md
+++ b/docs/architecture/transcription-quality-loop.md
@@ -47,7 +47,9 @@ We already have `tools/harnesses/transcription-eval-harness/` (audio -> transcri
 recall) and 5 recorded English fixtures. Transcription is now third-party, so the new harness focuses on
 the DICTIONARY CLEANUP step, holds transcription constant, and is multilingual. The design is lifted from
 established ASR-customization / contextual-biasing methodology (GenSEC/HyPoradise, LibriSpeech biasing),
-NOT invented.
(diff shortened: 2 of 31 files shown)


thefrederiksen:
## Rounds two and three: the deletion did not delete the design

The inference test kept failing after the method was gone, because the rejected approach was still **asserted as current behaviour** in live code and still **recommended as future work**. Three classes, handled three different ways per the owner's ruling.

### Active code, corrected to what ships

| Where | What it claimed |
|---|---|
| `RecordingEndpoints` agent-facing API help | the dictionary "biases transcription toward the user's terms" - the worst copy in the tree, because an agent reads it as the current contract |
| `RecordingEndpoints` x3 (tenancy comments) | the transcriber reads the tenant's glossary; a per-tenant edit biases that tenant's transcription |
| `TenantGlossary` | a term added on one path "biases every other path" |
| `DictionarySuggestionScreen` | told its screening **model**, in the prompt, that "the dictionary biases a speech model" |
| `AgentOptions` | a missing file means "no vocabulary bias" |
| Cockpit `DictionaryView`, client-core `dictionaryClient` / `dictionaryEdits`, phone `DictionaryPage` + `GatewayClient` | the vocabulary is "biased into speech-to-text" |
| ControlApi `VoiceUtteranceRequests` | true ("bias the transcript *cleanup*"), but made unambiguous rather than left to be re-flagged |

**Two of these were user-visible strings, not comments.** The Cockpit and the phone both told the user "Terms biased into speech-to-text". They now read "Terms corrected to this spelling after transcription", which is what the cleanup pass does.

The tenancy meaning is untouched - the glossary really is partitioned per tenant. Only the mechanism is corrected: the cleanup pass reads it, not the transcriber.

### Live architecture document, recommendation removed

`docs/architecture/transcription-quality-loop.md` is what `STATUS.md` points at as the current design, and its "Orthogonal wins" bullet actively proposed using the provider's own biasing channel - Whisper `initial_prompt`, Deepgram keywords, AssemblyAI `word_boost`, Speechmatics `sounds_like`. A live document recommending the banned approach is a bigger way back in than any dead method, so that bullet is now the ruling plus a do-not-propose-it-again.

### Historical documents, annotated at the claim and NOT rewritten

Dictation `STATUS` and `PLAN`, the dictation goal, the phone-recorder plan, the voice dictionary page plan, the transcription-speed research, the 2026-05-25 problem document, and the 2026-07-24 suggestions mockup. Each carries a short dated correction note **where the claim sits** - the mockup at both of its two claim sites, not just at the top, because a mockup gets copied section by section.

### Phase-0 script, tombstoned rather than deleted

Being off the build path is not safety while the file is executable and its own docstring says it verifies the design. It now carries a header naming the ruling and exits 2 immediately if run. Verified: it parses (`py_compile`), and running it prints the ruling and exits 2 - before any import or network call.

### Verified rather than pattern-matched

The glossary reaches `GatewayTranscriptionService` only as `dictionaryProvider`, consumed at the correction step; the transcription POST sends model, response format and language, and no prompt. Checked that no test pinned the screening-prompt sentence before changing it.

**Left alone because they are true:** the language hint in `languages.ts` / `voiceTestClient.ts` (the language parameter really is sent, and it is not a word-steering hint), and the two statistical uses of "bias" in `SessionHistoryStore` and `GatewayStatsStore`.

### Tests

- Web, all three workspaces: client-core 905, cockpit 274 (including `DictionaryView.test.tsx`), mobile 3. All green. No test asserted either changed string.
- .NET default gate: green, except one failure in `CcDirector.Avalonia.Tests` - `SessionRailStateTests.UncommittedBadge_CountChanges_RaisesTheRailsRepaint`, `System.InvalidOperationException: Call from invalid thread` in the Avalonia dispatcher. **Judged against the parent, not absolutely:** it passes 19/19 when re-run on the same commit, the suite passed 364/364 twice earlier on this branch, and the diff touches no Avalonia file. Flaky dispatcher race, not this change.
- Parked suites were run in round one (`Core.Tests` 159, `Gateway.Tests` 29, Dictation filter) and not re-run since; the later rounds add comments and prompt text, no logic.

thefrederiksen:
Cross-family review verdict: **PASS** at `1760cca47`.

I found no outstanding defect against the issue thefrederiksen/devthrottle_internal#1785 owner ruling: DevThrottle sends audio only to speech-to-text, and dictionary terms are applied afterward to the finished transcript. The dead `DictionaryLoader.BuildSttPrompt` path and its tests are gone; live request construction remains raw transcription only; active product copy, model-facing text, comments, plans, research, historical reports, and mockups now either state the ruling or visibly tombstone the rejected design.

Final verification:
- `git diff --check origin/main...1760cca47` is clean.
- `python scripts/sweep-bias-claims.py` exits 0 with zero unannotated matches and prints its limitations on every clean run.
- `python scripts/sweep-bias-claims.py --self-test` passes all named positive/negative cases, including the two negation-endorsement shapes, and finds the exact original claims after whole annotation blocks are removed.
- The detector's stated limits are honest: it explicitly says it is a net, not proof; calls out unknown paraphrases, cross-line claims, and visual/table implications; records that review stopped by instruction while narrower holes were still being found; and leaves human review as the backstop.
- The new code-shape rule introduces no match in the current repository outside the deliberately annotated/tombstoned material. Its token choice avoids bare C# `Dictionary<>`, and the generic provider-map negative case is pinned.

Test evidence reviewed: the established branch run had 4,265 passing tests and 2 skips, with targeted dictionary/reachability runs green; on the final head, Docs Drift, web CI, and Python tool contracts are green. The .NET CI job was still in progress when I posted this verdict. I did not make a live speech-provider call or manually render every historical HTML mockup on the final head; the latest rounds changed only documentation and the Python sweep, and I verified their text and executable checks directly.

PR #2578, opened 2026-08-12
voice: a failed transcript read is no longer reported as "nothing to say"

thefrederiksen:
Fixes thefrederiksen/devthrottle_internal#1838. Addresses the reporting half of thefrederiksen/devthrottle_internal#1850.

## The defect

A session could stop speaking forever with no error raised anywhere, its rail stuck on "Preparing voice".

The `turns` verb reports every FAILED read as a SUCCESSFUL command result carrying an empty widget list, with the real outcome in a `Status` field - `ok`, `unsupported`, `no_session_id`, `no_jsonl`, `parse_error`. **Nothing read that field.** `WingmanVoiceService.GenerateOnceAsync` looked at the widgets alone, found no text, and recorded `NothingToNarrate` - a NON-failure that is never retried, logs nothing, sets no state, and surfaces on no screen.

There was a sixth case that was worse: the non-Claude-Code branch stamped `Status = "ok"` unconditionally, while `SessionHistoryReader.ReadAll` returns an empty conversation whenever the transcript path cannot be resolved. So a Codex, Pi or Grok session whose transcript had not been located reported a successful read of an empty conversation, and the one field that could have said otherwise agreed.

Observed live on 12 August: a Pi session on `worktrees/oc-8403` sat silent for 48 minutes. The Gateway's own verdict for it was `nothingToNarrate` while the roster label said "Preparing voice".

## The change

**It generates, or it says why.**

- `SessionReadExecutor.Turns` - an unresolved transcript is now `no_transcript`, not `ok`. Gemini is deliberately exempt: it persists no transcript by design and reads its conversation from the terminal buffer, so a null path there is honest.
- `WingmanVoiceService.GenerateOnceAsync` - reads `turns.Status`. A null answer or any non-`ok` status records `HostedAiState.Retrying` and logs the status and error, so the voice sweep tries again and the screen has something to say. `NothingToNarrate` is now set ONLY when the read genuinely succeeded and carries no text reply - the one case where those words are true.
- The same read-failure guard on `POST /sessions/{sid}/wingman/explain`, which is the path the owner presses a button to reach. It used to answer "this session has not produced anything to summarize yet" - a confident, wrong sentence - on a failed read.
- A read that answers clears the read-failure `Retrying` it caused. Only `Retrying`: a standing `NeedsCredits` / `CapReached` / `NeedsKey` is a real condition that a successful transcript read says nothing about, and clearing it here would make the "add credit" message flap with every sweep.

**And the rail stops claiming work is in flight when there is none.**

- `SessionOrdering.StateLabel` renders the voice verdict the Gateway ALREADY folded (`SessionDto.VoiceDisplay`, from `VoiceDisplayFold`) instead of saying "Preparing voice" for every no-audio state. A row now reads "Nothing to read aloud", "Voice service down", "Voice needs credit", "Voice on its way" or "No narration yet" when that is what is true.
  - **The colour rule is untouched.** `IsVoicePreparing` still holds yellow across the gaps, per the 2026-07-19 ruling. Only the words moved, and they are words the fold already chose - so this is one answer rendered in a second place, not a second answer.
  - Deliberately narrow: a row carrying no verdict, or a verdict kind not listed, keeps "Preparing voice". A voice state added later does not silently change this label.
- `EnrichVoiceThenFoldForPush` now stamps `VoiceUnavailable` and `VoiceDisplay` as well as the two booleans, so the desktop push carries the same facts as the roster poll and the two surfaces cannot say different things about one session.

## Also

`packages/client-core/src/auth/AccountsPanel.test.tsx` - a zero-argument mock implementation typed the mock as zero-argument, so the one-argument call below it was a typecheck error. This had main's web workspace red, which blocked the local gate entirely. Chased forward here.

## Proof

Every new test was run against the UNFIXED code first: 14 failed. With the fix, all pass.

- `dotnet test` local gate: 9 projects, all `Completed`, 4,299 passed, 0 failed.
- Parked suites (`-Parked`): the gate flagged `Core.Tests` and `Gateway.Tests` as covering touched code, so both were run.
- Web: `@devthrottle/client-core` 88 files, 961 tests passed, no unhandled errors.

New tests:
- `TurnsVerbUnresolvedTranscriptTests` - Pi, Codex and Grok with no transcript yet report `no_transcript`; an agent with no history provider still reports `unsupported` (the two are different facts and stay different).
- `WingmanVoiceServiceTests` - each of the five failure statuses, and a non-answering tunnel, record `Retrying` and NOT nothing-to-narrate; a successful read with no text still records nothing-to-narrate and still raises no failure; a recovered read clears the read-failure `Retrying`.
- `SessionOrderingTests` - each verdict kind renders its own words while the dot stays yellow; a genuinely preparing session still reads "Preparing voice"; a row with no verdict, and a row with an unknown verdict kind, both fall back.

### DIFF
--- a/src/CcDirector.ControlApi/SessionReadExecutor.cs
+++ b/src/CcDirector.ControlApi/SessionReadExecutor.cs
@@ -107,8 +107,34 @@ internal sealed class SessionReadExecutor : ISessionCommandArea
                 return DirectorCommandResult.Success(SessionCommandExecutor.Serialize(resp));
             }
 
+            // AN UNRESOLVED TRANSCRIPT IS A FAILED READ, NOT AN EMPTY CONVERSATION.
+            //
+            // This branch used to stamp "ok" unconditionally. But SessionHistoryReader.ReadAll returns
+            // ConversationHistory.Empty whenever ResolveTranscriptPath returns null - which is exactly what
+            // the per-agent locators (PiSessionLocator, CodexRolloutLocator, GrokSessionLocator) do when they
+            // cannot find the session's transcript. So a Codex / Pi / Grok session whose transcript had not
+            // been located reported a SUCCESSFUL read of an EMPTY conversation, and the one field that could
+            // have told the caller otherwise said "ok".
+            //
+            // What that cost: voice narration reads this verb, sees no text widget, and records "there is
+            // nothing to read aloud" - a NON-failure that is never retried. A Pi session on worktrees/oc-8403
+            // sat silent for 48 minutes on 12 August with no error raised anywhere, and the roster showed it
+            // "Preparing voice" the whole time. See issue #2561.
+            //
+            // Gemini is the deliberate exception and must NOT be caught by this: it persists no transcript at
+            // all, so a null path is its normal state and its conversation is read from the session's own
+            // terminal buffer (see SessionHistoryReader.ReadAll). A null path for Gemini is honest; for every
+            // other agent here it means the transcript has not been found yet.
+            var transcriptPath = SessionHistoryReader.ResolveTranscriptPath(session);
+            if (transcriptPath is null && session.AgentKind != AgentKind.Gemini)
+            {
+                resp.Status = "no_transcript";
+                resp.Error = $"No transcript has been located yet for this {session.AgentKind} session.";
+                return DirectorCommandResult.Success(SessionCommandExecutor.Serialize(resp));
+            }
+
             var history = SessionHistoryReader.Read(session);
-            resp.JsonlPath = SessionHistoryReader.ResolveTranscriptPath(session);
+            resp.JsonlPath = transcriptPath;
             resp.LineCount = history.Messages.Count;
             resp.Widgets = ControlEndpoints.BuildTurnWidgetsFromHistory(history);
             resp.Status = "ok";
--- a/src/CcDirector.ControlApi/SessionReadExecutor.cs
+++ b/src/CcDirector.ControlApi/SessionReadExecutor.cs
@@ -1,4 +1,4 @@
-using System.Linq;
+﻿using System.Linq;
 using System.Text;
 using CcDirector.Core.Agents;
 using CcDirector.Core.Claude;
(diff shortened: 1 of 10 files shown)


thefrederiksen:
## Two review rounds, both by a different agent family, both acted on

**Round one** (four findings) is answered by commit two. **Round two returned DO NOT MERGE** and was right - the first fix round had introduced a regression - and is answered by commit three.

The regression, stated plainly because it is the interesting one: the sweep skip added in round two was an unbounded `continue` whose only escape was the Working transition. On the hosted push path that is observed by `TurnEndWatcher`'s 15-second sampler, which a quick turn can slip past entirely. Before the skip existed a missed edge still left the idle sweep able to retry; after it, a stale terminal verdict would have survived forever - the same permanent silence this pull request exists to remove, rebuilt one layer up. The verdict now carries a revalidation deadline.

Also from round two: the empty-read rule did not cover Claude Code (the agent this verb runs for most often), a terminal read verdict could hide behind a stale account condition nothing clears, the inverted rail label omitted the `working` kind, and the new status was undocumented on the contract.

## On the review's test findings

Checked rather than taken on trust. The three new discriminating tests were mutation-tested against their own fix removed, and each fails. The `ready` and `off` label rows pass either way - they are regression guards on retained behaviour, not evidence for the inversion, and the test now says so rather than implying otherwise.

## Deliberately out of scope, stated so it is not mistaken for coverage

- **`/exes/list` and the single-session route** fold without stamping the voice verdict, so their rows take the "Preparing voice" fallback and can disagree with the roster. They never stamped the two voice booleans either, so both were already wrong before this change and neither is made worse by it. `/exes/list` is a local diagnostics page; the single-session route has no shipped client.
- **No clock.** Nothing records when the no-audio state began, so no surface can yet say "this has been preparing for 44 minutes". That was asked for in thefrederiksen/devthrottle_internal#1850 and is not in this change.
- **No general give-up.** `unsupported` is terminal now, but there is still no "tried for N minutes, gave up" state. A session that keeps failing keeps saying "on its way".
- The resolved-but-empty case is **not** asserted for Copilot or opencode. Their readers open the developer's own database, so the verdict would move with the machine running the test - a first version of that test failed on exactly that. It is asserted through Gemini and Claude Code, which are deterministic, plus the status handling itself.

## Proof

- Full gate with both parked suites, run against commit three: 11 suites, all `outcome=Completed`, 0 failed (Gateway unit suite 3,030; Gateway.Tests 2,314; Core.Tests 4,284).
- Rebased onto main afterwards. The rebase touched ONE file - `AccountsPanel.test.tsx`, where main had landed the same web typecheck fix independently (#2577) and its version was taken. The C# tree is byte-identical to the tree the parked run went green on, and the default gate plus the workspace typecheck were re-run after the rebase and are green.
- Web: `@devthrottle/client-core`, 961 tests, no unhandled errors.

## Reaching users

The work splits across two deployables, and merging alone changes nothing anyone can see:

- The voice service, the explain route, the rail label and the display push run on the **hosted Gateway**.
- The `turns` statuses (`no_transcript`, `empty_history`) run on the **Director** - the desktop app. Until that ships, an unresolved transcript still answers `ok` with no widgets, so the originally reported session is not fixed by a Gateway deploy alone.

PR #2587, opened 2026-08-12
gateway: stop giving up on the database while the platform is still waiting

thefrederiksen:
Addresses part of thefrederiksen/devthrottle_internal#1856.

## The chain, from the platform's own log

```
13:00:24   Site started                        <- container A, serving fine
13:00:55   a SECOND container starts
13:02:42   Container has finished running with exit code: 1   (after 103.1s)
13:02:43   Failed to start site. Revert by stopping site.
13:02:43   Site: devthrottle-gw stopped        <- production dark
13:03:04   Stopping container: 35984bb...      <- container A, the HEALTHY one
13:03:29   production answers again            <- 46.7s
```

## The change

The database open gave up after 90 seconds. The platform allows 230. The container exited at 103.1s including boot, and the same database on the same image opened cleanly minutes later.

The 90 seconds rested on a premise this incident disproves - that exiting on our own terms beats being timed out. The platform makes no such distinction during site startup: both reach `Failed to start site. Revert by stopping site.`, and that stop kills the healthy container. So giving up early bought nothing and cost ~127 seconds of budget in which the database might have recovered.

Now sized against the budget it has to fit inside, with headroom for boot and for binding afterwards (the healthy boot that day bound and answered the probe 21s after container start, database open included).

No downside case: a genuinely dead database still fails, just later, and the site was going to be stopped in that case regardless.

Also corrects the `MustTerminate` comment, which claimed an exiting container is restarted while a silent one takes the healthy one with it. During site startup both do.

## What this does NOT do

**It is a mitigation, not the fix, and both files say so.** The fix is binding the port before the database work so site startup never depends on PostgreSQL - thefrederiksen/devthrottle_internal#1766's first recommendation, still unbuilt. This does not close thefrederiksen/devthrottle_internal#1856.

I have also not proved the database open was the reason this container exited. The 103.1s exit matches 90s of retry plus boot closely enough to name as the likely cause, but the Gateway logs to the file share rather than stdout, so the container's own reason is not in the platform log.

## Proof

Full gate with both parked suites: 11 suites, all `outcome=Completed`, 0 failed.

No test asserts the old constant, so nothing here is covered by a test that would have caught a wrong value - the evidence for the number is the platform log in thefrederiksen/devthrottle_internal#1856, not the suite.

### DIFF
--- a/src/CcDirector.Gateway/Data/GatewayDatabase.cs
+++ b/src/CcDirector.Gateway/Data/GatewayDatabase.cs
@@ -83,9 +83,23 @@ public sealed class GatewayDatabase : IDisposable
     //
     // The window is therefore sized against the budget it actually has to fit inside, with headroom for
     // the rest of boot and for binding the port afterwards: the healthy boot that day bound and answered
-    // the platform probe 21 seconds after the container started, database open included. There is no
-    // downside case. A database that is genuinely gone still fails - it just fails later, and the site was
-    // going to be stopped in that case regardless.
+    // the platform probe 21 seconds after the container started, database open included.
+    //
+    // THE COST, stated because an earlier version of this comment claimed there was none and that was
+    // simply false (found in review). GatewayDatabase's catch takes EVERY exception once the connection
+    // string has parsed, so a wrong password, an unreachable host, a missing or failed migration and a
+    // provider fault all traverse this window too - not just the transient contention it was written for.
+    // Every one of those now takes about eighty seconds longer to report and to restart. An operator who
+    // has mistyped a connection string waits nearly three minutes for the error instead of ninety seconds.
+    //
+    // That is a real cost and it is accepted deliberately, because the two sides are not comparable: the
+    // slow side costs an operator eighty seconds while they are already debugging a broken configuration,
+    // and the fast side costs every user a live outage on an ordinary deploy. A misconfiguration is
+    // noticed and fixed once; the deploy path runs every time we ship.
+    //
+    // What is NOT a cost: a database that is genuinely gone still fails, just later, and the site was
+    // going to be stopped in that case regardless. And a Migrate that HANGS is unaffected either way - the
+    // deadline is only tested after an attempt returns, so a blocked attempt never reaches it (see #2395).
     //
     // THIS IS A MITIGATION, NOT THE FIX. The fix is to bind the port BEFORE the database work so that
     // site startup never depends on PostgreSQL at all, which is #2383's first recommendation and is still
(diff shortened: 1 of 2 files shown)


thefrederiksen:
Review verdict was DO NOT MERGE, on two findings. Both correct, both fixed in `16ab4a957`.

**1. "There is no downside case" was false.** `GatewayDatabase`'s catch takes every exception once the connection string has parsed, so a wrong password, an unreachable host, a missing or failed migration and a provider fault all traverse this window - not only the transient contention it was written for. Each now takes about eighty seconds longer to report and restart; an operator who mistyped a connection string waits nearly three minutes for the error instead of ninety seconds.

The comment now states that cost and states it as ACCEPTED rather than absent, with the reasoning: the slow side costs one operator eighty seconds while they are already debugging a broken configuration, the fast side costs every user a live outage on an ordinary deploy. A misconfiguration is fixed once; the deploy path runs every time we ship.

**2. The GatewayWorker correction was incomplete.** `MustTerminate`'s own comment was fixed, but the block at `GatewayWorker.cs:117-122` still ended "...and killing its healthy neighbour", implying termination SAVES the healthy container. It does not - thefrederiksen/devthrottle_internal#1856's platform log shows an exiting container and a non-binding one both reaching `Failed to start site. Revert by stopping site.` That is the third time this comment has stated something untrue about the same behaviour, so it now says so and points at the evidence.

The reviewer also confirmed the author's claim that a hanging `Migrate` is unaffected by the window size (the deadline is only tested after an attempt returns, so a blocked attempt never reaches it - thefrederiksen/devthrottle_internal#1270). That is now written into the comment rather than left as an assertion in the pull request.

The 170-second change itself was not challenged. Re-running the full gate against the corrected commit before merge.

thefrederiksen:
Gate run against `16ab4a957`: ten of eleven suites green; `Gateway.UnitTests` reported one failure, `MissionNoteStoreTests.All_returns_a_stable_snapshot_detached_from_the_live_store`.

Judged rather than waved through. The stack is a SQLite handle crash inside `SqliteConnection.Open()` (`SafeHandle.DangerousAddRef` -> `sqlite3_create_function`) - the known teardown race, on the SQLite path. This change touches the PostgreSQL open-retry constant and two comments, so it cannot reach that code.

Re-run twice, because passing one test in isolation does not reproduce the parallel conditions it failed under: the test alone passes (10/10), and the WHOLE suite passes (3028 passed, 0 failed). The other ten suites were green in the original run, including both parked ones - `Gateway.Tests` 2314 and `Core.Tests` 4284.

Merging on that basis, and saying so here rather than reporting a clean green that did not happen.

PR #2588, opened 2026-08-13
voice: say how long a session has been waiting, and admit when the voice is not coming

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1850 - the two halves left unbuilt when the first part shipped in thefrederiksen/devthrottle#2578.

## What changes on the row

```
before:  Preparing voice                    (identical at 2 seconds and 48 minutes)
after:   Preparing voice (2m)               while it is genuinely being made
         Voice did not arrive after 48m     past three minutes, red
```

## The clock

`VoiceWaitingClock` stamps when a session's wait for voice begins, holds that moment for the episode, and clears it when audio arrives or a new turn starts.

**A SECOND clock beside `NeedsYouClock` deliberately.** That one is stamped only when the folded colour is RED, and a session waiting for voice is YELLOW - so the one elapsed-time fact the product had was, by construction, never running for exactly the sessions that got stuck. That is why nothing could say "48 minutes". Keyed by (tenant, session) like its neighbour, so one account's voice arriving cannot reset another account's wait.

## The terminal state

Past three minutes the verdict becomes `gaveUp`, red. Three minutes is not invented here - `SessionOrdering`'s note on `IsVoicePreparing` already said voice should average under a minute and that "anything over three minutes is an exception to be flagged and fixed". This renders a standard that was already written down.

Playable audio and a live attempt both outrank it, so nothing in flight is ever hidden behind it. The Gateway keeps trying - what ends is the promise, not the effort.

## A design decision worth flagging

I first computed the elapsed time on the phone. It did not typecheck: the client's types are generated from the API schema, so any new field needs a regeneration against a running Gateway - and I was not going to launch one beside the owner's live Gateway to get it.

So the wait is turned into words on the Gateway, beside every other voice string it already composes, and rides the existing `label`. No schema change, and it is what rule 7 asks for anyway. **The trade, stated plainly:** whole minutes refreshed at poll rate, rather than a second-by-second tick. Good enough to answer "is this stuck?", which is the only question this number exists for.

Both the roster and the display push stamp the clock from the SAME facts the fold reads, so the elapsed time and the words cannot disagree about whether a session is waiting at all.

## Proof

- Full gate, both parked suites: 11 suites, all `outcome=Completed`, 0 failed. Gateway unit suite 3030 -> 3050.
- **Mutation-tested.** Disabling the give-up arm and the rail wait fails four assertions; the negative controls (audio beats give-up, a live attempt beats give-up, not-waiting is unchanged, the ladder) stay green as they should.

## Not in this change

The give-up threshold is time-based, not attempt-based - the attempts come from two loops at rates neither controls, so a count means different things on a busy Gateway and a quiet one while a clock means the same thing to the reader. If we later want "gave up after N attempts", that is a different fact and would need its own counter.

### DIFF
--- a/src/CcDirector.Gateway.Contracts/SessionDto.cs
+++ b/src/CcDirector.Gateway.Contracts/SessionDto.cs
@@ -182,6 +182,25 @@ public sealed class SessionDto
     public DateTime? NeedsYouSince { get; set; }
 
     /// <summary>
+    /// Issue #2576: UTC timestamp this session last found itself a VOICE session with no playable
+    /// narration - the moment its wait for voice began. Gateway-owned, stamped during the same
+    /// aggregation that folds <see cref="VoiceDisplay"/>, held stable for the whole episode, and
+    /// cleared to null the moment audio arrives or the agent starts a new turn.
+    ///
+    /// It exists because <see cref="NeedsYouSince"/> could not answer this. That clock is stamped
+    /// only when the folded colour is RED, and a session waiting for its voice is YELLOW - so the
+    /// one clock the product had was, by construction, never running for exactly the sessions that
+    /// were stuck. A session sat on "Preparing voice" for forty-eight minutes and no surface could
+    /// say so, because no fact recorded when the wait started.
+    ///
+    /// Clients render the AGE of this, never a duration the Gateway computed: a number computed at
+    /// stamp time is wrong by however long it sat in transit and on the screen. In-memory only - a
+    /// Gateway restart re-derives it on the next waiting refresh. Always null in Director-local
+    /// responses, which have no Gateway voice state at all.
+    /// </summary>
+    public DateTime? VoiceWaitingSince { get; set; }
+
+    /// <summary>
     /// The absolute UTC time an ARMED snooze returns this session to "needs you" - the snooze clock's
     /// deadline, so a client can show "wakes in 3h 48m". Gateway-owned: the snooze registry
     /// (<c>SnoozeRegistry.SnoozeEntry.SnoozeUntilUtc</c>) is the sole source of the timer, stamped by the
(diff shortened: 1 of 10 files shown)


thefrederiksen:
Review returned DO NOT MERGE on three blockers. All three fixed in `80e9dc427`.

**1. The give-up verdict was hiding every actionable reason.** Past three minutes it preempted `ServiceDown`, `NeedsCredits`, `CapReached` and `NeedsKey`, so a member who needed to add credit was told "voice did not arrive" - the symptom in place of the sentence that says what to do. Same defect as a terminal read hiding a standing account condition, which this codebase fixed once already today.

It now displaces exactly two states, `Retrying` and `notReady` - the two whose entire content is "be patient", which are the sentences still being said at forty-eight minutes. Everything specific outranks it, as do audio, a live attempt, and nothing-to-narrate. New theory covers all five actionable states.

**2. The waiting rule existed twice**, hand-written in the roster aggregation and the display-push enrichment. They agreed character for character - the condition under which two copies stay wrong together and then diverge with nothing to notice. Now one predicate, `VoiceDisplayFold.IsWaitingForVoice`, called by both, with a test.

**3. A test proved nothing.** `StateLabel_GaveUp_ReachesTheRailWithItsOwnWords` passes unchanged on the parent, because the deny-list shipped earlier today already renders any unknown kind's label. Kept as a regression guard on that behaviour and now labelled as one rather than presented as evidence for this change. The clock's "strictly later" assertions asserted `>=`, so they are renamed to the not-earlier claim they actually make.

**Also corrected, from the same review:** the pull request said this needed "no schema change at all". False - it adds `VoiceWaitingSince` and `WaitedLabel`. The true, narrower statement is that no CLIENT consumes them, so the generated TypeScript needs no regeneration for this to work. `schema.ts` is therefore behind the contract until someone regenerates it against a running Gateway; that is a known, mechanical follow-up and not a blocker for this change.

**Gate against the corrected commit:** 11 suites, all `outcome=Completed`, 0 failed. Gateway unit suite 3050 -> 3057.

PR #2598, opened 2026-08-18
voice: a model reads the sentence before any unlisted word is changed

thefrederiksen:
Restores the design the dictation cleanup originally had, and closes the gap #1553 left open: unlisted corrections were switched off, not fixed.

## The shape

The matcher keeps its job and loses its authority.

1. Wrong forms the user listed by hand apply as before, deterministically. No ruling needed - they chose them.
2. `FuzzyDictionaryMatcher.ProposeCandidates` nominates spans - **every occurrence separately, with its offset**.
3. The existing `TranscriptEditEngine.Validate` gate still stands in front of the judge, so a candidate it should never see cannot be rescued by a permissive ruling.
4. A judge that reads the sentence rules on each nomination and answers with **ids only**: `{"acceptedCandidateIds":[1]}`.
5. Accepted rulings apply at the offset they were judged at.

## Why this is not the pass removed in July

That one was handed the transcript and asked what was wrong - open-ended, ~5s every turn. This is a closed question: one sentence, at most 12 candidates our own code isolated, answered with numbers.

- **Skipped entirely when nothing is nominated**, which is most utterances. No call, no cost.
- **1s deadline, no retry.** A retry doubles the worst case on the one path where the worst case is the whole problem, to recover a correction whose absence costs a wrong spelling.
- Routed through `devthrottle/dictation-cleanup` as the proven `IncludedModelId` type, so the deployment credential cannot be pointed at a catalog model that would bill credits for an internal feature.

## What the judge cannot do

Bounded by the contract, not by trust, on **both** sides:

- It returns `Task<IReadOnlyList<int>?>`, so it cannot reword, summarize, refuse into the transcript, or name a span nobody offered.
- It is handed a `ReadOnlyCollection` of fresh copies, and accepted ids resolve against a **private snapshot it never receives**. Review found that passing the same list we later read back would let an in-process judge swap an entry after validation and supply arbitrary text through the parameter - defeating the property through the door the return-type check cannot see. A judge attempting exactly that is now a test, and it asserts the write was refused.

Every unhappy path means **no ruling**, and no ruling means the transcript is untouched: no judge configured, non-success status, malformed body, prose, fenced JSON, a stall past the deadline, a cancelled turn, a transport failure, a thrown exception. An id that was never offered voids the **whole** ruling rather than being filtered out.

## Offsets

The judge ruled on one word in one sentence. The same word later in the same breath is a different use it never saw, so `TranscriptEditEngine.ApplyAt` rewrites only the judged span, right-to-left so earlier offsets are never shifted, skipping any span that no longer matches verbatim. Deferred from #1553.

## Shadow first

Ships in `Shadow`: rule on real dictation, record what would have changed, change nothing. `ShadowChanges` is a separate field from `ChangedWords` on purpose - a shadow run must be indistinguishable from no cleanup to anything reading the transcript.

Promotion is one environment variable (`DEVTHROTTLE_DICTATION_JUDGE_MODE=enforce`), needs no new build, and requires the **exact** value - case, padding and an undefined enum all shadow.

## The guard, and what it does not do

`TranscriptEditEngine` claimed since July that a `TranscriptionIntegrity` architecture test enforced the whole rule. It never existed.

Half of it is now genuinely enforced: the judge boundary, structurally, on both return type and parameter, backed by the runtime attack test above.

The other half - "only `TranscriptEditEngine` rewrites transcript text" - is **not** enforced, and now says so. Two attempts were written and deleted: a substring scan (evadable by one space; failed the build on the word inside a comment) and a Roslyn syntax walk (missed `text.Replace(...)`, `String.Concat`, `this.rawTranscript`, null-conditional calls and interpolated rebuilds, while flagging a `Substring` used for a log preview). A check that certifies an unconditional claim while missing the ordinary spellings is worse than none, because the green run is read as proof. `docs/CodingStyle.md` section 16 and the engine comment both now state which half is checked; the real check is scoped on devthrottle_internal#1556 with the controls it must satisfy.

## Tests

**256 Core dictation tests, 3094 Gateway unit tests.** Full solution builds clean.

Covering: the no-ruling invariant in every failure mode; the mutation attack; a mixed valid/unoffered ruling; shadow vs enforce and the exact-value switch; per-occurrence offsets; that the judge is *not* asked when the guessing is off, cleanup is disabled, nothing is nominated, or a listed alias matched; the candidate cap; ~30 malformed replies against the parser; and that neither the credential nor the transcript reaches a log on the judge paths.

Reviewed four times; five findings, one critical, all closed.

Refs devthrottle_internal#1554

### DIFF
--- a/docs/CodingStyle.md
+++ b/docs/CodingStyle.md
@@ -812,7 +812,7 @@ await Task.Delay(500);
 | Regex | Always set timeout | `TimeSpan.FromMilliseconds(50)` |
 | Brushes | Freeze for cross-thread | `brush.Freeze()` |
 | Collections | Snapshot before iterating across threads | `.ToList()` |
-| Transcription | Model may only LOCATE words; only `TranscriptEditEngine` may change them | Never round-trip a transcript through a model that returns free text |
+| Transcription | Model may only RULE on spans code isolated first, answering with candidate ids; only `TranscriptEditEngine` may change words | Never round-trip a transcript through a model that returns free text |
 
 ---
 
--- a/docs/CodingStyle.md
+++ b/docs/CodingStyle.md
@@ -826,7 +826,7 @@ This is an absolute, load-bearing rule (it traces to a real corruption incident,
 
 1. The raw speech-to-text transcript is the user's words. The ONLY permitted change to it is replacing a misheard term with the correct **dictionary** spelling of a term the user actually said (a single word, or a tightly-joined multi-word term like `cc-director`).
 
-2. A language model may be used ONLY to **locate** which spans are misheard dictionary terms, and it must return that judgment as a **JSON list of find/replace proposals** - never the transcript text itself.
+2. A language model may be used ONLY to **rule on** spans that deterministic code has already isolated, and it must answer with **candidate ids and nothing else** (`{"acceptedCandidateIds":[1]}`) - never the transcript text, and never a span it chose itself. It cannot name a candidate that was not offered, and an id that was never offered voids the whole ruling. This is tighter than the original rule, which let the model return its own find/replace pairs; `TranscriptEditEngine.ParseEdits` is the remnant of that protocol and has no production caller.
 
 3. Only deterministic code - `CcDirector.Core.Dictation.TranscriptEditEngine`, driven by `CleanupOrchestrator` - may change the words, and only by applying a validated proposal (the `find` must occur verbatim in the raw transcript, the `replace` must be an exact dictionary term, and it must be a plausible mishearing). Everything else fails open to the raw transcript.
 
--- a/docs/CodingStyle.md
+++ b/docs/CodingStyle.md
@@ -820,7 +820,7 @@ await Task.Delay(500);
 
 **When a user dictates by voice, the speech-to-text result is the user's words and is the source of truth. It must never be rewritten by a language model.**
 
-This is an absolute, load-bearing rule (it traces to a real corruption incident, issue #190). It has regressed before, so it is written here, enforced by an architecture test, and called out in the one engine allowed to touch the words.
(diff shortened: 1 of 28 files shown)


thefrederiksen:
**Status: complete and reviewed, blocked on one configuration change outside this repo.**

Not merged yet, so this is the written reason the branch is open rather than aging silently.

**Done:** four review rounds; five findings (one critical - a judge could mutate its own candidate list and supply arbitrary text through the parameter) all closed. 256 Core dictation tests, 3094 Gateway unit tests, clean solution build at `43cc2ef33`. Production code unchanged and confirmed sound since `a240c4a1d`.

**Blocked on:** `devthrottle/dictation-cleanup` has no verified mapping to a DeepInfra model on the hosted proxy. That mapping does not live in this repo - the id appears here only as a constant and two tests. Until it resolves, the judge gets an error, fails closed, and no unlisted correction is applied. Safe, but inert.

The Included AI mission record pairs the id with `o4-mini`, which is the wrong choice: it is a reasoning model, and reasoning is what made the removed cleanup pass cost ~5s per turn. `Qwen/Qwen2.5-72B-Instruct` is already proven on our proxy behind `devthrottle/wingman-fast` and is the better first candidate.

**Known risk to measure before enabling:** `docs/research/transcription-speed/FINDINGS.md:67` records 4-45s variability on DeepInfra for the latency-critical path. This judge runs on a 1s deadline with no retry. If that variability holds, it will time out constantly - failing closed each time, so correctness is unaffected, but the feature would do nothing. One real round trip through the route settles it.

**Owner decisions outstanding:**
1. Where the proxy model map is configured / which DeepInfra model sits behind the id.
2. Whether sending the utterance + candidate list to the hosted judge is acceptable - it is a change to a stated privacy property. `docs/architecture/transcription-quality-loop.md` previously claimed no transcription content leaves the machine; this branch corrects that text to say precisely where transcripts go. Nothing is sent to the speech-to-text provider; that constraint is unchanged.
3. Ship in shadow first, or straight to enforce.

Unlisted corrections remain OFF in production via #1553 regardless of this PR.

PR #2606, opened 2026-08-28
dictation: an empty recording no longer retries forever

thefrederiksen:
## What was wrong

A recording that captured no audio was written to the durable send queue like any other and then driven forever. With no bytes it plans no upload chunks and sends the Gateway `totalChunks: 0`, which the complete endpoint correctly refuses with a 400 (`GatewayDictationEndpoint.cs:266`). A 400 is not on the client's permanent-failure allow-list, so it fell through to the retryable arm and re-drove.

The phone showed *Saved - still sending* all evening about a recording that could never be sent. On the Gateway it left a staging directory holding a delivery record and no audio for every attempt - three of them registered in one second, none with a byte behind it, which is how this was found.

## Three guards, each at the level that owns the problem

**The door.** `DictationRecorder` builds its clip as `new Blob(this.chunks)`, which is zero bytes when no chunks were delivered. `backgroundTranscribeAndSend` now refuses that at the single point all four Send surfaces pass through, before the durable write - so an unsendable clip never enters the queue. It fails loudly and immediately: `phase: "failed"` (the red `role="alert"` strip that never auto-clears), the host's `onError`, and the typed text handed back. Nothing queued, so nothing to retry, nothing to cancel, no staging.

**The backstop.** `uploadDictationToSession` guards the clip it is about to send, for one queued whole whose copy degraded later. A copy that reads back successfully at zero bytes is permanent and parks. A read that *throws* stays held and keeps retrying - a failed read says nothing about the recording's length, so parking it would strand audio that is still there.

**The server.** `VoiceUploadStore` keeps the completeness gate's per-chunk measurements and compares each against what was actually read. On the hosted Gateway a chunk measured 871,724 bytes read back as zero twice inside five seconds; the empty assembly reached the caller, whose empty-audio arm calls `store.Delete(uploadId)` - so one bad read discarded audio already uploaded successfully. A mismatch is now reported incomplete: the client re-sends, the staged bytes are kept. Per chunk, not on the total, because two faults can cancel out in a sum.

## Testing

| Suite | Result |
|---|---|
| client-core | 969 pass |
| cockpit | 282 pass |
| mobile | 3 pass |
| Typechecks (all three workspaces) | clean |
| Gateway unit tests | 3,111 pass, 2 skipped |
| Core.Tests (parked) | 4,374 pass |
| Gateway.Tests (parked) | see below |

New tests cover the empty-capture gate (nothing queued, loud failure, immediate, and a real recording untouched), the backstop's split between a zero-length read and a thrown read, and `ReadBackVerdict` against the exact measured/read pair observed in production plus the compensating-error case.

The eight `HostedSchemaRefusesAnUnownedRowTests` failures seen on this machine were an absent Postgres rig, not this change: `CC_GATEWAY_TEST_PG_STATS_CONNECTION` was set in the User environment while nothing listened on the port, and the suite's skip-gate checks only that the variable is set, not that the rig is reachable. With a rig up (`scripts/pg-stats-proof-rig.ps1`) all eight pass.

Reviewed by an agent of a different family, which found two real defects that are fixed here: collapsing a thrown read into the permanent empty case, and comparing assembled totals rather than per-chunk lengths.

### DIFF
--- a/packages/client-core/src/api/client.ts
+++ b/packages/client-core/src/api/client.ts
@@ -1921,8 +1921,15 @@ async function sha256Hex(bytes: Uint8Array): Promise<string> {
  *  than re-drive forever. EVERYTHING outside this list (network, provider 502/504, busy session, out of
  *  credits) stays retryable/held. Default is retryable; only these reasons are permanent.
  *   - "audio-too-large": the clip is over the provider's per-request byte cap and cannot be split as-sent.
- *   - "unsupported-format": the audio is a format the Gateway cannot decode or split. */
-export type PermanentDictationReason = "audio-too-large" | "unsupported-format";
+ *   - "unsupported-format": the audio is a format the Gateway cannot decode or split.
+ *   - "empty-recording": the on-device copy was read SUCCESSFULLY and contained ZERO BYTES, so there is no
+ *     audio to send. Decided entirely on the device, before any network work - see the empty-audio guard in
+ *     uploadDictationToSession. Permanent because the bytes are the input: re-reading the same empty copy
+ *     produces the same nothing however often it is asked. A read that THROWS is deliberately NOT this - it
+ *     says nothing about the length, so it stays held and retryable
+ *     (DICTATION_HELD_UNREADABLE_MESSAGE). Do not widen this reason to cover it: parking a recording that a
+ *     later read would have recovered strands it, and tells the user it was empty when it was not. */
+export type PermanentDictationReason = "audio-too-large" | "unsupported-format" | "empty-recording";
 
 /** Outcome of one dictation delivery attempt.
  *  - `terminal` true: the server made a final decision (injected the turn, or deliberately dropped it as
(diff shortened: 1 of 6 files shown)


thefrederiksen:
Merged on the owner's explicit decision without waiting for Gateway.Tests to finish, so the record is straight about what was and was not run.

Run to completion on the merged code: client-core 969, cockpit 282, mobile 3, all three typechecks, Gateway.UnitTests 3,112 pass / 2 skipped / 0 fail, Core.Tests 4,374 pass. Reviewed twice by an agent of a different family; both rounds of findings are fixed in the commit.

NOT run to completion on the final commit: Gateway.Tests (65 minutes; stopped about 10 minutes in). It was run on the code one revision earlier - before the two review fixes - where all 65 of its dictation, upload, assemble and chunk tests passed, and its 97 failures fell entirely in classes that cannot reach the changed method (snooze, stats, spoken language, session proxy, device enrolment, Postgres proofs). Those failures were the absent Postgres rig: CC_GATEWAY_TEST_PG_STATS_CONNECTION is set at User scope on this machine while nothing listens on the port, and the suites' skip-gate checks only that the variable is set, not that the rig answers. With a rig up, Gateway.UnitTests went from 8 failures to 0 with no code change.

If Gateway.Tests is later found red for a reason that traces to this change, this comment is the place to start.

PR #2643, opened 2026-09-02
Add searchable mobile session picker

thefrederiksen:
## Summary

- redesign the mobile session starter as Director, agent, repository, and review steps
- keep five recent repositories as the immediate choices while searching the complete durable machine catalog
- record known repositories in Gateway storage with SQLite and PostgreSQL migrations
- prevent stale Director responses and rapid double submission from creating the wrong session

Closes thefrederiksen/devthrottle_internal#1853
Closes thefrederiksen/devthrottle_internal#1884

## Verification

- solution build completed with zero warnings and zero errors
- shared client: 972 tests passed
- mobile application: 10 tests passed
- SQLite and PostgreSQL migration snapshots report no pending model changes
- the PostgreSQL collation proof passed against a real database
- the complete flow passed in a real browser at 390 by 844 pixels with eight repositories and two enabled agents


thefrederiksen:
Independent review of `b18cc2548296c393cb8832eaa3aeeca668ea061d`, checked against issues thefrederiksen/devthrottle_internal#1853 and thefrederiksen/devthrottle_internal#1884 rather than the description. Everything below was read from that exact commit; where I say "reproduced", I ran it.

## What I ran

- `packages/client-core`: 90 files, 972 tests passed.
- `apps/mobile`: 2 files, 10 tests passed.
- `CcDirector.Gateway.UnitTests --filter KnownRepository`: 5 passed.
- Three throwaway probe tests against the real `NewSession` component, in a scratch copy of the workspace. The worktree was not modified.

## Findings

### 1. The manual path escape hatch fails silently

`chooseManualPath` sets `setCreateError("Enter a repository path before continuing.")` while `activeStep` is `"repository"`. But `createError` has exactly one render site, `apps/mobile/src/pages/NewSession.tsx:607`, and it sits inside `{activeStep === "review" && canReview && ( ... )}`. So the message is never mounted.

Reproduced: open the repository step, tap "Enter a path manually", tap "Use this path" with the field empty. The message is not on screen and there is no element with role `alert`. The button does nothing at all, which on a phone reads as a broken button.

The same render site makes the three early returns in `create()` invisible too, though those need a state change between render and tap to reach.

There is no test anywhere for the manual path field, neither the accepting case nor the rejecting one.

### 2. Retrying a repository source discards the user's agent choice, repository choice and typed search

"Retry repository sources" (and the "Retry" button on the agent error, which is wired to the same `setRepositoryReload`) re-runs the loading effect, and that effect calls `clearDirectorChoices()`. That resets `selectedAgentType`, `selectedRepository`, `repositoryQuery`, `manualPath` and `manualOpen`.

Reproduced: with the repository history failing, type `repository 1` into the search field, then tap "Retry repository sources". The field is empty afterwards. The chosen repository is gone with it.

Issue thefrederiksen/devthrottle_internal#1884 asks for explicit loading, empty, unavailable and error states. Throwing away what the user typed in order to recover from a partial failure is the opposite of that: the one action offered for the error costs the user their work.

Only the agent selection survives, and only by accident, because `rememberAgent` wrote it to browser storage.

### 3. The search result list is uncapped

Reproduced: with 400 repositories in the catalog, typing a single character `r` renders 400 rows. There is no cap and no virtualisation, and the Gateway deliberately does not cap either (`KnownRepositoryStore.ReadForMachine`: "There is deliberately no result cap").

That server side decision is right, but it makes the cap the client's job, and the client does not take it. Issue thefrederiksen/devthrottle_internal#1853 exists precisely because the catalog is large.

### 4. The SQLite backfill writes a machine key the read path can never match

`KnownRepositoryStore.NormalizeMachineKey` is `Trim().ToUpperInvariant()`. The SQLite migration backfills `MachineKey` with SQL `UPPER(TRIM("MachineName"))`. SQLite's `UPPER` folds ASCII only.

Demonstrated on SQLite:

```
sorens-pc with o-slash -> sqlite UPPER keeps \xf8 ; .NET ToUpperInvariant gives \xd8 ; match: false
cafe-build with e-acute -> sqlite UPPER keeps \xe9 ; .NET ToUpperInvariant gives \xc9 ; match: false
```

`ReadForMachine` narrows solely on `row.MachineKey == machineKey` in the database. So for any machine whose name contains a non-ASCII letter, every backfilled row is invisible to the picker, which is exactly the history the migration was written to preserve. It stays invisible until the recorder observes the repository again and inserts a second, duplicate row with the correct key.

The managed re-check in `Observe` does not rescue this: it runs on rows already narrowed by `MachineKey == machineKey` in the database, so it can only narrow further, never widen. Its comment ("the migration normalizes the machine key but preserves the display path available in history") is accurate about the path and misleading about the machine.

PostgreSQL's `UPPER` is locale aware and would fold these, so the two providers also disagree with each other here.

### 5. `MachineKey` is not pinned to `UseCollation("C")`

`GatewayDbContext` lines 900 to 990 state the rule and apply it to every ordinally compared identity string column, so that SQLite's BINARY collation and PostgreSQL agree exactly on equality. `KnownRepositoryEntity.MachineKey` is exactly that kind of column: it is the indexed lookup key, compared with `==` inside the database in both `Observe` and `ReadForMachine`, and normalized ordinally in managed code.

It is not pinned, and no reason is given. Note that the collation proof's count moving from 21 to 25 covers the four `session_turns` and `session_turn_heads` columns that arrived with thefrederiksen/devthrottle#2640; it is not evidence that this change's columns were considered.

### 6. A Director with a blank machine name returns 500 rather than an answer

`DirectorRegistry` (Discovery/DirectorRegistry.cs:345) permits `MachineName = ""`, and `POST /directors/register` validates only `DirectorId`. `ReadForMachine` calls `Required(machineName, ...)`, which throws `ArgumentException` on a blank value, and the new handler has no guard around it, so the request becomes an unhandled 500.

Every other per-Director route tolerates a blank machine name; `GatewayEndpoints.cs:2550` shows the defensive shape used elsewhere. An empty list, or an explicit stated error, is the right answer here.

### 7. `PathKey` is stored, documented as load bearing, and never read

The property comment says "Normalized repository path used by reads and in-process deduplication". It is used by neither. `Observe` compares `NormalizePathKey(row.Path)` in managed code and `ReadForMachine` groups by the same expression; nothing anywhere reads `PathKey`.

It is also inconsistent between row sources: new rows get the normalized key, backfilled rows get `TRIM("RepoPath")` unmodified. So the column carries two different meanings and a false description. Either make it the lookup key or drop it.

### 8. Every observation loads the machine's whole catalog

`Observe` narrows in the database by machine only, materializes the result with `.ToList()`, and then filters in managed code, all inside a process wide `lock`. That runs once per session per five minute freshness interval on the hosted Gateway, and each run materializes every repository row for that machine.

Filtering on `PathKey` in the database, which is what finding 7 says the column is for, turns this into a point lookup.

### 9. The only test of the HTTP route is in a parked suite

`KnownRepositoryEndpointTests` lives in `CcDirector.Gateway.Tests`, which the repository instructions (section 5a) name as parked and excluded from the default `scripts\test-local.ps1` run. That test is the only place the route, its ownership resolution and the JSON wire shape the mobile client parses are exercised. The verification list in the description does not say the Gateway suites were run. Please state whether `-Parked` was run, and on which commit.

Similarly `MobileViewportContractTests` (in `CcDirector.Core.Tests`, also parked) guards only `.terminal-screen`; the new pinned viewport screen is not covered. That matches how `.assistant-screen` was left, so it is an observation, not a defect.

### 10. The catalog is retained forever with no way to remove an entry

The entity comment makes this deliberate and thefrederiksen/devthrottle_internal#1853 asks for the complete historical set, so the intent is clear. But it does mean repository paths now outlive the ninety day session history retention indefinitely, per tenant, on the hosted service, with no delete path and no pruning. That is a retention decision worth an explicit ruling of its own rather than one that arrives inside a picker change.

### 11. Smaller points

- The agent error's "Retry" button calls `setRepositoryReload`. It works, because one effect loads both, but the name says otherwise to the next reader.
- Not verified, so treat as a question: `KnownRepositoryDto.LastUsed` is a bare `DateTime`. If the SQLite provider returns it with `DateTimeKind.Unspecified` it will serialize without a trailing `Z` while PostgreSQL emits an offset. The mobile client compares `lastUsed` as strings across two sources in `mergeRepositories`, so a format difference would only disturb tie-breaks, but a `DateTimeOffset` or an explicit kind stamp on read would remove the class of problem. I did not confirm the SQLite kind.

## What I checked and found correct

- Tenant boundary. The entity extends the Gateway minted key base, `ApplyTenantScope` is registered for it, `Observe` stamps `TenantId` explicitly, and the cross tenant and cross machine partition test is real and passes.
- Endpoint ownership. `TryResolveOwnedDirector` resolves in the caller's own tenant, and the machine name comes from the owned registration rather than from the caller, so a caller cannot name another machine. The second `ResolveReadTenant` call in the handler is redundant but harmless.
- The route is absent from `SessionKeyGuard`, which is correct: that is an allow list for agent session keys, and this is a device key read.
- Migration parity. Both providers create the same shape, both index `(tenant_id, MachineKey, LastUsedUtc)`, both back-fill from `session_history` with the same grouping, and both mint an identifier independent of any tenant supplied value. `Down` drops the table on both.
- Deduplication and normalization semantics. Trailing separators, separator direction and Windows case folding are handled, the Windows drive root is preserved, older observations never move facts backwards, and a blank name never overwrites a known one. The read path groups by the normalized path so duplicate rows, including any produced by two containers during a slot swap, collapse on read.
- Session history recording. `writeDue` is true on a session's first observation, so a short lived session is still catalogued. The swallow in `ObserveKnownRepository` matches the convention every other public entry point in that file already uses.
- Stale response handling. The generation counter plus the selected identifier check discards a previous Director's late agent and repository responses; the existing test proves it and I could not defeat it.
- Double submit. `createInFlightRef` is set synchronously before the first await, so two taps in the same tick produce one request.
- Narrow screen work. Tap targets are 44 pixels or more, paths use `overflow-wrap: anywhere`, the collapsed step summary ellipsizes rather than scrolling the page sideways, the screen follows the existing `.assistant-screen` pattern for the visible viewport and the top bars, the app bar correctly opts out of the sticky offset, and the footer respects the safe area inset.

## What I could not reach

- No real device or browser proof. I read the description's claim of a run at 390 by 844 pixels and did not repeat it.
- No PostgreSQL. The PostgreSQL backfill statement and the collation proof were not executed here; both are environment gated.
- `CcDirector.Gateway.Tests` did not finish inside my time budget; it takes the machine wide lock and was still waiting. So the endpoint test itself is unverified by me.

thefrederiksen:
> Posted as a comment rather than as a formal review state: GitHub refuses "request changes" on a pull request opened by the same account. Treat this as the review verdict.

## REQUEST CHANGES

Independent review of `b18cc2548296c393cb8832eaa3aeeca668ea061d`, verified against issues thefrederiksen/devthrottle_internal#1853 and thefrederiksen/devthrottle_internal#1884 rather than the description. The detailed findings are in the comment above; this is the verdict and the scope.

The design is the right one. The four step accordion is a genuine redesign rather than a search box bolted onto the old screen, which is what thefrederiksen/devthrottle_internal#1853 asked for; the Director, agent, repository and review sequence is what thefrederiksen/devthrottle_internal#1884 asked for; the durable Gateway catalog is the right place to answer "every repository this machine has ever run", and keeping it out of the ninety day history sweep is the correct call. The tenant and machine boundary is sound, the two migrations are genuinely parallel, and the stale response handling survived my attempts to break it. I am asking for changes on defects, not on the approach.

### What blocks

1. **The manual path field fails silently.** `createError` has exactly one render site, `NewSession.tsx:607`, inside the review section, but `chooseManualPath` sets it while the repository step is open. Tapping "Use this path" with an empty field produces no message and no alert. Reproduced.
2. **Retrying a repository source destroys the user's typed search, chosen repository and agent selection**, because the retry re-runs an effect that calls `clearDirectorChoices()`. Reproduced: the search field goes from `repository 1` to empty. The only action offered for an error costs the user their work, which is the opposite of thefrederiksen/devthrottle_internal#1884's "explicit unavailable and error states".
3. **The search result list has no cap.** Reproduced: a one character query over 400 repositories renders 400 rows. The Gateway deliberately declines to cap, which puts the cap on the client, and the client does not apply one. A large catalog is the premise of thefrederiksen/devthrottle_internal#1853.
4. **The SQLite backfill writes a machine key the read path cannot match** for any machine name containing a non-ASCII letter, because SQLite's `UPPER` folds ASCII only while `NormalizeMachineKey` uses `ToUpperInvariant`. `ReadForMachine` narrows on `MachineKey` in the database alone, so those backfilled rows are invisible: the migration's whole purpose is defeated for those machines. PostgreSQL's `UPPER` folds them, so the two providers also disagree.
5. **A Director registered with a blank machine name produces an unhandled 500** rather than an empty list or a stated error, because `ReadForMachine` throws `ArgumentException` on a blank value and the handler does not guard it. The registry permits a blank machine name and every other per-Director route tolerates one.

### What should be fixed but does not block

6. `MachineKey` is the indexed, ordinally compared lookup column and is not pinned to `UseCollation("C")`, against the rule stated and applied throughout `GatewayDbContext`.
7. `PathKey` is written, documented as used by reads and deduplication, and read by nothing; the backfill also puts an un-normalized value in it, so the column holds two different meanings.
8. `Observe` materializes every row for the machine on every write-due observation, inside a process wide lock, because it does not use the key column in 7.
9. The mobile page has no test at all for the manual path field, and no test for the retry path. Three throwaway probes against the real component found two defects in that untested area on the first attempt.
10. Dead style rules left behind by the rewrite: `.newsession-create`, `.newsession-manual` and `.newsession-manual-label` have no remaining references in any component.
11. Please state whether the parked suites were run, and at which commit. The route's only test lives in one of them.

### What I checked

Both migrations, including their backfill statements, indexes, identifier minting and `Down`; the tenant partition, the machine partition and endpoint ownership resolution; the absence of the route from the session key allow list, which is correct for a device key read; repository path normalization and deduplication on both the write and the read path; session history recording, including whether a short lived session is ever missed; the shared client's new call, its parsing and its error surface; the mobile page's stale Director response handling, its double submit guard and every loading, empty and error state; narrow viewport behavior in the stylesheet, tap target sizes, path wrapping, the pinned viewport pattern and the safe area inset; and the new and existing automated tests.

I ran `packages/client-core` (972 tests passed), `apps/mobile` (10 tests passed), and `CcDirector.Gateway.UnitTests` filtered to this feature (5 passed), plus three probe tests of my own against the real component. The worktree was not modified; `git status` is clean at the reviewed commit.

### What I could not reach

- **`CcDirector.Gateway.Tests` never ran.** It waited the full budget on the machine wide test lock, held throughout by process 14916 in `D:\ReposFred\devthrottle.wt-scrub-review`. So `KnownRepositoryEndpointTests` is unverified by me, and with it the live route, its ownership resolution and the JSON wire shape the mobile client parses.
- **No PostgreSQL.** The PostgreSQL backfill statement and the collation proof are environment gated and were not executed here.
- **No real device or browser proof.** I did not repeat the 390 by 844 pixel run described in the summary.
- The `DateTimeKind` question in point 11 of the findings comment is stated as unverified and should be treated as a question, not a defect.

thefrederiksen:
> Posted as a comment rather than as a formal review state: GitHub refuses a review verdict on a pull request opened by the same account. Treat this as the review verdict.

## PASS

Re-review of exact head `3bd4fdc277efa5e28e47d3c330781eabeecc148c`. All five blockers from the previous verdict are fixed, and I reproduced each fix rather than reading the diff and assuming.

Verified at that exact SHA in a detached worktree cut from it (`git rev-parse HEAD` = `3bd4fdc277efa5e28e47d3c330781eabeecc148c`, clean status) and, for the browser code, a `git archive` export of `apps` and `packages` at the same SHA. The earlier review worktree was deliberately left parked at the old head rather than switched, because switching a shared checkout is what the repository rules forbid; cutting a worktree from the wanted commit is the prescribed alternative. Every number below comes from `3bd4fdc`.

### The five blockers

1. **Manual path silent failure - fixed.** A render site now exists outside the review section (`activeStep !== "review" && createError !== null`), and typing into the field clears the message. My probe that failed against the old head now passes: the message is on screen after tapping "Use this path" with an empty field.
2. **Retry destroying user state - fixed, and fixed properly.** The single combined effect was split into an agent effect keyed on `agentReload` and a repository effect keyed on `repositoryReload`, each resetting only its own state, with `clearDirectorChoices()` removed from both. My probe confirms the typed search survives a repository retry. I also probed the direction nobody had tested - an agent reload - and it is clean: `getAgents` 2 calls, `getRepos` 1, `getKnownRepositories` 1. The two effects are genuinely independent, not independent in one direction only.
3. **Uncapped search - fixed.** Capped at 50 with the notice "Showing 50 of 400 matches. Type more to narrow the results." in my 400-repository probe, and every result stays reachable by narrowing. This is the shape agreed: the cap never silently hides a match.
4. **Non-ASCII machine key - fixed as agreed, and the fix does not depend on repair.** `CandidateMachineKeys` (invariant fold, legacy ASCII fold, exact stored name) is shared by `ReadForMachine` and `Observe`, so the predicate stays an indexed `IN` rather than becoming a tenant-wide managed scan. Reads are correct on their own, which matters because repair can never reach a retired machine. The PostgreSQL backfill moved from `UPPER` to `translate()`, so both providers now mint the same legacy key rather than two different ones. The new store test asserts `Assert.Single` after the observe, which is the real proof that `Observe` finds and repairs the legacy row instead of inserting a duplicate.
5. **Blank machine name - fixed.** An explicit 409 with a stated reason, covered by an endpoint test.

### The non-blocking points

- **Collation** - `MachineKey` and `PathKey` pinned to `"C"`, correctly inside the `Database.IsNpgsql()` guard. I confirmed the SQLite snapshot is untouched and the PostgreSQL snapshot carries `UseCollation("C")` on both columns, and the collation proof list moved 25 to 27 with exactly those two entries.
- **`PathKey`** - now genuinely load bearing: it is the database predicate on the fast path in `Observe`, and the property comment was corrected to say so.
- **`Observe` cost** - the ordinary path no longer materializes the machine catalog. The machine-wide fallback runs only when the fast path misses, and the row it finds is then repaired so the next observation is an index hit. There is still no index on `PathKey`, so this is an index seek on the machine plus a residual filter rather than a point lookup - fine, and strictly better than before.
- **Tests** - both tests I asked for are present and are the right shape: the real-migration test now uses a non-ASCII machine name and asserts `ReadForMachine` returns the backfilled row (it would have passed either way before), and the truncation test asserts the full match count is on screen rather than merely that fewer rows rendered.

### Unasked-for fix worth calling out

`repositoryKey` now case-folds only Windows paths, matching the server's `NormalizePathKey`. Before, it lower-cased everything, so two genuinely distinct POSIX paths such as `/repos/Project` and `/repos/project` were merged into one choice on the phone. That was a real latent defect nobody had reported, it was found and fixed in this round, and it has a test.

### Two notes for later, neither blocking

- The Directors effect still unconditionally resets the selection to `list[0]` and now also calls `clearDirectorChoices()`. That is unreachable today, because `directorReload` is only raised from the error state where no selection exists. But if a refresh path is ever added to that effect it becomes blocker 2 again, in a place that now has no test. A guard on "only adopt a default when nothing is selected" would close it permanently.
- Trivial leftover: `.newsession-create` is still in `apps/mobile/src/styles.css` with no reference in any component. `.newsession-manual` and `.newsession-manual-label` were removed; this one was missed.

### What I ran at this head

- `apps/mobile`: 14 of 14 passed.
- `packages/client-core`: 90 files, 972 of 972 passed.
- `CcDirector.Gateway.UnitTests --filter KnownRepository`: 6 of 6 passed, including the new non-ASCII legacy-key test.
- Five probe tests of my own against the real component: the three that failed against the previous head now pass, plus two new ones covering the rewritten asynchronous logic.
- `dotnet ef migrations has-pending-model-changes` on **both** providers: "No changes have been made to the model since the last migration." Independently confirmed, not taken from the summary.

### What I could not reach

- **`CcDirector.Gateway.Tests` did not run for me again.** The per-user suite lock was held throughout by a different holder than last time - process 54852, working directory `D:\ReposFred\devthrottle-turn-push` - for over 1,000 seconds, and my run aborted waiting. On the author's instruction I am not waiting a third time and am accepting the recorded 2,324-test completed result for that suite. So the endpoint tests and the PostgreSQL collation proof are author-reported, not reproduced by me.
- **No PostgreSQL database.** Worth stating plainly: the PostgreSQL backfill statement was *changed* in this round, from `UPPER` to `translate()`, and the SQLite migration test does not exercise it. Its only evidence is the collation catalog check and the author's own run. If anything in this change deserves a second look after merge, it is that statement.
- **No real device or browser proof.** I did not repeat the phone-viewport run.

The design was right the first time and the fixes are the ones that were agreed, not narrower substitutes. Passing.

thefrederiksen:
## Final review: PASS

Reviewed independently at exact head `3bd4fdc277efa5e28e47d3c330781eabeecc148c`, which I confirmed myself and which is level with `origin/main` (`168bdc7ab`). Read-only worktree; nothing modified. The review looked at the change's own diff (`168bdc7ab..HEAD`), not the merged-in turn-push work.

### The previous blockers, each checked in the code

**Visible manual-path validation.** `chooseManualPath` sets the error, and `NewSession.tsx` renders `{activeStep !== "review" && createError !== null && ...}` as a `role="alert"` banner inside the step list, so the message appears on the repository step where the user is standing rather than only on the review card. Covered by "reports a blank manual path and accepts a populated manual path".

**Retry state preservation.** The repository retry only bumps `repositoryReload`; the effect resets the two source lists and their errors and leaves `repositoryQuery`, `manualPath`, `selectedRepository`, the chosen agent and the chosen Director alone. The agent retry re-selects the same agent when it is still in the returned list. Covered by "preserves choices and typed input while retrying repository sources", which also asserts the agent route was not re-called.

**Fifty-result cap with the full count and narrowing guidance.** `SEARCH_REPOSITORY_COUNT = 50`, and when matches exceed what is rendered the page shows "Showing 50 of 60 matches. Type more to narrow the results." as a `role="status"` line, with every result still reachable by typing more. Covered by "caps broad search rendering, reports the full match count, and keeps every result reachable".

**Non-ASCII migration compatibility with bounded candidate keys shared by both paths.** The SQLite backfill writes `UPPER(TRIM(MachineName))` and the PostgreSQL backfill writes `translate(...)` over the ASCII alphabet only, so both leave a non-ASCII letter unfolded, while `NormalizeMachineKey` uses `ToUpperInvariant` and does fold it. `CandidateMachineKeys` returns exactly three bounded values (invariant-upper, ASCII-folded, raw trimmed) and is used by both `ReadForMachine` and `Observe`, so a migrated row is readable without waiting for a repair. `ReadAndObserve_LegacyAsciiFoldedNonAsciiMachine_FindsAndRepairsOneRow` proves the read on `Søren_North` against a legacy `SøREN_NORTH` key, and the migration test proves the same name end to end through the real migrator.

**Blank machine response.** The endpoint answers `409 Conflict` with "The Director has not reported a machine name." rather than an empty list, so the phone can tell "nothing known yet" from "the Director never said where it is". Proved by `GetKnownRepositories_DirectorHasNoMachineName_ReturnsExplicitConflict`.

**PathKey use and write performance.** The ordinary observe is one indexed query over `(tenant_id, MachineKey, LastUsedUtc)` with `PathKey` filtered in the database against four bounded candidates, so a repeat observation never materializes the machine catalog. The machine-scoped fallback is bounded by the repositories on one machine.

**PostgreSQL C collation.** `MachineKey` and `PathKey` carry `UseCollation("C")` in the model, in the PostgreSQL migration's column definitions and in the model snapshot, and both are added to the enumeration in `PostgresProviderProofTests`, whose count moves 25 -> 27 to match.

### Independent proofs I ran here

- `CcDirector.Gateway.UnitTests`, filter `KnownRepository`: 6 passed, 0 failed (store plus the real-migrator backfill test).
- `CcDirector.Gateway.Tests`, filter `KnownRepositoryEndpoint`: 3 passed, 0 failed (complete catalog served after the Director tunnel is closed, unknown Director 404, blank machine 409).
- Every web workspace, from a clean `npm ci`: client-core 972, cockpit 282, cc-assistant 106, mobile 14 - 1374 passed, 0 failed.
- `npm run typecheck` across all four workspaces: clean.

### Two notes, neither blocking

1. The comment above the candidate query in `KnownRepositoryStore.Observe` says the machine-scoped fallback "only runs when the original migration left an exact, unnormalized PathKey". That is not quite what happens: an exact unnormalized key is already one of the four candidates and is found by the first query, while the fallback in fact runs on every genuinely new repository and on a legacy row whose spelling differs in case or separator. The behaviour is right and bounded; the sentence describes it inaccurately, which is the kind of prose a later reader trusts.
2. `GET /directors/{id}/known-repositories` is not in `SessionKeyGuard`'s allow list, so a fleet session key is refused. The phone holds a device key, so the feature works as shipped; recording it so nobody later reads the refusal as a bug.

### What I did not verify

The on-screen layout. The new `newsession-*` rules key off `--app-vh` and `--topbars-h` and read as consistent with the rest of the mobile shell, but jsdom has no layout, so "the confirmation action stays above the software keyboard" is argued from the stylesheet and not observed on a device.

PR #2671, opened 2026-09-03
feat(gateway): say a rule in plain English and the Gateway writes it, grounded in a real screen

thefrederiksen:
Built and tested on the Session Rules mission worktree; this is the first of two slices landing it.

## What this is

Until now a standing instruction had to be assembled by hand - trigger words, checks, scope and the
two ceilings, written in the shape the store wanted. This adds the authoring conversation: you say
what you want in ordinary words, a model works out the rule, and you get it read back before
anything is stored.

## The three things that make it safe

**The words have to be on the screen.** A drafted rule's trigger words are checked against the real
captured terminal text, and a word that is not there is refused with the invented word named. A rule
watching for something never on the screen is one that sits in the list looking correct and never
fires.

**The rule is for one agent by default.** A usage-limit notice reads "Claude usage limit reached" on
Claude Code and something else elsewhere. When the screen came from a known session the rule is
scoped to that session's agent - a fact we hold, not a guess the model makes. Every agent is a star
you choose, never the default.

**A question back is a first-class answer,** as a decline is for the evaluator. A model that does not
know which sessions a rule is for says so and nothing is drafted. An empty scope is refused for the
same reason - reading omitted as widest would be a fail-open.

Authoring asks the thinking model (a person is waiting), wired to the brain directly rather than
through the evaluator's seam, which swallows every failure into "no answer" - right for an
unattended pass, wrong when somebody is watching a spinner and needs to know it timed out.

Storing a drafted rule still puts it in as a dry run. Promotion is still a separate act.

## The gate

Run locally on this commit's tree, not waited for on CI:

| Suite | Result |
| --- | --- |
| `.\scripts\test-local.ps1` (9 projects) | green, 4,848 tests, every outcome Completed |
| `Gateway.Tests` (parked) filtered to the two new classes | green, 7 tests |
| Every web workspace + `npm run typecheck` | green (979 + 106 + 292 + 14) |
| `pytest tools/cc-devthrottle/tests/test_rule_ops.py` | green, 10 tests |

The full parked `Gateway.Tests` run is the remaining coverage gap and runs before the second slice
merges; this slice touches `GatewayHost.cs`, so it is owed.

## What is NOT proven here

The grounding check is skipped entirely when no example screen is supplied, and nothing forces one
to be supplied. That is arguably right - a rule described from memory has no screen to check against
- but it means grounding is enforced only when a screen happens to be present. Called out for the
inspection rather than decided quietly.

### DIFF
--- a/docs/missions/session-rules-2026-09-02/guard-fix-report.md
+++ b/docs/missions/session-rules-2026-09-02/guard-fix-report.md
@@ -1,0 +1,229 @@
+# The rule routes reach the guard - fix report
+
+Branch `mission/rules-guard`, worktree `D:\ReposFred\devthrottle-rules-guard`, written 2026-09-03.
+Read it against the code, not against this document.
+
+---
+
+## What was broken
+
+`SessionKeyGuard` is a deliberate literal allow list of the routes a session key may call. Nothing
+under `/gateway/rules` was on it, so every rule route was refused with HTTP 403 and the code
+`session_key_out_of_scope`. The whole rule command group authenticates with a session key, so
+"set up a rule" could not work at all - not partly, not sometimes, at all.
+
+Every suite was green throughout, and that is the part worth keeping. The guard's own unit tests are
+written from the guard's own list, so they agreed with it. The command line's tests mock the
+transport, so they never sent a request. Nothing anywhere connected "a route was added" to "the guard
+was told about it". This is the third time the same mechanism has bitten this one file: the skill
+catalogue and the schedule surface each shipped refused-by-accident before it, and each was fixed by
+adding rows to the hand-kept list, which fixes the instance and leaves the mechanism running.
+
+---
+
+## What was built
+
+### The ruling, implemented exactly as the owner gave it
+
+An agent credential may do everything with rules EXCEPT move one out of dry run.
+
+| Verb | Route | Ruling |
+| --- | --- | --- |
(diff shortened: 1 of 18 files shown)


thefrederiksen:
DO NOT MERGE YET - an independent inspection returned FAIL with ten findings (one blocker, four high, five medium): `docs/missions/session-rules-2026-09-02/inspection-d.md`. All ten are accepted. The rulings are in `fix-round-d.md` and fix round D is in flight.

The blocker is product-level: the new command line authenticates with a session key and `SessionKeyGuard` authorizes none of the rule routes, so every rule command is refused before the endpoint runs - while every suite stays green, because the guard is tested against its own list and the command line's tests mock the HTTP. Fix in flight, and it carries a test that derives the route set from the built application so the next route cannot be forgotten.

The finding that changes the design is the grounding check. It was optional (an empty screen skipped it, and the page always sent one), caller-asserted (no session id, no server-side read), checked against different text than the model was shown, and defeatable by whitespace. Rather than patch four holes, the draft route stops accepting a caller-supplied screen: it takes a session id, the Gateway reads that session's screen itself, and the same check runs again at the create route, which is the one write gate. An ungrounded rule becomes structurally impossible rather than merely refused.

Two things the inspection cleared, both load-bearing: no path parses, compiles or evaluates generated program text, and a draft cannot promote itself.

PR #2672, opened 2026-09-03
feat(cockpit,cli): a Rules page and a rule command group, both thin clients of the one Gateway

thefrederiksen:
Second of two slices. Stacked on pull request 2671 (the Gateway half) so the diff shown here is only
the clients; it retargets to `main` when that merges.

## What this is

The authoring conversation is on the Gateway. This is the two ways of reaching it, and neither holds
any judgement of its own - both post what you said and render what comes back. That is the whole
reason "set up a rule from Claude Code" is the same feature as "set up a rule on the page" rather
than a second implementation of it.

**The Rules page** sits in the rail beside Workflows and Skills. It lists your rules with what each
wakes on, which sessions it acts on, its ceilings, and whether it is a dry run or live; and it takes
a sentence and shows you the drafted rule with its read-back before anything is stored. A refusal -
an invented trigger word, an unstated scope, a check that cannot be run - is shown in the Gateway's
own words, not as a generic failure.

**The command line** adds `cc-devthrottle rule` with `draft`, `add`, `screen`, `list`, `show` and
`delete`. It can pass a session, so an agent authoring a rule works from that session's real screen
rather than a paraphrase - which is what gives the grounding check something to check against.

Also lands the mission's own record: the implementation plan, the screen mockup, the measured timeout
diagnosis, the harness-first plan, and the authoring phase report.

## The gate

| Suite | Result |
| --- | --- |
| Every web workspace | green - client-core 979, cc-assistant 106, cockpit 292, mobile 14 |
| `npm run typecheck` (all four workspaces) | green |
| `pytest tools/cc-devthrottle/tests/test_rule_ops.py` | green, 10 tests |
| `.\scripts\test-local.ps1` (9 projects) | green, 4,848 tests |

Every workspace was run, not just `client-core` - a partial web run is a known false green in this
repository.

## What is NOT proven here

The Rules page has unit tests and a typecheck; it has NOT yet been driven in a browser against a
live Gateway. The mission's QA report carries that proof with screenshots, and this slice does not
claim it.


thefrederiksen:
DO NOT MERGE YET - an independent inspection returned FAIL with ten findings (one blocker, four high, five medium): `docs/missions/session-rules-2026-09-02/inspection-d.md`. All ten are accepted. The rulings are in `fix-round-d.md` and fix round D is in flight.

The blocker is product-level: the new command line authenticates with a session key and `SessionKeyGuard` authorizes none of the rule routes, so every rule command is refused before the endpoint runs - while every suite stays green, because the guard is tested against its own list and the command line's tests mock the HTTP. Fix in flight, and it carries a test that derives the route set from the built application so the next route cannot be forgotten.

The finding that changes the design is the grounding check. It was optional (an empty screen skipped it, and the page always sent one), caller-asserted (no session id, no server-side read), checked against different text than the model was shown, and defeatable by whitespace. Rather than patch four holes, the draft route stops accepting a caller-supplied screen: it takes a session id, the Gateway reads that session's screen itself, and the same check runs again at the create route, which is the one write gate. An ungrounded rule becomes structurally impossible rather than merely refused.

Two things the inspection cleared, both load-bearing: no path parses, compiles or evaluates generated program text, and a draft cannot promote itself.

PR #2681, opened 2026-09-04
fix(gateway): the turn log's staleness gap was flagging the healthy case

thefrederiksen:
Caught by the live corpus within two minutes of switching capture on: every record carried a "the conversation may be one turn short" gap, and every one was healthy.

The comparison was against the start of the capture. A push that landed **before** we read is the good case - it means the conversation contains the turn. The real risk is the reverse: the Director announces the state change before pushing the turn that caused it, so the conversation is short when its last push predates the **turn ending**. It compares against the session's last activity now.

A gap that marks good records as suspect is worse than no gap - it trains everyone to ignore the field, and then a real hole passes unnoticed.

Both directions have a test. Reintroducing the old comparison turns the healthy-case test red and nothing else; checked rather than assumed. Turn-log tests: 42 passing.

### DIFF
--- a/src/CcDirector.Gateway.UnitTests/TurnLog/TurnLogRecorderTests.cs
+++ b/src/CcDirector.Gateway.UnitTests/TurnLog/TurnLogRecorderTests.cs
@@ -219,24 +219,45 @@ public sealed class TurnLogRecorderTests
     }
 
     [Fact]
-    public async Task CaptureAsync_AConversationPushedBeforeTheCapture_IsNamedAsAGap()
+    public async Task CaptureAsync_AConversationPushedBeforeTheTurnEnded_IsNamedAsAGap()
     {
-        // The Director announces the state change BEFORE pushing the turn that caused it, so a capture can
-        // land between the two and store a conversation one turn short of the screen beside it. Unsaid, the
-        // record would claim to contain the turn it describes.
+        // The bad case: the last push predates the session's last activity, so the stored conversation stops
+        // short of the turn the screen beside it is showing.
         var env = new FakeEnvironment
         {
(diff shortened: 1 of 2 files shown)


thefrederiksen:
**Gate status.**

Fast gate green on this branch: Core.UnitTests 160, Avalonia 364, Engine 63, HostedAgent 88, Launcher 113, setup 25, setup-engine 456, all Completed with zero failures.

Two things in that run were NOT failures of this change and were checked rather than waved through:

- `CcDirector.Gateway.UnitTests` was stopped by the 120-second budget and produced no result file. Run alone it is **3,553 passing, 0 failed, 2m02s** - the known budget artefact, not a red.
- `Terminal.Avalonia.Tests` failed `Grok_path_link_does_not_flicker_on_footer_bytes` with a `RegexMatchTimeoutException` under load. Run alone: **24/24 passing**. A load flake, and it touches nothing this change goes near.

**The parked Gateway suite was not run, and the reason is not a shortcut.** Its per-user lock is currently held by a LIVE run in a different worktree (`devthrottle-2678`), acquired at 22:22 UTC and good for roughly fifty minutes. A queued run is the suite working as designed. That suite ran green on the parent of this change apart from the two pre-existing census failures filed as thefrederiksen/devthrottle_internal#1901, and this change is confined to one comparison inside `TurnLogRecorder` plus its tests.

Separately and still outstanding: the collation-census fix already on main (`turn_log_switches` account/machine) has NOT been verified by a run against real Postgres - it is queued behind the same lock. Its evidence so far is documentary: the recorded actual list from the parked run contained exactly those two tuples in exactly the sorted position they were added. That is strong, but it is not a run, and it should be confirmed once the lock frees.

PR #2684, opened 2026-09-04
Keep hosted Gateway logs off the durable file share

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1900

## Root cause

- The hosted Gateway's custom `FileLog` sink does not implement log levels, so changing informational logging settings cannot reduce this output. A live capture contained 138,708 lines and 25.8 megabytes in two hours and fifty minutes, which projects to about 218 megabytes per day.
- Every ten-second roster refresh re-sent every live session key. The Gateway rewrote the same database row and emitted two successful registration lines for each key. Those duplicate successes were 66,287 lines: 47.8 percent of all lines and about 53 percent of the bytes.
- There is no product background reader. The reads in the issue were the investigation itself: another troubleshooting run used Kudu at 20:06 and 20:07 universal time to execute `grep` across the dated log glob and then the exact current and old files in the storage evidence. Azure's one-minute metrics show the matching read spike and Server Message Block lease breaks. Those reads broke the active writer's lease and forced its cached writes across Azure Files, creating the apparent uptime step. Restarting opened a new file and restored the write lease; it did not reset an accumulating product reader.
- The traffic reached Azure Files because `FileLog` inherited its directory from `CcStorage.ToolLogs`, which is the durable mounted share in the hosted image.

## Change

- Configure the hosted image's process log beneath the container's temporary directory before the logger starts. Durable product state remains on Azure Files.
- Accept an unchanged session-key refresh without a database rewrite or success line while more than half of the key lifetime remains. Real rotation, shorter requested expiry, revocation, ownership refusal, and eventual refresh are unchanged.
- Remove the second routine success line from the hub while retaining explicit refusal records.
- Add focused tests for the hosted path boundary and the session-key refresh behavior.

This does not add retention, rotation, or cleanup.

## Verification

- Focused tests on the current head: 4 Core logger tests and 30 Gateway hosted-path, session-key, and log-capture tests passed, 0 failed.
- The full local gate built commit `d8c4e3b48` with 0 warnings and 0 errors, then ran all 11 test projects. Nine projects passed. Gateway reported 2,322 passed, 4 skipped, and exactly the two route-census failures tracked by thefrederiksen/devthrottle_internal#1901; `origin/mission/rules-fix-f` already owns their correction and tenancy proof. Core reported 4,385 passed, 8 skipped, and one transient failure in the unchanged `LifecycleSignalTests.EachRaise_RunsTheHandlerExactlyOnce`; an immediate exact-name rerun passed 1 of 1.
- The hosted continuous check independently passed that lifecycle-signal test and failed only the same two thefrederiksen/devthrottle_internal#1901 route-census assertions.

### DIFF
--- a/src/CcDirector.Core.Tests/Utilities/FileLogConsoleMirrorTests.cs
+++ b/src/CcDirector.Core.Tests/Utilities/FileLogConsoleMirrorTests.cs
@@ -10,8 +10,8 @@ namespace CcDirector.Core.Tests.Utilities;
 //   1. MirrorToConsole puts every line on standard output as well - the sink the
 //      container platform captures per container, and the one with no queue in front of
 //      it, so it survives the file-share stall that erased three startup records.
-//   2. UseUniqueInstanceId REFUSES to run after Start(), instead of quietly leaving the
-//      process writing to the shared file it was supposed to move off.
+//   2. UseUniqueInstanceId installs the caller-selected directory before Start(), and
+//      REFUSES to run after Start() instead of splitting one process record across files.
 // =====================================================================================
 public sealed class FileLogConsoleMirrorTests
 {
@@ -74,5 +74,30 @@ public sealed class FileLogConsoleMirrorTests
 
         var ex = Assert.Throws<InvalidOperationException>(FileLog.UseUniqueInstanceId);
         Assert.Contains("before FileLog.Start()", ex.Message);
+    }
+
+    [Fact]
+    public void UseUniqueInstanceId_WithDirectory_WritesToTheSelectedDirectory()
+    {
+        var dir = Path.Combine(Path.GetTempPath(), "cc-filelog-selected-" + Guid.NewGuid().ToString("N"));
+        try
+        {
+            FileLog.UseUniqueInstanceId(dir);
+            FileLog.Start();
+            FileLog.Write("[FileLog] selected directory proof");
+            FileLog.Stop();
+
+            var file = Assert.Single(Directory.GetFiles(dir, "director-*.log"));
+            Assert.Equal(Path.GetFullPath(dir), Path.GetDirectoryName(Path.GetFullPath(file)));
+            Assert.Contains("selected directory proof", File.ReadAllText(file));
+        }
+        finally
+        {
+            FileLog.Stop();
+            FileLog.UseUniqueInstanceId();
+            try { if (Directory.Exists(dir)) Directory.Delete(dir, recursive: true); }
+            catch (IOException) { }
+            catch (UnauthorizedAccessException) { }
+        }
     }
 }
--- a/src/CcDirector.Core/Storage/CcStorage.cs
+++ b/src/CcDirector.Core/Storage/CcStorage.cs
@@ -17,7 +17,7 @@ namespace CcDirector.Core.Storage;
 ///   CC_DIRECTOR_INSTANCES_DIR - Override the Director instance-discovery directory specifically
 ///
 /// NOTE: CcStorage methods intentionally omit FileLog.Write calls because
-/// FileLog.LogDir is initialized from CcStorage.ToolLogs(), creating a
+/// FileLog's default directory is initialized from CcStorage.ToolLogs(), creating a
 /// circular dependency at static initialization time.
 /// </summary>
 public static class CcStorage
--- a/src/CcDirector.Core/Utilities/FileLog.cs
+++ b/src/CcDirector.Core/Utilities/FileLog.cs
@@ -3,7 +3,8 @@ using CcDirector.Core.Storage;
 namespace CcDirector.Core.Utilities;
 
 /// <summary>
-/// Simple thread-safe file logger. Writes to cc-director logs/director/ directory.
+/// Simple thread-safe file logger. Writes to cc-director logs/director/ by default; a hosted process may
+/// select container-local storage before it starts.
 ///
 /// The actual dequeue/rollover/flush work lives in <see cref="FileLogWriter"/> so the day-rollover
 /// behavior can be unit-tested with an injectable clock (issue #171). This type is the thin static
@@ -12,7 +13,8 @@ namespace CcDirector.Core.Utilities;
 /// </summary>
 public static class FileLog
 {
-    private static readonly string LogDir = CcStorage.ToolLogs("director");
+    private static readonly string DefaultLogDir = CcStorage.ToolLogs("director");
+    private static string _logDir = DefaultLogDir;
 
     // What distinguishes this process's log file from another process's. The desktop uses the process id,
     // which is unique there and is what the desktop tooling globs for. A container must call
@@ -27,7 +29,7 @@ public static class FileLog
     // This comment used to say "Production never reassigns it", which was true and was the bug: a stopped
     // writer was restarted rather than replaced, so every later line threw. See Start.
     private static FileLogWriter _writer =
-        new(LogDir, _instanceId, () => DateTime.Now);
+        new(_logDir, _instanceId, () => DateTime.Now);
 
     private static int _started;
 
@@ -48,20 +50,34 @@ public static class FileLog
     /// leaving the process on the shared file.
     ///
     /// A hosted Gateway needs this because the default discriminator - the process id - is always 1 inside
-    /// a container, so every container computed the same path. Two of them then appended to one file on an
-    /// SMB share mounted <c>nobrl</c>, where the FileShare.Read request is not enforced across clients, and
-    /// clobbered each other mid-record. The token is generated per process rather than read from the
-    /// environment so its uniqueness does not depend on the platform supplying anything.
+    /// a container, so every container computed the same path. Two of them then appended to one file on a
+    /// Server Message Block share mounted <c>nobrl</c>, where the FileShare.Read request is not enforced
+    /// across clients, and clobbered each other mid-record. The token is generated per process rather than
+    /// read from the environment so its uniqueness does not depend on the platform supplying anything.
+    ///
+    /// A caller may also choose a different directory before the writer starts. The hosted Gateway uses
+    /// that seam to keep its process log on the container's temporary disk instead of the durable workload
+    /// share. Durable product state still uses <see cref="CcStorage"/>; a process log is not product state.
     /// </summary>
-    public static void UseUniqueInstanceId()
+    public static void UseUniqueInstanceId() => UseUniqueInstanceId(DefaultLogDir);
+
+    /// <summary>
+    /// Give this process a unique log file in a caller-selected directory. Like the parameterless overload,
+    /// this must be called before <see cref="Start"/>.
+    /// </summary>
+    public static void UseUniqueInstanceId(string logDirectory)
     {
         if (_started != 0)
             throw new InvalidOperationException(
                 "FileLog.UseUniqueInstanceId() must be called before FileLog.Start(); the writer is already " +
                 "running on the shared log file and moving it now would split this process's record in two.");
 
+        if (string.IsNullOrWhiteSpace(logDirectory))
+            throw new ArgumentException("A log directory cannot be blank.", nameof(logDirectory));
+
         _instanceId = Guid.NewGuid().ToString("N")[..12];
-        _writer = new FileLogWriter(LogDir, _instanceId, () => DateTime.Now);
+        _logDir = logDirectory;
+        _writer = new FileLogWriter(_logDir, _instanceId, () => DateTime.Now);
     }
 
     /// <summary>
@@ -129,7 +145,7 @@ public static class FileLog
             return;
 
         if (_writer.IsSpent)
-            _writer = new FileLogWriter(LogDir, _instanceId, () => DateTime.Now);
+            _writer = new FileLogWriter(_logDir, _instanceId, () => DateTime.Now);
 
         _writer.OnSinkHealthChanged = OnSinkHealthChanged;
         _writer.Start();
@@ -174,7 +190,7 @@ public static class FileLog
 
     /// <summary>Returns the current log file path (useful for display).</summary>
     public static string CurrentLogPath =>
-        Path.Combine(LogDir, $"director-{DateTime.Now:yyyy-MM-dd}-{_instanceId}.log");
+        Path.Combine(_logDir, $"director-{DateTime.Now:yyyy-MM-dd}-{_instanceId}.log");
 
     /// <summary>
     /// TEST-ONLY seam (issue #862). Redirects FileLog to a private, throwaway directory for the life
--- a/src/CcDirector.Gateway.UnitTests/Account/GatewayDeviceKeyStoreTests.cs
+++ b/src/CcDirector.Gateway.UnitTests/Account/GatewayDeviceKeyStoreTests.cs
@@ -9,6 +9,7 @@ namespace CcDirector.Gateway.Tests.Account;
 /// install id, that a different install id yields no key, that clearing removes it, and - DT-05 - that the
 /// raw key value is never written to the log even though it is persisted to disk.
 /// </summary>
+[Collection(CcDirector.Gateway.Tests.FileLogCaptureCollection.Name)]
 public sealed class GatewayDeviceKeyStoreTests
 {
     private const string Install = "install-857";
--- a/src/CcDirector.Gateway.UnitTests/FileLogCaptureCollection.cs
+++ b/src/CcDirector.Gateway.UnitTests/FileLogCaptureCollection.cs
@@ -1,0 +1,13 @@
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// FileLog.RedirectForTests swaps one process-wide writer, so every test that owns that seam must run with
+/// no other test class beside it. The Gateway unit assembly otherwise runs four classes in parallel.
+/// </summary>
+[CollectionDefinition(Name, DisableParallelization = true)]
+public sealed class FileLogCaptureCollection
+{
+    public const string Name = "FileLog capture";
+}
--- a/src/CcDirector.Gateway.UnitTests/HostedGatewayLoggingTests.cs
+++ b/src/CcDirector.Gateway.UnitTests/HostedGatewayLoggingTests.cs
@@ -1,0 +1,32 @@
+using CcDirector.Gateway;
+using Xunit;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// The hosted Gateway's process log is diagnostic output, not durable product state. Issue #2678 found that
+/// keeping it below the Gateway storage root made it most of the Azure Files traffic and let an unrelated
+/// reader break the active file's write lease. These tests pin the location boundary, not a cleanup policy.
+/// </summary>
+public sealed class HostedGatewayLoggingTests
+{
+    [Fact]
+    public void Hosted_process_log_lives_below_the_container_temporary_root()
+    {
+        var temporaryRoot = Path.Combine(Path.GetTempPath(), "hosted-log-root");
+
+        var directory = GatewayEntryPoint.ResolveHostedLogDirectory(temporaryRoot);
+
+        Assert.Equal(
+            Path.Combine(Path.GetFullPath(temporaryRoot), "devthrottle", "logs", "director"),
+            directory);
+    }
+
+    [Theory]
+    [InlineData("")]
+    [InlineData("   ")]
+    public void Hosted_process_log_refuses_an_absent_temporary_root(string temporaryRoot)
+    {
+        Assert.Throws<ArgumentException>(() => GatewayEntryPoint.ResolveHostedLogDirectory(temporaryRoot));
+    }
+}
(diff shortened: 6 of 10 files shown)


thefrederiksen:
## Review of commit `6f7c384c968344330d671f4ca1f9d20fba9ba930`

Read-only review of the whole change against issue thefrederiksen/devthrottle_internal#1900. Nothing was edited, committed or pushed.

**Scope reviewed:** the complete seven-file difference, both new test classes, the claimed production root-cause chain in the pull request body, and the surrounding code the change depends on — `FileLogWriter`, `GatewayHostedMode`, `SessionGatewayKeys`, the session-key indexes in `GatewayDbContext`, the Dockerfile, and the hosted provisioning workflow.

**Verdict:** the root-cause chain holds up and I found no correctness defect in the change. Six findings follow. None of them is a correctness blocker. Two of them are cheap and I would close them before merge, because they are exactly the coverage the local gate itself flagged on this commit.

---

### What I checked and confirmed

**Log placement is genuinely off the share.** The Azure Files share is mounted at `/home/gateway/cc-director` (`.github/workflows/provision-hosted-gateway-slots.yml:56`), which is also `CC_DIRECTOR_ROOT` (`Dockerfile:125`). Only that path is mounted, so `Path.GetTempPath()` — `TMPDIR` or `/tmp` on the Linux container — is container-local disk. `FileLogWriter.Start()` calls `Directory.CreateDirectory(_logDir)`, so the new directory is created rather than failing silently. Startup still writes `log: {FileLog.CurrentLogPath}` and hosted mode already sets the console mirror, so the new location announces itself on the platform-captured stream.

**Session-key ownership and rotation survive the coalescing branch.** I walked every path that could have been weakened:

- The row lookup is scoped by tenant, so a coalesced refresh cannot cross tenants, and skipping the `TenantId` write is safe because the row already carries that tenant.
- A revoked row is refused above the branch, so coalescing cannot resurrect a revoked key.
- The foreign-Director refusal also sits above the branch, and the branch re-checks the Director itself, so a takeover attempt cannot be coalesced into an acceptance.
- A changed hash bypasses the branch and rotates immediately; `A_directors_own_reseed_still_rotates_its_own_session_key` covers that with more than half the lifetime remaining.
- Skipping the "this hash belongs to another session" query on the coalesced path is safe: `KeyHash` carries a unique index (`GatewayDbContext.cs:1033`) and the session already owns the row holding that hash.
- The write cadence does not collapse in the second half of a key's life. `SessionGatewayKeys.RegistrationFor` recomputes the expiry as the Director's clock plus twelve hours on every send (`SessionGatewayKeys.cs:73` and `:80`), so the requested expiry keeps advancing and the condition stays satisfied. That is one write per session per six hours, not a return to one per ten seconds. I checked this case specifically, because an expiry fixed at mint time would have made the branch degenerate into rewriting on every refresh for six hours.

**Local gate on this commit.** I ran the default `scripts\test-local.ps1` in a worktree at `6f7c384c9`: nine suites, every one `outcome=Completed`, zero failures, and `CcDirector.Gateway.UnitTests` at 3556 passed — including the twenty-five focused tests the pull request cites. The two `ContextLessRouteCensusTests` assertions did not appear in this run because those tests live in the parked project.

---

### Findings

**1. The seam that actually moves the log has no test.** `FileLog.UseUniqueInstanceId(string)` is not exercised anywhere — I searched all three test projects. `HostedGatewayLoggingTests` covers `GatewayEntryPoint.ResolveHostedLogDirectory`, which is a path join plus an argument check; both of its cases would still pass if `UseUniqueInstanceId(string)` ignored its argument outright. Nothing pins that `CurrentLogPath` follows the supplied directory, or that the writer opens a file there. This is one small test, and it is the one that would catch a regression in the fix itself.

**2. The parked suites cover the changed code and the pull request does not report one of them.** On this commit the gate prints, verbatim: `COVERAGE GAP - this change touches code covered by PARKED suite(s) that did not run: CcDirector.Core.Tests, CcDirector.Gateway.Tests`. `FileLog` lives in `CcDirector.Core`, and its regression suite — `FileLogConsoleMirrorTests`, `FileLogRestartAfterStopTests`, `FileLogWriterTests` — is in the parked `CcDirector.Core.Tests`. The pull request names the two route-census assertions, which are in the parked `CcDirector.Gateway.Tests`, so that suite evidently ran; nothing says `CcDirector.Core.Tests` did. Please state its result at `6f7c384c9`.

**3. The expiry cap still writes a log line on every coalesced refresh when a Director's clock runs ahead.** The cap computation had to move above the coalescing branch because it feeds `effectiveExpiry`, but the `FileLog.Write` beside it did not (`SessionKeyRegistry.cs:141` to `:143`). A Director whose clock is ahead of the Gateway's now emits one cap line per session per ten-second roster refresh, and then the branch returns without writing anything else. That is the same shape of routine per-refresh line this change set out to remove, left standing on a neighbouring path. It is not a regression — today's code logs it just as often — but moving that single write below the branch would close it while the reasoning is fresh.

**4. The surviving success line reports the expiry that was asked for, not the one that was granted.** `SessionKeyRegistry.cs:186` still logs `expires={expiresAtUtc:O}`. With the hub's success line removed, this is now the only success record, and on a capped registration it states an expiry the Gateway declined to grant. Logging `effectiveExpiry` instead makes the sole remaining record true.

**5. "A shorter requested expiry is never coalesced" is asserted only in a comment.** The code is right, but `A_shorter_expiry_than_the_cap_is_honoured` registers once and never re-registers. No test re-registers a shorter expiry against a long-lived row, which is the branch that stops coalescing from ignoring a Director asking for less authority than it already holds.

**6. The post-mortem trade is real and is written down nowhere.** The process log no longer survives the container, and two things follow that a future responder will meet cold. The virtual file system Kudu exposes is rooted at `/home`, so it can no longer list or fetch the log at all; and once a container is replaced, the only remaining record is whatever the platform retained from standard output. `docs/incidents/gateway-roster-500-2026-07-30.md` reconstructed its entire timeline from that persisted file. The trade is defensible and the console mirror is already on, but the pull request presents the move as pure removal of traffic. One sentence, in the code comment or in the operational documentation, would stop the next reader concluding the log is broken.

**Residual, correctly out of scope.** Growth is now bounded by the container's local disk rather than by a 6.5 gigabyte share, at roughly a hundred megabytes a day with no rotation. The issue was right to forbid retention as the fix; now that the cause is named, rotation on the container-local path is the natural follow-up and is a smaller, safer change there than it would have been on the share.

**One hardening worth considering alongside finding 1.** The boundary rests on `TMPDIR` not pointing under the mount. Nothing sets it today, and `ResolveHostedLogDirectory` has no guard that would notice if something did — the test as written cannot fail in that case. Refusing a resolved directory that falls under the storage root would make the boundary self-enforcing rather than environmental.

---

### Verification limits

- **The read attribution is not independently verified.** That the 1.41 gigabytes of reads were the investigation's own Kudu `grep` at 20:06 and 20:07 universal time rests on Azure diagnostic evidence I cannot re-derive from the repository. I checked it only for internal consistency, and it is consistent: I found no product reader of `logs/director` in the Gateway. Only `CcDirector.Avalonia` enumerates that directory, and the container does not run it. As the issue says of its own search, that is a negative result, not an identification.
- **Nothing was run in a container.** The claim that `Path.GetTempPath()` resolves off the mount is derived from the mount path in the provisioning workflow and from the Dockerfile environment, not observed in a running hosted container.
- **I did not run the parked suites.** I started a `-Parked` run and then stopped it on request, so it would not queue behind the authoritative run already in flight for this commit. Finding 2 above is therefore a request for that run's `CcDirector.Core.Tests` result, not a claim about it.
- The unparked gate result quoted above is my own run in this worktree at `6f7c384c9`, not a reading of another session's report.

thefrederiksen:
### Follow-up on finding 2 (parked coverage)

Finding 2 asked for the parked `CcDirector.Core.Tests` result at `6f7c384c9`, because that is where the `FileLog` regression suite lives and the default gate reports it as a coverage gap. The result has now been supplied by the session that owns the authoritative parked run: **4,385 passed, 8 skipped, 0 failed.**

Recording it here so the answer sits with the review rather than in a message. To be exact about provenance: this is a reported result, not one I ran or read a result file for — my own evidence remains the default, unparked gate quoted in the review above. With that reported, finding 2 is answered.

Findings 1, 3 and 5 — the missing direct test of `FileLog.UseUniqueInstanceId(string)`, the placement of the expiry-cap log line relative to the coalescing branch, and the missing test for a shorter requested expiry — are being addressed on a new head, which I will re-review as a delta against `6f7c384c9`.

thefrederiksen:
## Delta review: `6f7c384c9..21321cacb`

Scope: the second commit only (`21321cacb`, "Prove hosted log sink and expiry behavior") - four files, two of them production. No tests were run, as instructed; everything below is read from the tree at `21321cacb`.

**Verdict: not a pass. Two findings, both in the new test code. The production change in this delta is sound.**

### The production change is correct

`SessionKeyRegistry.Register` moves both log lines to after `transaction.Commit()` and changes the success line to report `effectiveExpiry` instead of `expiresAtUtc`. That second part is a real correction, not just tidying: the old line named an expiry the database never held whenever the Gateway capped a skewed Director clock. Capping stays visible - the first registration writes and logs it, and a coalesced session writes again once it drops under `RefreshWhenRemaining` (six hours, half of the twelve-hour `MaxSessionKeyLifetime`), so a persistently skewed clock still leaves a record roughly twice a day instead of every ten seconds. The only thing lost is the capping record on a registration that is subsequently refused, and those refusals name their own reason on their own line. I am satisfied with this.

Widening the second `Register` in `A_roster_refresh_within_the_window_is_accepted_without_a_write` from `now + MaxSessionKeyLifetime` to `now + TimeSpan.FromDays(1)` is a genuine strengthening - it now exercises the capped-and-coalesced case rather than the exactly-at-ceiling one.

### Finding 1: the delta's central proof fails open

`src/CcDirector.Gateway.UnitTests/SessionKeyRegistryTests.cs:245` - `Assert.Empty(refreshLines)`.

This is the assertion the whole commit exists to make, and its pass condition is an absence. Nothing in this test, this class, or this assembly ever demonstrates that `FileLog.RedirectForTests()` is capable of capturing a `[SessionKeyRegistry] Register:` line at all. An empty capture is scored identically to a correctly-silent refresh.

Concretely: if the success line were later emitted through a different sink, or moved outside the scope's window, or the capture seam were disturbed by the race in finding 2, this test stays green while the exact noise regression it was written to prevent ships. The reader has no way to tell "the refresh was silent" from "the instrument recorded nothing".

The repository already has the right shape one directory over, in `src/CcDirector.Gateway.UnitTests/Account/GatewayDeviceKeyStoreTests.cs:86-88`, where the absence assertion (`Assert.DoesNotContain(lines, ... Key)`) is paired with a positive control (`Assert.Contains(lines, ... "stored per-device key for install_id=")`) that proves the capture worked.

Suggested fix, and it is cheap: inside the same scope, perform one registration that takes the write path and assert its success line **is** present, then assert the coalesced refresh contributed none. That converts the claim from "we saw nothing" to "we saw exactly the one line we expected and not the other".

### Finding 2: the seam's stated safety premise does not hold in this assembly

`FileLog.RedirectForTests` documents its own precondition at `src/CcDirector.Core/Utilities/FileLog.cs:200-203`: swapping the single process-wide static writer "is safe because the test assemblies disable parallelization, so exactly one test owns FileLog at a time."

That is true of `CcDirector.Core.Tests`, which sets `[assembly: CollectionBehavior(DisableTestParallelization = true)]`. It is **not** true of `CcDirector.Gateway.UnitTests`, which sets `[assembly: CollectionBehavior(MaxParallelThreads = 4)]` at `AssemblyParallelism.cs:21` - parallelization on, capped at four threads. `SessionKeyRegistryTests` is in no serial collection, so up to three other test classes run beside it.

The scope mutates two statics, `_writer` and `_started`. Two consequences follow:

- Any concurrently running class that calls `FileLog.Write` has its lines land in this scope's temporary directory. That makes `Assert.Empty` fail for a reason nobody caused - a flaky red.
- Two overlapping scopes each restore `_writer` and `_started` to what they captured on entry, so the later `Dispose` can leave the static writer pointing at a stopped writer with `_started` at zero. `FileLog.Write` then returns early, the registry's line is never written anywhere, and `Assert.Empty` passes for entirely the wrong reason - a false green, and the same fail-open direction as finding 1.

This condition is pre-existing, not introduced here - `GatewayDeviceKeyStoreTests` already used the seam in this assembly. But that test carries a positive control, so the false-green direction would surface there; this delta adds the first assertion in this assembly whose pass condition is emptiness, which is the one shape that breaks in both directions at once. I am reporting it because the delta increases the exposure, not to charge it to this change.

Cheapest correct fix: put the FileLog-capturing tests of this assembly into a `[CollectionDefinition(..., DisableParallelization = true)]` collection. A partial mitigation, worth doing regardless, is to filter the captured lines to those containing the session identifier before asserting - that removes the cross-talk red, though not the fail-open green, which finding 1 covers.

### The Core test reads correctly

`FileLogConsoleMirrorTests.UseUniqueInstanceId_WithDirectory_WritesToTheSelectedDirectory` mutates real global `FileLog` state rather than using the redirect seam, which is fine here because that assembly disables parallelization outright. The teardown is sound: the duplicated `FileLog.Stop()` is a no-op on the second call (guarded by `Interlocked.CompareExchange`), and the trailing `FileLog.UseUniqueInstanceId()` restores the default directory. The assertion is a presence, not an absence, and it names the directory explicitly. No objection.

### What I did not check

I did not execute any test, so I have not observed either finding fail in a run - both are read from the source and from the seam's own documented precondition. I did not re-review `6f7c384c9`, which is outside the assigned range, so the hosted temporary-directory placement, the `CcStorage` change, `GatewayEntryPoint`, `DirectorHub`, and `HostedGatewayLoggingTests` are not covered by this comment.

thefrederiksen:
## Delta review: `21321cacba47d7bfd5a34e9eadaace8320b7d79e..d8c4e3b483c917e84a494535a2f2589c12b22ed0`

**Reviewed head:** `d8c4e3b483c917e84a494535a2f2589c12b22ed0` ("Make FileLog capture proof fail closed"), the single commit in this delta.

**Verdict: PASS.** Both things this delta was asked to prove are proved, and I found no defect in it. Nothing was edited, committed or pushed.

---

### 1. The unchanged-roster log assertion now has a positive control and target-session protection - CONFIRMED

`SessionKeyRegistryTests.cs:229` to `:252`.

**Positive control.** The first `Register` moved from outside the redirect scope to inside it, so the scope now captures a line that must be there. `Assert.Empty(refreshLines)` became `Assert.Single(registrationLines)`. That is the change from an absence to a presence: the old assertion passed identically whether the ten-second refresh wrote nothing or the capture seam itself was broken and recorded nothing at all. The new one fails in the second case.

I checked the control is genuinely one line rather than one line by luck, by reading the production path at this head (`src/CcDirector.Gateway/Pairing/SessionKeyRegistry.cs`):

- The first registration asks for `now + MaxSessionKeyLifetime`, and the ceiling is `issued + MaxSessionKeyLifetime` with `issued == now`. The cap test is `expiresAtUtc > ceiling`, so equality does not cap, no cap line is written, and the success line at `:186` is the only line. Exactly one.
- The refresh ten seconds later coalesces and returns before any write: the hash matches, the Director matches, `effectiveExpiry` (capped to `2026-09-05T00:00:10Z`) is at or after the stored expiry (`2026-09-05T00:00:00Z`), and the remaining lifetime (twelve hours less ten seconds) exceeds `RefreshWhenRemaining` (six hours). So the second registration contributes zero lines, which is what `Assert.Single` is asserting - and the cap line no longer fires on that path, since it now sits below the coalescing branch.

**Target-session protection.** `Assert.Contains($"session={session}", recorded, OrdinalIgnoreCase)` pins the surviving line to this test's own session identifier, so a line produced by anything else cannot be mistaken for the control. `Assert.Contains($"expires={initialExpiry:O}", recorded, Ordinal)` additionally distinguishes the first registration from any later refresh line, so a refresh line standing in for the control also fails. Both assertions match the real line shape at `:186`, `Register: session={id}, director={row.DirectorId}, expires={effectiveExpiry:O} (key value never received)`.

### 2. Every `FileLog.RedirectForTests` caller in `CcDirector.Gateway.UnitTests` is in a `DisableParallelization` collection - CONFIRMED

I searched the whole assembly at this head rather than only the changed files. There are exactly two callers, and both carry the attribute:

| Caller | Line | Collection attribute |
| --- | --- | --- |
| `Account/GatewayDeviceKeyStoreTests.cs` | 77 | `[Collection(CcDirector.Gateway.Tests.FileLogCaptureCollection.Name)]` at line 12 (added by this delta) |
| `SessionKeyRegistryTests.cs` | 235 | `[Collection(FileLogCaptureCollection.Name)]` at line 18 (added by this delta) |

The definition is `[CollectionDefinition("FileLog capture", DisableParallelization = true)]` in the new `FileLogCaptureCollection.cs`. No class carries two collection attributes, and the new name collides with neither `"DirectorRoot"` nor `"serial-process-global-metrics"`.

The guard is load-bearing here rather than decorative: this assembly runs with parallelization enabled under `[assembly: CollectionBehavior(MaxParallelThreads = 4)]` (`AssemblyParallelism.cs`), and `FileLog.RedirectForTests` swaps the single static `_writer` for the whole process (`src/CcDirector.Core/Utilities/FileLog.cs:204` to `:225`). Without the marker, a class in another collection writing a log line during the scope would land in the capture, and the newly strengthened `Assert.Single` is exactly the assertion that would have started failing. The shape matches the existing precedent in the same assembly, `SerialMetricCollection`.

That also makes true, inside this assembly, the premise the `RedirectForTests` documentation comment already asserts ("the test assemblies disable parallelization, so exactly one test owns FileLog at a time"). For completeness, the other two assemblies that call the seam satisfy it a different way: `CcDirector.Core.Tests` and `CcDirector.Gateway.Tests` both carry `[assembly: CollectionBehavior(DisableTestParallelization = true)]`.

---

### Observations, none of them blocking

**A. Nothing stops the next caller from forgetting the attribute.** The guard is a convention held by two hand-written attributes. A new test class that calls `FileLog.RedirectForTests` without joining the collection reintroduces the hazard silently, and the symptom would be an intermittent failure in someone else's test rather than in the new one. This assembly already has the machinery to close that - the `Rules/` architecture tests - and a rule reading "every type that calls `FileLog.RedirectForTests` carries `[Collection(FileLogCaptureCollection.Name)]`" would make the guard self-enforcing. It is the same fail-open shape this delta just closed, one level up.

**B. The other capture assertion was already fail-closed.** `GatewayDeviceKeyStoreTests.Save_PersistsKeyToDisk_ButNeverLogsIt` pairs its `Assert.DoesNotContain` (the DT-05 secret check) with `Assert.Contains(lines, line => line.Contains("stored per-device key for install_id="))`. That positive control means a dead capture fails the test rather than certifying it. No change needed; recording it so the next reader does not have to re-derive it.

### Scope and limits

- **Scope was the delta only:** `21321cacba47d7bfd5a34e9eadaace8320b7d79e..d8c4e3b483c917e84a494535a2f2589c12b22ed0`, which is three files, twenty-five insertions and four deletions. I read the surrounding production code (`SessionKeyRegistry.Register`, `FileLog.FileLogTestScope`, `AssemblyParallelism.cs`) only as far as needed to judge whether the assertions are true.
- **Commit `21321cacb` is not re-reviewed here.** That is the commit carrying findings 1, 3 and 5 from the review of `6f7c384c9` - the direct test of the hosted log sink, the move of the expiry-cap log line, and the shorter-expiry test. This comment does not clear those; it clears only the fail-closed capture proof and the parallelization guard on top of them. I did observe, while checking assertion validity, that the cap line now sits below the coalescing branch and the success line reports `effectiveExpiry`.
- **Nothing was run and nothing was built.** The owner session holds the full Gateway gate, so I did not compile the delta or run any suite. Compilation is therefore reasoned, not observed: the attribute shape is identical to `SerialMetricCollection` which already compiles in this assembly, and `ImplicitUsings` is enabled so `StringComparison` and `IReadOnlyList<string>` resolve.
- **The no-overlap property of `DisableParallelization` is taken from the xUnit version 2 assembly runner's contract** - collections marked non-parallelizable run in their own sequential phase - not from a stress run measured here.
- All facts above were read from the named commit objects, not from a working tree.

thefrederiksen:
Final local verification at `d8c4e3b483c917e84a494535a2f2589c12b22ed0`:

- `scripts/test-local.ps1 -Parked` built with 0 warnings and 0 errors and produced test records for all 11 projects.
- Nine projects passed. Across the full run, 11,557 tests passed and 14 skipped.
- Gateway reported 2,322 passed, 4 skipped, and 2 failed. The failures were exactly `ContextLessRouteCensusTests.The_hosted_context_less_route_set_is_exactly_the_ruled_census` and `ContextLessRouteCensusTests.The_selfhost_context_less_route_set_adds_only_the_hosted_denied_families`, both on the existing `DELETE /gateway/rules/{id:guid}` census difference tracked by thefrederiksen/devthrottle_internal#1901.
- Core reported 4,385 passed, 8 skipped, and 1 failed in the unchanged `LifecycleSignalTests.EachRaise_RunsTheHandlerExactlyOnce`. An immediate exact-name rerun passed 1 of 1. This branch changes neither that test nor `LifecycleSignal`.
- Focused verification on this head passed 4 Core logger tests and 30 Gateway hosted-path, session-key, and log-capture tests with 0 failures.

The separate final review is recorded in the preceding comment and covers the exact pull request head.

PR #2712, opened 2026-09-06
The Architect is a human-facing seat again

thefrederiksen:
Closes thefrederiksen/devthrottle_internal#1920.

The owner, 6 September 2026:

> "parking the architect seat is wrong. the architect is always the session i talk to."
> "no the architect should push to me it is what i talk to."

That overturns his amendment of 9 July 2026, which the Gateway shipped on 3 September in commit 2a8679007 (#2667). An Architect stopped being visible: it parked as "Snoozed" the moment it stopped working, dropped out of the needs-you count, and the wingman stopped reading it aloud. The seat he addresses was the one seat that had been silenced.

## What changed

`Architect` leaves `SessionOrdering.IsSupervised`. That one predicate is read by the colour, the label, the triage bucket and the wingman's voice-mode sweep, so a single edit restores all three things being human-facing means: a stopped Architect goes red, it counts in the needs-you total, and it is narrated. A Worker with a live supervisor and a scheduled run are untouched and still park exactly as they did.

The wingman needs no repair pass. Its sweep enrols every session that is not supervised and is not already marked, so an Architect switched off while the earlier rule stood is switched back on by the next tick.

## The orphan arm, checked rather than carried over

`SupervisorLostNotice` already returned null for an explicit Architect carrying a controller. The behaviour is right; the reason it gave was not. It said "supervised by declaration, so losing a spawner strands nothing" - and an Architect is not supervised at all any more.

It still says nothing, for a reason that never depended on supervision: this notice explains a session that was quiet WHILE SUPERVISED and has suddenly started asking. An Architect never went quiet. It takes direction from the owner, not from whoever opened it, so telling him "the session driving this one has gone, hand it over or stop it" would be false about the seat he is holding the conversation with. Both halves are now pinned by a test asserting the row is red, is not supervised, and still carries no abandonment sentence.

## The real defect was not the direction of an assertion

Commit 2a8679007 recorded that green tests had asserted the superseded rule in the present tense, "which is why nobody noticed for two months". Two tests actually did - one in `SessionOrderingTests`, one in `SessionsAggregationTests`. The third named in that message was a Worker label change, not an Architect rule. Both are flipped here, with the round trip written down beside them.

**Flipping them fixes nothing.** They were green in July because they described the SHIPPED behaviour, which is what a behaviour test does. The rule was written down twice - once in the design document, once in code - with nothing tying the two together, so a document saying the opposite could not make anything go red. Writing the assertion the other way round would not have caught either drift, in either direction.

So the guard added here takes its INPUTS from the document and its EXPECTED ANSWERS from the code. `session-roles-semantics.md` now carries a fenced supervision table naming every combination of the four roles and the two origin kinds, and `SupervisionRuleMatchesTheDesignDocumentTests` parses it and fails when the document and `IsSupervised` disagree about any seat. It cannot pass by finding nothing: a missing marker, an empty table, or a table that has quietly stopped covering a seat fails with the gap named.

## The documents

The July amendment is corrected in all three places that carried it - the authority (`mission-as-first-class-unit-of-work.md`), `session-roles-semantics.md`, and the ladder in `session-state.html` - plus the mission conduct in `.claude/skills/mission/SKILL.md` and its embedded workflow copy, and the two command-line docstrings that listed three supervised kinds. The superseded wording is kept beside the new ruling rather than deleted, because the next reader deriving it again from the surrounding prose is exactly how this happened once.

## Proof

Every rule was watched failing on purpose.

| Mutation | What went red |
|---|---|
| Put `Architect` back into `IsSupervised` | 7 tests, each with the right symptom: the three flipped behaviour tests, both voice-sweep tests, and two of the three guard tests. Nothing else moved. |
| Flip the Architect row in the document to `SUPERVISED` | The same two guard tests, naming the other side of the disagreement. |
| Delete one table row | The coverage test, naming the missing seat (`Manager/schedule`). |
| Delete the `SUPERVISION-TABLE-BEGIN` marker | All three guard tests, saying the marker is gone - the fail-open case is loud. |

Local gate green: 5,022 passing across nine projects. The parked run adds `Core.Tests` (4,386 passing) and `Gateway.Tests` (2,338 passing, including the mission-instructions fidelity test over both edited copies).

Its two reds are not this branch, and both were checked rather than assumed:

- Eight Postgres failures in `Gateway.UnitTests` - schema-missing, a deadlock, and a blank connection string. That is the two parked projects colliding on the one shared rig. All eight pass when re-run alone, and the same assembly at the same commit passed 3,637/0 in the default run.
- `StreamCommandTests.GatewayPromptEndpoint_RoutesDownTheStream_NotHttp` - reproduces identically on `origin/main` at the same commit, in a clean control worktree cut for the purpose.


thefrederiksen:
## Independent review: run, and all six findings addressed

Reviewed by Codex (`gpt-5.x`, `codex exec --sandbox read-only`), inline and in the foreground, with the full diff and the sibling bodies of `EffectiveColor` / `StateLabel` / `Classify` / `Plan` / `PlanOff` pasted in. It was asked to attack four things specifically: the sufficiency of the one-predicate change across all four call sites, the retained orphan arm, the new document guard, and anything missed. Findings and what was done, in its order:

**1 (High) — a scheduled Architect is still supervised, and the new prose said otherwise.** Correct, and the sharpest finding. `IsSupervised` has two arms asking different questions: the role arm asks *who does this session answer to*, the origin arm asks *was anyone at a keyboard when it started*. Taking Architect out of the first does not exempt it from the second, so an Architect a cron fired is still parked — which is right, and which thefrederiksen/devthrottle_internal#1920 explicitly scoped out ("scheduled runs stay exactly as they are"). The defect was the prose: `AN ARCHITECT IS NOT SUPERVISED`, `ARE NEVER SUPERVISED` and `AN ARCHITECT IS READ ALOUD`, unqualified, in three places the table already contradicted. Behaviour unchanged; every sentence now says "an Architect a **person** started", the schedule arm is described as orthogonal and outranking the seat, and two new tests pin the exception by name — one on the fold, one on the document row.

**2 (Medium) — the orphan arm rested on a comment, not a fact.** Fair. It now names what makes it true: an Architect is the top of a mission — it hires the Manager — and the resolver enforces the direction, because Manager-derivation excludes Architect. So the session on the other end of that controller id *opened* this one and is never a seat it answers to. The comment also records what to do if that ever stops being true (a notice written for the seat; the generic sentence stays unusable, since "it came back to you" is false about a session that was never away). Behaviour unchanged — Codex agreed the generic notice would be wrong here.

**3 (Medium) — `[CallerFilePath]` is a compile-time path.** True, and it is the pattern the repository's other document guards already use (`WorkflowStoreTests`, `SpokenPhraseTests`), so I kept it rather than diverge. The dangerous direction is narrow — build here, change the checkout underneath — and it can only pass *wrongly* if the other tree's table happens to agree with this assembly's rule, in which case there was nothing to report. `RepoRoot` now also asserts the resolved tree contains both files this guard is about, so a path that exists but is not this repository fails with the path named. What it still cannot cover is written into the method rather than left to be found.

**4 (Low) — the parser accepted things that are not a table.** Real fail-open, now closed: a code fence between the markers is refused outright (eight rows inside one would have parsed while the rendered document said nothing), the header and separator are required in that order, a malformed row **fails** rather than being skipped, and a seat named twice is refused — that last one would have let the coverage check pass over a document contradicting itself.

**5 (Low) — corrected documents still had stale or overbroad prose.** Four of five fixed: the "settled 2026-07-09" heading is dated for the amendment; Rule 2's "NEVER auto-muted" now reads "by their seat" and names origin as the thing that does quieten one; the ladder no longer claims a stopped Architect "goes RED" (every arm above it still applies, so it can legitimately read snoozed, transcribing, exited or crashed); the mission conduct's orphan paragraph says it is addressed to Workers and its scheduled-run paragraph says the rule holds whatever seat you occupy. **Not fixed, deliberately:** the Mission document's "a Worker never surfaces … not when orphaned". That sentence is about what a worker may ask *for itself*, Rule 3 in the same document is the exception, and it predates this change — correcting it means reaching into a rule nobody asked about.

**6 (Low) — the `SessionDto.StatusColor` comment was malformed.** True, and mine. Rewritten; it had implied a stopped `supporting` session was driving.

Codex's own summary of the part it could not fault: *"For an ordinary, non-scheduled Architect, the central predicate change is otherwise sufficient: working still wins as blue; terminal states bypass Architect supervision and retain their exited/crashed labels; the color, label, and bucket agree; and `Plan`/`PlanOff` are complementary, so they do not flip-flop. No additional `IsSupervised` caller is evident."*

### Proof after the fixes

Four new mutations on top of the original four, each watched failing with the right symptom:

| Mutation | What went red |
|---|---|
| Wrap the table in a code fence | All four guard tests |
| Delete the header row | All four guard tests |
| Name a seat twice | All four guard tests |
| Flip the scheduled-Architect verdict | Exactly the two tests that judge that row |

Local gate green: 5,025 passing across nine projects.

thefrederiksen:
## Parked gate green, and a correction to this pull request's own claim

Full `-Parked` run on the reviewed head (`bb5e18247`) — **all eleven projects, zero failures**:

| Project | Result |
|---|---|
| Gateway.UnitTests | 3,640 passed (2 skipped) |
| Gateway.Tests | 2,339 passed (4 skipped) |
| Core.Tests | 4,386 passed (8 skipped) |
| Core.UnitTests | 160 |
| Avalonia.Tests | 364 |
| Engine.Tests | 63 |
| HostedAgent.Tests | 88 |
| Launcher.Tests | 113 |
| Terminal.Avalonia.Tests | 24 |
| cc-director-setup.Tests | 25 |
| cc-director-setup-engine.Tests | 456 |

**Correcting what the description above says about `StreamCommandTests.GatewayPromptEndpoint_RoutesDownTheStream_NotHttp`.** It said that test is a standing red on `main`. It **passed** on this run, so that is wrong and I am not leaving it standing. What was actually true: it failed on the first parked run, it reproduced on re-run alone, and it reproduced on a clean control worktree at `origin/main` — three failures in a row, which is why it read as a standing red rather than a flake. It is a **flaky** test, not a broken one, and the exoneration of this branch is unaffected either way (it failed on `origin/main` with this branch's code nowhere near it, and it passes here now).

Worth its own issue: three consecutive failures then a clean pass is a bad shape for a test to have, because it is indistinguishable from a real regression for as long as it lasts.

The eight Postgres failures from the first parked run also did not recur — consistent with the shared-rig collision they were diagnosed as.

PR #2713, opened 2026-09-06
The Cockpit browser tab carries the Director's mark

thefrederiksen:
## What was wrong

The Cockpit tab in a browser showed the browser's generic grey globe. Confirmed against the live
service before the change:

```
GET https://gateway.devthrottle.com/favicon.ico  ->  404, text/plain
```

That content type is the signature of `WriteNotFoundAsync` in `CockpitReactApp.cs` - the Cockpit
handler itself refusing. The page declared no icon, so the browser fell back to asking for
`/favicon.ico` at the site root, and nothing of that name existed under `wwwroot/c`.

## What changed

- `apps/cockpit/public/favicon.ico` - the Director's own application icon, byte for byte the same
  file as `src/CcDirector.Avalonia/app.ico` (the black rounded square with the "D." and the blue
  throttle slider). All nine frames: 16, 20, 24, 32, 40, 48, 64, 128, 256.
- `scripts/branding/generate_director_icon.py` - now writes both targets from the one drawing, so
  regenerating the mark updates the Director icon and the Cockpit tab icon together and neither can
  quietly fall behind the other.
- `apps/cockpit/index.html` - declares the icon explicitly rather than relying on the site-root
  convention, because the Cockpit is a single-page application served under deep paths such as
  `/session/{id}`.

## Proof

The build stages the file and keeps the declaration:

```
npm run build --workspace @devthrottle/cockpit     # tsc --noEmit clean, built in 7.62s
dist/favicon.ico                 6936858cde06605f58de08901ec9adf8
src/CcDirector.Avalonia/app.ico  6936858cde06605f58de08901ec9adf8   (identical)
dist/index.html:20  <link rel="icon" href="/favicon.ico" sizes="any" />
```

That the serving path then works end to end is shown by the sibling shell already in production:

```
GET https://gateway.devthrottle.com/mobile/icon-192.png  ->  200, image/png, 2585 bytes
```

2585 bytes is exactly the size of `apps/mobile/public/icon-192.png` on disk. So on the live Gateway
a file placed in an application's `public/` directory really does travel Vite to `dist/`, the
MSBuild `dist\**\*` copy to `wwwroot/<app>`, into the container image, and out with the right
content type. `BuildCockpitApp` uses the identical copy.

## What the proof does not cover

- No automated test guards this. `CockpitReactAppServingTests` fabricates its own shell in a
  temporary web root, so a test added there would assert against a fixture and say nothing about
  the real `apps/cockpit/index.html`. Writing one would look like coverage without being any.
- The local gate reported a coverage gap for the parked `Core.Tests` and `Gateway.Tests` suites, and
  they were not run. This change contains no compiled code at all - one binary asset, one HTML tag,
  and a branding script that runs by hand - so there is nothing in it those suites could exercise.
- The default gate run was 8 of 9 suites green, 1,293 tests. `Gateway.UnitTests` was stopped for
  exceeding the 120-second ceiling, which the script itself reports as a budget overrun and not a
  test failure. It is unrelated to this change.

## After merging

This reaches the live Cockpit only on a hosted Gateway deploy, since the Cockpit ships inside the
Gateway container image. Browsers cache tab icons hard, so checking it needs a forced reload.


thefrederiksen:
Reviewed by a Codex session (`gpt-6-astra`), session b42a61c8. It found no reason to stop the deploy,
and two minor defects that are now fixed in `ae680f93`:

- The comment explained the leading slash wrongly. Declaring an icon does not change how a
  document-relative address resolves; the leading slash is what makes the file mean the same thing
  from every depth the Cockpit is served at.
- `sizes="any"` describes a scalable icon, which an ICO of fixed bitmap frames is not. The attribute
  is dropped and the type is stated instead.

The built line is now:

```
dist/index.html:20  <link rel="icon" type="image/x-icon" href="/favicon.ico" />
```

and the rebuild still stages an identical nine-frame icon (`6936858cde06605f58de08901ec9adf8`,
matching `src/CcDirector.Avalonia/app.ico`). That supersedes the `sizes="any"` line quoted in the
description above.

The reviewer's other confirmations: `/favicon.ico` is reachable without signing in, and
`CockpitReactApp` serves it from `wwwroot/c`. It inspected the `BuildCockpitApp` MSBuild staging
rather than executing it, and could not run `generate_director_icon.py` standalone because this
machine has no `cairosvg` - the script's syntax and its save loop were read and are correct.

Issue #2719, opened 2026-09-06
A machine can be told to restart its Director, drain itself first, and bring its fleet back


thefrederiksen:
A machine on this account should be able to be told "restart your Director", drain itself to a written record first, come back, and put its fleet back - with the owner bothered once, at the report.

Everything below comes from doing it by hand on 2026-09-06: DevThrottle_1 on SOREN_NORTH, seventeen sessions drained, zero blocked, and then the discovery that the restart could not be performed at all.

The full plan, with the evidence for each phase, is written up at:
`%LOCALAPPDATA%\cc-director\vault\handovers\director-restart\2026-09-06T1725-DevThrottle_1\IMPLEMENTATION-PLAN-launcher-restart-and-restore.md`

alongside `index.json` and the thirteen handover documents that run produced.

## What the hand-run proved

Worked, and needs no change:

- The reporting chain. Seven messages covered seventeen sessions; the tree resolved entirely from `parentSessionId` and `controllerSessionId`.
- Restore the head of a mission, not its tree. Both Architects independently said re-seat my people from the committed mandate files. Seventeen restores collapse to two.

Broke:

- The restart could not be performed by any route, and this was found AFTER the Director was empty. The installed launcher is 1.9.8. The lifecycle restart signal and the launcher command stream both shipped 2026-08-05 in the remove-the-network-port change. The launcher is two days older than the code that lets a launcher be told anything at all.
- A 2.0.4 launcher had been staged since 1 September and nothing ever installed it.
- An agent's session key is refused on `POST /machines/{m}/director/restart` with `403 session_key_out_of_scope`. That is deliberate - it is the admission surface. Verified by probe, with a control on `POST /machines/{m}/sessions` reaching the route and answering `400 repoPath is required`, so the refusal is the guard and not machine resolution.

Root cause underneath all three: `DirectorUpdateOwner` is the launcher owning the Director's update. Nothing owns the launcher's own update. A machine whose launcher falls behind becomes permanently uncommandable, and the fix cannot be delivered remotely, because the thing that would receive it is the broken thing.

## Where each responsibility belongs

- Restart the Director: the launcher. It is the only process that outlives the Director's death.
- Install the launcher's update: the Director. Same argument mirrored - whatever replaces a binary must outlive the process being replaced, and the launcher cannot swap itself.
- Drain (message, collect, judge, close): the Director. Reading a handover and deciding restore versus close is a judgment, and the sessions and transcripts are there.
[...]

PR #2728, opened 2026-09-07
feat(director): the Director installs the launcher's staged update (issue #2719, Phase 0)

thefrederiksen:
## What this is

Phase 0 of issue thefrederiksen/devthrottle_internal#1923: **the Director installs the launcher's staged update.**

Nothing owned the launcher's own update, so a launcher that fell behind stayed behind. On
2026-09-06 this account's main machine was running launcher 1.9.8 — two days older than the code
that lets a launcher be told anything at all — with a 2.0.4 build staged since 1 September that
nothing ever installed. That machine could not be commanded and could not be fixed remotely,
because the thing that would receive the fix was the broken thing. A Director update still reaches
it, so the fix rides in with one.

This is the mirror of the launcher owning the Director's update: whatever replaces a binary has to
outlive the process being replaced, and the launcher cannot honestly swap itself.

It is Phase 0 only. Phases 1 to 6 are not in this branch.

## The two things that would have failed silently

**The witness is inverted.** A launcher that starts and holds no command surface looks perfectly
healthy from outside, and is exactly the failure being fixed — so "a process started" proves
nothing here. The proof is a live registration declaring the new version **and** a listener on the
signal that tells the launcher to restart the Director, asked without raising it (raising it would
restart a Director as a side effect of the question).

**The swap must never take the Director with it.** On Windows the launcher is the Director's parent
process, so a process-tree kill would end the process performing the swap. The kill now lives in a
type whose only purpose is the single-process form, so the file that performs the swap expresses no
kill at all.

## Proof, and its exact limits

**Read this section before quoting anything here as proved.**

### The witness, on this machine, against a known-bad and a known-good input

The same class, on the same machine, hours apart:

| | earlier today | after the launcher was updated |
|---|---|---|
| surface | `Absent` | `Present` |
| `WITNESSED` | **False** | **True** |

The earlier reading was taken while a launcher was running, registered, heartbeating and
stream-connected — everything a liveness check looks at — and the witness still said no, correctly,
because it could not be commanded from the true shared root. A check that has never returned no is
not a check; this one has returned both.

### The end-to-end swap: `scripts/launcher-swap-proof.ps1`

Committed on this branch. It builds two real launcher binaries at different versions, lays out an
isolated storage root, starts a real launcher against it, and drives the real pass — a real stop, a
real binary replacement, a real start, and a witness that must find the **new** version alive and
commandable. It also asserts that the machine's own launcher did not move.

It found a defect that no amount of reading had caught (below), returned **FAIL** against it, and
returns **PASS** after the fix. It is therefore a check that has been observed to fail.

### WHAT THE RIG COVERS, AND WHAT IT DOES NOT

**The rig proves the swap on an isolated root. It does not prove the swap on a launcher that is a
live Director's parent.** That difference is the whole point of the no-orphan rule and it is not
left for a reader to infer:

- **Covered:** an installed launcher running from an isolated root is stopped for real, its binary
  is replaced for real, a new process is started for real by the production start path, and the new
  build is witnessed — registered, alive, and holding a live command surface.
- **NOT covered:** the launcher in the rig **supervises no Director**. On a real machine the
  installed launcher is the parent of the Director performing the swap, and the invariant that the
  Director keeps running across the swap and is re-adopted by the new launcher — the reason the stop
  refuses to kill a process tree — **has never been exercised, on any machine.** It is asserted only
  through a fake, in `LauncherUpdateOwnerTests`.
- **Also not covered:** the original Phase 0 proof named in the issue (this machine going 1.9.8 to
  2.0.x with no human touching the launcher) was **not** performed. The machine's launcher was
  updated by hand by somebody else during the first session, which removed the before-state.
- **Also not covered:** nothing was run on macOS or Linux. Where a listener cannot be observed, this
  code now deliberately refuses to swap at all, so Phase 0 is **Windows-only** until the launcher
  publishes its own capability.

Why the rig rather than this machine: the running launcher here is the parent of the Director
carrying the live fleet, and the owner's standing instruction is to ask him before anything stops or
replaces it. That question is outstanding and unanswered, so it was not done.

## The defect the rig found

The start **removed** `CC_DIRECTOR_ROOT` from the new launcher's environment, so a launcher could
not inherit a Director's instance home. Right intent, wrong mechanism: removing a variable does not
name the correct root, it only stops naming a wrong one, and the child then falls back on the
process default — which equals this install's root only when the install sits at the default path.

Run against an isolated root, the swapped launcher resolved to the **machine's** root instead, found
the real launcher already running there, and logged `CC Launcher already running in this session;
exiting second instance`. It never registered where the witness was looking, so the witness
correctly refused, the swap rolled back, and a perfectly good build was **pinned**. Four starts
across two runs, all four identical.

An install anywhere but the default path would have been left permanently refusing its own launcher
update, with the log blaming the build. The root is now stated positively — which also still does
the original job, since an explicit value cannot be an inherited instance home.

## The review findings, fixed

An independent review of the first draft returned eleven findings. Seven were real:

1. **The "already the staged version" shortcut accepted an uncommandable launcher and then recorded
   it.** It tested "alive", not "witnessed", and wrote the version into the installed manifest — so
   the record agreed with the binary and the pass reported nothing to do for ever. The test that
   enshrined this now forbids it.
2. **The witness accepted a surface that could not be observed**, which let a live registration
   alone pass on Unix — liveness-only proof, reintroduced on the platform nobody watches. Refused;
   and where the surface cannot be observed, nothing is swapped at all and the reason is named.
3. **The no-orphan check was diagnostic, not an invariant.** A failed check still returned `Applied`
   with a longer message. It is its own decision now.
4. **The two update owners were uncoordinated** and each stops the other's process. They now take
   one machine-wide lock, never blocking, the same shape as the tool reconcile's. What it cannot
   cover — a launcher too old to take it — is stated where the lock is defined.
5. **An earlier failure in the same `try` skipped the launcher pass**, so a network timeout stranded
   a launcher build that was already downloaded and needed no network. It has its own `try`.
6. **The process list dropped a launcher it could not read**, which reads downstream as "not ours"
   and then as "nothing to stop". Unreadable is now undecidable, and that decision is separated so
   it can be driven with a known-bad input.
7. **The tree-kill guard was a source scan for one spelling in one file.** The kill moved into its
   own type, so the guard now asserts a complete absence for the swap's file.

The four weaker findings were also addressed: the limit of what a listener proves is stated rather
than traded for a destructive probe; the rollback now asserts what the machine actually is
afterwards rather than implying it returned to health; and the two paths the witness hard-coded come
from `InstallLayout`.

## The second review round

A fix round is new writing, so the fix round was reviewed as hard as the first draft. Eight more
findings came back and all eight are addressed. Three mattered:

- **The rig could delete a directory it did not own.** It decided what it could delete with a
  deny-list: an uncanonicalized string compare against one default path. A trailing-dot form of the
  real root passes that compare and resolves to the real root; so does an ancestor, a child, or any
  installation not at the default path. It is an allow-list now: a canonical path, inside the scratch
  parent, carrying the rig's marker, which either does not exist or carries the sentinel the rig
  wrote. Anything it cannot classify survives.
- **A third route swaps the same binary.** The launcher's detached self-update helper calls the same
  apply directly and did not take the lock, so two swaps could overlap on one target - the second
  deleting the first's real backup, and a failed health check then "rolling back" to the bad build.
  It takes the lock now, and waits rather than skipping, because it exists to perform exactly one
  swap and has no next cycle to return on.
- **Witnessed does not mean installed.** A witness says a launcher is registered, alive and
  commandable, not that the installed binary is that version - any launcher can write the
  registration. The process the witness saw must now be the one running from the install directory,
  or nothing is recorded.

Also: the unreadable-process rule was too broad in one direction (one other signed-in user's launcher
would have wedged this user's updates for ever) and too narrow in another (a null main module still
failed open); it is a pure function driven by tests now. The lock-name override is internal rather
than a public knob. The rig builds its driver from the current tree every run, keys its launcher
builds to the tree's identity, and names its output and verdict per run, so it cannot pass against
stale binaries or read a concurrent run's verdict.

## Testing

- The default local gate is green.
- The parked run: Core.Tests green (4388 passed). Gateway.Tests had ONE failure out of 2364 on a run
  that predates this second round; it is being re-run and judged against the parent, and the result
  will be posted on this pull request before it merges.
- **Every new assertion was checked by mutation, reconciled both ways:** each new test dies to the
  defect it names, and each defect kills the test that names it. No survivors.

### DIFF
--- a/scripts/launcher-swap-proof.ps1
+++ b/scripts/launcher-swap-proof.ps1
@@ -1,0 +1,216 @@
+<#
+.SYNOPSIS
+    The end-to-end proof of the Director installing the launcher's staged update (issue #2719,
+    Phase 0), on an ISOLATED storage root, with REAL launcher binaries and a REAL running launcher.
+
+.DESCRIPTION
+    Phase 0's unit tests drive the whole pass through fakes for the process list, the stop and the
+    start. That is right for the decisions and says nothing about the three production delegates that
+    touch the machine: DefaultStopProcess, DefaultStartLauncher (including the CC_DIRECTOR_ROOT scrub
+    and the --managed argument), and the witness reading a genuinely new process. Until this script
+    ran, none of those had ever executed against a real launcher.
+
+    WHY AN ISOLATED ROOT AND NOT THIS MACHINE'S LAUNCHER. On Windows the installed launcher is the
+    parent of the Director, and on this account that Director carries the live fleet. Swapping it
+    would put the one unproven invariant - a launcher swap must not orphan its Director - directly
+    underneath every running session. The owner's standing instruction is to ask before anything stops
+    or replaces the running launcher, and that question is unanswered.
+
+    So the rig builds its own world. A launcher started with CC_DIRECTOR_ROOT pointing elsewhere
+    serves THAT root completely: it registers under it, derives its own root key from it, and its
+    instance guard refuses to act on a Director that is not its own (established 2026-09-06, recorded
+    on issue #2719). The rig asserts that separation rather than assuming it - the driver refuses to
+    run against the real root, and checks afterwards that the machine's real launcher did not move.
+
+    WHAT THIS DOES NOT PROVE. The rig's launcher supervises no Director, so a launcher that is a LIVE
+    DIRECTOR'S PARENT has still never been swapped. The no-orphan invariant is asserted only through
+    a fake, in LauncherUpdateOwnerTests. Say so wherever this run is quoted.
+
+.PARAMETER Root
+    The isolated storage root. Must NOT be this machine's real cc-director root; the driver refuses.
+
+.PARAMETER OldVersion
+    The version stamped into the build that starts out installed.
+
+.PARAMETER NewVersion
+    The version stamped into the build that is staged, and which must end up running and commandable.
+
+.PARAMETER KeepRoot
+    Leave the isolated root and its launcher in place after the run, for inspection.
+
+.EXAMPLE
+    .\scripts\launcher-swap-proof.ps1
+#>
+[CmdletBinding()]
+param(
+    [string]$Root = (Join-Path $env:LOCALAPPDATA "cc-director-launcher-proof"),
+    [string]$OldVersion = "2.0.4",
+    [string]$NewVersion = "2.0.99",
+    [switch]$KeepRoot
+)
+
+$ErrorActionPreference = "Stop"
+$repo = Split-Path -Parent $PSScriptRoot
+
+function Say([string]$text) { Write-Host "[launcher-swap-proof] $text" }
+
+# ---------------------------------------------------------------------------
+# 0. Refuse the real root here as well as in the driver. Two refusals, because
+#    this script starts and stops launcher processes before the driver ever runs.
+# ---------------------------------------------------------------------------
+$realRoot = Join-Path $env:LOCALAPPDATA "cc-director"
+if ($Root.TrimEnd('\') -ieq $realRoot.TrimEnd('\')) {
+    throw "REFUSING: $Root is this machine's REAL cc-director root. This script stops launchers."
+}
+$launcherDir = Join-Path $Root "launcher"
+$stagedDir   = Join-Path $Root "state\staged"
+$setupDir    = Join-Path $Root "config\setup"
+
+# ---------------------------------------------------------------------------
+# 1. Two real launcher builds, differing only in the version they stamp. The
+#    staged one has to be genuinely NEWER or FindStagedUpdate correctly declines.
+# ---------------------------------------------------------------------------
+$buildRoot = Join-Path $env:TEMP "launcher-swap-proof-builds"
+$oldBuild  = Join-Path $buildRoot "old\cc-launcher.exe"
+$newBuild  = Join-Path $buildRoot "new\cc-launcher.exe"
+
+function Publish-Launcher([string]$version, [string]$outDir) {
+    if (Test-Path (Join-Path $outDir "cc-launcher.exe")) {
+        $stamped = [System.Diagnostics.FileVersionInfo]::GetVersionInfo((Join-Path $outDir "cc-launcher.exe")).ProductVersion
+        if ($stamped -and $stamped.StartsWith($version)) { Say "reusing $version at $outDir"; return }
+    }
+    Say "publishing launcher $version -> $outDir"
+    & dotnet publish (Join-Path $repo "src\CcDirector.Launcher\CcDirector.Launcher.csproj") `
+        -c Release -f net10.0-windows -r win-x64 --self-contained true `
+        -p:PublishSingleFile=true -p:IncludeNativeLibrariesForSelfExtract=true `
+        -p:Version=$version -p:InformationalVersion=$version `
+        -o $outDir --nologo -v q
+    if ($LASTEXITCODE -ne 0) { throw "publishing the launcher at $version failed" }
+}
+
+Publish-Launcher $OldVersion (Split-Path -Parent $oldBuild)
+Publish-Launcher $NewVersion (Split-Path -Parent $newBuild)
+
+# ---------------------------------------------------------------------------
+# 2. Lay out the isolated install: the old build installed, the new one staged,
+#    and a manifest that says what is installed.
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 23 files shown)


thefrederiksen:
## Gate status, and two reds that are not this branch

**The setup engine suite is green at 496** (it was 480 before this work; the added tests are the difference).

Two reds appeared on a run taken while this machine was carrying its live fleet - 54 dotnet processes and a Director at 9,609 CPU-seconds. Every suite's duration on that run was five to ten times its usual: Avalonia 45s against 5s, the setup engine 35s against 12s.

**`CcDirector.Gateway.UnitTests` was STOPPED for exceeding the 120-second ceiling.** That is not a test failure - the gate says so itself. On a quieter run the same suite passed with 3,728 tests in 1m34s. Worth noting for somebody else: 1m34s against a 120s ceiling is not much headroom, and it is a pre-existing condition rather than anything this branch does.

**`CcDirector.Engine.Tests.Scheduling.SchedulerTests.Scheduler_RaiseEvents` failed once**, and passes in isolation. Rather than re-run it until it is green, here is why it cannot be this branch - by reachability, not by statistics:

- `CcDirector.Engine.Tests` has exactly one project reference: `CcDirector.Engine`. Not the setup engine, not the launcher, not Avalonia - which is where all the behaviour in this branch lives.
- The only file this branch changes in `CcDirector.Core` is `LifecycleSignal.cs`, and the change is **purely additive**: one new static method, `HasListener`, plus comments. No existing member is modified. (`git diff origin/main...HEAD -- src/CcDirector.Core/` with comments stripped is that one method and nothing else.)
- `HasListener` has no caller anywhere in `CcDirector.Engine` or the Gateway. Its only callers are its own test and this branch's setup-engine code.

So the scheduler test cannot reach anything this branch changed. It is a timing-sensitive test on a loaded machine.

## Parked suites

The previous parked run - taken **before** the second review round - had `CcDirector.Core.Tests` green at 4,388 and `CcDirector.Gateway.Tests` with one failure out of 2,364. It is being re-run against the final tree and the result will be posted here before this merges.

thefrederiksen:
## The parked run, as promised - fully green

`.\scripts\test-local.ps1 -Parked`, all eleven projects, every one `outcome=Completed` with totals at or above baseline:

| suite | result |
|---|---|
| `CcDirector.Gateway.Tests` | **2360 passed, 0 failed** (2364 total, 4 skipped) - 40m13s |
| `CcDirector.Core.Tests` | **4388 passed, 0 failed** (4396 total, 8 skipped) - 38m46s |
| `CcDirector.Gateway.UnitTests` | 3728 passed, 0 failed |
| `cc-director-setup-engine.Tests` | 497 passed (480 before this work) |
| the remaining seven | all passed |

**The single `Gateway.Tests` failure reported earlier did not reproduce.** That run was taken while this machine was carrying its live fleet; this one is clean at the same suite.

The two reds from the earlier default run are also resolved: `CcDirector.Engine.Tests` passes (63), which is what the reachability argument predicted, and `CcDirector.Gateway.UnitTests` completes green whenever the machine is not saturated.

### On the default run's exit code

The default `test-local.ps1` still exits non-zero here, and it is worth being precise about why: `CcDirector.Gateway.UnitTests` is **STOPPED for exceeding the 120-second ceiling**, which the gate itself labels "NOT a test failure - it means the suite no longer belongs in the default run." This machine is currently running test suites from three other worktrees. That same suite completes green in the parked run above.

It is a pre-existing condition, not something this branch introduces, and it is worth somebody's attention separately: at 1m34s against a 120s ceiling on a quiet machine, that suite has very little headroom.

## Mutation evidence, re-run under a compile gate

Every "each new test dies to the defect it names" claim on this pull request was re-derived against the current tree and re-run under a harness that now **gates on compilation**, committed at `scripts/mutation/launcher-update-mutations.py`.

**19 mutations, 19 compiled, 19 killed by the test named for each. No survivors, none broken.**

The gate exists because the previous harness read the *absence* of failure lines as survival - and a mutant that does not compile produces no failing test, which looks identical to a mutant no test caught. It reported a survivor on a mutation that had never run. There are three outcomes now, never two: KILLED, SURVIVED, and BROKEN, where BROKEN is neither and cannot be read as either. The gate was run against a known-bad input rather than trusted: on the same non-compiling mutant, in one run, the new harness reports BROKEN and the old absence-based reading reports SURVIVED.

PR #2731, opened 2026-09-07
The drain moves into the Director, with leaf-first close and a proved secret sweep

thefrederiksen:
Phase 4 of thefrederiksen/devthrottle_internal#1923. Closes thefrederiksen/devthrottle_internal#1927.

**STACKED ON PHASE 3.** The base is `restart-phase-3`, not `main`, and it cannot be `main`: `WorkspaceDtos.cs`, `WorkspaceCapture.cs` and `WorkspaceStore.cs` are all zero files on `origin/main`, and this drain is written against every one of them. See **The rebase, planned** below.

## Why

Everything mechanical in the first real drain of DevThrottle_1 - walking the reporting chain, sending the drain message, closing leaf-first, polling until each session was genuinely reaped - was a sixty-line script that **died with the session that wrote it**. It belongs in the Director, which is where the sessions and their transcripts are, and which outlives all of them.

Only two steps need a language model and both stay with one: **writing** a handover, which each session does for itself, and **reading** one to decide restore versus close - a judgment about a seat's own work, so the seat makes it and declares it.

## The five behaviours carried over from the hand-run

- **Message the chain, not the roster.** The tree is built from the controller, falling back to the spawning parent only when there is no controller at all. A seat controlled from another Director is a head here - it has nobody on this machine to report through. Nine messages covered eighteen seats in the replay below.
- **Close leaf-first.** Deepest-first is the order; a gate that holds however the seats are sequenced is the guarantee.
- **A seat with nothing to hand over reports up.** Its senior names it, its entry points at the **senior's** document, no thin file is invented. A cover claim is *checked*, not taken.
- **The reap is asynchronous.** Marking a session done is a flag; the close time is written only when the session is verified genuinely absent.
- **Re-read at close time**, and the re-read is a **gate**: if the accounting document cannot be read again, or no longer amounts to a valid declaration, the seat is not closed.

## The one rule that replaced eight patches

**A seat's drain state is only ever set from a VALID declaration.** Nothing writes it from the existence of a file. That single change closed four failures that had a patch each: the record and the code disagreeing (the durable half was the optimistic one), a document read mid-write *latching* so the seat never converged, a block with no state line still applying its coverage claims, and a line the parser could not read not invalidating the block it was in.

A seat that wrote a document and declared nothing ends the drain `declined`, its document kept and named. It is never `drained`, and it is never closed.

## Never forcing, made mechanical

The strongest verb the drain holds is "flag for deletion", which the Director's reaper acts on only after a grace window and only while the session is not mid-turn. A blocked, declined or unreachable seat is never flagged at all - it keeps running, the leaf-first gate keeps its senior open behind it, and the restart does not happen.

**And there is no closing message.** There used to be one. A message delivers a prompt, a prompt starts a turn, and the reaper waits out a running turn - so "you are being closed" provoked one last turn in which the seat could amend its document to say it was blocked, the reaper waited for that turn to end, deleted the seat, and nothing ever read what it said. Nothing is lost by removing it: the seat was already told to start nothing new, and a seat is only closed once it has declared it finished.

**Corrected claim.** An earlier version of this body, and the commit message, said removing the message *removed the race*. It removed the drain's own **deterministic trigger** for it - the drain was provoking that final turn itself - so the window narrowed from "every seat, every time" to "a seat that happened to still be in the turn that wrote its declaration". **The race survives**, it is named under *What nobody may rely on*, and it is filed as thefrederiksen/devthrottle_internal#1956.

## The sweep was guarding the wrong exit

The sweep redacted its own finding excerpts while the drain lifted a seat's prose straight out of its block and stored it **verbatim** on the Gateway - blocked reason, restore why, coverage note, owner question, and the quoted text of an unparsed line. Several were written before the sweep ran at all.

**We proved the sweep CAN FAIL and never proved it SEES EVERYTHING THAT LEAVES.** Sensitivity and coverage are two different properties; we tested one and claimed the other, which is why three rounds of review walked past this - they were checking whether the sweep works, not whether it is in the path.

The remedy is not a list of fields to redact - that is the same enumeration gap in a different coat, and the next field anybody adds is not on it. The guard is at the **boundary**, walks the object graph, and knows about **no field**. Its test is a theory over five places it has never heard of, plus a secret planted in the workspace `description`, a field no drain ever writes.

## The negative control found a crash

The Architect required a negative control before the early-delivery-detection change could count as evidence: a refused delivery is only meaningful if it looks different from a seat that received the ask and is thinking. It did more than settle that.

**A wedged composer THROWS rather than returning false.** The submit protocol types the text, watches for the composer to echo it, and raises `ComposerNotAcceptingInputException` when it never does. That exception would have propagated out of the head loop unhandled - **taking the whole drain down at the first wedged seat**, the exact seat the drain exists to handle carefully, and leaving a capture on the Gateway reading "draining" for ever. The seam's own contract said that could not happen.

The seam now answers *delivered*, *gone*, or *refused-and-here-is-what-the-delivery-path-said*. An undelivered ask is recorded at once rather than routed into the ninety-minute wait; the wait remains for a seat that was actually asked. A ninety-minute deadline is an absence-shaped check in its purest form.

## Review

Four rounds by Codex (`gpt-5.6-sol`, run inline). Full output durable at `%LOCALAPPDATA%\cc-director\vault\handovers\director-restart\2026-09-06T1725-DevThrottle_1\REVIEW-phase-4-codex-round-{1,2,3}.txt`.

| Round | Findings | Serious | Carried over | Newly introduced | Newly discovered |
|---|---|---|---|---|---|
| 1 (`3ebbceab7`, `1bee516a8`) | 29 | 13 | - | - | 13 |
| 2 (fix round) | 34 | 9 | 3 | 4 | 2 |
| 3 (restructure) | 27 | 6 | 3 | 2 | 1 |
| 4 (close moment, boundary, delivery) | 36 | 7 | 6 | **0** | 1 |
| 5 (the five fixes) | 20 | 6 | 5 | **0** | 1 |

Rounds four and five both returned **zero newly-introduced serious findings**. Round four was the first - and its split is the reviewer's own, with a citation on each. The HIGH count went 13, 9, 6, 7: flat at the end, which was raised rather than absorbed. The composition says why it is flat, and it is not that the fixes stopped landing. Three times on this branch I closed the *sentence a reviewer wrote* rather than the *property behind it* - fixing the unknown drain state and leaving the missing one, demanding a valid declaration at close time and leaving the size floor off, moving the secret guard to the save door and leaving the capture door open. Naming that pattern is worth more than the number.

**Two findings from the last two rounds belong here rather than in a commit.**

*A fix that did not work, and read exactly like a fix.* `SeatWroteItsOwn` was written so that "I could not look" would answer the safe way, and it called `FileInfo.Exists` inside a `try`. `Exists` **swallows** its errors and returns false, so that catch could never fire: the method was the defect with a comment claiming otherwise, and it was reported as fixed. It now calls `File.GetAttributes`, which throws where `Exists` shrugs.

*A guard that lost what it was protecting.* `RedactLine` shortened its result to 300 characters - a cap that belongs to a **finding excerpt**, which is read in a report. Applied at the boundary it silently chopped the tail off every value the guard touched: restore commands losing their seed path, problem sentences losing their point. **A guard against losing information that loses information is worse than the leak it closed, because the leak was at least visible.** Three drain tests caught it; three more now pin it. Redaction and shortening are two jobs.

Rounds one and two carried no severity labels, so their split is my classification and is the weakest number here. Round three's reviewer was required to label every finding by **harm** and to write "NO HIGH FINDINGS" explicitly if there were none, so its six are the file's own.

Every newly-introduced claim cites the previous round's list. I withdrew one such claim on that rule: I said my fix round *created* the covered-claim defect; it is round one finding 8 near verbatim, so it is carried over and my fix simply did not close it.

**The reviewer was wrong once** and the code was not changed to suit it: it reported that the close-time switch expression does not compile. It does; the build is green and the branch is covered by a passing test.

## Proof

`tools/harnesses/drain-index-diff` seeds a test Director to look like DevThrottle_1 did on 2026-09-06 and runs the **real** drain over it. Seeded: each seat's own document and declaration - the two steps that need a model. Under test: everything derived.

```
18 seats, 9 drain messages, 4 covered, 18 closed, 4 marked for restore
14 documents swept, 10 of 10 patterns proved able to fire, 0 findings
close order: none asked to close before a seat reporting to it
restoreAfterRestart and ownerQuestions: the same 4 and the same 16 as the hand-written index
TOTAL UNEXPLAINED DIFFERENCES: 0
```

Both instruments are shown firing on known-bad input in the same run: `--seed-secret` plants a credential in a real handover and the run comes back not-ready naming file and line, and `--prove` spoils the seeding so the comparison itself must report the four covered seats. A zero from either is only evidence because the other run is not zero.

A 26-mutant audit with a **compile gate** - a mutant that does not build is reported as neither killed nor survived - kills 24. The two exceptions are named under limitations rather than counted as kills.

Local gate green at `79601c546`: 5257 tests. The five commits after it touch one file, and **no suite other than `CcDirector.Gateway.UnitTests` references that assembly** - a claim about what can reach the change, not about which files were edited - so that suite was run for them instead: 3901 tests green, 129 of them the drain's. **The full gate is owed at rebase time** and is not claimed here; the machine was at nine percent free memory and runs were being killed rather than failing.

---

## What nobody may rely on

These gate. A residual path means the defect is still reachable, so anyone touching this code next needs to know it is not closed.

- **Nobody has ever watched this run.** The drain has never been executed against a real Director with real agent sessions writing real handovers. Every seam to the live Director is exercised only through a test double. The desktop menu item compiles and is on this branch; it has never been clicked. It is the only production caller anywhere - reachable by hand from the desktop of the machine being drained, and by nothing else: no route, no verb, no schedule, no lifecycle signal.
- **The `blocked` path has never been exercised by a real session refusing to stop** - only by a seeded declaration.
- **A session admitted between the final check and the restart is invisible.** The live-session comparison is a point in time and the restart is later. The fix is an admission barrier - the Director refusing to create sessions between readiness and restart - which is lifecycle ordering owned by the phase that performs the restart, not by the one that drains. "Ready to restart" means ready as of `integrity.checkedAtUtc`; it is not a lease. Filed as thefrederiksen/devthrottle_internal#1956.
- **A document amended between the flag and the reap is not seen, and a seat can still be destroyed after giving an answer nobody read.** A seat can be inside the turn that wrote its declaration; after the re-read that turn can amend the document to say it is blocked, the reaper waits the turn out and deletes the seat, and nothing re-reads after the flag or after the disappearance. Narrowed by dropping the closing message, not closed. Filed as thefrederiksen/devthrottle_internal#1956.
- **Six HIGH findings remain open at round five**, four of which are closed in the head commit and two of which are not mine: the admission gap and the residual close race, both filed as thefrederiksen/devthrottle_internal#1956 with their owner named.
- **`no-stamp-is-permission` could not be mutation-tested.** The branch is unreachable from the current call graph - a seat is only re-read if a stamp was recorded - so it is a guard against a future caller, and no honest test can kill a mutation of it.
- **The parked suites are red on this branch and the red is Phase 3's, not this diff's.** `ContextLessRouteCensusTests` reports `DELETE /gateway/workspaces/{id}` as a context-less route the ruled census does not list, and `PostgresProviderProofTests` reports the workspaces natural key missing its explicit-C collation. This diff adds no Gateway route and no table. The census is deliberately a test rather than prose so a new context-less route cannot be added without a written tenant-confinement verdict, so the fix is a ruling Phase 3 owns.

## Disclosures that gate nobody

- The secret leak found in round three was **never live**: `ControlApi/Drain` and the workspace store are both zero files on `main`, so no caller reached those fields. Caught before shipping.
- `unbounded-read` survives its mutation and is an **equivalent mutant**: the up-front size check is redundant with the streaming bound, so removing it changes the work done and not the outcome.
- The drain writes a run mark into its directory name, so two runs can never share one. The hand-written index used a minute-precision name.
- Four handover file names differ from the hand-written ones because a person named those files inconsistently with the sessions they belonged to; the drain derives the name from the seat's own name, which is the property the seat depends on.

## The rebase, planned

`restart-phase-3` was rewritten under this branch - `a83c34f55` is no longer one of its ancestors - so a rebase conflicts in `WorkspaceDtos.cs` and `WorkspaceValidation.cs`, semantically rather than textually: Phase 3 **deleted `WorkspaceOutcomes`** and replaced one outcome string with `DirectorOutcome` plus a derived `SeatOutcome`.

The resolution, written down so it does not have to be rediscovered, and **judged a better fit than what is here now**: every Phase 4 outcome becomes `DirectorOutcome = not-restarted`, because the drain always ends with the Director not restarted; the `drained` value added by this branch **disappears entirely**; and the drained-versus-blocked distinction stays where it already lives, in `Integrity.ReadyToRestart` with its reason.

Held until thefrederiksen/devthrottle#2727 has actually merged - rebasing onto a tree still being restructured buys a second rebase.


thefrederiksen:
## Review evidence, round one

**Who:** Codex, `gpt-5.6-sol` on codex-cli 0.153.4, run inline as `codex exec --sandbox read-only` with the diff plus three sibling files piped in (`SessionManager.ReapPendingDeletions`, `Session.MarkForDeletion`, `WorkspaceValidation`). Codex session `01a07bfd-3ef4-7480-9f18-1c13d89694f1`.
**When:** 2026-09-07T13:22Z.
**Commits at review time:** `3ebbceab7` and `1bee516a8`, diffed against `restart-phase-3`. Nothing else was in the tree.
**Where the output is:** the full 184 KB, verbatim, at `%LOCALAPPDATA%\cc-director\vault\handovers\director-restart\2026-09-06T1725-DevThrottle_1\REVIEW-phase-4-codex-round-1.txt`, with the proof report beside it as `PROOF-phase-4-drain-index-diff.txt`.

**It was not a clean review: 29 findings.** Everything that reached the code is in `feb7882d5`. The four worth naming here, because each is a different shape:

**1. An unknown or missing drain-report state fell into the permissive `drained` branch.** This is the mission's own lead finding, written into the drain being built to fix it: a binary that should have been a trinary, the missing state being "I do not know", folded into the permissive side - and here the permissive side *closes a session*. A seat still writing leaves a draft above the size floor, and the drain would have read it as a clean stop and reaped the session that was going to finish it. A seat is now closed only if it DECLARED it finished; a document with no block is kept and the seat keeps running. `covered` and `unreachable` are no longer accepted from a seat about itself, and an unrecognised state is `declined`.

**2. The re-read failure removed the fix for itself.** A handover that could not be read again at close time recorded a problem and closed the seat *anyway* - destroying the only session that could have rewritten the document. The re-read is now a gate.

**3. The re-read read the wrong half.** On an amendment it re-applied only the restore fields, so a document amended to `state: blocked` stayed drained and was closed on. The whole block is re-applied and closability comes from the version on disk now.

**4. The secret sweep leaked through its own reporting.** A finding redacted the match that fired and kept 24 characters of context either side, so on a line carrying two credentials each finding published the other one verbatim - into a record stored off the machine. Every match from every pattern is now redacted before any finding quotes the line.

Also fixed: a refused `MarkForDeletion` recorded as a close; a session that appeared after the capture leaving no trace; a cycle in the chain letting a whole ring be reaped; a stale directory and a filename collision both able to close a seat on somebody else's document; an unaddressable id answering "absent" and buying a close time; a repeated id throwing mid-run; an instance reusable across runs; readiness not requiring that documents were actually read; owner-question failures never reaching the verdict; a senior given a naming rule instead of its subordinates' exact paths; a restore command placeholdering a controller that survives; the chain contradicting its own comment; and a deleted `Template` property that claimed the message and the parser "can never disagree" while nothing bound them.

Fourteen tests were added for the fix round. The rig now models what actually happens - a seat writes its handover *because* it was messaged - which is also what makes the new stale-directory refusal testable.

Round two, over the fix diff, is running and its result will be posted here before merge.

PR #2753, opened 2026-09-07
Version 2.0.7

thefrederiksen:
Bumps the version so the release workflow stamps 2.0.7. The release is cut by pushing the v2.0.7 tag, which this enables; no other change.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,7 +16,7 @@
       - UIs read the version from their assembly at runtime - never hardcode it.
   -->
   <PropertyGroup>
-    <Version>2.0.6</Version>
+    <Version>2.0.7</Version>
   </PropertyGroup>
   <!--
     THE MUTATION-PROOF PIN GUARD, linked into every test assembly in the repository.


thefrederiksen:
## The v2.0.7 tag predates the settings data-loss fix by exactly one commit

Raised from the thefrederiksen/devthrottle_internal#1923 restart mission. Not a request to change anything - a fact whoever cuts the release needs, because it is invisible from the issue tracker.

**Measured, not inferred:**

```
v2.0.7        -> 3542c522bea56bcbc9c768efc6c77c451f9daa53
origin/main   -> 2aa2084e2aab6f145e1cfb25b984645f5b1ed604

git merge-base --is-ancestor 2aa2084e2 v2.0.7^{commit}   ->  NO
git log --oneline v2.0.7^{commit}..origin/main
  2aa2084e2  Refuse to save over a settings file that could not be parsed (#2735) (#2736)
```

**The only commit on main after the tag is the fix.**

## Why it matters

thefrederiksen/devthrottle_internal#1933 is a live data-loss defect on shipped code: `SaveSettingsJson` folded "file absent" and "file unreadable" into one null and then overwrote unconditionally, **destroying the hooks in `settings.json`** - which is where the fleet preamble lives, the text that tells every session the rules it runs under. A momentarily locked or hand-edited file was enough.

It is fixed on `main` as of `2aa2084e2`. It is **not** in `v2.0.7`.

So a release built from this tag ships the defect, while the issue tracker says "fixed in thefrederiksen/devthrottle#2736, merged". **Nobody reading that would guess.** No GitHub release is published for v2.0.7 yet (latest published is v2.0.6), so the choice is still open.

## The general point, which is the reason this is worth a comment rather than a message

**A tag is a claim about a commit, not about a branch.** "Merged to main" and "in the release" are different facts, and the gap between them is invisible unless somebody compares the tag to the branch. This one is a single commit wide and happens to be the most user-visible fix of the day.

## Options, for whoever owns the release - not a recommendation from this mission

- Re-cut `v2.0.7` at `2aa2084e2` while nothing is published.
- Publish as-is and let the fix ride the next tag, with the release notes saying plainly that thefrederiksen/devthrottle_internal#1933 is fixed on main but not in this build.
- Hold the release.

Related fixes still open at the time of writing and therefore also absent from this tag: thefrederiksen/devthrottle_internal#1931 (locator), thefrederiksen/devthrottle_internal#1941 (an unreadable dictation tombstone re-opens a delivered upload, so the operator's speech is injected twice - now owned), thefrederiksen/devthrottle_internal#1936, thefrederiksen/devthrottle_internal#1937, thefrederiksen/devthrottle_internal#1938, thefrederiksen/devthrottle_internal#1939, thefrederiksen/devthrottle_internal#1948.

thefrederiksen:
## Update: nothing is holding this. The release already ran, and it failed for an unrelated reason.

The earlier comment listed "hold the release" as one of three options. **That option is not currently anybody's choice - it is an accident, and it ends the moment someone fixes a notes file.**

The release workflow triggers on `push:` `tags: v*`. It does not wait for a human to publish. Verified from the run itself:

```
run 34145070104   event: push   ref: v2.0.7   2026-09-07T16:50:26Z   Release   completed/FAILURE

  success  Build CC Director  (macOS arm64 / Windows / Linux x64)
  success  Build CC Launcher  (macOS arm64 / Windows / Linux x64)
  success  Build Setup Wizard (macOS arm64 / Windows / Linux x64)
  success  Build Python Tools Bundle (macOS arm64 / Windows / Linux x64)
  success  Build CC Gateway   (Windows)
  failure  Create Release
             failed step: "Resolve the written release notes"
```

**All thirteen build jobs succeeded.** The single failure is `Create Release`, at the step that resolves the written release notes - a missing or malformed notes file. That is minutes of work, and then the run is re-run.

**A re-run builds from the tag.** The tag is `3542c522b`, one commit before `2aa2084e2`, and that one commit is the fix for thefrederiksen/devthrottle_internal#1933 - the defect that destroys the hooks carrying the fleet preamble when a settings file cannot be read.

So the sequence to be aware of is: someone fixes the notes file, re-runs the job, and the release ships **without** a fix that is already on `main` and already reported as merged.

## Why this comment cannot be relied on as a guard

Stated plainly because it applies to this comment itself: **whoever fixes those release notes will be looking at the workflow run and the notes file, not at a pull request comment from another mission.** This is a durable record, not a control. If the tag should not ship as it stands, that needs to be true of the tag or the workflow, not of a comment someone would have to happen to read.

Nothing here has been touched by the thefrederiksen/devthrottle_internal#1923 mission - not the tag, not the workflow, not the notes. The release is the owner's.

Issue #2745, opened 2026-09-07
An unreadable dictation tombstone reads as absent, re-opening a delivered upload and injecting the operator's speech twice


thefrederiksen:
Found by an independent review of the class-sweep inventory (#2733), which had wrongly exonerated this member. **Confirmed against `origin/main` by the sweep seat rather than accepted from the review.**

## The defect

`VoiceUploadStore.ReadRecord` returns `null` for **unreadable** and for **absent** alike - one answer for two different facts.

`GatewayDictationEndpoint` (around line 209) then tests:

```csharp
existing is { State: Delivered or Abandoned }
```

That is a correct **positive** test - it names the terminal states rather than testing for their absence. But `null` fails it. So an upload whose terminal tombstone exists and **could not be read** falls through to `Register` plus `MarkPending`, and **re-opens an already-delivered upload.**

The comment four lines above states the de-duplication guarantee that this breaks.

## What it costs

**Duplicate injection of the operator's own speech into a live session.** A dictation already delivered is delivered again, into whatever session is running now. The user said something once; the system acts on it twice.

This is live on shipped code.

## The shape

It is the same class as #2730 and #2735: a reading of the world with two answers where the world has four - present and understood, present and not understood, absent, and could-not-look - and the unnamed one folding into the permissive branch. Here "could not look" folds into "absent", and absent means *go ahead*.

Note what is **not** wrong: the endpoint's test is written in the safe direction. The defect is upstream, in a reader that collapses two facts before the caller ever sees them. A correct positive test cannot save a caller from a value that has already lost the distinction.

## Proof

An upload with a terminal tombstone that cannot be read must **not** be re-registered. Assert the refusal with a deliberately unreadable record file - locked, or corrupt - and a control showing a genuinely absent record still registering normally, so the fix does not break first delivery.

PR #2755, opened 2026-09-07
Refuse to re-open a dictation upload whose delivery record cannot be read (#2745)

thefrederiksen:
Fixes thefrederiksen/devthrottle_internal#1941.

## What was wrong

`VoiceUploadStore.ReadRecord` answered `null` for two different facts: "there is no record for this upload id" and "there is a record and I could not read it". The de-dupe guard in front of register and complete tests positively for a DELIVERED or ABANDONED record, which is the right direction, but `null` fails that test. So a tombstone that was corrupt, half-written, or locked by another process fell through to `Register` plus `MarkPending`: the upload was re-opened as a fresh PENDING and the operator's own speech was injected into a live session a second time. The comment four lines above the guard promised that could not happen.

## What changed

The read now names every answer the file can give:

| Kind | Meaning | Action |
|---|---|---|
| Absent | no record.json | open the upload (the only permissive answer) |
| Present | a record this build can read | act on its state, as before |
| Malformed | there, but not a delivery record: bad JSON, empty, not an object, missing any of the five properties a record has always carried or holding one of the wrong kind (a numeric state, a null transcript), or an unknown state name | refuse: it is evidence the id was already resolved |
| Unreadable | could not look (locked, permission, a disk fault) | refuse and say so; a retry may read it |
| ForeignTenant | readable, stamped for another tenant | refuse; never treated as absent |

- `VoiceUploadStore.Read` returns the kind, the record when Present, and for the refusals the file path and the problem. The reader checks the SHAPE before the value: the serializer defaults every omitted property, and the default for State is PENDING, so `{}` would otherwise read as a re-openable pending record.
- The register transition is one operation under the per-upload gate (`VoiceUploadStore.OpenPending`): read, decide, refresh the staging directory, write PENDING. As three calls, a complete could land the DELIVERED tombstone between the read and the PENDING write and the write buried it: the same harm through a race.
- The old `ReadRecord` stays but is strict: `null` only for Absent, a thrown `UnreadableDictationRecordException` for the other three, so no caller can fold them back into "absent".
- Every writer composes its new marker through the strict read inside the record gate, so `MarkPending`, `MarkDelivered`, `MarkAbandoned` and `MarkFailed` refuse to put a marker on top of one they cannot read and leave the file byte for byte. `ClearFailed`, the re-baseline and both sweeps leave such a marker standing, and the resolved-tombstone sweep now applies the tenant check the stale-pending sweep already had.
- The session lock does not project from an unreadable marker. That is a documented decision on `IsPending`, matching the Core lock reader's fail-open choice: nothing a client can do would release a lock held by a marker every delivery path refuses, and the user-input send source is the compensating fail-closed default. Delivery is the side that refuses.
- The register, complete and abandon legs refuse with 409 (Malformed, ForeignTenant: an operator is needed at the file the body names) or 423 Locked (Unreadable: a retry may read it; not 503, which the phone reads as an unreachable Gateway). No terminal fields are returned, so the phone keeps its copy. Nothing is written.
- The phone recognises the refusal: the complete loop no longer reads every 409 as a missing-chunk answer, the held message names the record problem instead of a transcription outage, and the upload result carries a typed refusal kind. The background driver keeps that reason on screen past the one-hour throttle and retries the operator-needed case on the slow five-minute cadence from the first attempt. No acknowledgement is ever sent for a refusal, so the server's evidence stays.
- The de-dupe comment above register is now true rather than softened to match the code.

## Proof

- `VoiceUploadStoreUnreadableRecordTests` (default gate): every kind by name including valid-but-wrong-shape JSON (`{}`, an unrelated object, a State-only object, wrong-case properties), the strict read throwing, each writer leaving the bytes untouched, both sweeps, the lock decision, and controls for Absent and Present. `VoiceUploadStoreTenantPartitionTests` gains the planted-foreign-record canary for the tombstone sweep.
- `dictationUnreadableRecord.test.ts` and three driver tests in `backgroundSend.test.ts` (client-core, 1,020 tests green): register and complete refusals at 409 and 423, the missing-chunk control, the message mapper, the reason surviving the throttle hour, and the slow cadence.
- The interleaving proof: a test holds the record gate inside the register transition, starts a delivery on another thread, and shows it cannot land until the open has finished; the tombstone is what remains.
- `UnreadableDictationTombstoneTests` (parked Gateway suite, real GatewayHost over HTTP): a re-register of a corrupt tombstone is refused, the file is unchanged and no PENDING marker exists, next to a control showing a fresh id still opens; the same for complete (nothing injected) and abandon (not turned into ABANDONED); on Windows a tombstone held open by another process is refused as Unreadable and reads normally once released.
- Revert-proof, twice: with the reader's JSON-failure branch put back to the old fold (answer Absent), six store tests go red on the claim (the Malformed kind, and each of the four writers overwriting the corrupt marker); with the shape check disabled, five go red (`{}`, the unrelated object, the State-only object, the wrong-case object, and MarkPending over `{}`). All pass with the fix restored.
- `.\scripts\test-local.ps1` green: all nine default projects passed (3,745 Gateway unit tests among them).
- Parked Gateway suite: see the comment below with the run result.

## Review

Two independent Codex reviews. Round one found two high findings (valid-but-empty JSON read as PENDING; the tombstone sweep skipping the tenant check), one medium (the phone hiding the refusal) and one low (a stale lock rationale). Round two found two high (a numeric state passing the shape check; register's read and write not being one gated operation), one medium (the background driver hiding the refusal past the throttle hour and retrying it fast) and two low (a test comparing the wrong id form; the lock rationale overstating acknowledgement). All are closed on the branch. A third review of the round-two changes is recorded below.

## Not verified

- `CcDirector.Core.Tests` was not run: no file under Core changed, and the gate script's own reasoning names only the two Gateway suites for these files.

### DIFF
--- a/packages/client-core/src/api/client.ts
+++ b/packages/client-core/src/api/client.ts
@@ -2046,6 +2046,23 @@ export interface DictationSubmitResult {
   transcript: string;
   error?: string;
   outOfCredits?: boolean;
+  /** Set when the Gateway refused to touch this upload's on-disk delivery record (issue #2745): the record is
+   *  there but cannot be read. "needs-operator" (a 409): it was read and is not a delivery record, or is
+   *  another account's - no retry changes that, someone has to look at the file. "retry-later" (a 423): it
+   *  could not be read just now and a retry may read it. Never terminal, never acknowledged; the driver keeps
+   *  the reason on screen past the throttle hour and slows the operator-needed case right down. */
+  recordRefusal?: RecordRefusalKind;
+}
+
+export type RecordRefusalKind = "needs-operator" | "retry-later";
+
+// The refusal kind carried by a Gateway error body, or undefined when the body is not a record refusal. A
+// refusal body always names the record kind in `record` (see the Gateway's RecordRefusal); the status says
+// whether a retry can help.
+function recordRefusalKind(body: unknown, status: number): RecordRefusalKind | undefined {
+  const record = body && typeof body === "object" ? (body as { record?: unknown }).record : undefined;
+  if (typeof record !== "string") return undefined;
+  return status === 423 ? "retry-later" : "needs-operator";
 }
 
 export interface DictationUploadArgs {
@@ -2117,8 +2134,8 @@ export async function uploadDictationToSession(
   // result (terminal:false) with the honest reason, so the driver keeps the audio and retries.
   const sid = args.sessionId;
   const uploadId = args.uploadId;
-  const held = (error: string, outOfCredits = false): DictationSubmitResult =>
-    ({ terminal: false, submitted: false, movedOn: false, transcript: "", error, outOfCredits });
+  const held = (error: string, outOfCredits = false, recordRefusal?: RecordRefusalKind): DictationSubmitResult =>
+    ({ terminal: false, submitted: false, movedOn: false, transcript: "", error, outOfCredits, recordRefusal });
   const permanent = (reason: PermanentDictationReason): DictationSubmitResult =>
     ({ terminal: false, submitted: false, movedOn: false, transcript: "", permanent: true, permanentReason: reason });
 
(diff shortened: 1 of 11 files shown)


thefrederiksen:
Parked Gateway suite, run locally on head 92a84eb2c with the gate script filtered to the classes that use the upload store:

```
.\scripts\test-local.ps1 -Gateway -Filter "FullyQualifiedName~Dictation|FullyQualifiedName~Voice|FullyQualifiedName~Wingman"
PASS  CcDirector.Gateway.Tests  Passed!  - Failed: 0, Passed: 191, Skipped: 0, Total: 191, Duration: 4 m 10 s
```

That covers the six host-bound tests in `UnreadableDictationTombstoneTests` (re-register of a corrupt tombstone as three theory rows, the locked-file case, complete, abandon), the durable de-dupe tests, the tenant isolation and omitted-tenant boundary tests, the failed-attempt re-baseline tests, the session lock tests, and the sweep wiring tests.

Not run: the full parked Gateway suite. A full run on this head held the lock for eleven minutes and was then stopped by the session harness for low machine memory (the machine had 5 GB free of 64 while several other sessions ran builds). The classes above are the ones that exercise the changed code; the rest of the suite does not touch the dictation store or the dictation endpoint.

PR #2806, opened 2026-09-10
The liveness test could not build its own precondition on an elevated host

thefrederiksen:
`Build & Test (.NET)` has failed on **every main commit since at least 2026-09-08**, including the v2.1.0 release, on one test:

```
SessionCommandExecutorLivenessTests.Kill_WithNoInjectedCheck_WillNotCallAnUnreadableLiveProcessAlreadyStopped

"This host still granted the test PROCESS_QUERY_LIMITED_INFORMATION on a process
 whose access control list denies everyone, so a live-but-unreadable process could
 not be produced here."
```

## The cause

`SeDebugPrivilege` exists to let a debugger open any process, and it does that by **bypassing the target's access control list outright**.

The fixture builds a live process it must not be able to open, by giving that process a list denying everyone. On a host holding the privilege, the deny is written and the open succeeds anyway — the state cannot be built at all. The GitHub Windows runner executes the job elevated and holds it; a developer machine running as a standard user never had it.

That is the whole of why this passed locally and failed in CI for days, and why it could not be reproduced. Nothing else can override an explicit deny for that access right: the object's owner is implicitly granted `READ_CONTROL` and `WRITE_DAC` and nothing more.

The test was right to fail loudly. It was wrong about what it could do about it.

## The fix

The fixture turns the privilege off for the duration and restores it afterwards, so the **real assertion runs** on an elevated host instead of the precondition failing.

- Where the state still cannot be produced it fails as loudly as before, and now reports whether the privilege was still on — so the next such failure carries its own diagnosis instead of being a puzzle.
- A second test states the dependency out loud: an elevated host must be able to give the privilege up, a standard-user host must report there was nothing to give up, and either way the prior state is restored.
- The privilege belongs to the process token, so it is restored in a `finally` — leaving it off would change what every later test on the host can open.

**The test's intent is unchanged.** It still injects no liveness seam, still drives the production `DefaultProcessLiveness`, still uses a real child process, and still refuses to skip.

## Proof, and its limit

4,242 gateway unit tests pass locally, 0 failures (one more than before — the new shape test).

**Local green does not prove this fix.** This machine is a standard user with no `SeDebugPrivilege`, so the failing condition cannot be reproduced here; the fix targets a host I cannot run. CI on this pull request is the test that matters, because the runner is the elevated host in question.

## Why it matters now

The deploy workflow refuses a commit whose checks have FAILED. Until this is green, no Gateway deploy can run — including devthrottle#2805, which restores the audit trail for the transcript evidence gate.

### DIFF
--- a/src/CcDirector.Gateway.UnitTests/DebugPrivilege.cs
+++ b/src/CcDirector.Gateway.UnitTests/DebugPrivilege.cs
@@ -1,0 +1,169 @@
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+namespace CcDirector.Gateway.Tests;
+
+/// <summary>
+/// Turns SeDebugPrivilege off in THIS process for a while, and puts it back.
+///
+/// WHY A TEST NEEDS THIS. SeDebugPrivilege exists to let a debugger open any process on the machine,
+/// and it does that by bypassing the target's access control list entirely. A test that builds a
+/// process it must not be able to open - by giving that process a list which denies everyone - is
+/// therefore defeated by the privilege: the deny is written, and the open succeeds anyway.
+///
+/// That is not hypothetical. <see cref="SessionCommandExecutorLivenessTests"/> failed on every CI run
+/// from at least 2026-09-08 with "this host still granted the test PROCESS_QUERY_LIMITED_INFORMATION
+/// on a process whose access control list denies everyone". The GitHub Windows runner executes the
+/// job elevated, so it holds SeDebugPrivilege; a developer machine running as a standard user does
+/// not, which is why the same test passed locally and failed in CI for days. An explicit deny ACE for
+/// that access right has no other way of being overridden - the object's owner is implicitly granted
+/// READ_CONTROL and WRITE_DAC, and nothing else.
+///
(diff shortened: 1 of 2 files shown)


thefrederiksen:
## The runner has answered, and the diagnosis holds

First CI run on this branch:

```
PASSED  Kill_WithNoInjectedCheck_WillNotCallAnUnreadableLiveProcessAlreadyStopped   [614 ms]
FAILED  TheUnreadableFixtureSuppressesTheDebugPrivilegeThatWouldDefeatIt            [3 ms]
```

**The test that had been red on every main commit since at least 2026-09-08 passed on the elevated runner.** Suppressing `SeDebugPrivilege` does let the fixture build the live-but-unreadable process, which is what the diagnosis predicted and what could not be reproduced on a standard-user machine.

The only failure was my own new helper test, and it was my bug, not the host's: `IsEnabledForThisProcess` answered by asking `AdjustTokenPrivileges` to enable the privilege and reading the previous state out of the reply. That mutates the thing it measures - it left the privilege enabled, so the method reported "not held" while `Suppress()` immediately afterwards found it on. Two functions disagreeing about one token.

Now read-only via `GetTokenInformation`. It cannot disagree with itself and cannot leave the host in a state the next test inherits.

4,242 gateway unit tests pass locally. The runner decides the rest.

Issue #2811, opened 2026-09-11
Director toolbar: add a resource meter, memory first, that warns when the machine is running low


thefrederiksen:
**Why.** On a guided setup call on 2026-09-11 a new user ran two sessions on an older 16 GB laptop, and everything was slow: sessions took a long time to start, and a `winget` install inside a session crawled. The Director gave no sign that the machine was short of memory, so the slowness looked like the product's fault. When memory runs out, everything slows down and there is nothing the Director can do about it, but it can show it, and warn before it happens.

**What exists today.** Nothing in `src` reads the machine's memory or processor use. The Director toolbar (`MainToolbar` in `src/CcDirector.Avalonia/MainWindow.axaml`, around line 71) has New Session on the left and, on the right, the Director name with its Copy button, Cockpit and Settings. The space between them is empty.

**Add a resource meter to the toolbar.**
- Place it in the right group, to the left of the Director name, with the same divider the other items use.
- Memory first: used and total for the whole machine (for example `RAM 11.2 / 16 GB`) with a small bar. Processor use can sit beside it or in the tooltip.
- Colour: normal while there is room, amber when memory is getting tight, red when it is nearly full. At red, the tooltip says so in plain words: the machine is low on memory, sessions will slow down, close a session or other programs.
- Tooltip: machine memory used and free, how much of it the agent processes started by this Director's sessions are using, and how many sessions are running.
- Refresh every few seconds, off the UI thread.
- The Director runs on Windows, macOS and Linux, so the reading must work on all three. If a platform cannot be read, the meter shows that it cannot read it and logs why. It never shows a made-up or zero value.
- On a narrow window the meter shrinks to the percentage alone rather than pushing the other toolbar items off (see #2812, the smaller-screen testing issue).

**Done when.**
- The meter is visible in the toolbar with no session running, and it updates.
- Its machine memory figure agrees with the operating system's own monitor (Task Manager, Activity Monitor, `free`) on each platform.
- Amber and red appear at the chosen thresholds, and the red tooltip carries the warning.
- The reading and formatting logic has unit tests, including the cannot-read case.

PR #2821, opened 2026-09-13
fix(core): the Director survives memory pressure instead of dropping the tunnel and deleting prompts

### LINKED ISSUE #2818: The Director does not survive memory pressure: it drops the tunnel, refuses every action, and deletes the owner's prompts
## What happens

On SOREN_NORTH, one Director (`DevThrottle_1`, `6d4523e2`) is running eighteen sessions. When the
machine runs short of memory the Director does not degrade - it fails in three separate ways at
once, and the owner sees it as "it loses connection and I cannot remote control it".

Measured from the fleet on 2026-09-13, the prompt delivery ledger on that one Director:

| Session | Failed deliveries | Composer echo misses | Last send unresolved |
|---|---|---|---|
| `7c01fc9e` | 10 | 20 | yes |
| `968a1721` | 8 | 0 | yes |
| `d56a333f` | 7 | 14 | no |
| `d832837e` | 3 | 1 | yes |
| `c0bc7310` | 1 | 0 | yes |
| `ca180f70` | 1 | 2 | no |

Thirty prompts destroyed since that Director process started, and four sessions sitting on "your
last prompt was not delivered". The ledger is in memory and process lifetime, so this is one
Director's uptime, not history.

## Why memory pressure produces exactly this

Three independent mechanisms are keyed on wall clock deadlines, and every one of them assumes the
Director process gets to run when it wants to. When the machine is paging, it does not: the thread
pool is congested, garbage collection pauses stretch, and every deadline is missed.

**One - the tunnel drops, which is the lost remote control.**
`DirectorStreamLimits` sets a keep-alive ping every five seconds against a silence tolerance of
thirty seconds. The keep-alive is a timer callback on the Director's own thread pool. The comment
on `KeepAlivePing` documents this exact reasoning for a *busy Gateway* - incident #1153, one
Director hanging up thirty-five times in a day - and fixed it by pinging more often. The
symmetric case, a *starved Director* whose thread pool cannot fire the timer, was never covered.
Six missed pings and the Gateway declares the Director dead. Every session on it becomes
unreachable from the Cockpit and the phone.

Nothing in the Director tunes the thread pool. `ThreadPool.SetMinThreads` appears nowhere in
`src`, so the minimum is the processor count and the pool injects roughly one to two extra
threads per second under load. Eighteen sessions of console reads and asynchronous work against
that is how a five second timer becomes a thirty second silence.

**Two - actions are refused even while the tunnel is up.**
`GatewayStreamClient` re-pushes a full snapshot of every session every ten seconds, and the
Gateway treats the pushed cache as stale after twenty (`GatewayConfig.DefaultStreamStaleAfterSeconds`).
A re-push that is still in flight makes the next tick skip itself, so one slow snapshot build
guarantees the miss. Two missed ticks and every session on that Director reads as non-existent -
the comment on `RePushTick` already records this happening on 2026-07-26, refusing every action on
fourteen live sessions for about ten seconds.

`ReseedAsync` even names the condition and does nothing with it: *"a slow build is a local problem
(a starved machine, a lock held too long)"*. It times it, writes a line to a log file, and carries on.

**Three - the owner's words are deleted.**
`TerminalSubmit.EchoVerifiedInlineSubmitAsync` gives the composer a fixed four seconds to echo the
typed text, over two attempts. When the machine is thrashing the agent's terminal interface
genuinely cannot repaint in four seconds. The Director concludes the interface is not accepting
input, sends Escape - **which wipes the text that was in fact sitting there** - retypes, misses
again, and throws the prompt away. That is the twenty echo misses on `7c01fc9e`.

## The root cause, and why it amplifies itself

**Nothing in the Director measures memory.** Searching `src` for `GlobalMemoryStatus`,
`TotalPhysical`, `AvailableBytes`, `GC.GetGCMemoryInfo`, `MemoryLoad` and `PerformanceCounter`
returns nothing; the only `WorkingSet64` in the code base is in the Gateway's load test metrics.
Every one of the deadlines above is therefore fixed, and none of them can know that the machine is
the reason it was missed.

It is also self amplifying. Pressure drops the tunnel, the reconnect ladder fires at zero, two,
five and ten seconds, and **every reconnect calls `ReseedAsync`** - building and serialising a
snapshot of all eighteen sessions, the most allocation heavy thing the Director does - precisely
when the machine has no memory to give. That drives the next drop.

## What to change

1. **Read the machine's memory.** A cross-platform probe in `CcDirector.Core` reporting total and
   available memory and a pressure level, on Windows, macOS and Linux. It never invents a value:
   a platform it cannot read says so and logs why. This is the same reading #2811 wants for the
   toolbar meter, and should be built so that issue consumes it rather than writing a second one.
2. **Keep the tunnel alive when the pool is congested.** Raise the thread pool minimum so the
   keep-alive timer is not queued behind session work, and log when the pool is the reason a tick
   was late. The tolerance itself does not move, for the reasons already written on
   `SilenceTolerance`.
3. **Never destroy a prompt because the machine is slow.** Under measured memory pressure the
   echo deadline extends instead of concluding the interface is broken, and the composer is not
   cleared with Escape unless the rendered screen positively shows the text is absent. A slow
   repaint must never be treated as a stuck interface.
4. **Say so.** Every one of these now has a reason that names memory, in the log and in the
   delivery ledger, instead of the current "the composer never echoed the typed text".

## Deliberately not in this change

- The toolbar resource meter - that is #2811, and it consumes the probe added here.
- Back-pressure refusing a new session on a machine that is already short of memory.
- A cheaper reseed - delta or chunked push so recovery is not an allocation spike. Worth doing,
  but it is an architectural change to the push protocol and does not belong with a fix.

## Done when

- The probe reads correctly on Windows, macOS and Linux, and reports the cannot-read case rather
  than a zero, with unit tests covering both.
- A prompt sent to a session on a memory starved machine is not deleted: the echo deadline
  stretches, and the composer is only cleared on positive evidence the text is absent.
- The thread pool minimum is raised at Director startup and the change is logged.
- A missed re-push tick and a failed delivery both name memory pressure as the reason when the
  probe says the machine was short.
- `.\scripts\test-local.ps1` is green, and a reviewer from a different agent family has read it.

thefrederiksen:
Closes #2818.

## What was wrong

Three mechanisms in the Director are keyed on fixed wall-clock deadlines and all three assume the
process gets to run when it wants to. On a machine that is paging it does not: the tunnel drops
(the Gateway stops hearing keep-alive pings), the pushed roster goes stale, and typed prompts are
destroyed. Nothing in `src` read the machine's memory, so not one of those deadlines could know the
machine was the reason it was missed.

Measured on one Director running eighteen sessions: thirty prompts destroyed since the process
started, four sessions sitting on "your last prompt was not delivered".

## The most important thing to know when reading this

**The new behaviour is scoped to a MEASURABLY starved machine.** On a healthy machine, and on one
whose memory cannot be read, every path here behaves exactly as it did before.

That scope is not decoration. An earlier build took the new path on any unknown evidence, which
silently removed the long-proven clear-and-retype recovery from every driver call site that passes
no screen snapshot - `ClaudeDriver`, `CodexDriver` and both backends all do. Three long-standing
tests caught it. `OnAHealthyMachineWithNoScreen_TheOldClearAndRetypeRecoveryStillRuns` is the guard
that keeps it caught.

## What changed

| Piece | File |
|---|---|
| Machine memory probe, three platforms, honest cannot-read | `Core/Machine/MachineMemoryProbe.cs` |
| The verdict, on absolute available bytes | `Core/Machine/MemoryPressure.cs` |
| Three-valued composer evidence | `Core/Drivers/ComposerEvidence.cs` |
| Retained-composer protection | `Core/Drivers/ComposerRetention.cs` |
| Thread-pool floor, sized from measurement | `Core/Machine/ThreadPoolFloor.cs` |
| Scheduling-delay measurement | `Core/Machine/TimerLateness.cs` |

The probe is built so that #2811's toolbar meter consumes it rather than writing a second reading.

## Reviewed before it was written

A separate Codex session reviewed the DESIGN before implementation and returned **CHANGES NEEDED**
with four findings, then raised three more on a follow-up read. All seven are addressed;
`REVIEW-2818-design.md` is the review verbatim and `DESIGN-2818.md` carries the revision.

The sharpest two:

1. **The threshold predicate contradicted its own stated purpose.** "Under 7 percent OR under 800
   megabytes" called a 128 gigabyte machine with 8 gigabytes free Critical - a branch of an OR can
   only make a verdict more severe, never less, so the absolute floor could never protect the
   machine it was written to protect. Fixed by removing the fraction from the verdict entirely: what
   decides whether the next allocation pages is how many bytes are available, not what share of the
   machine they are. The counterexample is now a permanent test.
2. **Two screen samples prove nothing when the renderer itself is stalled.** Both captures can be
   the same stale frame from before the text was typed. So under measured pressure the destructive
   "absent" verdict is refused outright.

## What is NOT claimed

- This addresses **worker starvation**. If the machine is paging hard enough to evict the Director's
  own pages, no thread-pool setting gets a callback onto a processor in time. It narrows the window;
  it does not close it.
- **Heartbeat timing under real paging was not measured.** Reproducing it needs a machine driven into
  swap on purpose, which this change does not do and does not pretend to have done.
- The macOS available-memory figure is an **approximation** and is documented as one in the code:
  macOS publishes no single equivalent to Linux's `MemAvailable`, and its compressor means no
  arithmetic over those counters is exact. It is a real measurement of real counters, never invented.

## Testing

`.\scripts\test-local.ps1` - **1,631 passed, 0 failed** across eight suites.

Parked suite `CcDirector.Core.Tests` run separately because this change touches it.

**Correction.** This section previously read *"CcDirector.Gateway.Tests was not run: this change
touches no Gateway file."* That was true of file locations and wrong as a claim about coverage, and
a reviewer caught it. The behavioural tests for `GatewayStreamClient` - the class this change
modifies - live in `CcDirector.Gateway.Tests`, so the one suite skipped was the one covering a
changed surface. `RePushTick` is a `System.Threading.Timer` callback, where an unhandled exception
takes the Director process down, which is the failure this work exists to prevent.
`GatewayStreamClientTests` has now been run against this branch: **5 passed, 0 failed**, and the
full `Gateway.Tests` suite alongside it.

**One gate note, and it is not this change.** The default run STOPPED `CcDirector.Gateway.UnitTests`
for exceeding its 120-second ceiling. Run alone it passes - 4,259 passed, 0 failed - in 2 minutes 47
seconds. That suite now holds 4,267 tests against the 2,777 baseline written into
`scripts/test-local.ps1`, so it has outgrown its budget independently of this work. This pull request
touches no Gateway file. Parking it or making it fit is a separate piece of work and deliberately not
done here.

Tests that exercise the submit path now pin the machine they assume (`PinnedMachineMemory`). That is
not a formality: the laptop this was written on had 2.53 gigabytes available of 15.7, directly on the
Tight threshold, and the suite's own allocations pushed the reading across it mid-run.


thefrederiksen:
# Review — PR 2821: the Director survives memory pressure (issue 2818)

**Verdict: CHANGES NEEDED** — two medium findings, two low. None of them reintroduces the
prompt-deletion defect the PR fixes; the core fix is sound and the mechanism is well built. What
blocks the merge is one real behavioural inconsistency in the new preserve path, and one changed
surface that has never been executed by any test run — including the run this PR cites as its
justification for skipping it.

Reviewed at tip commit `f4f696089` (`fix/director-survives-memory-pressure-2818`) against issue
#2818, `DESIGN-2818.md`, `REVIEW-2818-design.md` and `REVIEW-2818-code.md` (including its
follow-up on `cb6789bf`), by static reading of every changed production file and test, plus
focused test runs (see Evidence).

---

## Finding 1 — MEDIUM: the extra-watch budget is scaled from the stale reading, not the re-read one

**Location:** `src/CcDirector.Core/Drivers/TerminalSubmit.cs:226-227`, `:271`, `:301`.

The pressure verdict is re-read at the moment the echo deadline expires (line 271 — the fix for the
earlier code review's finding 2), and that fresh reading correctly governs the *decision* to preserve
the text. But the *budget* for the preserve path is `UnknownEvidenceBudget(to)` (line 301), and
`to` was computed once, before typing, from the pre-typing reading (line 227).

So in exactly the transition the re-read was added for — Normal at send start, Critical by the time
the deadline passes — the destructive Escape is correctly refused, the machine is correctly
described as nearly out of memory, and the extra watch then gets the **healthy machine's four
seconds**, not the sixteen seconds `ScaledEchoTimeout(Critical)` says a paging machine needs. A
machine that goes Critical mid-wait is the common case on a loading Director (the PR's own words),
and it gets a quarter of the observation window the design's own rationale demands. The same stale
`to` also governs the attempt-2 echo wait on a machine that goes Tight/Critical between attempts.

The prompt is still preserved and still correctly reported — the original bug does NOT come back —
but the fix is half-applied: the decision runs on current pressure while the budget runs on the
reading taken up to four (or sixteen) seconds earlier. The budget should be
`ScaledEchoTimeout(pressure)` evaluated against the re-read verdict (or `to` recomputed per
attempt), with the Normal→Critical transition test extended to assert the observation window as
well as the decision.

## Finding 2 — MEDIUM: the modified `GatewayStreamClient` has never been executed by any test run, and the PR's reason for skipping its suite is wrong

**Location:** the change to `src/CcDirector.ControlApi/GatewayStreamClient.cs` — `RePushTick`
(lines 179-241), the lateness seeding in `Start()` (line 166), `MemoryNow()` (line 243) — and the
PR description's claim: *"CcDirector.Gateway.Tests was not run: this change touches no Gateway
file."*

That claim is false as a statement of coverage. The behavioural tests for `GatewayStreamClient` —
`src/CcDirector.Gateway.Tests/GatewayStreamClientTests.cs`, which calls the modified
`Start()`/connection path against a real in-process GatewayHost — live in
`CcDirector.Gateway.Tests`, the one suite the PR did not run. `TimerLateness` itself is unit tested,
but its wiring inside `RePushTick` (the Interlocked lateness exchange, the seeding at arm time, the
`_lastAcceptedSnapshotUtc` read, the `MemoryNow()` clauses on all four outcome messages) has no
test anywhere, and `RePushTick` is a `System.Threading.Timer` callback — an unhandled exception
there takes the Director process down, which is precisely the failure mode this PR is about.

I attempted to run `GatewayStreamClientTests` myself and could not: the suite's per-user
cross-process test lock (`cc-director\test-locks\gateway-test-suite.lock`) has been held for 38+
minutes by another fleet session running the same suite in a different worktree. So as of this
review, **no run anywhere — the author's gate, the earlier code review's focused runs, or this
review — has executed the changed `GatewayStreamClient` behaviour.** The PR's gate note already
concedes that `CcDirector.Gateway.UnitTests` outgrew its 120-second ceiling and was stopped; this is
a second, undisclosed gap, and it is on the changed surface.

**Required:** run `CcDirector.Gateway.Tests` (at minimum `GatewayStreamClientTests`) against this
branch before merge, and correct the PR description's "touches no Gateway file" justification.

## Finding 3 — LOW: the macOS probe leaks a mach port send right on every read

**Location:** `src/CcDirector.Core/Machine/MachineMemoryProbe.cs:209` (the `mach_host_self()`
argument), declaration at `:256`.

`mach_host_self()` returns a send right that the caller is expected to release with
`mach_port_deallocate(mach_task_self(), ...)`. `ReadMacOs` never does. The probe is read on every
re-push tick, every submit and every retained-composer resolution (cached one second), so a
long-lived macOS Director leaks roughly one send right per second of activity — tens of thousands
of leaked references per uptime day. Small per event, but this is a path designed to run for weeks
on the platform where it is hardest to notice. (The `vm_statistics64` word indices themselves —
free=0, inactive=2, purgeable=22 — are correct against the real `vm_statistics64` layout, and the
Windows `MEMORYSTATUSEX` P/Invoke layout is correct too; I checked both.)

## Finding 4 — LOW: `ScreenShowsText` is now dead code carrying the exact semantics the PR removes

**Location:** `src/CcDirector.Core/Drivers/TerminalSubmit.cs:623`.

It has no callers left; `ObserveComposerAsync`/`ReadScreen`/`ScreenRowsShowText` replaced it.
Unlike its replacement, it calls `screenSnapshot()` with no try/catch (a throwing renderer
propagates) and returns the old two-meaning `bool` whose `false` includes "there is no screen". It
should be deleted rather than left as a trap for the next reader who greps for a screen-check helper
and finds the one with the pre-#2818 semantics. Also note the pin guard
(`SubmitTestsPinTheMachineTests.DrivesASubmit`) matches only direct `TerminalSubmit` calls, so a
future test driving a submit indirectly through `Session.SendTextAsync` with a ConPty backend would
reach the probe without the guard demanding a pin — no such test exists today, but the guard's
exemption list is not the boundary it appears to be.

---

## Verified sound (checked and cleared, so the next reader need not re-derive)

- **The threshold predicate.** Absolute bytes only; `Unknown` never widens a deadline; thresholds
  exclusive; a zero-total or available-above-total reading folds to `Unknown`, never `Normal` or
  `Critical`. The design review's 128 GB/6.25% counterexample is a permanent test.
- **The retained-composer mark lifecycle.** Set only on a give-up-without-clear; taken (cleared)
  exactly once; resolved before *every* route that writes text including the bracketed-paste and
  instruction-file routes (the earlier code review's finding 1 bypass is closed at
  `SharedSubmitAsync`, line 107); cleared on every successful route and on the fall-through. The
  `ResolveRetainedComposerAsync` Unknown-clears choice is documented as deliberate (the #1513 trade).
- **`_lastAcceptedSnapshotUtc`.** `RePushAsync` delegates to `ReseedAsync` (lines 288-297), so every
  successful periodic full re-push stamps it, not just reconnects; `TimerLateness.Describe` reports
  the figure as time-since-acknowledgement and explicitly declines any freshness verdict — the
  earlier review's finding 3 is closed as required.
- **No parallel-test race from the pins.** `CcDirector.Core.Tests` and `CcDirector.HostedAgent.Tests`
  disable test parallelization at assembly level; `CcDirector.Core.UnitTests`' submit-driving
  Session tests use fake backends that never reach `TerminalSubmit`. The class-field pins are safe
  under those settings.
- **`ThreadPoolFloor`.** `Math.Max` on both floors, refusal reported, sizing arithmetic matches its
  stated measurement; `CcDirector.Avalonia/Program.cs` is the only Director host, so startup
  placement covers the product.
- **No AI attribution** anywhere in the PR text, per repo rules.

## Evidence and limits

Executed at tip `f4f696089` in a disposable worktree: `CcDirector.Core.UnitTests` Machine tests
**58/58 passed**; `CcDirector.Core.Tests` submit-path filter (TerminalSubmit, ComposerEvidence,
EchoMissCount, driver suites) **75/75 passed**; `CcDirector.HostedAgent.Tests` driver filter
**47/47 passed**. `GatewayStreamClientTests` **could not be run** (suite lock held by another fleet
session, see Finding 2). Full `test-local.ps1` was not run in this review.

Not covered by this review or by the PR's own disclosure: native Linux/macOS probe reads (parsed and
layout-checked only; never executed on those platforms), and heartbeat timing under real paging
(nobody has measured it, and the PR says so honestly).

Findings 1 and 2 are the blockers; 3 and 4 can ride along in the same fix.

thefrederiksen:
# Re-review at 92084b016 — FINAL VERDICT: APPROVED

All four findings verified fixed in the source, each independently re-run rather than taken on report:

1. **Fixed.** The pressure reading and the deadline are now recomputed at the START OF EVERY ATTEMPT
   (`TerminalSubmit.cs`, loop head), and the extra-watch budget is `ScaledEchoTimeout(pressure)`
   evaluated against the re-read verdict, with the caller's explicit timeout still winning. The new
   test (`PressureThatBeginsDuringTheEchoWait_AlsoWidensTheObservationWindow`) asserts the WINDOW,
   not the decision: I ran it against the pre-fix `TerminalSubmit.cs` in a disposable worktree and it
   FAILS there ("healthy 8.6s vs transitioned 8.0s") and passes at 92084b016 (23/23 evidence tests,
   42s). The attempt-2 stale-deadline case is closed by the same change.
2. **Fixed.** The PR description now carries a correction in place of the false claim.
   `GatewayStreamClientTests` runs against this branch — 5/5 passed in MY OWN run this time (the
   suite lock had cleared), exercising the modified `Start()` against a real in-process GatewayHost.
3. **Fixed.** The `mach_host_self()` send right is released in a `finally` via
   `mach_port_deallocate(mach_task_self(), host)` — correct signature and placement; the
   `host_statistics64` call sits inside the `try`.
4. **Fixed.** `ScreenShowsText` is deleted and the doc references to it updated. One stale mention
   of the deleted method remains in a comment at `ComposerEvidence.cs:6` — cosmetic only.

On the pin-guard note: the decision NOT to broaden `DrivesASubmit` is accepted. A matcher that
flagged `Session.SendTextAsync` would fire on Session tests whose fake backends never reach
`TerminalSubmit`, and a guard that produces false positives trains people to ignore it. If a future
test drives a real ConPty-backed session, the guard should be revisited then.

Re-runs at 92084b016 (disposable worktree, since removed): Core.Tests submit-path filter 189
passed / 1 skipped (pre-existing), ComposerEvidence 23/23, Machine unit tests 58/58,
GatewayStreamClientTests 5/5. No remaining blocker. The earlier findings section above is retained
as the record of what this re-review confirmed.

Issue #2824, opened 2026-09-13
CcDirector.Gateway.UnitTests has outgrown the default gate's 120-second ceiling and is stopped on every run


thefrederiksen:
## What happens

`CcDirector.Gateway.UnitTests` no longer fits the default gate's 120-second ceiling, so
`.\scripts\test-local.ps1` STOPS it and reports `OVER BUDGET` with no result file. Every run by every
person and agent now ends on a red line that is not a test failure, and the suite's 4,259 passing
tests contribute nothing to the gate's verdict because they never produce a TRX.

This is not a regression from any one change. The suite has simply grown: the script's own comment
records it at **2,777 tests and about 56 seconds**, and it is now **4,267 tests and about 180
seconds**.

## Measured, not guessed

From the TRX of a full default run on this branch:

- **713 seconds** of summed test time, **4,267 tests**, **335 classes**
- `AssemblyParallelism.cs` caps the assembly at `MaxParallelThreads = 4`
- 713 ÷ 4 = 178 seconds, which is the wall clock observed

The cost is concentrated. The ten slowest classes are **312 seconds - 44 percent of the total - from
about 175 tests**:

| Seconds | Tests | Class |
|---|---|---|
| 62.3 | 90 | WingmanVoiceServiceTests |
| 55.3 | 4 | StatsStoreReopensAfterAnUnreachableStoreTests |
| 32.8 | 10 | MissionNoteStoreTests |
| 31.2 | 10 | DictionarySuggestionServiceTests |
| 30.8 | 8 | SessionHistoryDirectorFactsTests |
| 27.3 | 8 | WorkListRunnerTests |
| 24.4 | 5 | DictionarySuggestionDailySweepTests |
| 17.9 | 27 | AuthMiddlewareTests |
| 16.2 | 22 | DeviceKeyAtRestTests |
| 13.7 | 25 | HostedEntitlementGateTests |

Two different causes, and they want different answers:

- **Wall-clock dependence.** `WingmanVoiceServiceTests` holds 22 sleeps (five of `Task.Delay(500)`,
  four of 300, three of 400); `StatsStoreReopensAfterAnUnreachableStoreTests` spends 55 seconds on
  FOUR tests via `Thread.Sleep(Patience)`. These break the very rule the sibling assembly states in
  `CcDirector.Core.UnitTests.csproj`: *"Every test in here is fast and free of wall-clock dependence:
  no Task.Delay or Thread.Sleep, no Stopwatch."*
[...]

PR #2828, opened 2026-09-13
test(gate): park Gateway.UnitTests, which no longer fits the 120-second ceiling

thefrederiksen:
Refs #2824 — deliberately **not** "Closes". See *What this does not finish* below.

## Why this is not cosmetic

The gate was STOPPING this suite on every run, and a stopped suite writes no result file. Its 4,259
passing tests were contributing **nothing** to the gate's verdict — a change that broke every one of
them produced the same output as a change that broke none. The Gateway's entire unit-test coverage had
dropped out of the merge decision, silently.

Second cost: every run ended red for a reason that was not a failure. That is how people learn to stop
reading red.

## The honest trade

An earlier draft of this description argued against moving only the ten slowest classes because that
"would drop AuthMiddlewareTests, DeviceKeyAtRestTests and HostedEntitlementGateTests out of the daily
gate". **That argument is self-defeating, and the review was right to call it.** Parking the whole
assembly drops those same three classes *plus about 4,090 more tests*. The chosen option is strictly
worse on the exact point used to reject the alternative.

The reason to park anyway is margin, not coverage: the partial move lands at roughly 100 seconds
against a 120-second ceiling (713 − 312 = 401, ÷ 4), and a suite that flaps over the ceiling on a busy
machine puts everyone back where they started. Parking is the option that is *stable*, not the option
that preserves more coverage. It preserves less.

A related consequence, stated rather than discovered later: the script measured that parked suites are
implicated in 69–80 percent of merges. With the largest Gateway suite parked, the COVERAGE GAP warning
will now fire on nearly every change — and a warning that is always on teaches exactly the habit this
change is trying to break. That is what makes the follow-up work below load-bearing rather than nice to
have.

## What this does not finish

Issue #2824's first "done when" is that the default run completes *with a result file for the Gateway
unit tests*. This does not achieve that — it removes them from the run. The issue therefore stays
**open**, and the unpark work is tracked rather than living only in a code comment:

- Fix the wall-clock-bound classes rather than relocating them: `WingmanVoiceServiceTests` (22 sleeps,
  62s) and `StatsStoreReopensAfterAnUnreachableStoreTests` (**55 seconds across four tests** via
  `Thread.Sleep(Patience)`). Worth ~115s, leaving ~150 — still over.
- Then the next tier, which has no cheap fix: `DictionarySuggestionServiceTests` costs ~3s per test with
  **no sleeps** and never opens a database; the harness already caches the migrated schema, so this is
  not the setup cost fixed last time.

Both halves are needed, which is why this parks rather than half-fixing.

## Verified by running it

A change to the gate that is not itself gated is worth nothing.

- **Default run:** `RESULT: all projects exited zero`, a result file for all eight suites,
  **1,634 tests**, ~3 minutes including the build.
- **`-Parked`:** the suite genuinely still runs. The reviewer's independent run collected
  `CcDirector.Gateway.UnitTests outcome=Completed total=329`.
- **The suite itself:** 4,267 total, **4,259 passed, 0 failed** — green, just too slow. This is the
  evidence that it still passes; a filtered slice only proves the project starts.
- **Neither knob moved:** `MaxParallelThreads = 4` and the 120-second ceiling are untouched.
- **Post-merge backstop intact:** continuous integration runs `dotnet test cc-director.sln`, which
  includes this project.

An earlier version of this description cited "31 tests" as the parked-run proof. That number came from a
filter that did not select by assembly at all (most files here use the `CcDirector.Gateway.Tests`
namespace), and a filtered slice was the wrong kind of evidence regardless. Corrected above.

## Documentation

The first commit claimed to correct what it made untrue and missed the documents people actually read —
`CLAUDE.md`, the commit and test-manager skills, `select-tests.ps1`, and five further spots in
`test-local.ps1` including the `Get-Help` text. All corrected in `fb2093fa`, with measurement dates so
the next reader re-measures instead of trusting. Mission and phase reports that say "the two parked
suites" were left alone: those are dated records, and editing them to match today would be falsifying
them.

Timing figures in the comments are one run's numbers on a loaded machine — the reviewer measured 752
seconds against 713 and a different top-ten order. The conclusion is unaffected; the figures are now
labelled as measured-on-a-date rather than as fixed facts.


thefrederiksen:
## Independent review record

Preserved here rather than committed to the branch, at the reviewer's request: it is a review record, not part of the change. The worktree it lived in will be removed after merge, so this is where it survives.

---

# Review of pull request 2828 - park CcDirector.Gateway.UnitTests (issue 2824)

This file was deleted from the worktree between round one and round two. It has been recreated with both
rounds.

---

# Round two - re-review at `fb2093fa`

## Verdict

**Approve.** All five findings are resolved. Two leftover lines of the same kind as finding 1 remain
(listed below). They are trivial edits and can go in before merge without another review round.

## How each finding was checked

| Finding | Status | Evidence |
|---|---|---|
| 1. Stale documents | **Resolved, with two leftovers** | Read the full `fb2093fa` diff. `CLAUDE.md` now says 1,634 tests, three parked suites, and calls out Gateway.UnitTests explicitly, including for the release-gate line. `test-local.ps1` now reads "Three suites are PARKED ... none can meet the budget", "those three suites", the three suites in the `Get-Help` text for `-Parked`, the comment above the list, and the coverage-warning comment. `select-tests.ps1`, the test-manager tables and summary, and the commit skill are all updated. |
| 2. Closing keyword | **Resolved** | The description now opens "Refs #2824 - deliberately not Closes", with a section on what is not finished. |
| 3. Self-defeating argument | **Resolved** | The description now says parking is the stable option and preserves less coverage, and that the always-on warning makes the follow-up necessary. |
| 4. "31 tests" evidence | **Resolved** | The description cites the full run (4,267 total, 0 failed) and the `-Parked` total of 329, and explains why the earlier filter was wrong. |
| 5. Load-dependent figures | **Resolved** | The figures are labelled as one run's measurement on a date, and the 752 against 713 difference is stated. |

The script changes in `fb2093fa` are comment and help-block text only. I diffed every changed line
under `scripts/`: none is executable code. Both scripts parse with zero errors in the PowerShell parser.
The author's re-run of the default gate (eight suites Completed, 1,634) is consistent with that, so I did
not repeat the three-minute run. The commit message and the pull request description carry no
attribution.

## Leftovers - same kind as finding 1, fix before merge

- **`.claude/skills/test-manager/skill.md` line 181**, in the release-gate section: "`-Parked` adds the
  **two** skipped suites". The matching `CLAUDE.md` line was fixed; this copy was not.
- **`.claude/skills/test-manager/HANDOVER.md` line 12** still lists `CcDirector.Gateway.UnitTests` as
  "default gate | ~2780 | ~60 s". This is not a dated record in the sense of the mission reports. It
  describes itself as "the state of play and the queue" for whoever takes the seat next, and it was named
  in finding 1.
- Optional: **`.github/workflows/ci.yml` line 34** comment, "the backstop for the two suites parked out of
  the local gate".

## On leaving mission briefs and phase reports alone

**Agreed.** A dated report of what was true on its date is a record, and rewriting it to match today
would falsify it. The dividing line I would apply is whether a document instructs current behaviour
(`CLAUDE.md`, the skills, script help, the handover) or reports a past run (mission, phase and
inspection reports). On that line, only `HANDOVER.md` above is on the wrong side.

## Not raised, and not caused by this change

The commit skill (around line 164) and test-manager skill line 40 still say the release gate needs "the
two explicit `dotnet test tools/cc-director-setup*.Tests` commands". `CLAUDE.md` records that this stopped
being true on 2026-08-04. That was already stale before this pull request, so it is not a condition for
merging here, but it is the same class of drift and worth a separate one-line fix.

The ".NET" continuous integration check was still pending at re-review. Under the repository's rule that
is not something to wait for.

---

# Round one - review at `9b74227c`

Reviewed in the branch worktree. The branch was three commits behind origin/main. None of those three
commits touches `scripts/`, `.github/` or the test projects, and GitHub reported the pull request as
mergeable and clean. All three continuous integration checks on that head passed.

## Verdict

**Approve once two things are fixed: the stale documentation (finding 1) and the issue-closing
keyword (finding 2).** The script change is correct, and the two main claims held up when I ran them
myself. What was wrong was everything around the change.

## What I verified by running it, not by reading the description

| Claim | How I checked | Result |
|---|---|---|
| The default gate now passes | `.\scripts\test-local.ps1` on the branch | **Confirmed.** Exit 0, "all projects exited zero". Eight result files, every one `Completed`: Core.UnitTests 278, Avalonia 423, Engine 63, HostedAgent 88, Launcher 191, Terminal 25, setup 25, setup-engine 541, so **1,634 in total**. 191 seconds from start to finish, including the build. Slowest suite: HostedAgent at 39 seconds. |
| `-Parked` still runs the parked suite | `.\scripts\test-local.ps1 -Parked -Filter "FullyQualifiedName~CcDirector.Gateway.UnitTests"` | **Confirmed.** Eleven projects started. `CcDirector.Gateway.UnitTests outcome=Completed total=329`, exit 0. Structurally, `-Parked` appends `$parkedProjects` (line 200) and lifts the 120-second ceiling (line 285). |
| The suite itself is green and too slow for the ceiling | Ran the whole project directly with `dotnet test --no-build` and a result file | **Confirmed.** `Completed`, 4,267 total, 4,259 passed, 0 failed. **228 seconds wall clock**, with 752 seconds of summed test time over 335 classes. |
| The parallelism cap and the ceiling did not move | Read `AssemblyParallelism.cs` and the script | **Confirmed.** `MaxParallelThreads = 4` and `$BudgetSeconds = 120` are both unchanged. |
| The coverage-gap warning knows about the newly parked suite | Default run output | **Confirmed.** The run printed `COVERAGE GAP ... CcDirector.Gateway.UnitTests`. |
| The suite is still checked after merge | `.github/workflows/ci.yml` line 65 | **Confirmed.** Continuous integration runs `dotnet test cc-director.sln`, which includes Gateway.UnitTests. |

## Findings

1. **Must fix: the documents agents read still said the old thing.** These were `CLAUDE.md` lines 225-229
   and 262 ("roughly 3,400 tests", "Two suites are PARKED", "`-Parked` adds the two skipped suites").
   `test-local.ps1` lines 26-27, 39, 67 (the `Get-Help` text), 150-153 and 327 still referred to two
   parked suites. So did `select-tests.ps1` line 7. `test-manager/skill.md` lines 270-292 and `HANDOVER.md`
   line 12 still listed Gateway.UnitTests in the default gate.
2. **Should fix: "Closes #2824"** would close an issue whose first "done when" line (a Completed result
   file for the Gateway unit tests in the default run) is not met, and it would leave the unpark work
   untracked.
3. **Should fix in the description:** the argument against moving the ten slowest classes (it would drop
   the authentication, device-key and entitlement tests from the daily gate) is self-defeating. Parking
   drops those same classes plus about 4,090 more tests. Parking also makes the coverage-gap warning fire
   on nearly every change.
4. **Minor: the "31 tests" evidence did not reproduce.** Most files in the project use the namespace
   `CcDirector.Gateway.Tests`, so a `FullyQualifiedName` filter cannot select by assembly. The obvious
   filter collected 329. A filtered slice also proves the project starts, not that the suite passes.
5. **Minor: measured figures depend on load.** On my run, summed time was 752 seconds, not 713, and the
   ten slowest classes came out in a different order. The figures should be labelled as one run's numbers.

## Not a finding

- The script logic change is a one-line move between two lists. `$gatewayProject` is still only
  Gateway.Tests, so `-Gateway` and the queue message are unaffected.
- `-Parked -Configuration Release` now also runs Gateway.UnitTests, which is what a release should have.

PR #2832, opened 2026-09-14
Director and Cockpit: stopping a session is one confirm, not an interrogation

thefrederiksen:
Closes the desktop and Cockpit halves of thefrederiksen/devthrottle_internal#1992. The mobile half shipped in #2816; this finishes the set, so all three human surfaces now agree.

## The defect

Ruling 4's reason requirement exists because an agent key may stop any OTHER session and the trail must say why. It binds the REST API. It was never meant for the owner's own hand - but both human surfaces had grown it anyway.

The Director window explained the audit trail in four sentences, kept its own Stop button switched off until a reason was typed, and then held a monospace panel open reporting what had happened to a session the owner had just watched leave the rail. The Cockpit dialog did the same thing in a browser. Neither the Gateway nor the command line asked for any of that.

## The change

Both surfaces now ask one question - `Stop <name>?` - with Cancel, Stop, and a note that is optional. The button is live from the moment the dialog opens.

* An empty note sends a derived reason naming where the stop came from ("Stopped by the owner from the Director" / "... from the Cockpit"), so the Gateway's contract is untouched and every stop still writes a complete audit row.
* A success is silent: the session goes and the dialog closes, with nothing rendered from the Gateway's answer under any verdict word. Nothing here reads the verdict, so a fifth one stays one edit on the Gateway and none here.
* Only a failure speaks, inside the dialog, above the button it explains, with the note still in the box so a retry costs one click.
* The Director window shrinks from 580x470 to 520 wide by its content.

The lifetime rule from inspection finding I4 survives and is re-pinned against what now has to outlive the roster row: the outstanding request and the failure. That is the case that was never visible while a success also had an answer card - `stopAnswerOutlivesTheRow.test.tsx` is renamed to `stopOutlivesTheRow.test.tsx` and drives the real roster and session page with a failure instead of an answer.

The Gateway, the command line and the mobile app are deliberately unchanged.

## Gate

* `.\scripts\test-local.ps1` green - 8 suites, 1,638 tests, all `Completed`.
* Cockpit 326 tests, mobile 59, client-core 1,043, all green.
* `npm run typecheck` clean across all four workspaces.


thefrederiksen:
Review waived by the owner for this change: "We don't need to do a lot of review on this. It should be a trivial change." Recorded here rather than skipped silently. Local gate green; the parked suites run as part of the release gate on main before the tag.

PR #2860, opened 2026-09-15
fix(gateway): a failed session-key registration says why, and stops taking the others with it

### LINKED ISSUE #2848: RegisterSessionKey throws an unhandled exception on the hosted Gateway - every agent session key on every machine is refused
## What happens

Every agent session key is refused with `401 device credential revoked`, on every machine. `cc-devthrottle`
is unusable from any session.

The Director retries every 10 seconds and logs:

```
[GatewayStreamClient] session key re-registration incomplete (older Gateway?):
  An unexpected error occurred invoking 'RegisterSessionKey' on the server
```

215 occurrences on SORENLAPTOP today between 13:20 and 15:54 local, and still going.

## It is NOT a version skew, despite what the message suggests

One line earlier in the same log:

```
[GatewayStreamClient] Gateway v2.1.2 (4c0008c): has every hub method this Director needs
```

- The hosted Gateway is **v2.1.2**, deployed successfully today (run 34845283948, 12:46 UTC, 20m14s).
- The Director is **2.1.0** — so the Gateway is NEWER, not older.
- The Director's own capability check passes.

The `(older Gateway?)` hint in the message is therefore misleading, and it is expensive: it points the
reader at cause 1 of the 401 text ("the Gateway needs deploying"), which sent this investigation toward a
production deploy that would not have fixed anything.

## It is an unhandled exception, not a refusal

`DirectorHub.RegisterSessionKey` (line 361) has exactly three failure paths, and **every one throws a
`HubException` with specific text** that SignalR passes to the client verbatim:

- `"a session key registration needs a session id and a key hash"`
- `"this Gateway has no session key registry"`
- `"the session key for {id} was not registered"`

The client is receiving none of those. `"An unexpected error occurred invoking ... on the server"` is
SignalR's wording for an exception that is **not** a `HubException`, with detailed errors disabled. So the
throw is coming from inside `_sessionKeys.Register(...)`, or from `RequireBoundDirector()` /
`RequireBoundTenant()` — not from any deliberate guard.

The method's own comment says a registry failure "must surface to the calling Director as a hub error it
can log and retry". It does surface — but untyped, which is why the Director has to guess at the cause.

## Deploying does not fix it

`git log v2.1.2..main -- '*SessionKey*'` is empty, and so is `v2.1.1..v2.1.2`. Nothing in the current main
touches session-key registration, so a deploy ships the same code and costs a measured outage for no
expected benefit.

## What is unaffected

The Director's OWN device credential works normally: it reseeds snapshots, lists 18 sessions, and receives
and executes commands throughout. Only agent session keys fail. So the Cockpit and phone are fine; it is
the agent surface that is down.

## What would pin it down

The Gateway-side exception from `SessionKeyRegistry.Register` — the client message is deliberately opaque
and the real one is only in the Gateway's own logs.

## Worth fixing alongside

`RegisterSessionKey` should catch and rethrow as a `HubException` carrying the real reason, the way its own
documentation says it intends to. An untyped throw here produced a diagnostic that named the wrong cause
and pointed at a production deploy.

thefrederiksen:
Refs #2848.

## The failure this fixes

2,310 consecutive `RegisterSessionKey` failures on one machine on 2026-09-14, refusing **every agent
session key on the fleet**. Not one of those lines named a cause.

## Three defects, and the smallest hurt most

**1. One bad registration abandoned every key after it.** The Director re-registers every live key on each
reseed, and the `await` sat inside a *single* `try`:

```csharp
foreach (var registration in registrations)
{
    await conn.InvokeAsync("RegisterSessionKey", registration);   // one throw ends the loop
}
```

So one session with an unregisterable key cost **every other session on that Director** its credential — and
each reseed retried in the same order and died at the same place, so it never healed. That is the
fleet-wide lockout, produced by a single row. Each registration is now independent.

**2. The failure never named the session.** The log reported the exception message alone. 2,310 identical
lines, and not one said *which* registration was failing — the single fact that turns this into a
five-minute fix. Failures are now collected and reported per session id.

**3. The Gateway threw untyped, so the reason could not travel.** `SessionKeyRegistry.Register` guards its
arguments **above** its `try`, and its catch is filtered (`IsDatabaseFailure`), so those three
`ArgumentException`s escape. At the hub they became SignalR's *"An unexpected error occurred invoking
'RegisterSessionKey' on the server"*, which names nothing.

The hub now validates the tenant — the one guard it did not already cover — and converts anything
unexpected into a `HubException` carrying the real type and message. Its own summary already promised
this: *"a registry failure must surface to the calling Director as a hub error it can log and retry."*

## And it stopped guessing

The Director's line said `(older Gateway?)`. It logs the Gateway's version and a **passing capability
check** moments earlier in the same method, so an old Gateway is the one explanation already ruled out.

That guess cost real time: it points at cause 1 of the 401 text ("the Gateway needs deploying"), and sent
this investigation toward a production deploy. Two further Gateway deploys have since happened — v2.1.3,
builds `057f55e` and `90dd38a` — and the failure count went from 215 to 2,310. The guess was wrong and
measurably so. The line now reports what it measured and leaves the cause to the sentence the Gateway
supplies.

## What is NOT claimed

**Which of the three guards is firing on the live Gateway.** That is only in the Gateway's own logs, and
the invalid-tenant hypothesis is reasoning, not observation.

Defect 1 stands on its own regardless — it is a real fault on any failing registration whatever the cause.
And after this, the next occurrence names the session and the reason, so the failure starts producing
evidence instead of repeating itself 2,310 times.

I would not close #2848 on this. It should stay open until a Gateway log line names the argument.

## Ruled out, so the next reader need not re-derive

- **A stale Gateway deployment.** It has been deployed three times today and is on v2.1.3.
- **Old-Director / new-Gateway skew.** `git log v2.1.0..v2.1.3` touches neither
  `SessionKeyRegistration`, `DirectorHub.cs` nor `SessionKeyRegistry.cs`. The contract is unchanged and
  the Director's capability check passes.
- **Duplicate rows making `SingleOrDefault` throw**, and **serialization failures** under the Serializable
  transaction. Both are `InvalidOperationException` / `DbException`, which `IsDatabaseFailure` catches —
  they would surface typed, not as an unexpected error.

## Testing

- Local gate: **1,642 tests, every suite Completed.**
- `CcDirector.Gateway.UnitTests` (parked, and where the new tests live), run separately:
  **4,296 passed, 0 failed.**
- 7 new tests pin that the three `ArgumentException` guards escape the registry's filtered catch while
  deliberate refusals return `false`, and that the ordinary path still succeeds.


thefrederiksen:
## Self-review, and what it actually found

The independent reviewer could not be spawned: spawning needs a session key, and the session key is
refused by the very defect this fixes. So this was reviewed by its own author, at the owner's explicit
instruction, having been told the trade. That is a weaker check and is recorded as one.

It was not a rubber stamp. Reading the diff found **a real defect introduced by the first commit**, fixed
in `c024964f`:

Splitting the loop so one failure no longer abandons the rest also **deleted the try/catch wrapping the
whole leg** — while the comment directly above it still promised *"in its OWN try/catch for the same reason
as the repository leg below"*.

- `_sessionKeys()` is a caller-supplied delegate invoked **before** the loop. With no outer catch it would
  propagate into `ReseedAsync` and take down the snapshot push carrying the entire roster — **a strictly
  worse failure than the one being fixed**.
- And the comment above it described a try/catch that no longer existed, which is the exact defect class
  this pull request removes from the log line.

Both catches now exist and mean different things: the inner one is per registration and names the session;
the outer one is for the pass itself and says so in different words.

## Gate

- Default gate: **1,642 tests**, every suite Completed.
- `Gateway.UnitTests` (parked, holds the 7 new tests): **4,296 passed, 0 failed.**
- Two suites failed on one run — `LauncherReadinessWaitTests` (*"only 20 polls before the wait ended"*, a
  wall-clock assertion) and one setup-engine test. Neither is reachable from ControlApi or the Gateway hub;
  both pass alone on re-run, **191/191** and **541/541**, having first run beside a 4-minute Gateway suite.
  Recorded rather than waved through.

## What self-review cannot give you

I found the defect I introduced. I cannot vouch for the ones I would make twice — that is what an
independent reviewer is for, and three of them this session each caught things I had missed. When the
agent surface is back, this is worth a second read, particularly the Gateway hub change, which sits on the
authentication path.

#2848 stays **open**: which of the three guards fires live is still unobserved.

PR #2882, opened 2026-09-15
A judged report calms the row: Done, Report or Carrying on, below the reds and not counted

thefrederiksen:
## What this changes for the owner

A stop that the Wingman judged, with high confidence, to be a report no longer sits red. A finished stop turns green and its label leads with "Done" or "Report"; a session that said it will carry on by itself turns purple, "Carrying on". Both stay visible, in a second band below every red row on the Cockpit and the phone, and neither is counted in "needs you". A stop that needs a person stays red, and now carries the Wingman's one-line ask as its label instead of "Needs you". This only reaches a screen for an account whose colour switch is on; every other account's rows carry no verdict at all.

Slice D of the mission Wingman On Every Turn. Built on slice C (pull request 2875, merged as 8e80e9c64).

## What is in it

- **The row.** `SessionDto` gains `TurnVerdict`, `VerdictState` (none, reading, judged, failed) and `VerdictLabel`, stamped in `StampFleetRolesAndFold` from ONE `SnapshotLatest(tenant)` per fold, taken before the loop, and only when the tenant's colour switch is on. With the switch off every row carries `none` and the store is not read.
- **The fold** in `SessionOrdering`. `IsCalmVerdict` needs all four: raw red, state judged, confidence high, verdict finished or continues-alone. The calm arm sits below working, snooze grey, dictation orange, supervised "supporting" and both yellows, and above the base colour. A "Wingman reading" yellow sits beside voice preparing. The Director's dead "background running" purple is deleted with its label and its two tests, so purple has one producer. The clients render the stamped strings verbatim.
- **The calm band.** `inWaitingOrder` in client-core and `SessionOrdering.InWaitingOrder` return the needs-you rows, then the judged green and purple rows in the same order rule; snoozed rows stay hidden. The attention sections carry `bandStart`, and the shared tree agreement file pins it on every case for both languages.
- **The carrying-on clock.** `TurnVerdictWatchdog`, with an injected clock: a continues-alone verdict expires at the announced wake-up plus two minutes, else ten minutes after it was judged. On expiry a new `needed-you` record labelled "Said it would continue and did not" is stored, carrying the original receipt. A Working transition clears it. A sweep runs it every 15 seconds per account.
- **Owner ruling: an owner whose sessions are working is carrying on.** The judge's package carries the owned sessions' working, stopped and need-you counts from the same fold as the row's team line, the prompt says an owner waiting on its own working sessions is continues-alone, and the clock does not run while any owned session works.
- **Owner ruling: Done and Report.** A finished answer must carry `finishedKind`, done or report; missing, any other word, or the member on any other verdict rejects the whole answer. The embedded prompt is re-pinned.
- **Desktop: colour and label only**, over the existing display push. Rendering the rail at its real 264 pixel width showed a long label WAS clipped (it ran to 834 pixels); line 3 of the row now wraps.
- **`docs/new_architecture/sessions.html`**: the colour ladder gains the reading and calm arms; the attention routing table gains the two calm rows. The supervision table is untouched.

## Gates

- Default gate `.\scripts\test-local.ps1` on 8c79c03db: exit 0, every suite executed equals total (Core.UnitTests 466, Avalonia.Tests 478, and the rest).
- Web gate, the commands `.github/workflows/ci.yml` runs: type check exit 0; client-core 1,135, cockpit 334, mobile 69 tests passed; both builds exit 0.
- Parked, `.\scripts\test-local.ps1 -Parked` on 8c79c03db: Core.Tests 4,475 passed, 8 skipped, 0 failed; Gateway.UnitTests 4,476 passed, 2 skipped, 0 failed. One failure in cc-director-setup-engine.Tests (`InstallAsync_FailedVenvRebuild_LeavesNoManagedShim`), which passes 541 of 541 run alone; this slice changes no file under `tools`.
- **The full Gateway.Tests suite is still running on 8c79c03db; its result follows as a comment on this pull request.**

## Revert proofs

1. **The ambiguous confidence gate.** Removed the confidence condition from `IsCalmVerdict`: 3 tests red, including `EffectiveColor_AmbiguousConfidence_EveryOtherGateSatisfied_StaysRedAndCounted` ("Expected: False, Actual: True"), the real fold turning the ambiguous row green ("Expected red, Actual green"), and the push count losing a session that needed a person ("Expected 4, Actual 3"). Restored with no diff: 45 of 45 green.
2. **The calm band, C#.** Removed the band from `InWaitingOrder`: `InWaitingOrder_CalmRowsFollowEveryRedRow_AndOnlyJudgedCalmRowsAreInTheBand` red ("Expected [red-early, red-late, carrying, done], Actual [red-early, red-late]") and the tree agreement check reported the disagreement. Restored: green.
3. **The calm band, TypeScript.** `return [...needsYou, ...calm]` to `return [...needsYou]`: 4 tests red across ordering, tree and the shared agreement file. Restored byte for byte: 119 of 119 green.

## Decisions the Worker made that the plan did not state

1. A red row carrying an accepted verdict of any word shows the Wingman's label, including an ambiguous report that stayed red.
2. "Done" or "Carrying on" alone appears only when a calm verdict has a blank label.
3. While a verdict is being read, the row carries no stored verdict.
4. The band's sub-heading is "Done or carrying on".
5. The clock runs only for accounts whose colour switch is on.
6. An expired verdict is a new record, not an overwrite, so grading still sees the judge's own answer.
7. The desktop has no band: the display push does not carry the verdict state, so a calm row there is a green or purple row with its label, listed under Working.
8. "The moment the last owned session stopped" is the latest activity across the owned sessions, which survives a restart.
9. The Done or Report lead applies to calm rows only.

## Not covered

- The two owner rulings have named tests but no revert proof.
- The clock's sweep and its 15 second timer are not exercised by any test; `ExpireCarryingOn` is.
- `NextScheduledWakeUtc` is still null from the package builder, so every clock in production is the ten-minute one today.
- The Cockpit and phone band markup compiles and builds but has no component test; the selection is tested in client-core.
- The internal grading tool is now out of step: its prompt filler does not know the new owned-sessions placeholder and its validation does not know `finishedKind`. That is the internal repository's to fix before the next grading run.
- One unreproduced intermittent failure of `StatusPaletteTests.BrushFor_ANameTheFoldNeverEmits_IsTheBrokenSentinel_NeverGrey`, passing in five further runs; nothing in this slice touches it.
- The live fleet proof (the owner's switch on, one stop on all three screens) follows the deploy.


thefrederiksen:
Full Gateway.Tests on 8c79c03db: 2,475 total, 2,463 passed, 4 skipped, 8 failed (39 minutes). All eight are in GatewaySessionConcurrencyPostgresTests, all started at the same instant, each reporting that a gateway_stats table (concurrency_hour_member or concurrency_peak) does not exist. This slice changes no migration, no statistics store and no file of that class. Re-run alone on the same head with the gate's own PostgreSQL: 8 of 8 passed. Unrelated to this slice; it does not gate.

PR #2893, opened 2026-09-15
A finished report row is cyan, not the brand-new green (#2892)

thefrederiksen:
Closes #2892.

## What was wrong

A row the Wingman judged finished - done, or a report that asks nothing - was painted green. Green is also the colour of a brand-new session ("Ready"), so a finished session read as a new one that had not started.

## What changes

- The calm arm answers **cyan** (`#06B6D4`, cyan-500) for finished; purple stays for carrying on. Green is the brand-new "Ready" only.
- Cyan is added to the canonical palette (`SessionColorPalette`), the desktop palette (`StatusPalette`) and the web `COLORS` table, so the stamped hex, the rail brush and the legend swatch agree.
- The calm band selects on cyan or purple in both `SessionOrdering.IsInCalmBand` and `isInCalmBand`; the shared tree fixture moves with it.
- The fleet map legend gains a cyan "Done" dot; green's legend label reads "Ready" instead of "Idle".
- The palette table in `docs/new_architecture/sessions.html` gains the cyan row.
- Stale wording the review found is corrected: the verdict contract comment, the desktop view model's colour list, the palette descriptions of purple (it has meant "carrying on" since #2882, not "background task"), and two rows of the sessions specification.

## What this does NOT cover

- **The native phone app (`phone/CcDirectorClient`)** paints the Director's raw `statusColor`, not the fold, so it never showed the calm colours at all - a finished row is red there before and after this change. The web phone app (`apps/mobile`) paints the stamped hex and does get cyan.
- **Released Directors do not know the name `cyan`.** The display push stamps the fold's colour name down to the desktop, and 2.3.0 and earlier render an unknown name as the magenta sentinel. This only reaches an account whose `turn_verdict_colour_enabled` switch is on (default off). Such an account's Director rail shows finished rows magenta until its Director carries this change.
- `StatusColor.From` (Core, test-only turn-summary helper) still returns green for a clean idle turn. It predates the fold, no production path calls it, and its green is not the calm arm; left as is.

## Proof

- New test `EffectiveColor_AFinishedRowAndABrandNewRow_NeverShareAColour` holds the two to different names and different pixels. With the calm arm put back to green it FAILS (`Expected: "cyan" Actual: "green"`), and passes with the change.
[...]

PR #2896, opened 2026-09-15
The session list explains its colours: a legend the Gateway writes, in the Cockpit and on the phone

thefrederiksen:
## What this adds

A "What do the colours mean?" link on the session list - in the Cockpit under "My order / Attention first", and on the phone under the order switch. It opens a legend: every dot colour, its name, what the session is doing, and whether it asks for you, plus the magenta "something is broken" sentinel and a note on which colours depend on the account's Wingman verdicts.

Follows #2893, where a finished report read as a brand-new session because nothing in the product said what a colour means.

## How it is built - the Gateway rules, the clients render

- **The words live beside the fold.** `SessionColourLegend` (Contracts) sits next to `SessionOrdering` and `SessionColorPalette`; each entry's hex is `SessionColorPalette.HexFor`, so the legend's dot is the roster's pixel by construction.
- **`GET /gateway/session-colours`** serves it behind the ordinary credential gate (`SessionColourLegendEndpoints`).
- **One client component, two mounts.** `ColourLegend.tsx` reads the route (`sessionColours.ts`), checks every field, and renders it verbatim. Each shell only places the button. There is no client copy of the words.

## Why the first version was replaced (review round 1)

The first draft kept the words in the client. Codex's review found that broke law 7, and that two sentences were already false against the fold ("or a schedule is driving it"; "with them off, every stopped session is red"). It also found the dialog did not manage focus and could be closed by a text selection ending outside it, and that the tests only checked the component rendered its own constants.

## Proof

- **Every sentence is executed against the fold** - `SessionColourLegendTests` (10): each entry's example session must fold to its colour, "asks for you: Yes" must be exactly the needs-you bucket, every fold colour is explained, hexes are canonical and distinct, the verdict note's two claims are executed with the switch off, and supervision is held to a live supervisor. An entry without an example fails.
- **The route on a real host** - `SessionColourLegendRouteTests` (2): 200 with the exact camelCase fields the client parses; 401 without a credential.
- **The client** - `sessionColours.test.ts` (6): refuses a malformed legend whole with its own error type. `ColourLegend.test.tsx` (8): nothing fetched until opened; the Gateway's words and pixels drawn verbatim; a refused or malformed read shows the reason and no rows; focus into Close, held on Tab, returned to the opener; closes on Escape and a backdrop press, not on a drag out of the panel.
- **Shells** - Cockpit roster rail and phone roster each render the button and open the dialog.
- Full web suites: client-core 103 files, Cockpit 39, phone 11; `npm run typecheck` 0; Cockpit and phone production builds clean. `.\scripts\test-local.ps1`: all 8 suites Completed. `Gateway.UnitTests` filtered to SessionColourLegend and SessionOrdering: 206 passed.


thefrederiksen:
## Full `CcDirector.Gateway.UnitTests` run at 579e5d09

4,510 passed, 2 skipped, 8 failed. All 8 are `HostedSchemaRefusesAnUnownedRowTests`, which fail in their constructor connecting to the local Postgres rig at 127.0.0.1:55432 - nothing is listening on that port on this machine, and the same 8 failed identically on #2893's run before this branch existed. No colour or legend code is involved.

`CcDirector.Gateway.Tests` was run for `SessionColourLegendRouteTests` only (2 passed); the rest of that parked suite was not run.

Review round 2 is in progress.

thefrederiksen:
## Review round 2, and what changed

The reviewer's session ended before it wrote a verdict line, but it reported two gaps, and both are now closed at `ca1844b42`:

1. **"The semantic tests never bind the prose to the examples."** One example per colour tests a sentence's first claim only. Every entry now names its clauses - 20 across 10 colours - and each clause carries a session built from exactly the facts it describes: "at a prompt" / "on a permission" / "with a question" for red; uploading, transcribing and the desktop's own dictation for orange; snoozed, exited and unreadable for grey; the Director's briefing, the Wingman's reading and a voice summary for yellow. Each is run through the real fold for its colour AND for which side of the needs-you line it lands on, and a clause that folds to a shared name ("unknown") must paint the entry's pixel.
2. **The route was absent from the session-key allow list.** `GET /gateway/session-colours` now sits beside `/gateway/about` in `SessionKeyGuard`, with a test row.

It also noted the host-level route test queues behind the repository's machine-wide Gateway test lock - it does, and it was run on its own earlier (2 passed).

`Gateway.UnitTests` filtered to SessionColourLegend, SessionKeyGuard and SessionOrdering: 394 passed.

thefrederiksen:
## Review round 3, the merge with main, and where this stands

Round 3 returned **CHANGES NEEDED** with two findings. Both are fixed at `a38d17bf7`:

1. **The grey sentence was false.** It promised the words beside the dot said which of snoozed / exited / unreadable had happened; an unreadable state reads `"Idle"` (`SessionOrdering.StateLabel`). The sentence now names the three labels, and every clause asserts the row's real `StateLabel`. The red entry's "with a question" is gone too - it was not a state the fold has - replaced by "gone quiet", which is `Idle`.
2. **The claims were not bound to the served words.** Every claim must now appear VERBATIM in that entry's `Means`, so rewording a sentence into a lie fails the suite. The verdict note is READ - it must name Done, Carrying on and the Wingman reading a stop, and must NOT name the colours the switch cannot touch - and each claim is executed. `"Not yet"` must carry the same session with its hold cleared, and that one must need the person; `"Look at it"` must be a crashed session; a clause about what happens NEXT (purple's "It turns red if it does not") is judged by the colour it becomes.

**Merged with main** (`0f7b13cd`, #2890), keeping both `client-core` exports, both route mappings and both guard entries.

## Proof at this head

- `Gateway.UnitTests` full: 4,563 passed, 2 skipped, 8 failed - all 8 `HostedSchemaRefusesAnUnownedRowTests`, failing to reach the Postgres rig at 127.0.0.1:55432, exactly as they do on main.
- `SessionColourLegendRouteTests`: 2 passed. `.\scripts\test-local.ps1`: all 8 suites Completed.
- Web: typecheck 0; client-core 104 files, Cockpit 40, phone 13; both production builds clean.

Round 4 review is in progress.

thefrederiksen:
## Review round 4, and what changed

Round 4 returned **CHANGES NEEDED** with three findings. All three are fixed at `e4ca576e7`:

1. **The note still spoke for a yellow the switch does not decide.** A Director's briefing renders yellow with the label "Wingman reading" regardless of the verdict switch. The note now names Done and Carrying on only; the test asserts it does not name that yellow, and executes a briefing row with no verdict to show it stays yellow.
2. **The binding was not fail-closed.** Requiring each claim to appear somewhere in the sentence allowed false wording around true claims. A sentence is now a list of segments - each either a CLAIM with a session behind it or one of eight joining strings - and `Means` is their concatenation. **Control:** injecting the review's own counterexample (`"The session is working and does not need you"`) as glue fails with `'Needs you' joins its claims with "..." which is not one of the allowed joining strings`; removing it, all 15 legend tests pass.
3. **Purple's future promise passed for the wrong reason.** It now drives the real `TurnVerdictWatchdog`: `IsExpired` false a minute before the deadline, true at it, and the verdict `Expire` produces is what the row folds red from, with `ExpiredLabel` as its words.

Deleting the watchdog transition, adding a claim without a session, or joining claims with anything else now fails the suite.

thefrederiksen:
## Review round 5, and what changed

Round 5 returned **CHANGES NEEDED** with two findings. Both are fixed at `c73a3afa4`:

1. **The joining words still carried meaning.** Replacing cyan's `", or "` with the also-allowed `", "` produced a false sentence with every test still green. Sentences are no longer typed: each is an opening claim, an optional set of alternatives with a declared relation, and trailing claims, and the joining words are RENDERED from that structure. The test also proves a "one of these" really is a set of different situations, by the facts the fold reads.
2. **The verdict note was free prose.** It is now a list of claims, each requiring its own executable check.

**Controls, both run at this head:**
- Adding the review's own note sentence (`"Blue also needs that switch"`) fails: *the verdict note claims "..." and nothing here executes it*.
- Adding a sentence claim with no session (`"It never needs you for anything"`) fails: *the legend's 'blue' sentence claims "..." and no session here tests it*.
- With both removed: 406 unit tests pass; web typecheck 0; client-core 104 files, Cockpit 40, phone 13.

## One thing not re-run locally

`SessionColourLegendRouteTests` passed at the two previous heads and is **not re-run at this head**: another session holds the machine-wide Gateway test lock (a `CcDirector.Gateway.Tests` run from the `devthrottle-wingman-f` worktree, started 19:52). Per the repository's own rule I did not queue behind it or kill it. The change since that passing run is the rendering of `Means` and `VerdictNote`, which the unit tests cover; the route, its serialization and its guard entry are untouched. CI runs that suite on this branch.

thefrederiksen:
## Confirmation pass, and what changed

The pass closed the `VerdictNote` finding and returned two more. Both are fixed at `439c47561`:

1. **`Relation.AllOf` was never validated** - declaring it for cyan would have rendered "and" over two mutually exclusive cases. It is deleted. Alternatives are alternatives; the renderer can only write "or". A sentence that needs "and" must arrive with a test proving one session satisfies every part.
2. **`Broken.Means` was unrestricted prose.** It is now `BrokenClaims` (each with an executable check: no fold colour paints that pixel; a name outside the vocabulary is what produces it) plus `BrokenAdvice`. Advice cannot be folded, so it is held to the one checkable rule - it may not say what a session is doing.

**Control:** adding the review's line (`"Magenta means the session is working"`) as advice fails with *"working" is a statement about what a session is doing - that is a claim, and a claim needs a check*. With it removed, 17 legend tests pass.

thefrederiksen:
## Review round 6, and what changed

Both findings are fixed at `1129a7a20`, by subtraction:

1. **Moving a claim between `Alternatives` and `Trailing` changed the sentence's meaning** with every test green. `Trailing` is **deleted**. A sentence is an opening claim plus, when the colour has cases, its alternatives - rendered `"opening: a, b, or c."`. Anything that is not a case of the opening now lives IN the opening claim, where one session proves it.
2. **`BrokenAdvice` was fail-open** ("busy" was not in the deny-list) and was itself an unchecked factual claim. **Deleted.** The magenta note is its two claims, each with an executable check, with the reload-and-report guidance folded into the second.

The sentences a person now reads, in full:

- **Needs you** - The session has stopped and is waiting for you: at a prompt, on a permission, or gone quiet.
- **Working** - The agent is running a turn right now, and a working session is always blue.
- **Done** - The session stopped and the Wingman judged it finished: the work is done, or it is only reporting something and asks you nothing.
- **Carrying on** - The session stopped, but the Wingman judged it will continue on its own, and it turns red if it does not.
- **Being read** - The session stopped and is being looked at before it is shown to you: the Wingman or the Director is reading the stop, or its voice summary is not ready yet.
- **Ready** - A brand-new session at its first prompt, which has not done anything yet.
- **Transcribing** - Your dictation is on its way: it is still uploading from your phone, or it is being turned into text.
- **Supervised** - Stopped, but another live session is driving it, so it waits on that session instead of you.
- **Snoozed or exited** - This session is resting: you snoozed it, its agent exited, or its state could not be read.
- **Crashed** - The agent process died.
- **Magenta** - Not a state. The screen received a colour this app does not understand, so reload and report it if it stays.

Every one of those claims has a session behind it, run through the real fold for its colour, its label and its bucket.

thefrederiksen:
## Review round 6, final pass: one finding closed, one recorded and left open

The magenta note is confirmed closed - it is built only from `BrokenClaims`, and every claim needs an exact `CheckBrokenClaim` case.

**Left open, in the reviewer's own words:** *"Tests execute each claim separately but never prove that an alternative's session satisfies its opening. Rearranging the existing cyan claims can render 'the work is done' over a Report session while all 16 focused tests still pass. A person can therefore still read a sentence nobody executed."*

**Why this is merging anyway, stated plainly so the trade is on the record:**

- Every sentence served today is TRUE, and every claim in it is executed against the real fold for its colour, its label and its bucket. The residual risk is a FUTURE edit that moves text between an opening and its cases, producing a garbled sentence - not a false statement in the product now.
- The defect class this pull request was really about - words a person reads that no test executes - is closed in every place it was found: the words moved to the Gateway, one session per claim, claims bound to the served text, joining words rendered from structure, both notes' claims checked, and the "all of these" relation and the free advice deleted.
- Six review rounds found eleven defects and all eleven are fixed, each with a control that fails when the fix is removed. This last one is worth doing when someone next edits a sentence; holding the legend out of the product for it costs the owner the thing he asked for.

Filed as follow-up work rather than carried silently.

PR #2899, opened 2026-09-15
A snooze that ends comes back honest: nothing new is calm, an outstanding ask is still an ask

thefrederiksen:
## What this does

A snooze that ends no longer manufactures a red on the clock. This is ruling 10 of the mission plan, and it is the last behaviour the Wingman owed the roster.

Expiry is not an event. `SnoozeRegistry.IsExpired` is a pure computation the fold already stamps onto every row, so a plain condition would fire on every roster poll, every display sweep and every accepted push. The trigger is therefore an EDGE: `SnoozeExpiryReJudge` keeps the previous fold's answer per account and session, the way `NeedsYouClock` keeps its stamp, and acts only on the transition from not expired to expired.

Three cases at that edge:

1. Nothing ended while the snooze ran and nothing had judged that stop - the row comes back cyan, label "Snooze ended, nothing new". No read, no model call, no stored record.
2. A stop happened while it ran and a verdict covers it - that verdict rules, calm or red.
3. A stop happened and no verdict covers it - the seat is asked to judge the current screen now, and the row keeps its red while it waits.

The edge is WON, not observed. The roster, the single-session read and every accepted push fold through one shared memory concurrently; reading "was it expired" and then writing it is two steps another fold can slip between, and both would ask the judge about one stop. The transition is a compare and swap and only the winner asks.

## Decisions the plan did not state, for the Architect to rule

1. **An accepted verdict rules the row whatever its age**, so "nothing new" also means "nothing had judged that stop". Ruling 10 never says what happens to a verdict the row already carried. Read literally today it quietens an outstanding question, because slice C made every owned stop judged, so a snoozed row almost always carries one: an unanswered "needed you" would come back cyan while the session still sits on that question. That makes a snooze a delete button rather than an alarm clock, and it empties ruling 10's own last sentence, "only needed-you brings it back red", by leaving nothing that can bring it back red. Ruling 10 predates universal judging; this is the reconciliation, and it is the safe side. Pinned as behaviour. To flip to the literal reading: delete one line in `SnoozeExpiryDecision.AtExpiry` and one clause in `SessionOrdering.IsSnoozeEndedNothingNew`; three tests change.
2. **Cyan, not the plan's green** - already ruled by the Architect during the build and recorded here as he asked. Green now means brand new, so a row returning from a snooze painted green would say the opposite of what is true about it.
3. **The account's colour switch gates the whole slice**, so an account in shadow does not re-judge on a snooze expiry. Noted against the slice D ruling that reversed the carrying-on clock onto the judge switch; the difference is that the clock reads the store while this reads the row, which is empty in shadow by design. Extending it would break a pinned slice D invariant, so it was not done.
4. **The snooze's set time is observed by the fold, not stored.** No column, no migration. The consequence is that a Gateway restarting mid-snooze decides nothing at that expiry and the row shows the red it would have shown before.
5. One new ledger event, `turn-verdict-snooze-expiry`, with three cause words, because the acceptance row for this slice reads "a staged snooze and its ledger rows".
6. A new trigger, `TurnVerdictTrigger.SnoozeExpiry`, rather than reusing turn end - the detector observed no turn end, so that word in the ledger would be untrue.
7. The fold takes its moment as an optional parameter so both ends of a snooze can be folded without a sleep in every test. Production passes null and reads the clock, unchanged.
8. `TurnVerdictRowStamp.Stamp` now returns whether this account's verdicts reached the rows, so the two stamps cannot disagree about one account.
9. The ladder gates the arm on the row not already carrying a judged verdict. Found by the author's own test, not by review: without it a row carrying both went cyan.

## Revert proofs

Each mutation was applied to committed work, so the restore is exact and a clean status after it is evidence of that.

1. **The edge** - the previous-fold comparison removed so it fires every fold. `TheEdgeFiresOnce_AThousandFoldsLaterNothingHasBeenAskedTwice` went red with exactly the reported symptom, twelve fires and twelve reads where there should be one, and `TheCalmStandsAcrossEveryLaterFold_AndIsStillNotRead` with it. The other 37 stayed green. Restored, 39 of 39 pass.
2. **The turn-ends-since-the-snooze comparison** - removed so the two cases cannot be told apart. Three tests went red, naming the read directly, `Expected: NothingNew, Actual: ReadRequested`. Restored, 39 of 39 pass.
3. **The compare and swap** - replaced with a blind write. `ManyFoldsRacingOnOneExpiry_AskExactlyOnceBetweenThem` went red with 32 reads for one expiry, three times out of three, never flaky. Restored, 40 of 40 pass.

## Tests

40 new tests in the parked `CcDirector.Gateway.UnitTests`, driven through the real fold over the real snooze registry and the real verdict store. Nothing hand-sets the stamp it asserts on, and every test first folds the snooze while it is still running so the arming is observed the way production observes it. All five tests the plan names are present, plus the colour pinned to cyan specifically, the outstanding-ask case pinned as behaviour, the restart case, the shadow case, the re-snooze case and a 32-thread race.

## Gate

The default gate on the final head: eight suites, 1,882 executed, zero failures, no suite executed zero.

`Core.Tests` parked, run in the foreground: 4,160 passed, 0 failed. It ran one commit earlier and stands for the final head by construction rather than assumption - the only later changes are in the Gateway project and its tests, and the Gateway project is not in that suite's closure.

`Gateway.UnitTests` parked, run in the foreground: 4,587 passed, 8 failed. The 8 are one class this change does not touch, `Stats.HostedSchemaRefusesAnUnownedRowTests`, every one failing in the class constructor before any product code runs, unable to reach a PostgreSQL server on the loopback address. The same 8 failed identically before this change. They were NOT observed passing; the run that provisions that rig is the parked gate below.

`Gateway.Tests` and the PostgreSQL-backed rig: running as the release-shaped parked gate on this exact head, in its own tracked terminal. Its result is added to this pull request when it lands, and this change is not to be merged before it does.

No migration. Nothing under `apps/` or `packages/` is touched, so the web gate does not apply; the calm band's TypeScript needed no change because a snooze-ended row carries no accepted verdict and the band selects on one, which is pinned by a test.

## What is not proven

No live proof on the three surfaces; that follows the deploy. The switching design's count path is written and pinned at the decision function but is unreachable in production because nothing produces the field. A Gateway that restarts mid-snooze decides nothing at that expiry. An account in shadow gets none of this. "Red stands" in case 3 is true at the moment of the request, not for the whole flight - once the seat asks, slice E's ruling makes the row yellow while it reads. No measurement of how often each case fires on a real day.

Full evidence, including the gaps in the author's own words: `docs/missions/wingman-on-every-turn/proofs/slice-f-proof.md` in the internal repository.


thefrederiksen:
## The base moved by one commit after the gate head, and it was checked

`origin/main` advanced to `6f932061c` (the launcher self-update fix, pull request 2900) after this branch was cut and after the parked gate started on `3f086672a`.

This branch is deliberately NOT rebased onto it. The reason is not convenience:

- The two changes have ZERO file overlap, checked mechanically rather than by eye. The new commit lives entirely in the launcher, the launcher tests and the installer engine; this slice is the Gateway roster fold, the verdict service, the session contract and the Gateway unit tests.
- Rebasing would move the head out from under a parked gate run that is more than half an hour in, holding the machine-wide Gateway suite lock, and that run would have to start again from nothing to take in a change that cannot reach any code this slice touches.

The staleness is recorded here rather than ignored, because the rule exists so that nobody builds on a base they have not examined. This base was examined. The merge is clean and the merge commit picks the launcher fix up.

One thing worth knowing for later: `6f932061c` does touch `tools/cc-director-setup-engine`, which is where the intermittent installer test tracked as issue 2878 lives - the one this gate hit again and which is written up in the gate section above. If that suite starts behaving differently after this lands, that commit is the first place to look.

thefrederiksen:
## The parked gate has landed, on this exact head, and it is green

`.\scripts\test-local.ps1 -Parked -Filter "FullyQualifiedName!~HostedImagePublishedArtifactTests"` on `3f086672a`, in its own tracked terminal. Counts read from the eleven result files, not from the console.

| Suite | outcome | executed | passed | failed | skipped |
|---|---|---|---|---|---|
| CcDirector.Gateway.Tests | Completed | 2475 | 2475 | 0 | 4 |
| CcDirector.Gateway.UnitTests | Completed | 4595 | 4595 | 0 | 2 |
| CcDirector.Core.Tests | Completed | 4475 | 4475 | 0 | 8 |
| CcDirector.Core.UnitTests | Completed | 467 | 467 | 0 | 0 |
| CcDirector.Avalonia.Tests | Completed | 479 | 479 | 0 | 0 |
| CcDirector.Engine.Tests | Completed | 63 | 63 | 0 | 0 |
| CcDirector.HostedAgent.Tests | Completed | 88 | 88 | 0 | 0 |
| CcDirector.Launcher.Tests | Completed | 194 | 194 | 0 | 0 |
| CcDirector.Terminal.Avalonia.Tests | Completed | 25 | 25 | 0 | 0 |
| cc-director-setup.Tests | Completed | 25 | 25 | 0 | 0 |
| cc-director-setup-engine.Tests | **Failed** | 541 | 540 | 1 | 0 |

**No suite executed zero.** That is checked explicitly and not inferred from the absence of a red line: a suite whose test host dies waiting on the machine-wide lock reports zero executed and does not fail the run by itself, which is the defect this mission files as a follow-up at slice H.

### Two things this settles

**`Gateway.Tests` ran.** Slice E never got this suite on any head - one attempt hung on the publish test tracked as issue 2871, and the retry was cancelled with no result file. It ran here, on this head, with that one test excluded by name, and 2,475 tests passed.

**The eight PostgreSQL failures are resolved, and were the rig, exactly as reported.** The author's own foreground run of `Gateway.UnitTests` saw 8 failures in `Stats.HostedSchemaRefusesAnUnownedRowTests`, every one failing in the class constructor on a refused socket, and said plainly in the proof that it had NOT watched them pass. This run provisions the throwaway PostgreSQL those tests need: 4,595 executed, 0 failed. The claim is now an observation.

### The one failure does not gate

`PythonToolsHealAndShimTests.InstallAsync_FailedVenvRebuild_LeavesNoManagedShim`, in the installer engine, a project this slice does not touch.

- Re-run alone on the same worktree: passed, 1 of 1, in 385 milliseconds.
- It is already tracked as issue 2878, filed earlier the same day from the v2.2.0 release gate: same test, same assertion at line 204, same message, same 540 of 541 counts, same heavy-parallel-load conditions. A second occurrence has been added to that issue, which turns "under load" from one coincidence into a repeated correlate. The cause remains unobserved and nothing here identifies it.

By the gate policy ruled for this mission, a failure in a project the slice does not touch that passes in isolation is written down rather than merged on. It is written down here.

thefrederiksen:
## The green parked gate above is SUPERSEDED - do not read it as this pull request's result

The parked gate result posted earlier passed on `3f086672a`. That head no longer exists on this branch: the branch has been rebased onto `origin/main` over pull request 2894, and the inspection's five findings are being fixed on top of that.

So the 2,475-test `Gateway.Tests` run, and every other count in that comment, describes a commit nobody will merge. It is left in place as a record of what was run and when, not as evidence about what is here now.

The parked gate will be run again, once, on the final head, and its result posted the same way. This pull request is not to be merged on the earlier one.

thefrederiksen:
## The parked gate on the final head is GREEN

`.\scripts\test-local.ps1 -Parked -Filter "FullyQualifiedName!~HostedImagePublishedArtifactTests"` on `97498ba1d`, in its own tracked terminal. Counts read from the eleven result files, not from the console.

| Suite | outcome | executed | passed | failed |
|---|---|---|---|---|
| CcDirector.Gateway.Tests | Completed | 2487 | 2487 | 0 |
| CcDirector.Gateway.UnitTests | Completed | 4786 | 4786 | 0 |
| CcDirector.Core.Tests | Completed | 4478 | 4478 | 0 |
| CcDirector.Core.UnitTests | Completed | 469 | 469 | 0 |
| CcDirector.Avalonia.Tests | Completed | 479 | 479 | 0 |
| CcDirector.Engine.Tests | Completed | 63 | 63 | 0 |
| CcDirector.HostedAgent.Tests | Completed | 88 | 88 | 0 |
| CcDirector.Launcher.Tests | Completed | 195 | 195 | 0 |
| CcDirector.Terminal.Avalonia.Tests | Completed | 25 | 25 | 0 |
| cc-director-setup.Tests | Completed | 25 | 25 | 0 |
| cc-director-setup-engine.Tests | **Failed** | 553 | 552 | 1 |

**No suite executed zero.** Checked explicitly rather than inferred from the absence of a red line.

### The test that had never run

`AFilteredRosterReadNeverPrunesTests` is new in this slice and, being in the parked host-bound suite, had never been executed before this run - which was said out loud in the proof rather than glossed. It ran here: **2 tests, 0 failed.** It drives the real `GET /sessions` routes, with and without a machine filter, against a real push store and snooze memory, and asserts both that a filtered read never prunes AND that the unfiltered read still does - so "never prunes" cannot be satisfied by never pruning at all.

### The 35 hosted failures from the earlier run: all six classes GREEN

An earlier parked run on a different head failed 35 tests across six classes that stand up a hosted Gateway. Each of those classes ran here and passed:

| Class | ran | failed |
|---|---|---|
| HostedVaultDenyTests | 11 | 0 |
| StatisticsFailureIsContainedOnTheHotPathTests | 7 | 0 |
| HostedContentReadDenyTests | 9 | 0 |
| HostedLauncherHubTenantTests | 7 | 0 |
| HostedAccountTrialEndpointTests | 8 | 0 |
| NetDiagEndpointTenantIsolationTests | 7 | 0 |

That is the second observation, and it is the one that was wanted: the same suite, the same machine, the same lock, on a head carrying MORE of this work, not less. The failures did not recur.

**What that does and does not establish.** It removes the worrying reading - that something on main breaks those six classes - because they pass here on a superset of the code. It does NOT establish the cause of the earlier run's failures, and nothing in this run observed one. The earlier failures clustered in two short windows totalling about two minutes, all in classes that stand up a hosted Gateway, with 503 Service Unavailable and null values, while 2,450 tests around them passed. Contention remains the shape of the evidence and is still not a proven cause.

### The one failure does not gate

`PythonToolsHealAndShimTests.InstallAsync_FailedVenvRebuild_LeavesNoManagedShim`, in the installer engine, a project this slice does not touch. It is tracked as issue 2878 and has now failed on five consecutive parked runs while passing alone every time; both earlier occurrences are commented there with the evidence. By the gate policy a failure in an untouched project that passes in isolation is written down rather than merged on.

PR #2906, opened 2026-09-16
docs(release): notes for v2.3.1

thefrederiksen:
Release notes for v2.3.1, covering all six changes that have landed since v2.3.0 - not only the launcher fix.

`new-release.ps1` refuses to cut a release without `docs/public/release-notes/v2.3.1.md` and applies a 200 non-whitespace-character floor, so this has to be on main before the release can be cut. 2773 characters, ASCII only.

The launcher section is deliberately written against what v2.3.0's notes claimed - *"a launcher updates itself, so machines pick this up on their own"* - which was true only on Windows. That sentence is what a reader has in mind, so this one says plainly that a machine already behind catches up on its own, with nothing to run by hand.

Implemented in #2900.


thefrederiksen:
Superseded by the v2.4.0 release (#2939), which carries these changes and describes the launcher fix.

PR #2907, opened 2026-09-16
A wrong verdict can be reported, and answering a row no longer deletes the record

thefrederiksen:
## What this does

Two things the owner asked for on the first live evening, and the loop that connects them.

**Answering a row no longer deletes the record.** A Working transition used to DELETE the session's stored verdicts, so the moment the owner answered a red row the verdict that made it red was gone - it could not be examined, reported wrong, or graded against what he actually did. He hit this on his own Architect row. `Invalidate` now SUPERSEDES: the row keeps its record with a moment stamped on it, the fold and the latest read ignore superseded records, the history returns them, and retention still ages everything out after seven days.

**A wrong verdict can be reported.** `POST /sessions/{sid}/turn-verdict/feedback` takes the account's correction, joins account, session and verdict, refuses a word outside the vocabulary, and replaces an earlier correction of the same verdict. The panel's "This is wrong" action, left disabled in slice E, is now live on both surfaces.

**And the correction reaches the corpus.** `GET /gateway/admin/turn-verdict-feedback` is the read the daily pull uses, and in the internal repository `pull_feedback.py` writes the rows to `corpus/labelled/owner-<date>.jsonl` as owner labels, which the grader ranks above a two-family label. The grader also catches up with three product rules it was behind on, and the stored answers are re-scored under them.

## Read this before reading the re-scored numbers

The re-score moves the fast tier's accepted answers from 222 of 341 to 185, false calm from 6.5 percent to 5.2, and the green "finished" false-calm row from 1.3 percent to **0.0**.

**That zero is not an improved judge and must not be read as one.** Every `finished` answer in that run was asked under the version one prompt and carries no `finishedKind`, so it is now refused rather than judged calm. The honest reading of the green false-calm rate today is NOT MEASURED. The report says so in its own words, and slice H is under written instruction not to copy the old numbers or the new zero into the product charter without the Architect's re-read of the slice 0 ruling.

The receipt-refusal rate is 63 of the turns the fast tier answered, about one in five. It is REPORTED, not reduced: the prompt's quoting instruction is the lever and it is deliberately untouched here.

## Decisions for the Architect to rule

1. **An administrator read route was added.** The plan says the daily pull reads "through the administrator surface" and none existed for these rows; the worker's brief did not name it. RULED IN SCOPE by the Manager, conditional on it being watched - see the tests and revert proof 4 below. It sits on the existing `AdminTrialEndpoint.ServiceTokenDenial` gate, called rather than copied, and serves one closed word, two moments, two identifiers and the note. Never a screen, a reply, a label or a summary, so it keeps the line the turn-log route draws.
2. The corrected word may be any of the SEVEN shared vocabulary words, not the six the Wingman may answer with. The destination is the corpus, whose vocabulary is the seven, and "not a turn end" is a real reading of a real failure none of the six can express.
3. `FindById` still finds a superseded verdict; the latest read does not. A report is made after the owner answers, and answering is what supersedes, so a lookup that skipped superseded rows would refuse almost every real report.
4. The supersede moment is served on the history read, so a superseded record and one in force do not look alike to a reader.
5. A re-store that mints a NEW verdict identifier clears the supersede stamp, as it already cleared the answered mark; one that keeps the identifier keeps both.
6. The feedback route writes no activity-ledger line. The ledger records acts that leave the Gateway; a correction writes a durable row carrying its own moment, word and note, which is the record.
7. The shadow rule is applied to the write exactly as the reads and the answer route apply it.
8. The panel owns the reporting flow in client-core rather than each shell passing a callback.
9. The client keeps its own list of the seven words, because a picker must hold them before one is chosen. A guard in the DEFAULT gate reads that file and fails when it differs from the server's list in content or order.
10. `Core.Tests` was run as eight foreground chunks because it does not fit one foreground call and backgrounding is forbidden. The partition is exhaustive and disjoint and was verified against the test list afterwards.

## Revert proofs

1. **The supersede** - the delete put back. Seven tests red of 171, with the verdict GONE: the history empty after the Working transition, the store's own supersede test reading 0 where it expects 2. Restored, 171 pass.
2. **The verdict-belongs-to-this-session join** - removed. Exactly one red of 27, and the symptom is that THE FEEDBACK WAS ACCEPTED for another session's verdict. Restored, 27 pass.
3. **The guard entry** - deleted. Exactly one red of 197, naming the method and path. Restored, 197 pass. What this leg does NOT prove is stated below.
4. **The administrator read's gate** - both token checks removed. Two red of 6: the gate test got **200 with the rows in the body** for a caller with no authorization header, and the unset-token test got 400 instead of 503 because the query was read before the token was looked at. Restored, 6 pass.

## The migration

Both providers, same pull request, generated with the documented commands following slice E's pair exactly. Additive and nullable on both, correctly typed per provider, schema-qualified only on PostgreSQL, reversible.

Designers match the snapshots on both - "No changes have been made to the model". APPLIED FRESH on both rather than merely generated: SQLite against a deleted design-time database through the whole chain, and PostgreSQL against a throwaway container, with the column then read back off the live table beside the one slice E added.

## Gate

Default on the final head: eight suites, no suite executed zero.

`Gateway.UnitTests` parked, on the final head: 4,594 executed, 4,594 passed, 0 failed, 2 skipped by their own conditions. **The first run of this suite was RED and it was the author's own bad invocation, recorded rather than quietly re-run** - a bare test command with no PostgreSQL rig reported 17 failures across three classes this change does not touch. The correct invocation provisions the rig, which the gate script does; provisioned, the suite came back green in full and the rig was torn down.

`Core.Tests` parked, on the final head: 4,475 executed across eight chunks, 0 failed. Coverage was checked rather than assumed - the test list names 4,481 tests and every one appears in the union of the eight result files.

Web, the exact commands from the continuous integration workflow: type check clean, all three test workspaces green (1,153 plus 339 plus 77), both builds built.

`Gateway.Tests`: running as the parked gate on this exact head, in its own tracked terminal, queued behind slice F's run for the machine-wide lock. Its result is added here when it lands, and this is not to be merged before it does.

## What is not proven

- **`Gateway.Tests` had not run when this was opened.** Seven new hosted tests and an extension to the PostgreSQL provider proofs were unrun. Among them is the only proof that the guard entry matches the route's REAL mapped path - a missing guard entry answers 403 in production while every other test stays green, which this repository has been bitten by before.
- **The round trip is proved in LEGS, not end to end.** Panel to route, route to store, store to administrator surface, surface to corpus file - each has a test, but no live Gateway served a real response to the pull, and no owner label produced this way has been graded. What stands in for the missing wire is that the field names are pinned from both sides.
- **No correction in the corpus can currently be joined.** Every bundle captured before 15 September has no observed-moment field, so the join key is absent. The tool counts and names every correction it could not label rather than silently writing fewer labels than it fetched. The first real owner label arrives with a turn captured after the field exists.
- Nothing here is evidence about the version two prompt; it has never been asked. Nothing is deployed.

Full evidence, including the author's own account of what it does not cover: `docs/missions/wingman-on-every-turn/proofs/slice-g-proof.md` in the internal repository.


thefrederiksen:
## The parked gate named in the body was STOPPED, and this branch is being rebased

The body says a parked gate is running on `8ac243e24`. That run was stopped deliberately, with a written reason: this branch must rebase onto `origin/main` (pull requests 2894, 2896 and 2900), so the run was testing a head nobody will merge while holding the machine-wide Gateway suite lock that the rebased run needs.

`2bfaa2a24` (pull request 2894) touches both migration snapshots, so the `AddTurnVerdictSupersededAt` pair is being REGENERATED on top of the current snapshot on both providers and applied fresh on each, rather than merely having its conflict resolved.

The parked gate runs once on the rebased head and its result is posted here. Nothing in the gate section of the body should be read as a result until then.

PR #2910, opened 2026-09-16
The narration names the right session, and stops reading identifiers aloud

thefrederiksen:
The owner audited the narration across all twenty of his sessions on the phone and called the result
ridiculously bad. Two of the three causes are here. Both are rules the judge prompt **already states**
and the model does not keep.

Placed by the Architect (`64181388`) to miss slice F and slice G: nothing here touches
`TurnVerdictService`, `SessionOrdering`, or the answer route.

## The model was writing the session name, and getting it wrong

| session | narrated as |
|---|---|
| `Wingman Inspector - Cockpit tab` | "Wingman Inspector **mobile**" |

Not mangled punctuation - a different name. The listener cannot see a screen and identifies the session
by that name, so a wrong one is worse than none.

The name is now prepended **in code** from the row's own title, at the point the spoken string is
assembled for synthesis, and the prompt's SPOKEN section tells the judge not to write one. A job the
model demonstrably fails is taken off it.

The title resolver has been supplied to this service since it was written and was **never read on the
narration path** - the prompt asked the model for the title instead. It is read now.

## It was reading identifiers aloud

Both heard live, against a prompt that says NEVER VOICE AN IDENTIFIER OR A HASH and REFERENCE NUMBERS
ARE NOT SPOKEN NUMBERS:

> "...confirm cutting the release **at 2bfaa2a24**. Say the word and it will be cut."
> "Agent recommends doing **issue 2905** first."

A deterministic scrub now runs at the same assembly point. A hash goes with whatever carried it so the
sentence still parses ("Merged at commit d0630a5." becomes "Merged."), and a reference number becomes
the work without the number ("doing that issue first") rather than the ungrammatical remains of
deleting the digits.

## What it deliberately does NOT do

The tests say this as loudly as they say the rest:

- **Ordinary English survives.** "defaced" and "effaced" are seven hex characters each - which is why
  the pattern demands a digit. A rule that eats real words to catch hashes is worse than the hashes.
- **A number that IS the answer survives.** "All 340 tests passed and 3 findings are fixed."
- **A version or a duration is not an identifier.** "Version 2.1 shipped and it took 46 seconds."
- **Only the SPOKEN string is touched.** Label, summary, evidence and options are untouched - they are
  read on a screen, where an identifier is useful and can be copied.

## Contract version

Stamped **v2.1**. The prompt file keeps its `v2` name and the grading tool keeps its path, because the
SHAPE and the validation did not change - only the wording.

The test that pinned version and filename equal is restated rather than loosened: the **major** part
names the file (the shape, which the grader reads off disk by path), and the **full** version is the
stamp on a stored record and must still move for a wording change, or a record cannot say which prompt
produced its words. The half that protects the grader is exactly as strict as it was.

## Verification

- **17 new tests**, two of them the owner's own utterances quoted verbatim from the live Gateway.
- Gateway unit suite **4648 pass**, Core **467 pass**. The 8 failures are
  `HostedSchemaRefusesAnUnownedRowTests` against a Postgres that is not running locally (nothing on
  5432); this change touches no file on that path.

## Still not fixed, and it is the third cause

Two sessions narrate **nothing at all**. Their verdicts are stored `failed` with
`failureReason: "the evidence is not found verbatim in the reply or on the screen"` - the model
fabricated its receipt, the receipt check correctly rejected it, and because a stop is never re-judged
those sessions are silent permanently while sitting in the owner's Needs you queue.

The receipt check is right and is not touched here. What is unresolved is that a rejected judgement
costs the owner the narration outright. That is the owner's ruling to make.


thefrederiksen:
Review result: changes requested at `a8086bbc4`.

1. **[P1] The identifier scrub deletes ordinary numeric answers, dates, versions, durations, decimals, and percentages.** [`IdBody` accepts any seven-or-more characters from digits/A-F/hyphens](https://github.com/thefrederiksen/devthrottle/blob/a8086bbc4a3b0e6e6395989c687df6b6bb3bc7eb/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L30-L51), so a digit-only value is necessarily treated as a hash, and a dotted value is matched one component at a time. I ran the branch implementation against adversarial inputs; these are the actual results:

   ```text
   1000000 tests passed.          -> tests passed.
   Pi is 3.1415926.               -> Pi is 3..
   Version 2.1234567 shipped.     -> Version 2. shipped.
   Release 2026-09-15 shipped.    -> Release shipped.
   The value is 12345678%.        -> The value is %.
   ```

   This directly violates the stated rule that a number which *is* the answer, a version, or a duration survives, and it changes the substance heard by the owner. A bare digit run is indistinguishable from a legitimate numeric answer; it cannot safely be scrubbed as a hash without a carrier. Reference nouns should be handled before the bare-hash pass, and a bare hash needs evidence beyond "contains a digit."

2. **[P1] `WithTitle` is not word-matching: it accepts title words as prefixes and deletes real content.** The inter-word and trailing expressions are `*`, with no word boundary/lookaround at either side ([lines 125-131](https://github.com/thefrederiksen/devthrottle/blob/a8086bbc4a3b0e6e6395989c687df6b6bb3bc7eb/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L121-L131)). Actual branch outputs:

   ```text
   title "Dev", body "Development is complete."
     -> "Dev. elopment is complete."
   title "Dev Manager", body "DevManager finished."
     -> "Dev Manager. finished."
   title "Go", body "Go-live failed."
     -> "Go. live failed."
   ```

   This is especially reachable for one-word/common-word titles. The match needs real word boundaries and must not consume punctuation that belongs to the body's first sentence merely because that sentence begins with the same word.

3. **[P1] The original wrong-name defect is still spoken; the right name is only prepended to it.** The code explicitly leaves a different model-written name in the body ([lines 109-112](https://github.com/thefrederiksen/devthrottle/blob/a8086bbc4a3b0e6e6395989c687df6b6bb3bc7eb/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L106-L112)). For the exact reported case, `Assemble` returns:

   ```text
   Wingman Inspector Cockpit tab. Wingman Inspector mobile. Merge complete with new table and fixes.
   ```

   The test named `TheWrongNameTheModelInvented_IsReplacedByTheRealOne` only asserts `StartsWith`, so it passes while the false name remains. The test should pin the complete spoken value (or at least assert the invented name is absent); otherwise the defect this PR exists to remove survives immediately after the correct title.

4. **[P2] Hash removal still leaves ungrammatical narration in the requested edge cases.** `CarriedIdentifier` is case-sensitive, requires whitespace before the carrier, and cannot cross quotes; the later bare pass then removes only the identifier ([lines 46-51 and 73-83](https://github.com/thefrederiksen/devthrottle/blob/a8086bbc4a3b0e6e6395989c687df6b6bb3bc7eb/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L46-L83)). Actual outputs:

   ```text
   At commit d0630a5f2b1c, we merged.       -> At, we merged.
   Merged at commit "d0630a5f2b1c".         -> Merged at commit "".
   Merged d0630a5f2b1c and 2bfaa2a24.       -> Merged and.
   ```

   The current positive tests cover only lowercase, unquoted carriers in mid-sentence and do not exercise a sentence-start carrier, quotes, or two hashes.

5. **[P2] The revised contract-version test no longer guards the full version.** `major` is derived from `Version`, so `Assert.StartsWith(major, Version)` is tautological; only `v2` and the v2 filename remain pinned ([lines 1634-1639](https://github.com/thefrederiksen/devthrottle/blob/a8086bbc4a3b0e6e6395989c687df6b6bb3bc7eb/src/CcDirector.Core.UnitTests/Wingman/TurnVerdictContractTests.cs#L1634-L1639)). I mutated `TurnVerdictContract.Version` back from `v2.1` to `v2` while leaving the revised prompt in place, and `ContractVersion_NamesThePromptFileByItsSHAPE_AndStillMovesForWording` still passed. Keeping the major/file assertions is sound, but the test also needs to pin the current full stamp (`v2.1`) so the next wording revision has to update it deliberately.

I also checked the requested wiring: both spoken assembly points call `SpokenForEar.Assemble`, and both production `WingmanVoiceService` constructions in `GatewayHost` supply `ResolveSessionTitle`; that resolver reads the tenant-scoped pushed-session record. The existing 17 `SpokenForEarTests` pass, as do the two targeted contract/resource tests. Eleven review-only adversarial cases above fail against the compiled PR implementation. I did not review the excluded `TurnVerdictService`, session ordering, or answer-route work, and this evidence does not exercise the external TTS provider or playback path.

thefrederiksen:
Review result: changes requested at `7dcb6c72b`.

1. **[P1] A "carrier" still turns ordinary dates, durations, and numeric results into identifiers.** [`CarriedIdentifier` treats the general prepositions `on`, `from`, `to`, and `for` as sufficient evidence for a digits-only identifier](https://github.com/thefrederiksen/devthrottle/blob/7dcb6c72be37810fb52bc6ef8eb71d8eed573140/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L97-L100). Those words also introduce exactly the legitimate values the first review found were being destroyed. Compiled against this head, the actual results are:

   ```text
   Deployed on 2026-09-15.                         -> Deployed.
   It ran for 12345678 seconds.                    -> It ran seconds.
   The population grew from 1000000 to 2000000.   -> The population grew.
   ```

   Requiring a hex letter fixes only a *bare* number. The carried pass deliberately drops that requirement, so original finding 1 remains open whenever normal prose puts one of these prepositions before the value. A digits-only run needs an identifier-specific carrier such as `commit`/`sha`/`hash`/`revision`; a general preposition is not evidence that a date or quantity is an identifier.

2. **[P1] `DropTheModelsOwnName` deletes short real instructions and status sentences.** [The heuristic removes the first sentence when only one title word (for a one-word title) or two title words agree and its word count is at most title length plus one](https://github.com/thefrederiksen/devthrottle/blob/7dcb6c72be37810fb52bc6ef8eb71d8eed573140/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L247-L260). Those conditions do not distinguish a name from a real sentence. Actual outputs:

   ```text
   title "Go", body "Go now! The deploy needs approval."
     -> "Go. The deploy needs approval."

   title "Dev Manager", body "Dev Manager failed. Retry queued."
     -> "Dev Manager. Retry queued."

   title "Construire", body "Construire maintenant! Le deploiement attend."
     -> "Construire. Le deploiement attend."
   ```

   The existing guard is long enough to clear the bound, so it cannot exercise the dangerous half of the heuristic. Common one-word titles are especially unsafe: any two-word imperative/status sentence beginning with that word vanishes.

3. **[P1] The invented-name defect still survives a colon-separated opening.** [`DropTheModelsOwnName` searches only for `.`, `!`, `?`, or `;` as the end of the supposed name](https://github.com/thefrederiksen/devthrottle/blob/7dcb6c72be37810fb52bc6ef8eb71d8eed573140/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L247-L254). A title-as-label is a natural model output, but the colon makes the matcher count the following clause as part of the "name" and reject it as too long:

   ```text
   title "Wingman Inspector - Cockpit tab"
   body  "Wingman Inspector mobile: Merge complete with fixes."
     -> "Wingman Inspector Cockpit tab. Wingman Inspector mobile: Merge complete with fixes."
   ```

   The listener again hears both the real name and the false one, which is the original defect this helper is meant to close.

4. **[P2] The dangling-conjunction sweep deletes legitimate words from narration that contained no identifier.** [`DanglingJoiner` removes every `and` or `or` immediately before clause punctuation](https://github.com/thefrederiksen/devthrottle/blob/7dcb6c72be37810fb52bc6ef8eb71d8eed573140/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L123-L125), and `Tidy` runs unconditionally. `and` and `or` can themselves be the spoken names of operators:

   ```text
   The condition uses AND, not XOR.   -> The condition uses, not XOR.
   The operator is OR.                -> The operator is.
   ```

   The sweep needs evidence that this particular joiner was orphaned by a removal; its punctuation position alone is not enough.

I re-ran the 32 checked-in `SpokenForEarTests` on `7dcb6c72b`; all 32 execute and pass, covering the five exact cases from the first review. I also re-ran the contract-version test, then mutated `Version` from `v2.1` to `v2`: the test reddened with expected `v2.1` / actual `v2`, so original finding 5 is closed. Nine review-only executable cases covering the four findings above all fail against the compiled head. I did not review the excluded `TurnVerdictService`, session ordering, or answer route, and this evidence does not exercise the external text-to-speech provider or playback path.

thefrederiksen:
Review result: changes requested at `f6e62fcbb`.

1. **[P1] The strictly-shorter heuristic deletes real status sentences under normal long session titles, while still missing padded invented names.** [`DropTheModelsOwnName` compares the candidate's word count only with the record title's word count](https://github.com/thefrederiksen/devthrottle/blob/f6e62fcbb31b80091b7c0a0e9febe939751aa02e/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L318-L333). Long fleet titles make concise real sentences satisfy that rule. Compiled against this head:

   ```text
   title "Dev Manager - Inspector - 2910 round three"
   body  "Dev Manager found four defects."
     -> "Dev Manager Inspector 2910 round three."

   title "Wingman Inspector - Cockpit tab"
   body  "Wingman Inspector mobile client view. Merge complete."
     -> "Wingman Inspector Cockpit tab. Wingman Inspector mobile client view. Merge complete."
   ```

   The first case loses the entire result because five real-sentence words are fewer than the six title words. The second keeps the false name because the model padded it beyond the four-word title. An unpunctuated `Wingman Inspector mobile` also survives because the helper requires a delimiter. The checked-in negative cases use one- and two-word titles, where a sentence is necessarily longer; they do not exercise the dangerous long-title surface. Length alone is not a discriminator in either direction.

2. **[P1] `commit` and `hash` are treated as naming nouns even when they are ordinary verbs, so real numeric content is deleted.** [`NamedIdentifier` accepts any seven-digit token after one of four spellings](https://github.com/thefrederiksen/devthrottle/blob/f6e62fcbb31b80091b7c0a0e9febe939751aa02e/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L89-L99); the regex has no evidence that the word is functioning as a noun. Actual outputs:

   ```text
   We will commit 1000000 rows tonight.
     -> We will rows tonight.

   The database will hash 1234567 records before upload.
     -> The database will records before upload.
   ```

   These are ordinary batch sizes, not identifiers. The split fixes digits behind general prepositions, but moves the same false-positive class behind ambiguous English words.

3. **[P1] Common parenthesized and colon-labelled commit references bypass the naming carrier and are still spoken.** The same pattern requires either start/whitespace immediately before the noun and whitespace immediately after it ([lines 89-99](https://github.com/thefrederiksen/devthrottle/blob/f6e62fcbb31b80091b7c0a0e9febe939751aa02e/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L89-L99)). Digits-only tokens then cannot fall through to the bare-hash pass. Compiled results are unchanged:

   ```text
   Merged (commit 1234567).   -> Merged (commit 1234567).
   Merged commit: 1234567.    -> Merged commit: 1234567.
   ```

   Both are natural model formatting for a real identifier, and both violate the scrub's core contract.

4. **[P1] The removal gate is global, so one real removal licenses destructive repairs against unrelated sentences.** `removedSomething` is one boolean for the whole narration, and every repair then runs over the whole string ([lines 169-205](https://github.com/thefrederiksen/devthrottle/blob/f6e62fcbb31b80091b7c0a0e9febe939751aa02e/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L169-L205)). Actual outputs:

   ```text
   The operator is OR. Merged deadbeef1.   -> The operator is. Merged.
   Wait... Merged deadbeef1.               -> Wait. Merged.
   ```

   The identifier removal is valid, but it causes `DanglingJoiner` to erase the real operator in an earlier sentence and `DoubledPunctuation` to collapse an unrelated ellipsis. The new no-removal tests close the exact round-two examples but do not make the repair local to the hole it is repairing.

5. **[P2] The capitalization helper mistakes punctuation characters for sentence boundaries and misses boundaries inside opening punctuation.** [`Determiner` inspects only the previous non-whitespace character](https://github.com/thefrederiksen/devthrottle/blob/f6e62fcbb31b80091b7c0a0e9febe939751aa02e/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L209-L223). Compiled outputs:

   ```text
   "Run 35047040578 failed."                  -> "that run failed."
   (Run 35047040578 failed.)                  -> (that run failed.)
   Use a prior run, e.g. run 35047040578.     -> Use a prior run, e.g. That run.
   ```

   A quote or bracket at the start hides the real boundary, while the period in `e.g.` invents one. The helper therefore gets both sides of its stated sentence-start rule wrong.

All nine earlier exact findings are closed at this head: the five numeric-answer cases, three prefix-title cases, exact invented-name case, three carrier/grammar cases, three general-preposition cases, three short-real-sentence cases, colon label, and no-removal `AND`/`OR` cases all execute in the checked-in suite; `72/72` `SpokenForEar` tests pass. The contract-version test also passes, and mutating `Version` from `v2.1` to `v2` makes it fail with expected `v2.1` / actual `v2`, so the first review's fifth finding remains closed.

The review-only compiled probes above fail outside that inventory. I did not review `TurnVerdictService`, session ordering, or the answer route, and this evidence does not exercise an external text-to-speech provider or playback.

thefrederiksen:
Round four review of `7fe9180d5fa7663aebd4ed25678a901ce6343128`: **REQUEST CHANGES — this is not safe to ship yet.**

1. **[P1] The remaining regex still deletes real narration that is not a leading exact title.** [`WithTitle`](https://github.com/thefrederiksen/devthrottle/blob/7fe9180d5fa7663aebd4ed25678a901ce6343128/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L85-L91) joins title words with `[^\p{L}\p{N}]{0,4}`: zero characters means `Dev Manager` matches `DevManager`, while the trailing comma/dash class treats punctuation inside a real sentence as proof that the title was a standalone clause. Executed against this head:

   - `Assemble("Dev Manager", "DevManager. Release complete.")` produced `Dev Manager. Release complete.`
   - `Assemble("Dev", "Dev, Manager found four defects.")` produced `Dev. Manager found four defects.`
   - `Assemble("Dev", "Dev—short for development—is complete.")` produced `Dev. short for development—is complete.`

   In all three, body text was eaten even though the exact title was not a standalone opening clause. That breaks the one remaining safety boundary of this implementation.

2. **[P1] `SpeakableTitle` corrupts valid Unicode names and can erase a nonblank title.** [`TitleNoise`](https://github.com/thefrederiksen/devthrottle/blob/7fe9180d5fa7663aebd4ed25678a901ce6343128/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L43-L57) permits letters and numbers but not combining marks (`\p{M}`), so it changes the words it claims never to change. Executed results: `नमस्ते` became `नमस त`, decomposed `équipe` became `e quipe`, and `+++` became empty, causing the narration to omit the session name entirely. The same rule reduces a semantically named `C++` session to `C`. The feature exists to say the real session name; saying a different name or none is a blocker.

3. **[P2] The `{0,4}` bound and trailing punctuation class also under-match titles the model genuinely copied.** The existing supported title `build (nightly) [v2]` followed by `build (nightly) [v2]. Result.` produced `build nightly v2. build (nightly) [v2]. Result.` because `]` is not accepted before the clause punctuation. `Alpha ----- Beta` likewise remains duplicated because its separator is longer than four characters. Regex metacharacters in word content are escaped correctly, but normalization means matching reconstructed words cannot reliably recognize the original written title.

4. **[P2] The prompt still gives the model contradictory title instructions.** The JSON shape says [`spoken` opens with the session title](https://github.com/thefrederiksen/devthrottle/blob/7fe9180d5fa7663aebd4ed25678a901ce6343128/src/CcDirector.Core/Wingman/Prompts/turn-verdict-v2.txt#L39), while the changed SPOKEN section says [do not write the title](https://github.com/thefrederiksen/devthrottle/blob/7fe9180d5fa7663aebd4ed25678a901ce6343128/src/CcDirector.Core/Wingman/Prompts/turn-verdict-v2.txt#L156-L160). A judge following the schema can reproduce the duplicate/invented-name defect this PR is meant to remove. The contract tests pass because they do not assert that the old instruction is gone.

5. **[P2] The claimed byte-preservation property is false even before the regex fires.** [`var body = (spoken ?? "").Trim()`](https://github.com/thefrederiksen/devthrottle/blob/7fe9180d5fa7663aebd4ed25678a901ce6343128/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L78-L83) changes leading and trailing bytes for every caller. `Assemble("session", "  Result.  \t")` produced `session. Result.`. `SpeechContract.Finish` currently masks this at the two production call sites by trimming first, but the public method, header, and supposed property test all promise more than the code proves; the test uses one fixed already-trimmed body and is not a property test.

What did check out:

- The identifier/reference scrub is actually gone from `SpokenForEar`; the prompt's denial of an identifier-specific downstream safety net matches runtime. Generic Markdown cleanup and the 12,000-character audible cap still exist, so “exactly as you wrote it” is broader than the shipped pipeline. The stale `WingmanVoiceService` comment at lines 1509–1511 still says identifiers are removed there and should be corrected.
- Both `WingmanVoiceService` call sites compile against the reduced API. Both pass `_sessionTitleResolver(tenant, sid)`, and both production construction sites wire that resolver to `ResolveSessionTitle`, which reads the pushed session record's real `Name`.
- Focused existing tests: Gateway `SpokenForEarTests` **33/33 passed**; Core `TurnVerdictContractTests` **84/84 passed**.
- An added-and-removed adversarial fixture ran 14 cases: **5 passed, 9 failed**. The pass controls covered an RTL Hebrew title, a regex-metacharacter title, title-equals-whole-body, punctuation-only body, and a 2,000-word title. The failures are the concrete outputs above.

Scope: I reviewed only the requested speech helper/tests, the two voice-service assembly call sites and their production title wiring, and the prompt/contract assertions. I did not inspect `TurnVerdictService`, session ordering, or the answer route, and I did not run the full Gateway suite.

thefrederiksen:
MERGE DECISION: **NOT SAFE TO SHIP** at `f362062fc218452b29a2927c0e3b648496db4c1b`.

### [P1] The production callers still delete narration words before `SpokenForEar`

Both callers pass `verdict.Spoken` through `SpeechContract.Finish` before `Assemble` ([turn-end path](https://github.com/thefrederiksen/devthrottle/blob/f362062fc218452b29a2927c0e3b648496db4c1b/src/CcDirector.Gateway/Wingman/WingmanVoiceService.cs#L1514-L1516), [on-request path](https://github.com/thefrederiksen/devthrottle/blob/f362062fc218452b29a2927c0e3b648496db4c1b/src/CcDirector.Gateway/Wingman/WingmanVoiceService.cs#L1644-L1646)). That pre-pass is not word-preserving: it deliberately removes fenced blocks whole. Running the exact head with a body of `The fix is:` followed by a fenced block containing `Use AND, not XOR.` produced `Dev Manager. The fix is:`; the decisive words were gone. The new unit tests call `SpokenForEar.Assemble` directly, so they prove the helper preserves the already-sanitized input but do not cover either production caller. This directly falsifies the design's single invariant and the adjacent `NOTHING ELSE about the narration is changed` comment.

### [P2] A valid separator-only session name still disappears

[`SpeakableTitle`](https://github.com/thefrederiksen/devthrottle/blob/f362062fc218452b29a2927c0e3b648496db4c1b/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L52-L56) maps `_-/:|` to the empty string, and `Assemble("_-/:|", "Build passed.")` therefore returns only `Build passed.`. This is reachable: explicit session-name validation rejects blank names and the bare repository name, but accepts this non-whitespace name. The method consequently fails to do its one job for the exact separator-only title class the review requested. The same regex also does not actually collapse a single embedded whitespace character: `Alpha\tBeta` retains code unit 9 because `\s{2,}` only matches runs of two or more.

### [P2] Repository contracts still promise the deleted behavior

The changed v2.1 history in [`TurnVerdictContract.cs`](https://github.com/thefrederiksen/devthrottle/blob/f362062fc218452b29a2927c0e3b648496db4c1b/src/CcDirector.Core/Wingman/TurnVerdictContract.cs#L39-L45) says the judge is told identifiers and reference numbers "are stripped after it answers," while the shipped prompt correctly says nothing downstream will catch them. Separately, [`TurnVerdictDto.Spoken`](https://github.com/thefrederiksen/devthrottle/blob/f362062fc218452b29a2927c0e3b648496db4c1b/src/CcDirector.Gateway.Contracts/TurnVerdictDto.cs#L80-L82) still claims that stored field opens with the session title, although v2.1 now stores the judge's title-free value and only the audio assembly prefixes it. These are stale executable-contract-adjacent claims about the behavior this PR deletes.

Checked clean: the full prompt's JSON-schema line and SPOKEN section both tell the judge not to write the session name; the prompt explicitly says no downstream identifier scrub exists. Both `WingmanVoiceService` paths compile, and both production constructions inject `ResolveSessionTitle`, which reads `located.Session.Name` from the tenant-scoped pushed record. The 32 `SpokenForEarTests` and the two prompt/resource-version tests passed at the exact head. Those tests do not cover the production `SpeechContract.Finish` pre-pass described above.

thefrederiksen:
MERGE DECISION: **NOT SAFE TO SHIP** at `b1a3674eb351bbef0559777a5ddae72a944eb0c8`.

1. **[P1] `SpeakableTitle` still says a different name for valid punctuation-bearing titles, and the separator-only fallback sends raw punctuation to synthesis.** [`Separators` is applied everywhere](https://github.com/thefrederiksen/devthrottle/blob/b1a3674eb351bbef0559777a5ddae72a944eb0c8/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L43-L45), not only where a separator joins two words, while [the empty-result fallback returns the original punctuation](https://github.com/thefrederiksen/devthrottle/blob/b1a3674eb351bbef0559777a5ddae72a944eb0c8/src/CcDirector.Gateway/Speech/SpokenForEar.cs#L62-L70). These names are accepted by the product because explicit-name validation rejects only blank names and the bare repository name. Compiled against this head, review probes produced:

   ```text
   UTC-5       -> UTC 5
   Phase 1/2   -> Phase 1 2
   -1          -> 1
   _-/:|       -> _-/:|. Build passed.
   ```

   The first three change the user's word and can reverse or blur its meaning; the file's own motivating rule says a wrong name is worse than none. The last case avoids an empty .NET string but does not make the name speakable: [`StoreSpokenAsync` passes the assembled text to `TtsAsync`](https://github.com/thefrederiksen/devthrottle/blob/b1a3674eb351bbef0559777a5ddae72a944eb0c8/src/CcDirector.Gateway/Wingman/WingmanVoiceService.cs#L1135-L1148), which [forwards `spoken.Text` as the provider input](https://github.com/thefrederiksen/devthrottle/blob/b1a3674eb351bbef0559777a5ddae72a944eb0c8/src/CcDirector.Gateway/Wingman/WingmanVoiceService.cs#L2002-L2039). The provider can therefore voice those punctuation names or ignore them and leave the session unnamed. I did not call the external provider, so the exact audio outcome is not covered; the raw punctuation reaching its input is.

2. **[P2] The repository still contains current TurnVerdict contracts that say the judge-written spoken section opens with the title.** [`TurnVerdictPackage.SessionTitle`](https://github.com/thefrederiksen/devthrottle/blob/b1a3674eb351bbef0559777a5ddae72a944eb0c8/src/CcDirector.Core/Wingman/TurnVerdictPackage.cs#L60-L66) says exactly that, and [`GatewayWingmanVoiceEndpoint`](https://github.com/thefrederiksen/devthrottle/blob/b1a3674eb351bbef0559777a5ddae72a944eb0c8/src/CcDirector.Gateway/Api/GatewayWingmanVoiceEndpoint.cs#L180-L181) says the verdict package carries the title that `spoken` opens with. Both now contradict contract v2.1 and the code-side prefix. The older `WingmanTranslator.FidelityPrompt` and its tests also still require a model-written title; grep found no production caller of `TranslateAsync` or `TranslateWithAsync` at this head, so I am not counting that legacy path as a runtime regression, but it is still a remaining repository description of the deleted behavior.

What checked clean: `Assemble` does not change any narration-body word; both production call sites run `SpeechContract.Finish` and then `SpokenForEar.Assemble`, and their two-step comments are true; both production `WingmanVoiceService` constructions inject `ResolveSessionTitle`, which reads `located.Session.Name` from the tenant-scoped pushed-session record. A serial Gateway build completed with 0 warnings and 0 errors. The 34 checked-in `SpokenForEar` cases plus 8 review probes passed (42/42), and `TurnVerdictContractTests` passed 84/84. I did not review the excluded `TurnVerdictService`, session ordering, or answer route, and I did not re-raise the fenced-block loss tracked in #2911.

thefrederiksen:
MERGE DECISION: **NOT SAFE TO SHIP** at `0311905e30f98f1c49ef94d9301b374c2ce06412`.

1. **[P2] The changed v2.1 contract and shipped prompt still overclaim what happens downstream.** [`TurnVerdictContract` says that "NOTHING downstream strips an identifier"](https://github.com/thefrederiksen/devthrottle/blob/0311905e30f98f1c49ef94d9301b374c2ce06412/src/CcDirector.Core/Wingman/TurnVerdictContract.cs#L39-L44), and [the prompt says both that nothing downstream will catch one and that whatever the judge leaves is spoken "exactly as you wrote it"](https://github.com/thefrederiksen/devthrottle/blob/0311905e30f98f1c49ef94d9301b374c2ce06412/src/CcDirector.Core/Wingman/Prompts/turn-verdict-v2.txt#L169-L177). Both production callers run the judge's text through [`SpeechContract.Finish`](https://github.com/thefrederiksen/devthrottle/blob/0311905e30f98f1c49ef94d9301b374c2ce06412/src/CcDirector.Gateway/Speech/SpeechContract.cs#L131-L141) before `Assemble`; that pass removes a fenced block whole, including any identifier inside it, and performs other Markdown/whitespace normalization. The new `SpokenForEar` header and call-site comment correctly acknowledge that fact, so the changed files contradict one another. This does **not** ask for the fenced-block behavior back or reopen #2911: make the prose say the narrower true thing, such as "there is no downstream identifier-specific scrub" / "no downstream identifier scrub catches one."

2. **[P2] Two active production comments still cite the retired model-written-title mechanism.** Both `GatewayHost` construction comments say the title behavior is from [`WingmanTranslator.FidelityPrompt v5.2`](https://github.com/thefrederiksen/devthrottle/blob/0311905e30f98f1c49ef94d9301b374c2ce06412/src/CcDirector.Gateway/GatewayHost.cs#L2832-L2837) (the duplicate is [here](https://github.com/thefrederiksen/devthrottle/blob/0311905e30f98f1c49ef94d9301b374c2ce06412/src/CcDirector.Gateway/GatewayHost.cs#L3715-L3720)). The outcome in those comments remains true, but the cited mechanism is now the legacy path: production prefixes the pushed-record name in `SpokenForEar`. These are the live, untouched-file leftovers from the requested repository-wide grep.

The other requested checks are clean:

- `Assemble` preserves every body word; only leading/trailing body whitespace is trimmed. A .NET 10 probe against this exact head preserved embedded controls and surrogate pairs, preserved a two-million-character name byte-for-byte at the UTF-16 string level, collapsed 500,000 whitespace runs as specified, and returned empty for an all-Unicode-whitespace title.
- `SpeakableTitle` performs only `\s+` to one ASCII space plus edge trim. Unicode letters/marks, punctuation, non-whitespace controls, valid and unpaired surrogates all survived unchanged in the compiled probe.
- Both `WingmanVoiceService` assembly sites compile and pass `_sessionTitleResolver(tenant, sid)`. Both production constructions inject `ResolveSessionTitle`, which reads `located.Session.Name` from the tenant-scoped pushed-session record.
- Focused execution: `SpokenForEarTests` 37/37, `TurnVerdictContractTests` 84/84, and `TurnVerdictVoiceMergeTests` 4/4 passed.

Repository-wide wording inventory: `WingmanTranslator.FidelityPrompt`, `BuildPrompt`, and their tests still describe a model-written/repunctuated title and model-side identifier dropping. [`SpokenPaths` explicitly marks that whole-answer translator as a draft path with no production route and test-only callers](https://github.com/thefrederiksen/devthrottle/blob/0311905e30f98f1c49ef94d9301b374c2ce06412/src/CcDirector.Gateway/Speech/SpokenPaths.cs#L44-L56), so it is legacy, not a runtime regression. The `SpokenForEar` source/tests mention the scrub, title de-duplication, and repunctuation only as deleted history. The old issue-368 plan files describe a separate, historical backtick-cleanup bug. No live title de-duplication, identifier scrub, or code-side name repunctuation remains on this path.

Scope stayed on the five requested questions. I did not review `TurnVerdictService`, `SessionOrdering`, the answer route, or the fenced-block defect itself.

thefrederiksen:
MERGE DECISION: **NOT SAFE TO SHIP** at `79a7f01a45bc924027f290bcec2ec27b40c1ae25`.

1. **[P2] A third active `GatewayHost` title description still attributes the behavior to the retired model-written-title path.** [`GatewayHost.cs` still says the title is for `WingmanTranslator.FidelityPrompt v5.2` and that "the prompt rule no-ops on a missing title"](https://github.com/thefrederiksen/devthrottle/blob/79a7f01a45bc924027f290bcec2ec27b40c1ae25/src/CcDirector.Gateway/GatewayHost.cs#L2212-L2221). At this head the turn-verdict prompt says **not** to write the title, `SpokenForEar.WithTitle` is what handles a missing title, and both repaired construction comments correctly say `FidelityPrompt` is the legacy translator path with no production caller. Those claims still disagree. The stale block is also followed immediately by a second `<summary>`, so it documents `AmbientSnapshotFresh`; `ResolveSessionTitle` itself has no header. Move/update this description onto `ResolveSessionTitle` and name the current mechanism: the resolver reads the pushed-record name and `SpokenForEar` prefixes it during assembly.

The other requested checks are clean:

- The rebase did not alter this feature's seven existing patches: `git range-diff` matched all seven old/new commits one-for-one with `=`. Comparing the old head with the rebased pre-wording-fix head changed only slice-F files; within the reviewed path, only unrelated slice-F sections of `GatewayHost` differ. The prompt, contract history, `SpokenForEar`, and `WingmanVoiceService` are byte-identical across the rebase.
- At the current head there are exactly two production `SpokenForEar.Assemble` calls, both pass `_sessionTitleResolver(tenant, sid)` and `SpeechContract.Finish(verdict.Spoken)`; there are exactly two production `WingmanVoiceService` constructions, both inject `ResolveSessionTitle`; that resolver reads `located.Session.Name`; and assembly still delegates to `WithTitle`.
- The round-seven prompt and `TurnVerdictContract` correction, the `SpokenForEar` header, both call-site comments, and the two repaired construction comments consistently say the narrower true thing: the generic Markdown pass runs first and can drop fenced blocks, no downstream step looks for an identifier, and the record title is prefixed afterward. The remaining `GatewayHost` block above is the contradiction.

Scope stayed on the requested rebased narration/title path and wording set. I did not review `TurnVerdictService`, session ordering, the answer route, or reopen the fenced-block defect in #2911. I did not exercise the external text-to-speech provider or playback.

PR #2916, opened 2026-09-16
The Wingman's charter says what the judge actually is, and an unmeasured rate never reads as zero

thefrederiksen:
## What this does

The last slice of the Wingman-on-every-turn mission. It makes the written record true: the product's own charter, the settled specification, and the dead code the build replaced.

**The charter, `docs/wingman/WINGMAN.md`.** Invariant 1 is reworded from "strong model only - NEVER a cheap model" to "the judge that scored best on the labelled corpus", with the score and the reply times written into the document and the models in force named there rather than in a hard-coded set inside a test. A new section describes the turn verdict: why it exists, the one question and its contract, the free checks in order, which way every gate errs, what the verdict may NOT decide, what it does to a row, and the activation route as the owner typing. Section 7 says why a model judge is back and what makes it a different object from the one that was deleted. Section 8 separates the detector's open items, which belong to the switching design, from this build's.

**The specification.** `docs/architecture/wingman/TURN_VERDICT.md` is new, with four sanitised twins. `TURN_BRIEFING.md` is marked superseded at the top and in its status line, in the past tense, naming the document that replaced it - not deleted, because a superseded document that says so is a record and a deleted one is a hole.

**The dead code, as one closed set**: `TurnBriefContract`, `ITurnBriefGenerator`, `StubTurnBriefGenerator`, `TurnBriefTests.cs`, `WingmanGoldenTests.cs` and the three doc-comment references. `TurnPackage` and `TurnPackageBuilder` are KEPT - `DictatedPromptResolver` uses them in real code, and the first version of the plan named the wrong consumer. Plus the stale comment in `GatewayEndpoints.cs` about the pushed-session store being null when stream mode is off, which the code has contradicted for longer than this mission.

## The number this slice most had to get right

The charter carries the 15 September figures for the fast tier **labelled in the document as measured under contract version one**: every call answered, ninety-fifth percentile 20.2 seconds, 222 of 341 accepted, false calm 10 of 154 (6.5 percent), of which `finished` is 2 of 154 (**1.3 percent**).

It then records what the re-score under contract version two did - acceptance 185 of 341, false calm 8 of 154 (5.2 percent) - **and the `finished` row as NOT MEASURED, never as 0.0 percent**, with the reason beside it: every `finished` answer in that run was asked under version one, which never asked for `finishedKind`, so version two refuses all of them. There is no accepted answer of that kind for a mistake to be counted among, the numerator cannot rise however the judge behaves, and a rate over no accepted answers is not a rate.

Printed as 0.0 percent it would read as a perfect score on the row that gates the colour switch. The charter says the row becomes a number again on the first grading run whose `finished` answers were asked under version two.

The figures were read from the grading report **as it stands on the internal `origin/main`**, not from the copy in the mission record worktree, which was five commits behind and still carried the pre-re-score numbers. That hazard is named in the proof.

## Revert proofs

On the charter audit pin, done twice - once for each direction it now binds. Every run built; no `--no-build` anywhere, including the restore runs, and both mutations were made after the fix was committed so a checkout could not eat it.

1. **The code moves away from what the charter names.** `WingmanService.Model` mutated from `opus` to `sonnet`: one test red, naming the model and quoting the line the charter must carry. **`sonnet` is the deliberate choice of mutant and it is the whole point of the change** - the OLD pin was a set containing both `opus` and `sonnet`, so this mutation would have passed it silently. The new pin catches it. Strictly tighter than the one it replaced, demonstrated rather than asserted.
2. **The charter loses the sentence that keeps an unmeasured rate from reading as zero.** `NOT MEASURED, not zero` replaced with `0.0 percent`: one test red. The safety number is now guarded by a test, not by prose.

## Decisions for the Architect to rule

1. **`docs/architecture/mobile/voice-mode.html` was re-pointed, not left dangling.** Its table named the deleted contract and the superseded document; the row now names their replacements and says in the cell what it used to name and why, so a reader of the old document is not left wondering.
2. **The `wingman-brief-reviewer` skill was marked RETIRED, not deleted.** It reviewed briefs from a store nothing has written to since issue 549 and names the deleted class as the contract under review. It carries a banner and a rewritten description saying what replaced the thing it reviewed and the loop it ran. Kept because its cold-reader framing is the reasoning behind the verdict's label, summary and receipt rules, and deleting it leaves a hole where the argument is. It is NOT one of the four built-in skills, so no Gateway deploy is involved. If you would rather it were deleted outright, that is one `git rm`.

The rest are in the proof, numbered.

## Gate

The default gate green. `Core.Tests` 4434 over nine foreground runs whose partition was verified exhaustive. `Gateway.UnitTests` 4786, **with its 8 PostgreSQL-rig tests watched passing under the provisioned rig** rather than explained away. The parked `Gateway.Tests` run is going on this exact head - dead-code deletion can break a test project's compilation, so it is not a formality here - and its result is posted when it lands. This is not to be merged before it does.

## Issues filed

2912, 2913, 2914 and 2915. Issue 2878 was RAISED with new evidence rather than re-filed, because it already existed and a duplicate would have split the record.

Full evidence, including what it does not cover: `docs/missions/wingman-on-every-turn/proofs/slice-h-proof.md` in the internal repository.


thefrederiksen:
## The parked gate is GREEN on this head

`.\scripts\test-local.ps1 -Parked -Filter "FullyQualifiedName!~HostedImagePublishedArtifactTests"` on `d378833b8`, in its own tracked terminal. Counts read from the eleven result files.

| Suite | outcome | executed | passed | failed |
|---|---|---|---|---|
| CcDirector.Gateway.Tests | Completed | 2487 | 2487 | 0 |
| CcDirector.Gateway.UnitTests | Completed | 4786 | 4786 | 0 |
| CcDirector.Core.Tests | Completed | 4426 | 4426 | 0 |
| CcDirector.Core.UnitTests | Completed | 469 | 469 | 0 |
| CcDirector.Avalonia.Tests | Completed | 479 | 479 | 0 |
| CcDirector.Engine.Tests | Completed | 63 | 63 | 0 |
| CcDirector.HostedAgent.Tests | Completed | 88 | 88 | 0 |
| CcDirector.Launcher.Tests | Completed | 195 | 195 | 0 |
| CcDirector.Terminal.Avalonia.Tests | Completed | 25 | 25 | 0 |
| cc-director-setup.Tests | Completed | 25 | 25 | 0 |
| cc-director-setup-engine.Tests | **Failed** | 553 | 552 | 1 |

13,596 tests executed. No suite executed zero - checked explicitly, not inferred from the absence of a red line.

### The deletion is visible in the counts, which is the point

`Core.Tests` executed **4,426 here against 4,478 on the previous slice's gate - 52 fewer**, which is `TurnBriefTests` and `WingmanGoldenTests` no longer existing. That is positive evidence that the dead-code deletion took effect, rather than the weaker claim that nothing broke. A deletion slice whose test count did not move would be the thing worth worrying about.

And `Gateway.Tests` compiled and ran in full at 2,487, which is the check this gate was really for: deleting a type that a parked, host-bound test project still referenced would have shown up as a compile error, not as a failing test, and a run that never built is not a run that passed.

### The one failure does not gate

`PythonToolsHealAndShimTests.InstallAsync_FailedVenvRebuild_LeavesNoManagedShim`, in the installer engine, a project this slice does not touch. It is tracked as issue 2878, has now failed on six consecutive parked runs, and passes alone every time. The evidence and the argument for fixing it rather than continuing to write it down are on that issue.

PR #2932, opened 2026-09-16
CI: run the cc-devthrottle tests on Windows, macOS and Linux

thefrederiksen:
Part of #2922 - the first of the six pull requests in the implementation brief. Nothing about the output changes here. The point of this one is to make the tests run everywhere first, so the output work that follows has a place to prove itself.

## What changed

**The Python tool-contracts job now runs on `windows-latest`, `macos-latest` and `ubuntu-latest`.** It ran on Windows alone, which meant the `cc-devthrottle` suite proved the tool worked on one operating system. `fail-fast` is off on purpose: when one operating system fails, the other two finishing is exactly the information the matrix exists to produce.

**The first macOS run found a real product defect, not a test defect (corrected after inspection).** `test_doctor_uses_recorded_python_tool_scripts` stands up a healthy install and put only the install bin directory on `PATH`. On macOS and Linux the shims are links in `~/.local/bin`, and that is the ONLY directory the installer puts on `PATH` (`InstallFinalizer.EnsureMacUserBinOnPath`, `tools/cc-director-setup-cli/Commands.cs`). The first version of this pull request "fixed" the test by putting both directories on `PATH` - which hid the fact that the setup doctor still demanded the install bin directory, so every healthy Mac install reported `install bin directory is not on PATH` and `needsRepair: true`. The inspection caught it. Now:

- The doctor requires the platform's shim directory on `PATH`: the install bin directory on Windows, `~/.local/bin` on macOS and Linux. `--json` keeps every existing key (`binDirOnPath` still means the install bin directory) and adds `shimDir` and `shimDirOnPath`; the human output shows both.
- The fixture models the `PATH` the installer really leaves: install bin directory on Windows, `~/.local/bin` only elsewhere.
- New test `test_doctor_requires_the_platforms_shim_directory_on_path`: a healthy install with that `PATH` has no problems, and taking the shim directory off `PATH` is reported.

**The setup-tool download no longer hands a Windows exe to macOS or Linux.** When the setup tool is not installed, `setup install/update/repair/autostart` downloads it from the latest release - and the selector listed only `devthrottle-setup-cli-win-x64.exe`, so a Mac or Linux machine would download that and try to run it. The release workflow (`.github/workflows/release.yml`) publishes `devthrottle-setup-cli-win-x64.exe`, `devthrottle-setup-cli-mac-arm64` and `devthrottle-setup-cli-linux-x64`. Selection is now keyed by operating system and processor using exactly those names (Windows on Arm takes the x64 build, which Windows runs under its built-in emulation). A platform with no published setup tool (Intel Mac, Arm Linux, anything else) fails with an error naming the platform and what is published, BEFORE any network call. The downloaded tool is made executable on macOS and Linux, since a release asset carries no executable bit. New tests cover macOS and Linux selection, refusal of a Windows-only release on both, the unpublished-platform error, no download for an unpublished platform, platform name normalisation, and a download that fetches the Linux asset and leaves it executable.

**Second inspection: two more defects, both fixed.**

- **The doctor demanded a directory Unix never creates.** On macOS and Linux the installer links the shims into `~/.local/bin` (`PythonToolsInstaller.WriteUnixShims`) and never creates the install bin directory, yet the doctor reported `install bin directory is missing` and `needsRepair: true` on a healthy install. The test fixture created that directory everywhere, which hid it. Now only the directory that holds the shims must exist (`tool shim directory is missing: <path>`): the install bin directory on Windows, `~/.local/bin` elsewhere. The fixture creates the install bin directory on Windows only. New tests: `test_doctor_healthy_install_needs_only_the_shim_directory` (asserts the install bin directory is absent off Windows and there are no problems) and `test_doctor_reports_a_missing_shim_directory`.
- **The setup-tool download asked the wrong repository, and hid the failure.** The release repository defaulted to `devthrottle/devthrottle`, which answers 404; `_latest_release()` turned that 404 into a silent "nothing found", so no setup tool could ever be downloaded. The default is now `thefrederiksen/devthrottle`, the same as `tools/cc-director-setup-engine/GitHubRepositoryDefaults.cs`, with `DEVTHROTTLE_GITHUB_OWNER` and `DEVTHROTTLE_GITHUB_REPO` still overriding it (blank means the default, as in the engine). Any HTTP or network failure now raises an error naming the address and the status, and `setup install/update/repair/autostart` prints it and exits 1. New tests: the default release address, the overrides, an HTTP 404 raising with status and address, and an HTTP 403 reaching the user's screen.

**A Windows path literal in `test_session_stop.py` is now a raw string.** It was a plain string, so `\a` became a bell character and `\R` raised a `SyntaxWarning` on newer Python. The test was not carrying the path it meant to carry. The two other path literals in that file were already raw.

## Proof

**Each fix can fail.** Reverting the doctor rule to demand the install bin directory, with the corrected fixture:

```
FAILED tests/test_setup_ops.py::test_doctor_uses_recorded_python_tool_scripts
FAILED tests/test_setup_ops.py::test_doctor_requires_the_platforms_shim_directory_on_path
2 failed, 20 passed
```

Reverting asset selection to the Windows-only list for every platform:

```
FAILED tests/test_setup_ops.py::test_select_setup_cli_asset_picks_mac_asset_on_macos
FAILED tests/test_setup_ops.py::test_select_setup_cli_asset_picks_linux_asset_on_linux
FAILED tests/test_setup_ops.py::test_select_setup_cli_asset_never_offers_windows_exe_to_macos_or_linux
FAILED tests/test_setup_ops.py::test_select_setup_cli_asset_unpublished_platform_raises[macos-x64]
FAILED tests/test_setup_ops.py::test_select_setup_cli_asset_unpublished_platform_raises[linux-arm64]
FAILED tests/test_setup_ops.py::test_select_setup_cli_asset_unpublished_platform_raises[freebsd-x64]
FAILED tests/test_setup_ops.py::test_run_setup_cli_unpublished_platform_fails_before_any_download
FAILED tests/test_setup_ops.py::test_download_setup_cli_fetches_this_platforms_asset
8 failed, 14 passed
```

Restored: `22 passed`.

Second inspection fixes, each reverted in turn:

```
# doctor back to "install bin directory is missing"
FAILED test_doctor_uses_recorded_python_tool_scripts
FAILED test_doctor_requires_the_platforms_shim_directory_on_path
FAILED test_doctor_healthy_install_needs_only_the_shim_directory
FAILED test_doctor_reports_a_missing_shim_directory
4 failed, 24 passed
# owner default back to "devthrottle"
FAILED test_latest_release_url_defaults_to_the_product_release_repository
FAILED test_latest_release_url_honours_environment_overrides
2 failed, 26 passed
# 404 back to a silent None
FAILED test_latest_release_http_error_raises_with_status_and_url
FAILED test_run_setup_cli_release_lookup_failure_reaches_the_user
2 failed, 26 passed
```

Restored: `28 passed`. Full suite on this Mac, with colour forced on (the way the job runs it) and with colour off:

```
306 passed in 37.64s     # FORCE_COLOR=1
306 passed in 37.62s     # NO_COLOR=1
```

Live, from this branch on the Mac: `_latest_release()` returned `v2.3.0` with `devthrottle-setup-cli-linux-x64`, `devthrottle-setup-cli-mac-arm64` and `devthrottle-setup-cli-win-x64.exe`; with `DEVTHROTTLE_GITHUB_OWNER=devthrottle` it raised `GET https://api.github.com/repos/devthrottle/devthrottle/releases/latest failed with HTTP 404 Not Found`.

`tools/test_shipped_tools_contract.py`: `36 passed`.

**Run for real against the live fleet from the Mac**, from this branch's source rather than the installed tool:

```
$ python -c "from src.cli import app; app()" session list
+------------------------------------------------------------------------------+
| NO. | ID        | NAME      | MACHINE   | REPOSI... | MODEL     | STATUS     |
|-----+-----------+-----------+-----------+-----------+-----------+------------|
| 136 | ba8b2ead  | The Dev   | SORENL... | cc-con... | [model].. | Waiting... |
|     |           | Manager   |           |           |           |            |
| 107 | 52d1fc34  | Wingman   | SOREN_... | devthr... | no model  | Waiting... |
|     |           | Inspector |           |           | yet       |            |
|     |           | - Cockpit |           |           |           |            |
|     |           | tab       |           |           |           |            |
```

(The model column is redacted to a placeholder of the same width; every other cell is verbatim.)

That output is the problem #2922 is about, captured before any of it is changed: the name, the machine, the repository and the state are all cut, and no row can be read back whole. The next pull request in the series is the shared output helper that replaces it.

**Linux** is proven by the `ubuntu-latest` runner in this job and by nothing else - there is no Linux machine in the fleet to run the tool against. Stating that plainly rather than implying a live Linux run.

## What is not proven here

- Nothing about output shape, `--json`, filters or recoverability. Those are the later pull requests.
- The macOS run above used a live Gateway connection; it does not exercise the offline path.
- The release lookup was run live against the real repository (above), but no release asset was downloaded and run on macOS or Linux; the download itself is tested with a stub.
- On Linux the setup tool does not add `~/.local/bin` to `PATH` itself (only Windows and macOS have that step); the doctor now reports it when a distribution does not already have it there.


thefrederiksen:
## Inspection

Inspected PR head e23c4d8671296b6e8f9326e566168da02b76122e.

1. **High — tools/cc-devthrottle/tests/test_setup_ops.py:56-59 (and tools/cc-devthrottle/src/setup_ops.py:235-237).** The revised fixture puts the install root's bin directory on PATH on macOS and Linux. A real macOS install places the command links in ~/.local/bin, and tools/cc-director-setup-cli/Commands.cs:524-529 adds only ~/.local/bin to the shell PATH. The doctor still requires the install root's bin directory on PATH, so a healthy macOS install reports "install bin directory is not on PATH" and needsRepair=true. Adding that directory to the fixture makes the test green while hiding this product defect. The fixture should model the installer-created PATH, and the doctor or installer should agree with it.

2. **Medium — tools/cc-devthrottle/tests/test_setup_ops.py:108-116 (and tools/cc-devthrottle/src/setup_ops.py:25-28, 291-334).** The remaining setup asset test asserts selection of a Windows .exe on every matrix platform. The selector itself only lists Windows assets; if the local setup CLI is absent on macOS or Linux, the fallback downloads that .exe and tries to execute it there. All three test jobs can pass without exercising a usable Unix fallback. This predates the matrix change, but it is a Windows-only assumption still in this suite.

Matrix check: .github/workflows/ci.yml:169-198 uses checkout, Python 3.11, dependency installation, then unconditional pytest steps in each matrix job. The job logs for this head show 36 shipped-tool tests and 287 cc-devthrottle tests passed on each of Windows, macOS and Ubuntu, with no skipped tests. Pytest's no-tests-collected exit would fail these steps. I found no shell or path step that silently bypasses the suite.

thefrederiksen:
﻿## Inspection - re-check

Inspected head `1e848b596f28620df7cbf656ef80928370d0c60d`.

1. **High — `tools/cc-devthrottle/src/setup_ops.py:262-263` (fixture: `tools/cc-devthrottle/tests/test_setup_ops.py:30-31`).** The new shim PATH rule fixes the original PATH mismatch, but `doctor_data()` still treats a missing install-root `bin` directory as a repair problem on every operating system. On macOS and Linux, `PythonToolsInstaller.WriteUnixShims` creates links in `~/.local/bin` (`tools/cc-director-setup-engine/PythonToolsInstaller.cs:596-606`); it does not create the install-root `bin`. The Avalonia wizard also leaves Python tool provisioning to first launch (`tools/cc-director-setup-avalonia/Services/EngineInstallRunner.cs:132-139`). A healthy Unix tool install can therefore have working commands and no install-root `bin`, yet doctor returns `needsRepair: true` with “install bin directory is missing.” The test helper creates that directory unconditionally before its Unix branch, hiding this case. The doctor should require the directory that actually holds the shims on Unix, and the fixture should exercise a healthy install without the unused directory.

2. **High — `tools/cc-devthrottle/src/setup_ops.py:23,73-88,355-357`.** The new per-platform asset selector is correct, but the download fallback cannot reach it with the shipped defaults. `REPO_OWNER` defaults to `devthrottle`, so `_latest_release()` asks for `devthrottle/devthrottle`, receives HTTP 404, and silently returns `None`; `_download_setup_cli()` then exits before asset selection. The actual release is `thefrederiksen/devthrottle` (also the default in `tools/cc-director-setup-engine/GitHubRepositoryDefaults.cs:12-16`): its latest release has all three setup CLI assets. I checked the two GitHub release endpoints: the former returned 404 and the latter returned v2.3.0 with the Windows, macOS and Linux assets. The tests stub `_latest_release()`, so they remain green while `setup install/update/repair/autostart` cannot download a missing local setup CLI on any supported platform unless an environment override is set. Use the product release repository as the default and test the URL construction.

The three primary asset names match `.github/workflows/release.yml:891,894,904` exactly. The new tests would fail if the old install-bin PATH rule or Windows-only asset selector returned; they do not exercise the Unix install-root directory state or the default release endpoint. I found no other product reader of `binDirOnPath` or the old problem text. On Linux, the CLI and Avalonia wizard do not add `~/.local/bin` to PATH; the doctor reports its absence, while the fixture assumes it is already present. The Python tool-contract checks passed on Windows, macOS and Ubuntu for this head; the .NET check was pending when I inspected it.

PR #2934, opened 2026-09-16
Shared AXI output helper for the command-line tools

thefrederiksen:
Step 2 of #2922: one shared place that renders command-line output to the AXI standard (`docs/axi-standard.md`). No command uses it yet - step 3 (`session list`) is the first. No `--json` output changes.

## What it adds

`tools/cc_shared/axi_output.py`, pure functions with no third-party dependencies:

- `render_list(name, fields, records)` - the `name[N]{f1,f2}:` header and one row per record, indented two spaces. A value is quoted when it contains a comma, a double quote, a control character (newline included), leading or trailing whitespace, or any non-ASCII character, or when it is empty. Inside quotes: `\\`, `\"`, `\n`, `\r`, `\t`, `\uXXXX`, `\UXXXXXXXX`. None is written as nothing (`a,,c`); the empty string is written `""`, so the two stay distinct. A backslash on its own does not force quoting (escapes are only read inside quotes), so a Windows path stays readable. A record missing a requested field raises rather than rendering as empty.
- ASCII safety: non-ASCII is escaped reversibly, never dropped or turned into `?`. `write_blocks` refuses any non-ASCII text that did not come through the renderers.
- `format_count(shown, total=None, breakdown=None)` - `count: N`, `count: N (needs-you 3, working 4)` in the caller's order, `count: M of N total`. A breakdown that does not add up to the count raises. Empty results print `count: 0` / `count: 0 of N total`; an empty list still prints its header.
- `format_help(commands)` - `help[N]:` and one indented command per line, exactly as the caller wrote them (placeholders included). Multi-line or non-ASCII commands raise.
- `parse_fields(requested, valid, default)` raises `FieldsError` for an unknown, empty or repeated name, with a message listing every valid name. `parse_fields_or_exit(...)` writes that message to standard error and exits 2; it works from inside a typer command (tested). Nothing is dropped or corrected.
- `parse_list(text, name=None)` - reads a list back out of a whole command's output, exactly. Strict: the row count must match the header, rows must be indented exactly two spaces, and malformed quoting or escapes raise. This is what step 3's recoverability test will use.

Numbers and booleans are written as text and come back as text; the list format does not carry types (that is what `--json` is for). This is stated in the module docstring.

## Continuous integration

The `tool-contracts` job in `.github/workflows/ci.yml` gets one step that runs `tools/cc_shared/tests/test_axi_output.py` - only this file, because the other `cc_shared` tests need dependencies that job does not install. Rebased onto `origin/main` after #2932 merged, so this step now runs inside the three-system job, on Windows, macOS and Linux.

## Proof

Run locally on the Mac (Python 3.14, pytest, typer):

```
$ python -m pytest tools/cc_shared/tests/test_axi_output.py -q
......................................................................   [100%]
79 passed in 0.04s
```

The tests cover quoting of commas, quotes, newlines and surrounding whitespace; non-ASCII (including characters above U+FFFF, a lone surrogate, and literal `\u00e9` text) rendered ASCII-only and parsed back exactly; 500 randomly generated record sets round-tripping exactly; twelve kinds of malformed input that the parser must refuse; the count line in every form; the empty state; `help[]`; and `--fields` exiting 2 with every valid name listed, both directly and from inside a typer command.

**I proved the quoting test can fail.** I disabled the comma and double-quote check in the quoting rule, and 5 tests went red (`test_format_value_Comma_Quoted`, `test_format_value_DoubleQuote_QuotedAndEscaped`, `test_format_value_BackslashInQuotedValue_Escaped`, `test_render_list_TrickyValues_RoundTripExactly`, `test_render_list_RandomRecords_RoundTripExactly`) while the other 65 stayed green. Then I restored it and all 70 passed again.

## Inspection fixes

The inspection found three defects. All three are fixed, each with a regression test, and each test was run with its fix reverted to prove it fails.

- **A name ending in a newline passed validation.** The name pattern used `$`, which also matches just before a trailing newline. List names, field names and count labels are now checked with `fullmatch`, and so is the header pattern `parse_list` uses. New tests: `test_render_list_BadListName_Raises`, `test_format_count_BadLabel_Raises`, and an `"id\n"` case in `test_render_list_BadFieldName_Raises`. With the fix reverted, exactly those three newline cases failed (3 failed, 76 passed).
- **A header with a repeated field name lost a value.** `parse_list` now raises `ListParseError` for `sessions[1]{id,id}:`. New case in `test_parse_list_Malformed_Raises`; with the fix reverted, 1 failed, 78 passed.
- **An escape beyond U+10FFFF raised `ValueError`.** `\U00110000` now raises `ListParseError`. New case in `test_parse_list_Malformed_Raises`; with the fix reverted, 1 failed, 78 passed.

Not covered by a test: the header pattern's switch to `fullmatch`. `parse_list` splits the text into lines before matching, so a line can never end in a newline, and the old `$` could not be tricked through it. The change is there so every pattern in the module is checked the same way.

The branch is still one commit, now on top of the current `origin/main`.

## Not proven here

- On Windows and Linux the tests have run only through continuous integration, not on a real machine.
- No command uses the helper yet, so there is no live fleet output to paste; that comes with step 3.

Ready for inspection.


thefrederiksen:
## Inspection

- **P2 - A final newline passes name validation and breaks the output shape.** `tools/cc_shared/axi_output.py:66` uses `$`, which matches just before a final newline. `render_list("sessions\n", ["id"], [{"id": "x"}])` succeeds but emits a split header that `parse_list` cannot find. A field named `"id\n"` does the same; `format_count(1, breakdown=[("ready\n", 1)])` splits the count line. Require a full-string match.

- **P2 - Duplicate fields are silently lost while parsing.** `tools/cc_shared/axi_output.py:253-267`: `parse_list("sessions[1]{id,id}:\n  first,second")` returns `(["id", "id"], [{"id": "second"}])`. The first value disappears. `render_list` rejects duplicate fields; the parser should reject this malformed header too, since it is used to check recoverability.

- **P3 - An out-of-range Unicode escape raises the wrong exception.** `tools/cc_shared/axi_output.py:201`: `parse_list("sessions[1]{id}:\n  \"\\U00110000\"")` raises `ValueError` from `chr()` instead of `ListParseError`. The parser promises to report malformed escapes as `ListParseError`.

I checked whitespace, lone backslashes before quotes, leading quotes, U+2028, lone surrogates, a 100,000-character value, and a header-shaped value; each valid value returned exactly through one ASCII row. The 70 helper tests ran on all three CI systems. The package configuration includes `cc_shared`, so the new module is included when that distribution is built. Filtered command behavior begins in the next pull request.

thefrederiksen:
## Inspection - re-check

At head bf4b9bdb48cd6daf90e6c5047f9033be3214a9ad, the three earlier defects are fixed. A final newline in a list name, field name, or count label now raises ValueError; a repeated header field and an escape beyond U+10FFFF now raise ListParseError. I checked each regression test against an in-memory revert: the three newline cases produced split output, the repeated-field case overwrote "first" with "second", and the out-of-range escape leaked ValueError. Those tests therefore fail with their respective fixes removed. The Python tool-contract checks passed on Windows, macOS, and Ubuntu. I also round-tripped 6,270 string/None cases, including controls, Unicode line separators, and surrogates; each rendered as ASCII and parsed back exactly.

- **P3 — A non-ASCII count digit passes the strict parser.** tools/cc_shared/axi_output.py:69,257. parse_list("sessions[\u0661]{id}:\n  x") returns (["id"], [{"id": "x"}]): \d accepts the Arabic-Indic digit and int converts it. This allows a non-ASCII header that render_list would never write, so a parse-back check alone could certify output that breaks the ASCII contract. The malformed-input tests cover non-ASCII rows but not header counts. The same count conversion leaks ValueError, rather than ListParseError, for "sessions[" + "9" * 4301 + "]{id}:" on Python 3.11.

- **P3 — An empty supplied breakdown bypasses count reconciliation.** tools/cc_shared/axi_output.py:311. format_count(5, breakdown=[]) returns "count: 5" even though the supplied breakdown sums to zero. The function promises to raise when a breakdown does not add up to shown; None already represents no breakdown. A caller can omit an aggregate accidentally and receive apparently valid output.

PR #2944, opened 2026-09-16
AXI step 3: cc-devthrottle session list - plain state, filters, full ids (#2922)

thefrederiksen:
Step 3 of the AXI mission (#2922): `cc-devthrottle session list` in the AXI shape.

**Stacked on pull request #2934** (the shared output helper). This branch carries that commit too. After #2934 squash-merges, this branch is rebased onto origin/main so that commit drops out.

## What changed

- Default output through `cc_shared/axi_output.py`: fields `id,name,state,repo`. Full ids and full names, never cut. ASCII only.
- `--fields` validated by `parse_fields_or_exit`. Valid: id, name, state, repo, machine, number, model, agent, mission, path. `--fields` together with `--json` exits 2 rather than being ignored.
- `--state` (one state or several, comma separated), `--repo` (folder name or full path, ignoring case and slash direction) and `--machine` (ignoring case). All three also narrow `--json`: the same bare array, only fewer rows. Unfiltered `--json` is byte-for-byte unchanged and does not depend on the state fold at all.
- One plain state, folded exactly per the Architect's ruling 1: `crashed == true` -> crashed; `needsYou` -> needs-you; `onHold` -> snoozed; `active` + `Working` -> working; `active` otherwise -> ready; a missing or unknown bucket exits 1 naming the value and the session. `lastStatusReason` is never read.
- Counts by state (zeros left out), `count: 0 of N total` when a filter matches nothing, `help[]` next steps with placeholders only.
- An unknown `--state` value or field exits 2 listing the valid values. An unknown flag exits 2 (Typer).
- The roster caveats from #1051 are kept, printed after the rows; a caution sentence from the Gateway that is not ASCII is escaped.

The `MODEL` and `(crashed)` table tests were rewritten against the new output: the model is now a `--fields` field read back exactly, and a crashed row reads `crashed`.

## Tests

`tools/cc-devthrottle/tests/test_session_list_axi.py` (43 tests):

- Recoverability: a roster with a comma, quotes, non-ASCII, a very long name, surrounding spaces and no name; every id, name and state is read back exactly with `parse_list`, and the states are pinned independently of the fold.
- **The same check fails against the old Rich table** (a frozen copy of the renderer from main, at the 80 columns an agent reads it at): `test_recoverability_check_OldRichTable_Fails`. The table has no list to read, full ids are cut to eight characters and the long name is elided. A second negative control, a correctly shaped list that shortens names, fails too.
- `--json` unfiltered byte-for-byte equal to the roster (including a row with an unknown bucket); filtered is the same bare array narrowed; a filter matching nothing gives `[]`; a `--state` filter over an unknown bucket exits 1.
- `count: 0` on an empty roster, `count: 0 of 6 total` on a filter that matches nothing.
- Usage errors exit 2 with the valid values: unknown state, unknown field, `--fields` with `--json`, unknown flag, empty `--repo`.
- The fold for every bucket, crashed over every bucket, a string "true" not being a crash, `lastStatusReason` ignored, and unknown, wrong-case, empty and missing buckets raising with the value named.

Local run on macOS: `python -m pytest tools/cc-devthrottle/tests tools/test_shipped_tools_contract.py tools/cc_shared/tests/test_axi_output.py` -> 456 passed.

## Continuous integration

**Green on Windows, macOS and Linux.** Run [35127534705](https://github.com/thefrederiksen/devthrottle/actions/runs/35127534705), job "Tool contracts (Python)" on `windows-latest`, `macos-latest` and `ubuntu-latest`: each ran the `cc-devthrottle` suite with colour forced and reported **350 passed** (the 307 existing plus the 43 new), plus the shipped-tool contract guard (36 passed) and the output helper tests (70 passed). Read from the job logs, not only the check marks. "Build & Test (web)" passed. "Build & Test (.NET)" was still running when this was written; this pull request changes no .NET code.

## Live run against the fleet, from the Mac, with this branch's code

The command is this worktree's code installed editable into a scratch environment (`pip install -e tools/cc_shared -e tools/cc-devthrottle`), talking to the hosted Gateway with this session's own key. The runs are a few seconds apart on a live fleet, so a count can move between two of them.

```
$ cc-devthrottle session list
count: 35 (needs-you 16, working 8, ready 6, snoozed 5)
sessions[35]{id,name,state,repo}:
  (live data rows withheld - this repository is public)
help[4]:
  cc-devthrottle session list --state needs-you
  cc-devthrottle session list --fields id,name,state,repo,machine,number,model,agent,mission,path
  cc-devthrottle session list --json
  cc-devthrottle session whoami
(exit 0)

$ cc-devthrottle session list --state working
count: 8 of 35 total (working 8)
sessions[8]{id,name,state,repo}:
  (live data rows withheld - this repository is public)
help[3]:
  cc-devthrottle session list --fields id,name,state,repo,machine,number,model,agent,mission,path
  cc-devthrottle session list --json
  cc-devthrottle session whoami
(exit 0)

$ cc-devthrottle session list --machine devthrottle-mac-mini --fields id,name,state,machine,number,mission
count: 11 of 35 total (needs-you 4, working 3, ready 2, snoozed 2)
sessions[11]{id,name,state,machine,number,mission}:
  (live data rows withheld - this repository is public)
help[2]:
  cc-devthrottle session list --json
  cc-devthrottle session whoami
(exit 0)

$ cc-devthrottle session list --machine NO_SUCH_MACHINE
count: 0 of 35 total
sessions[0]{id,name,state,repo}:
No session matches the filter.
help[2]:
  cc-devthrottle session list
  cc-devthrottle session list --help
(exit 0)

$ cc-devthrottle session list --state waiting
Error: unknown --state value 'waiting'. Valid states: needs-you, working, ready, snoozed, crashed
(exit 2)

$ cc-devthrottle session list --fields id,colour
Error: unknown field for --fields: colour. Valid fields: id, name, state, repo, machine, number, model, agent, mission, path
(exit 2)

$ cc-devthrottle session list --state working --json | python3 -c 'import json,sys; print(len(json.load(sys.stdin)))'
8

$ cc-devthrottle session list --json | python3 -c 'import json,sys; print(len(json.load(sys.stdin)))'
35
```

## Inspection fixes

All six findings from the inspection (comment 5702701451), fixed in 9a10b112:

1. **Empty results.** With no rows the breakdown is now `None`, so `count: 0` and `count: 0 of N total` print instead of raising. This is what turned the three Python jobs red.
2. **Missing ids and empty names.** A roster row with no session id (absent, empty, whitespace or null) exits 1 with `Error: the Gateway returned a session with no session id (row N, ...)` and prints no list. An empty-string name renders as `""` and reads back as the empty string. The recoverability fixture now carries an empty-string name next to the absent one, and four no-id rows are tested.
3. **Absent roster.** `gateway.get_fleet` (in `tools/cc_shared/gateway.py`) now raises `GatewayError` when the answer is an envelope with no list of sessions, or is not a list or an envelope at all. A bare array from an older Director is still accepted. Every other caller already catches `GatewayError`: `cc-status` and `cc-history` print it and exit 1, the spawn mission lookup warns and continues, and the other `session_ops` verbs go through `_get_fleet`, which exits 1.
4. **ASCII warnings under `--json`.** Both standard-error warnings are escaped the same way as the plain output.
5. **Stronger `--json` filter tests.** Added: `--repo` alone excludes a working row from another repository; `--state working --repo devthrottle` returns only the row in that repository; `--machine` matches a real machine and returns only its row; `--machine soren_north` excludes the other machine.
6. **`docs/cli-reference.md`.** The session list section now covers the default fields, `--fields` and its valid names, `--state`, `--repo` and `--machine`, filters applying to `--json`, full ids, empty answers, and has a sample of the output.

**Proof that each test can fail.** Each fix was reverted in turn, the tests were run, and the file was restored:

```
== 1 empty breakdown: 3 failed (EmptyRoster, FilterMatchesNothing, NonAsciiCaution)
== 2a no id check: 4 failed (RowWithNoSessionId x4)
== 2b empty name to None: 1 failed (DefaultOutput_EveryIdNameAndStateReadBackExactly)
== 3 absent sessions as empty: 3 failed (EnvelopeWithNoSessionsField, get_fleet x2)
== 4 json warnings raw: 1 failed (JsonNonAsciiCautions_StderrIsAscii)
== 5a --json ignores --repo: 2 failed (JsonRepoFilterAlone, JsonStateAndRepo)
== 5b --json ignores --machine: 3 failed (JsonMachineFilter x2, JsonFilterMatchesNothing)
== 5c --json --machine always []: 2 failed (JsonMachineFilter x2)
```

**Full suites, on this Mac:**

```
tools/cc-devthrottle, FORCE_COLOR=1:  360 passed
tools/cc-devthrottle, NO_COLOR=1:     360 passed
tools/cc_shared/tests:                219 passed
tools/test_shipped_tools_contract.py:  40 passed
```

**Live against the fleet, from this Mac, running this branch:**

```
$ cc-devthrottle session list
count: 28 (needs-you 5, working 12, ready 9, snoozed 2)
sessions[28]{id,name,state,repo}:
  (live data rows withheld - this repository is public)
help[4]:
  cc-devthrottle session list --state needs-you
  cc-devthrottle session list --fields id,name,state,repo,machine,number,model,agent,mission,path
  cc-devthrottle session list --json
  cc-devthrottle session whoami
exit 0

$ cc-devthrottle session list --state crashed --machine NO_SUCH_MACHINE
count: 0 of 28 total
sessions[0]{id,name,state,repo}:
No session matches the filter.
help[2]:
  cc-devthrottle session list
  cc-devthrottle session list --help
exit 0

$ cc-devthrottle session list --state working --machine devthrottle-mac-mini --json | python -c "import json,sys; print(len(json.load(sys.stdin)))"
5
```

Ready for inspection.


thefrederiksen:
﻿## Inspection

1. **High — empty results crash before `count: 0` is written.** `tools/cc-devthrottle/src/session_ops.py:384-386`. With zero rows, `breakdown` is `[]`, but `axi_output.format_count` rejects an empty breakdown. Reproduce with `python -m pytest -q tools/cc-devthrottle/tests/test_session_list_axi.py::test_list_sessions_EmptyRoster_PrintsCountZero tools/cc-devthrottle/tests/test_session_list_axi.py::test_list_sessions_FilterMatchesNothing_PrintsCountZeroOfTotal`: both raise `ValueError: breakdown was supplied but is empty`. The current head's tool-contract jobs fail on Windows, macOS, and Linux; the Linux log reports seven failures from this path. An ordinary filter with no match cannot return the promised empty state.

2. **Medium — missing ids are accepted and empty names lose their value.** `tools/cc-devthrottle/src/session_ops.py:277-278`. With `get_fleet` returning `[{'name':'orphan','triageBucket':'active','activityState':'Working'}]`, `list_sessions(False)` exits 0 and emits `,orphan,working,` with a blank id. A row with `{'sessionId':'id-1','name':''}` emits a blank name that `parse_list` reads as `None`, so the exact name cannot be recovered. The recoverability fixture covers `None` but not an empty string or a missing id.

3. **Medium — an envelope missing its session list becomes a successful empty answer.** `tools/cc_shared/gateway.py:375-383`, reached from `tools/cc-devthrottle/src/session_ops.py:351`. Set `gateway.get_json = lambda _: {'rosterComplete': True}` and call `list_sessions(True)`: it exits 0 and prints `[]`. The response supplied no `sessions` field, so the command cannot establish that the roster is empty. The existing `or []` fallback leaves this path unguarded.

4. **Medium — JSON warnings can contain non-ASCII text.** `tools/cc-devthrottle/src/session_ops.py:375-380`. Set `gateway.get_fleet = lambda: ([], False, 'SØREN', None)` and call `list_sessions(True)`: stderr contains `WARNING: ... SØREN`. The plain-output path escapes cautions at line 410; this JSON path writes both Gateway cautions verbatim. The bare-array stdout remains intact, but the command's output is not ASCII-only.

5. **Medium — the JSON filter tests can pass after a filter regresses.** `tools/cc-devthrottle/tests/test_session_list_axi.py:281-295`. In `test_list_sessions_JsonFiltered_SameBareArrayNarrowed`, the selected states already identify exactly the two asserted rows; bypassing `--repo` only for `--json` leaves that assertion green. The sole JSON `--machine` test expects `[]`, so a JSON-only path that always returns `[]` for a machine filter also passes it. Add positive JSON cases with same-state rows in different repositories and a matching machine.

6. **Low — the CLI reference describes the removed table.** `docs/cli-reference.md:684-694`. It lists only `--json` and says the output has short ids, table columns, and a `(you)` marker. The new default emits full ids with `id,name,state,repo` and the added filters. Following that reference now gives the wrong output contract.

I also checked the state fold, unfiltered JSON serialization, the old-table recoverability negative control, and repository callers. The implementation uses the specified state precedence and the non-JSON filters do narrow rows. I found no script that parses the former text table; callers I found either invoke the command for display or consume its bare JSON array.

thefrederiksen:
﻿## Inspection - re-check

1. **Low — the missing-ID regression tests are red on a supported Windows dependency set.** `tools/cc-devthrottle/tests/test_session_list_axi.py:155` (the runner is constructed at line 36). On Windows with Python 3.11, Typer 0.20.0 and Click 8.1.8, `cd tools/cc-devthrottle; python -m pytest -q` reports **9 failed, 351 passed**. All four new `RowWithNoSessionId` cases produce the intended exit code 1 and error text, but `result.stderr` raises `ValueError: stderr not separately captured`: this `CliRunner()` merges standard error into its output. Five older CLI assertions fail for the same reason. These four tests are red with the fix present, so they cannot serve as a passing regression guard on this allowed install. A direct invocation confirmed the command itself exits 1 and reports the missing ID. The continuous integration jobs passed on Windows, macOS and Linux after installing the latest test dependencies; they do not cover this dependency set.

The six earlier findings otherwise appear fixed. Empty results now produce `count: 0`; an empty-string name survives the list round trip; a missing session list raises `GatewayError`; both JSON cautions are ASCII escaped; the JSON repository and machine tests have positive matches that would fail if either filter were ignored; and the CLI reference describes the new output. I checked the assertions against the removed code paths without changing the checkout. The documentation change has no automated regression assertion, so reverting that hunk alone would leave the suite green.

I traced every Python `get_fleet` caller, including the self-test and spawn mission lookup: each handles `GatewayError` through its local error path. The Gateway endpoint returns a lowercase `sessions` array in its envelope; a read of the live endpoint returned that shape, and the older bare-array shape remains accepted. The eight focused `get_fleet` tests passed. I found no additional production defect involving truncation, non-ASCII output, guessed state, ignored filters or exit codes.

PR #2947, opened 2026-09-16
Output helper: ASCII-only header counts and no empty breakdown

thefrederiksen:
Part of #2922

Fixes the two low-severity defects from the re-check inspection on pull request #2934.

## Fixes

1. **Header count is ASCII digits only.** `parse_list` matched the header count with `\d`, which also accepts non-ASCII digits such as U+0661, and `int()` converts them. So `sessions[١]{id}:` parsed, even though `render_list` never writes that. The pattern is now `[0-9]+`. Also, a count too long for `int()` to convert (for example, 4,301 nines) now raises `ListParseError` instead of leaking `ValueError`.
2. **An empty breakdown is refused.** `format_count(5, breakdown=[])` returned `count: 5`, which skipped the check that the breakdown adds up to `shown`. An empty breakdown now raises `ValueError`. `None` still means no breakdown.

## Proof

- New test `test_parse_list_HeaderCountNotAsciiDigitsOrTooLong_RaisesListParseError` (cases `non-ascii-digit` and `too-many-digits`). With fix 1 undone, both cases fail. With the fix back, both pass.
- New test `test_format_count_EmptyBreakdown_Raises`. With fix 2 undone, it fails. With the fix back, it passes.
- `python -m pytest tools/cc_shared/tests/test_axi_output.py -q`: 82 passed (Python 3.14, macOS).

- Continuous integration: Tool contracts (Python) passed on Windows, macOS and Linux.

Ready for inspection.


thefrederiksen:
## Inspection

No findings at head `38dc6bf4166dc4bbebe46225d2b872c6cfdd0a2f`.

I checked both findings from the re-check of #2934 against the changed code and its parent. The new non-ASCII header-count and oversized-count test cases pass at this head and fail against the parent: the old parser accepts the Arabic-Indic count and leaks a plain ValueError for 4,301 digits. The new empty-breakdown test passes at this head and fails against the parent: the old formatter returns `count: 5` for `format_count(5, breakdown=[])`. I ran these test methods against both revisions in memory, with no checkout changes.

I also checked an ordinary `render_list`/`parse_list` round trip and `format_count` with no breakdown and a valid nonempty breakdown on both revisions. The diff changes only `tools/cc_shared/axi_output.py` and its tests. Python tool-contract checks passed on Windows, macOS, and Ubuntu. The .NET check was pending at inspection time.

Issue #2933, opened 2026-09-16
Fleet Manager: one session you talk to that runs the fleet (replaces the Assistant)


thefrederiksen:
## What

A **Fleet Manager**: one place, in the Cockpit, where the user talks about all their work. The user says what they want. The Fleet Manager starts the sessions, gives each one instructions in the user's own words, and watches them. It comes back with only three kinds of news: something is **ready**, something was **found**, or something needs **a decision**. When things are waiting on the user, it takes them through one at a time, and each answer goes straight to that session.

It never does the work itself, so it is always free to talk. It does not read screens or summarise sessions either: the Wingman already does that on every turn, and the Fleet Manager uses the Wingman's verdicts.

It **replaces the Assistant** (Cockpit page, phone page and Settings tab).

## Why

Users running many sessions become the router. They open each session, name it, carry results from one session to the next, and open red sessions that turn out to need nothing. The machinery to own and quiet sessions already exists. What is missing is the one session on top that uses it on the user's behalf.

## Decided

1. The Fleet Manager is a **real session** (not the Assistant's Gateway model), one per account, pinned at the top of the session list, with every session it starts owned by it.
2. **Where it runs is a user setting:** agent and computer, in a Fleet Manager tab in Settings (replacing the Assistant tab). If no Director is running on the chosen computer, the launcher starts one. An unreachable computer is shown as unavailable; it never silently starts somewhere else. Moving it starts a new Fleet Manager that rebuilds its picture from its records.
3. The **Fleet Manager page is the Cockpit's landing page**, replacing `/assistant` (the old address redirects).
4. **Cockpit only** for now. The phone's Assistant page is removed at the same time.
5. It **uses the Wingman on every turn** for what a session needs, the session's own words, the answer options and the spoken text. It reads a screen itself only when the Wingman cannot tell or a session is stuck and needs a person.
6. It may **open a Mission on its own** for big work, and says so in one line.

## Its rules

1. Never do the work: every change, fix, review and investigation goes to a session.
2. Never summarise a session: use the Wingman's verdict.
3. The user is the only authority: no merge without the user's word unless the repository allows merge on green, and never a red merge. Product, scope, money and anything irreversible go to the user.
4. Never close a session with unlanded work.
5. The user's words and the sessions' quoted words are passed on verbatim.
6. Report outcomes, not machinery.
7. Everything visible: every session it starts is named, owned, and logged.
8. Small work stays small: one session, no ceremony.

## What to build

[...]

PR #2951, opened 2026-09-16
The Wingman judges the sessions the account's Fleet Manager owns

thefrederiksen:
Part of #2933 (the Fleet Manager), step 2 of phase 1.

## What changes

The Wingman now judges the sessions that the account's Fleet Manager owns. Before this, the Wingman skipped every session held by a live owning session, which would have left the Fleet Manager reading and summarising its own sessions - the thing it is designed never to do.

- **One Fleet Manager per account, by an explicit mark.** The account stores which session is its Fleet Manager (a per-account setting), set, cleared and shown with the new `cc-devthrottle fleet-manager show|set|clear` command. A session is the Fleet Manager only while it is the marked session and nothing owns it. The workflow a session runs on does not decide it.
- **Only the marked Fleet Manager's sessions are judged while held.** Every other held session - a Worker under an Architect or a Manager - keeps the old rule and is still never read automatically. A test proves that an unmarked, unowned Architect on the same workflow does not make its Workers readable.
- **Still quiet for the owner.** A Fleet-Manager-owned session is never narrated automatically, never counted as needing the owner, and never shown on the badge. A person pressing Explain can still narrate any held session, as before.
- **Not in this change:** delivering the stored verdict to the Fleet Manager session (a later step), and pinning the marked session first in the list (a later step). The code and documentation say so.

## Proof

- Independent inspection by a different agent family, two rounds; round one failed on the identity rule and a test that stayed green with the wrong identifier, round two passed.
- Tests watched failing: substituting the old held rule turns 6 of the 16 new judging tests red, including the unmarked-Architect and direct-child cases; changing the production setting key turns its independently pinned test red; changing the command's route turns 6 of its 10 Python tests red.
- Gateway unit suite on this branch: 4,829 passed, 7 failed. The same 7 fail on the base commit on the machine it ran on and are unrelated. Solution build: 0 errors, 0 warnings.

## Not proven

- No live hosted Gateway was exercised; the proofs are in-process.


thefrederiksen:
The two failing .NET tests (GatewayStatsStoreContainmentTests, the statistics store self-host checks on the Windows runner) fail on main without this change - runs 35107865877 and 35084756981 - and pass 5 of 5 locally on this branch and on main. They are not caused by this change.

PR #2996, opened 2026-09-17
Throwaway: a job fails on purpose (CI result proof)

thefrederiksen:
Throwaway proof for pull request 2986. Will be closed.


thefrederiksen:
Throwaway proof for 2986, done.

PR #2997, opened 2026-09-17
The Fleet Manager's outcome records, preferences and digest live in the Gateway

thefrederiksen:
Part of #2933 (the Fleet Manager), step 3 of phase 1.

## What changes

The Gateway now keeps the Fleet Manager's outcome records, so its news survives a restart, a reset or a move.

- **Outcome records** - Ready, Finding and Decision - filed from the command line, stored per account, and open until answered. An answer is final: a conditional update means exactly one answer wins across Gateway instances, and a second caller gets a clear conflict. Each answer records who gave it.
- **Standing preferences** - what the owner has told the Fleet Manager to do without asking.
- **The digest** - one command that prints what the Fleet Manager needs at start: open records, the sessions it owns, and the preferences. Nothing is silently truncated: the outcomes list pages with an opaque, stable cursor that the commands follow, and totals come from the database.
- **Authority.** With a session key, only the account's marked Fleet Manager may file, read or answer records or use preferences; every other session key is refused with a reason. The owner, through his own signed-in credentials, may read and answer records and manage preferences.
- **A replacement Fleet Manager** sees the sessions owned by earlier ones: the Gateway keeps the 20 most recently marked sessions, and the digest includes an earlier one while it still owns a live session. Actually handing ownership over is a later step.
- **Storage** is additive, with SQLite and PostgreSQL migrations that match column for column, and a guard test that fails if the PostgreSQL migration is missing.
- **Commands:** the `cc-devthrottle fleet` group, documented in `docs/cli-reference.md`.

## Proof

- Independent inspection by a different agent family, three rounds. Round one found six blocking defects (no test through the real host, any session could read the records, silent truncation at 200, a replacement Fleet Manager lost its sessions, answers could race, no migration guard); round two found that records past 200 could not be listed; round three's only blocking finding was colour codes in command test output, already fixed on main.
- Every route is tested through the booted in-process Gateway host and its middleware with real credentials, allowed and refused cases both.
- Tests watched failing before each fix, including following the cursor past the largest page, answering between pages, a cursor this Gateway did not issue, and two store instances answering the same record.
- Gateway unit tests for the Fleet Manager, session-key guard, store, migration and built-in classes: 1,199 passed, 2 failed (two rename-failure tests in other stores that fail the same way on main on the machine that ran them). Command-line tool: 3,177 passed with colour forced on; contract and output-style suites pass.

## Not proven

- The PostgreSQL migration has not been applied to a real PostgreSQL database by this change; the guard test checks it is present, and the two migrations were compared column by column.
- No live hosted Gateway was exercised.


thefrederiksen:
The one failing .NET test on the first run (TenantSessionMapTests.Contend_GetOrAddOnOneKey_HasExactlyOneWinnerPerRound, a timing race in Core that this change does not touch) passed when the failed job was re-run.

PR #3026, opened 2026-09-17
Cockpit: the Now view of the Wingman tab, and the type it renders

thefrederiksen:
The Wingman tab's Now view draws the live stop of one session and nothing else, in all ten states the approved version 3 mockup draws: needs you; needs you and the Wingman is not sure; the Wingman reading; working; just answered; carrying on; done; report; the Wingman failed; and the Wingman switched off.

This is the first slice of item 3. It is the type and the pure component. Mounting it on the tab, with the read that feeds it, is the next slice.

## The view decides nothing

It renders the strings and flags of one `WingmanNow` object, showing each piece exactly when the Gateway sent it. There is no branch anywhere in the file on what a state means - no colour rule, no wording of its own beyond the buttons, three small headings, and the clock and "how long ago" phrasing the Gateway cannot write because it does not know the reader's time zone. Adding a state to the product is a change in the Gateway's fold, not here. That is rule 7 in the product instructions.

Every action - play the narration, turn voice on, answer by option, send a reply, snooze, open the terminal, open Settings, go to the next session that needs you - is handed in by whatever mounts the view. An action with no handler is not drawn.

## What is not proven

The Gateway route `GET /sessions/{sid}/wingman-now` is being built separately and does not exist yet, so nothing here has been checked against a real answer. `wingmanNowRead.ts` is the settled contract written down field for field, not an observed one. That is said plainly at the top of both files.

Ratings, History and Debug are later items and are not touched.

## How it was checked

- Twenty-two tests: one per state, each handing the view an object and asserting the words it draws, plus answering by option, sending a reply, the four voice controls and the timed sentences. All green.
- The tests were watched failing on purpose. Three separate breakages were made - hard-coding the pill words, dropping the carrying-on deadline sentence, and making the calm card depend on the state name - and six tests went red with the right symptoms. Restoring the file put all twenty-two back green.
- The web typecheck is clean across all four workspaces, and the new files lint clean.
- The real component was rendered in every state to a local page and looked at against the mockup. Two defects were found that way and fixed in the second commit: the cards were capped too narrow, and a locale that writes "7:19 a.m." was getting a doubled full stop.


thefrederiksen:
Superseded by 3057, which merged the whole chain.

PR #3027, opened 2026-09-17
Cockpit: the Wingman tab opens on Now, with the version 1 list kept as History

thefrederiksen:
Stacked on pull request 3026, which holds the Now view and the type it renders. Read that one first; this one mounts it.

The Wingman tab's default view is now the live stop of the session and what it needs from you. The version 1 list of every stop the Wingman judged is unchanged and sits beside it as History. Redesigning History, and adding Debug, are a later item and are not touched here.

## The temporary bridge, and when it goes

Now is fed by `GET /sessions/{sid}/wingman-now`, re-read every five seconds while the tab is open. That route is being built separately and does not exist on any Gateway yet, so while it answers 404 the tab shows the version 1 list on its own, with no view buttons - exactly the screen that shipped. Nothing is worse than it is today, and the new screen appears by itself the moment the route lands.

This is scaffolding for one route that is still being built, not a fallback design. It is marked as temporary in the tab and in the test that covers it, and it is deleted when the route ships everywhere: after that, a Gateway that cannot serve Now should say so like any other failure rather than quietly showing an older screen.

A refusal that is NOT a missing route is shown as the Gateway's own sentence, never as the older screen. There is a test for exactly that, because swallowing every refusal into the bridge is the obvious way to get this wrong.

## What Now can do, and what it deliberately does not offer

The actions are wired by the Cockpit, because the shell owns the tabs and the router: send a reply to the session, answer by option with the Gateway's own verdict identifier and option index, open the terminal, go to the next session that needs you, and open Settings.

An action the shell does not wire is not drawn. So Now currently offers no voice control, no snooze and no "Why this colour?" link - the voice screen, the snooze call and the Debug view are not wired to it yet. A button that would do nothing is worse than no button.

## How it was checked

- Seven new tests for which view is shown, including the bridge, the refusal that must not use it, and coming back to Now when the screen moves to another session. The nine existing tests for the version 1 list are unchanged and now name the component directly.
- All 284 client-core tests and all 93 Cockpit tests pass. The web typecheck is clean across all four workspaces and the changed files lint clean.
- The new tests were watched failing on purpose. Making the bridge swallow every refusal turned the refusal test red; making History the view that is restored turned four tests red, including the new one. Restoring the file put all seven back green.
- Two teardown errors appear when the whole Cockpit session suite runs together, from an interval in `waiting.ts` raised by `attentionOrder.test.tsx`. They were confirmed to be there with this change reverted, so they are not from this work.


thefrederiksen:
Superseded by 3057, which merged the whole chain.

PR #3028, opened 2026-09-17
Wingman tab: the Gateway serves one session's live stop, in finished words

thefrederiksen:
The Now view of the Wingman tab asks one question - what is going on in this session, and what do I
need to do about it - and until now nothing on the Gateway answered it. This is the read it will
render, and it is the first of six slices for the Wingman tab's version 3.

## What it adds

`GET /sessions/{sid}/wingman-now` answers one session's live stop as finished strings and flags:
which state it is in, the pill's words, when it stopped, the Wingman's headline and story, the
agent's own decisive sentence and its whole last reply, what the session needs with each option and
its note, and whether an option may still be tapped. The Cockpit turns the instants into local time
and lays the rest out. It decides no words.

Every word lives in one pure fold, `WingmanNowFold`, so each state is proved by handing it inputs
rather than by booting a Gateway. The handler gathers and nothing more: the session's folded roster
row from the same fold the roster route serves - so the pill's colour and the dot beside the same
session in the Sessions list cannot disagree - its stored verdicts, and its stored conversation.

This slice rules on the stopped states the Wingman explained: needs you, needs you but not sure,
done, report and carrying on. Every other row folds to "other" and wears its own label, so the view
is never blank. Reading, failed and switched off, the carrying-on deadline sentence, working, just
answered, and the voice control each arrive in their own slice.

## The refusals

It takes the refusals `GET /sessions/{sid}/wingman-stops` takes, for the same reason: this answer
carries the agent's own words and its whole reply, so it is served to a device signed in with its
own device key and never to a session key, whatever the account's colour switch says. The handler
refuses on its own rather than leaning on the session-key guard's allow list, and the route is
deliberately absent from that list as well - both halves are tested.

## Two things named as gaps rather than approximated

The carrying-on card carries its heading without the deadline sentence, because that instant must
come from the same function the carrying-on clock expires on, and an approximation written here
would be exactly the disagreement it is meant to prevent. It arrives in slice 3, with the clock.

When the verdict in force is not in the history window the one-tap path closes, rather than offering
a tap the answer route would refuse.

Every guard was broken on purpose, watched going red with the right symptom while the rest stayed
green, and restored:

|---|---|
| The session-key refusal removed | the session-key test, 9 controls green |
| The device-identity refusal removed | both callers with no device, 8 controls green |
| The other-account existence check removed | the cross-account 404 test |
| `wingman-now` added to the session-key allow list | the allow-list census case, 255 controls green |
| "Not sure" stops reading an ambiguous answer | the ambiguous-answer wording test |
| An answered verdict left tappable | the fold's rule and the route's answered test |
| The whole reply allowed to carry tool output | the whole-reply test |
| The pill computing its own colour | the pill-colour test and three more |
| The conversation's account scope removed | the new scope test (below) |

That last one is the finding. The conversation is read inside the caller's own account scope, and
none of the route's other tests could see that line: they build the tenant boundary over the
single-account context, whose scope is a no-op, so with the line deleted all eleven stayed green.
A guard nothing can see is a guard nobody is keeping, so one test now builds the boundary the hosted
Gateway is built with, over a real ambient account which refuses a read taken outside a scope, and
it goes red where the whole reply should be.

**The same line, unguarded, is on `GET /sessions/{sid}/history` in
`SessionConversationEndpoint.cs`** - shipped, correct, and with no test that can see it. It is not
touched here.

Proven: `CcDirector.Gateway.UnitTests` - 5,544 passed, 8 skipped, and the 37 tests of this slice all
green.

Not proven: `Gateway.Tests` and the rest of the parked suites have not been run against this branch,
so nothing here says anything about them. The eight PostgreSQL-backed proofs skip without a database
and prove nothing on a run without one.

One failure in that suite is pre-existing and not caused by this branch:
`TurnsVerbUnresolvedTranscriptTests.Turns_SupportedAgentWithNoTranscriptYet_ReportsNoTranscript_NotOk(agent: Grok)`.
It was reproduced on clean `origin/main` in a separate worktree with none of this code. Its premise
is machine-dependent: the test roots a session on the system temporary directory and assumes no
locator can resolve a transcript for it, but this machine has a real Grok session directory for that
exact path, so the locator resolves one and the verb correctly answers "ok". The product is right
and the test's assumption is wrong. Filed separately rather than fixed here.


thefrederiksen:

PR #3030, opened 2026-09-17
Wingman tab: Now says so when there is no explanation to show

thefrederiksen:
Slice 2 of six for the Wingman tab's Now view. **Stacked on pull request 3028** (slice 1), which is where its base branch points; merging that one first retargets this to main automatically.

## What it adds

Three of the ten states Now draws are the ones where the Wingman has nothing to say about the stop: it has not finished reading it, its answer was refused, or it was never switched on for this account. Until now all three folded to "other" and wore the row's own label - true, but nothing a person can act on, and on a switched-off account it read "Needs you", which is a judgement nothing had made.

Each now shows **the session's own last words** instead of an explanation the Wingman does not have, so the owner can answer without waiting for a judgement he may not need. They are shortened on a word boundary and folded onto one line, because they stand where a headline stands.

- **Being read** says what is happening, that it usually takes a few seconds, and offers the reply box anyway ("You can answer now without waiting.").
- **A refused answer** says the Wingman could not explain this stop, then its own reason verbatim, then what the refusal leaves the row - "the row stays red because the session stopped" - and offers the last stop it DID explain, from the stored history.
- **Switched off** says so in one sentence, names the way to switch it on, and is the one state that hides the "why this colour?" link: there is no verdict behind the colour, so the link would open an explanation of a rule that did not run. Either switch does it, per the Architect's ruling - and that is also what the screen can tell apart, because with the colour switch off no verdict reaches a screen at all.

The words are the ones in the mockup the owner approved, taken from it rather than rewritten.

## Three places it deliberately says less

- A refusal that recorded **no reason** gets no story at all, rather than a sentence about a colour that explains nothing on its own.
- A session whose **first** stop is the unreadable one is told there is no earlier explanation, rather than handed back the record that was just refused.
- A Gateway that was handed **no settings resolver** claims nothing about the account's switches either way. An assumed "on" would put "the Wingman could not explain this stop" on a session nothing was ever going to explain.

## Working still outranks all of it

A working session on a switched-off account is not "Stopped". The product's law is that a working session is blue, always, and nothing goes above that check - so it keeps the row's own state until slice 4 draws working properly. There is a test whose only job is that.

The account's switches reach the fold through the **same per-tenant resolver the row source reads them through**, so what Now says about them and what the roster does about them cannot be two answers.

Broken on purpose, watched red with the right symptom while the rest stayed green, restored, and the assembly rebuilt each time:

|---|---|
| The reading arm deleted | the reading tests and the last-words tests, 59 controls green |
| The refused arm deleted | the eight failure tests, 56 controls green |
| The switched-off arm deleted | three fold cases, the hidden-link test, and all three route cases |
| Switched off allowed to outrank working | the working test alone |
| A refusal with no reason given the colour sentence anyway | the no-reason test alone |
| The record in force offered back as the last good one | that test alone |
| A refused record allowed to be the last good one | **nothing - see below** |
| The excerpt cut at the character count, not the word | the word-boundary test alone |
| The "why this colour?" link shown in every state | the hidden-link test and all three route cases |
| The route never asking for the account's switches | all three route cases |

**The finding, and it is a real one.** With the rule that skips REFUSED records removed from the last-good-explanation read, all sixty-four tests stayed green. The only refused record in any of them carried no verdict word, so a different rule skipped it for a different reason and the rule under test was never exercised. That is not hypothetical: a refused record **can** carry a decision, because the invented-menu correction that landed in #3024 salvages one out of a refusal. An older refusal could therefore have been served back under the words "last good explanation" - a judgement the Gateway threw away, offered as the reassuring one. A test with exactly that shape was added, and it goes red without the rule.

One earlier test changed rather than being added to: it pinned a refused verdict folding to "other", which was slice 1's interim answer. Its real subject - that nothing from a REJECTED record is ever served as though it were live - is unchanged, and is now pinned harder, against the refusal's own words beside it.

Proven: `CcDirector.Gateway.UnitTests` - 5,572 passed, 8 skipped. 65 of those are this route's, all green.

Not proven: `Gateway.Tests` and the other parked suites have not been run against this branch. The eight PostgreSQL-backed proofs skip without a database and prove nothing on a run without one.

The one failure in that suite is pre-existing and filed as issue 3029 - `TurnsVerbUnresolvedTranscriptTests...(agent: Grok)`, reproduced on clean `origin/main` with none of this code.


thefrederiksen:

PR #3031, opened 2026-09-17
Wingman tab: a session carrying on says when it will turn red if it does not

thefrederiksen:
Slice 3 of six for the Wingman tab's Now view. **Stacked on pull request 3030** (slice 2), which is stacked on 3028 (slice 1); GitHub retargets each one as the one below it merges.

## What it adds

A session that says it will keep going by itself turns the row purple on its own word, and the carrying-on clock holds it to that word: when it has not worked again by a deadline, the row goes red and says so. Until now Now showed the calm card with no sentence under it - the owner was told nothing is needed and left to guess what happens if nothing then happens.

The card now names the moment: *"If it has not worked again by 12:05 PM, and none of the sessions it owns is still working, this turns red and says so."*

## The moment comes from the function the clock expires on

Nothing here works out ten minutes from anything. The sentence makes the owner a promise about a specific time, and a second piece of arithmetic in a second file would be free to drift from the first - the drift he would notice being the row going red at a time the card told him it would not.

So the card asks `TurnVerdictWatchdog.DeadlineFor`, and the test hands the instant the card printed **straight back to the watchdog**: it must not be expired one tick earlier, and must be expired at it. A test that recomputed the deadline itself would pass just as well with the card and the test wrong together.

## A session with one of its own still running names no time

No clock is counting then, so there is no moment to name, and the card says so in words instead. That is the clock's own rule, and **alive** is the test rather than **working**: a Worker eight minutes into one long silent command is running, whatever the terminal's silence suggests. Once they have all stopped the clock runs from the **last** of them, and the card names that moment rather than the judging moment.

**A client never has two sentences to choose between.** When a deadline is running the sentence is the deadline - words, instant, words - and the card's body is empty. When no clock is counting the body carries the whole sentence and there is no deadline object. Exactly one is ever present, and a test says so.

The route reads the session's own sessions the way the clock's own seat reads them - the same call over the same fresh snapshot - so the card and the clock cannot be looking at two different rosters.

|---|---|
| The deadline worked out locally as "judged plus ten minutes" instead of asked of the clock | the announced-wake-up test, the last-owned-stop test, both live-session cases and both route cases |
| The deadline taken with the owned sessions ignored | both live-session cases, the last-owned-stop test, both route cases |
| The card silent instead of saying why no clock is counting | both live-session cases and both route cases |
| The deadline written into every calm state | the done and report card tests, and two more |
| The route never gathering the owned sessions | both route cases, and nothing else - which is exactly the wiring they exist for |

The plausible duplicate was the first one, not a silly mutation: "judged plus ten minutes" **is** the clock's own answer in the common case, so only the announced-wake-up case and the owned-sessions cases can tell the two apart. Both do.

## One earlier test restated

It pinned the carrying-on card having no sentence, which was the placeholder while the clock was unwired. It now pins the rule that replaced it - the deadline or a body, never both.

## One sentence worth a second look

The undated form - *"It turns red if it stops working and none of the sessions it owns is still working."* - is the design's own wording, kept verbatim rather than rewritten. Its "it" is the session, as in the dated form. If the Architect wants it plainer it is one constant.

Proven: `CcDirector.Gateway.UnitTests` - 5,584 passed, 8 skipped. 77 of those are this route's, all green.

Not proven: `Gateway.Tests` and the other parked suites have not been run against this branch. The eight PostgreSQL-backed proofs skip without a database and prove nothing on a run without one. The one failure in that suite is pre-existing and filed as issue 3029, reproduced on clean `origin/main` with none of this code.


thefrederiksen:

PR #3032, opened 2026-09-17
Wingman tab: a working session says what it is doing and who last asked it

thefrederiksen:
Slice 4 of six for the Wingman tab's Now view. **Stacked on pull request 3031** (slice 3). The chain is 3028 → 3030 → 3031 → this one; GitHub retargets each as the one below it merges.

## What it adds

Most of a session's life is spent working, and until now Now had nothing to say about any of it: a working session fell through to the row's own label and the tab showed one word. There is no stop to explain in that state, so it says **what the session is doing** instead.

- **How long it has been working** measures forward, and the sentence is the elapsed time alone - *"Working for 6 minutes"*, with no clock time in it. The moment it started is the moment its last stop stopped being the live one, which is what the store records when a Working transition invalidates the stop. So the number is the length of **this working stretch**, not the age of a stop that is over. A session that has not stopped since this Gateway learned of it has no such moment: the pill still says Working and no duration is invented.
- **What it was last asked**, from the stored conversation, with who asked when that is known.
- **The stop it has just come from**, as one line.

## Who asked is never guessed

This is the whole of the slice's care, and it has three outcomes.

1. **A fleet message** is recognised by reading back the frame this Gateway itself wrote. `FleetMessaging` gains `TryParseFrame` beside `BuildFramedMessage`: it returns the sender's name and the message with the frame **and the reply hint** removed, so the owner reads what the other session said rather than our delivery wrapper. It is **round-trip tested against the builder**, not against a hand-written copy of the format - a parser pinned to a copy keeps passing the day the frame changes. Two of the three header forms carry no name at all; those are recognised as fleet messages that name nobody, because a short identifier is not a name.
2. **The owner**, but only when the Director's record of his last turn in this session sits **at or after the previous stop** *and* **within half a minute** of the message. Both halves matter.
3. **Nobody.** The card says what was asked and when, and nothing about who.

## One flag, not a second shape

The timed sentence gains `elapsedOnly`. The client is *told* whether it is showing a clock time with an optional "ago" or an elapsed time alone, rather than working that out from the state - which is exactly the branch the dumb-client rule exists to remove.

Ten mutations, each red with the right symptom, controls green, restored and rebuilt:

|---|---|
| Working no longer a state of its own | 13 fold tests and both route tests |
| The stretch measured from the stop, not from going back to work | both working-moment tests and the route test |
| The elapsed-only flag dropped | the working sentence test and the route test |
| The owner named on his stamp alone, with no previous-stop check | the before-the-stop test |
| The half-minute tolerance removed | **nothing at first - see below** |
| The frame left on the text | the fleet-message test |
| A fleet message allowed to fall through to the owner check | **nothing at first - see below** |
| A refused record allowed to be the stop it came from | that test |
| A short identifier handed back as a sender's name | both unnamed-sender cases and the unnamed-fleet-message test |
| The reply hint left on | three round-trip cases and both fleet-message tests |
| Anything at all parsing as a frame | two round-trip cases and four fold tests |

**Two real gaps, both found this way and both closed with a test.**

The **half-minute tolerance** was replaced with "always true" and every test stayed green: in all of them the owner's stamp also failed the previous-stop half of the rule, so the tolerance was never reached. What that leaves unprotected is the ordinary case - he types into a session, walks away, and another session messages it twenty minutes later, which would be shown to him as his own words.

The **fleet branch** was allowed to fall through to the owner check, which also left everything green. Two of the three frame forms carry no sender name, so an unnamed fleet message arriving beside his own turn would have been labelled "You" - another session's words under his name. A message that came through the fleet came through the fleet, and the owner check is not consulted at all.

Proven: `CcDirector.Gateway.UnitTests` - 5,617 passed, 8 skipped. 110 of those belong to this route and the frame parser, all green.

Not proven: `Gateway.Tests` and the other parked suites have not been run against this branch. The eight PostgreSQL-backed proofs skip without a database. The one failure in that suite is pre-existing and filed as issue 3029, reproduced on clean `origin/main` with none of this code.

`TryParseFrame` is new public surface on `FleetMessaging`, which lives in the Contracts assembly; nothing else calls it yet.


thefrederiksen:

PR #3033, opened 2026-09-17
release: v2.6.0 - a message waits in an inbox instead of interrupting the session it is sent to

thefrederiksen:
## What this is

The release notes for v2.6.0 (`docs/public/release-notes/v2.6.0.md`) and the version bump in
`Directory.Build.props`. No product code changes.

## What v2.6.0 carries

The whole of the Message Load mission (pull requests 2970 and 3016): a fleet message is a record in a
Gateway inbox rather than text typed into a running session; the send gate; the doorbell, typed as one
fixed line only when the session is idle and its composer is verified empty; rings, stuck marks and the
sender being told; replies asked for without blocking; the row line every screen renders; the snooze a
doorbell does not end; restore as the Director's own act with the new `cc-devthrottle director restore`;
and the removal of the ask-and-wait command. The notes also state the limits plainly - the doorbell reaches
only the two agents whose screens can be read, the millisecond gap between the last look and the first
keystroke, the read that is marked before the text is sent, and what has not been exercised live.

## Why the notes also cover cc-secrets and Linux

v2.5.0's bump and notes merged to main this morning as `fc323b8b`, but the `v2.5.0` tag was never pushed
and no release was created - the latest published release is still v2.4.0. A user updating from v2.4.0
therefore receives the cc-secrets toolbelt and the Linux app menu and updater as part of v2.6.0, so those
entries are folded into this page rather than left on a page nobody was ever served.
`docs/public/release-notes/v2.5.0.md` is untouched and still stands as its own record.

## Checks

`docs/public/release-notes/` is exempt from the retired-messaging-words tree scan by design (a release
page is dated history), so naming the removed command here does not trip that guard.


thefrederiksen:
Added the Fleet Manager to these notes, plus dev reports, the worktree pool and the Director id rule - all of it merged to main after this pull request was opened, and all of it ships under this same tag. The version bump is unchanged.

PR #3035, opened 2026-09-18
Wingman tab: what you answered, that the session took it, and where to go next

thefrederiksen:
Slice 5 of six for the Wingman tab's Now view. **Stacked on pull request 3032** (slice 4). The chain is 3028 - 3030 - 3031 - 3032 - this one; GitHub retargets each as the one below it merges.

## What it adds

The state that closes the loop. He answers a stop, and the screen says what he sent, that the session took it, and where to go next.

- **What he answered**, in his own words and his own order.
- **That the session took it**: *"The session started working again 2 seconds later."*
- **The next session waiting on him**, in the Sessions list's own order.

It holds for five minutes and then the session is simply working again.

## The decision was not recorded anywhere, and now it is

Two paths answer a stop and only one of them wrote anything down.

Tapping an option marked the verdict answered - **the moment, never which option**. The activity ledger deliberately keeps `chosen=2` and not which two: it is a record that an activation happened, not of what was decided. So a minute later nothing anywhere remembered the decision.

`turn_verdicts` gains **`AnsweredWith`**, written in the same call as the moment, holding the options' own `Key` values in the order he picked them. **The keys, not the bytes**: what goes to the session is `31`, what he DID is *"the seed rows, the schema"*, and only the second is worth anything to a reader. A migration for each provider, generated last, with `has-pending-model-changes` clean on both.

A **typed reply** touches no verdict at all - the terminal, the phone, voice, the reply box - so it is read from the stored conversation instead: the first thing he said after the stop. From his side the two are one act, and the card draws them as one.

Three ways it can be wrong, and each is refused rather than guessed:

- **A moment with no words.** Every stop answered before this column carries exactly that. "You answered:" with nothing after the colon is worse than silence, so the state is not entered.
- **A fleet message after the stop.** Another session answered it; he did not. It stops the search rather than being skipped past, because skipping would find a later message of his own and present it as the answer to a stop somebody else had already dealt with.
- **A return to work before the answer.** The session picked itself up on its own, so nothing is claimed about his answer causing it.

## Where to go next

`nextNeedsYou` is `SessionOrdering.InWaitingOrder` over the folded roster the handler **already read for this session's own row** - one read, one fold, one order. It filters to the needs-you bucket, because the waiting line carries the calm band after its reds and a finished session is not waiting on anybody.

It appears only on the answered state. On a session that is merely working he asked nothing and is owed no answer; a standing pointer at another session would read as this one nagging him about it.

## The clock

The fold stays pure and now takes a moment in, which is the design's own shape ("rows and a moment in, the finished view out"). It is needed for exactly one rule. **Null means the rule is not evaluated** and the session reads as plainly working - never as an answer that might be an hour old. The five minutes lives in one named constant; the other half of the rule needs no number, since the state is only reachable while the session is working and a session that stops again leaves it by itself.

Fifteen mutations, each red with the right symptom, controls green, restored and rebuilt:

|---|---|
| Just answered never reached | 4 fold tests and the route test |
| The five-minute rule ignored | the older-than-the-window test |
| A missing clock replaced with the real one | **nothing at first - see below** |
| A moment with no words shown as an answer | that test |
| A fleet message counted as his answer | that test |
| A message from before the stop counted | that test |
| Working again claimed anyway | the went-back-first test |
| The gap rounded in whole minutes | both answer tests ("0 minutes later") |
| The stop's lead never changing | the option test and the route test |
| The next session including this one | the skip test |
| The next session including the calm band | **nothing at first - see below** |
| Next offered while merely working | that test |
| The store dropping what was chosen | 2 store tests and both route tests |
| The bytes recorded instead of the words | the service test |
| His order thrown away | the service test |

**Two real gaps, both found this way and both closed.**

The **missing-clock** test used the fixture's own morning in September, which is hours past the window whichever way the fold reads a missing moment - so it passed with the rule deleted. It now places the answer twenty seconds ago by the real clock, with a control proving the same stop IS just answered when the clock is supplied.

The **calm band** was never in any roster a test built. The waiting line carries it after the reds, so "the first row of the line" and "the first row that needs him" differ whenever no red exists - and the snoozed and working rows already in the tests are absent from the line altogether, so neither could tell. A finished session would have been offered as the next one that needs him.

Proven: `CcDirector.Gateway.UnitTests` - **5,636 passed, 8 skipped**; 189 of those belong to this route, the store, the answer service and the frame parser. The default local gate is green across its eight suites (2,057 tests).

Not proven: `Gateway.Tests` and `Core.Tests` have not been run against this branch. The eight PostgreSQL-backed proofs skip without a database, so **the PostgreSQL migration has been generated and checked against the model, not applied to a server**. The one failure in the unit suite is pre-existing and filed as issue 3029, reproduced on clean `main` with none of this code.


thefrederiksen:

PR #3044, opened 2026-09-18
Wingman tab: one voice stamp for every surface, and the control the Now view offers

thefrederiksen:
Slice 6 of six - the last - for the Wingman tab's Now view. **Stacked on pull request 3035** (slice 5). The chain is 3028 - 3030 - 3031 - 3032 - 3035 - this one; GitHub retargets each as the one below it merges.

## Two stamps became one, and they had already drifted

A row's voice verdict was stamped in **two places** - the roster route and the Director push - each calling `VoiceDisplayFold` with its own argument list. The push path's list is **missing the backup-voice fact** the route's carries. Nothing detected that, because nothing compared them: two answers to one question do not disagree loudly, they just disagree.

The Now view needs the same verdict on the same row, and a third copy of that call is how one session comes to be "preparing audio" on one screen and silent on another. So there is now **one stamp**, `VoiceRowStamp`, and all three call it. It gathers, in one order, once:

- the two readiness booleans the Director never sets,
- the reason there is no audio (no credit, nothing to narrate, a Director too old to send its conversation, a narration given up on),
- the backup-voice notice,
- the waiting clock,
- and the folded verdict itself.

**It decides nothing.** Every word and flag is still `VoiceDisplayFold`'s.

**A fact a caller cannot see is left unstamped rather than invented as false.** That is what makes the move behaviour-preserving: the push path is handed no backup-voice fact and passes none, exactly as before, instead of a confident "no". Each path stamps precisely what it stamped before, which is why `DisplayPushVoiceEnrichmentTests` - the suite that exists because that seam already shipped once folding without two of the roster's inputs - is green untouched.

## The control the Now view offers

One of four things, and the fourth is nothing.

| | |
|---|---|
| **Play** | The voice fold says there is a clip. Copied, never recomputed. |
| **Preparing audio...** | There is not one yet. Not a button - there is nothing to press. |
| **Read this session aloud from now on** | Voice is off for this session. **From now on**, because that is what the switch does: it is a setting, not a request to narrate what is on the screen. |
| **Nothing** | Being read, working, just answered, or the Wingman switched off. |

Nothing at all is the honest answer in those four states: three have no stop on the screen to read aloud, and the fourth has no Wingman to read it. A control there is an affordance for something that is not there.

What follows the turn-on offer says whether he will hear **this** stop or **the next** one, and it is not decoration: promising him this stop when there is none to read is the same broken promise as a dead-end button, in smaller print. That is the defect this whole record exists for - the phone, handed no reason, guessed a hopeful "Generate narration now" beside a red "Voice unavailable" badge.

**A row carrying no voice verdict at all offers nothing rather than claiming there is no audio.** A row arrives without one when its caller cannot see the voice service, and that is not a claim the caller has earned.

## One difference from the mockup, stated rather than buried

The mockup's own script checks voice-off **before** the no-audio states, so it shows the turn-on offer on a working or just-answered session too. The design's table puts `none` for those four states first, and that is what this implements. Worth a ruling either way; changing it is one line.

Ten mutations, each red with the right symptom, controls green, restored and rebuilt:

|---|---|
| A control offered where there is nothing to read | that test |
| Play offered whatever the voice fold says | the preparing test and its route test |
| A row with no verdict read as "no audio" | that test and its route test |
| This stop promised when there is none | the next-stop test |
| The turn-on offer withheld | both voice-off tests |
| A fact the caller cannot see invented as false | that test |
| The backup-voice notice dropped | that test |
| Two reasons swapped on the way to the fold | 2 stamp tests **and a push-seam test** |
| The clock told the row as it arrived, not as it is | 8 tests across the stamp, the push seam and the route |
| The Now route never stamping the verdict | both route tests |

No gaps this time: every mutation was caught by a test written before it.

Proven: `CcDirector.Gateway.UnitTests` - **5,655 passed, 8 skipped**. The default local gate is green across its eight suites (2,057 tests). The push seam's own suite is green with no change to it, which is the evidence that the move preserved that path.

Not proven: `Gateway.Tests` and `Core.Tests` have not been run against this branch. Nothing about the Cockpit rendering any of it. The one failure in the unit suite is pre-existing and filed as issue 3029, reproduced on clean `main` with none of this code.


thefrederiksen:

PR #3049, opened 2026-09-18
Fleet Manager steps 5 to 9: where it runs, its page, the walkthrough, pinning and hand over, and the Assistant removed

thefrederiksen:
The Fleet Manager mission's steps 5 to 9, on top of steps 1, 3 and 4 which are already on main.
Seventeen commits: the five steps, the fixes each review round demanded, and the migration work
that puts every migration this branch adds after every migration now on main.

## What each step gives the owner

**Step 5 - where the Fleet Manager runs, and starting, restarting and moving it.**
Where the account's Fleet Manager runs is a per-account setting with its own Settings tab on the
desktop and the phone. One read tells the page everything: the agent and the computer, every
computer the account can reach and its state, whether the Fleet Manager is running, and which
actions are offered. Starting, restarting and moving it are single actions. A restart never takes
a Fleet Manager mid-turn: the new one is started and recorded as the successor, the old one stays
the marked, authoritative Fleet Manager until its Director reports it idle, and only then does the
mark move. If the new one fails to start, the old one keeps running and keeps the mark. If the
Gateway stops in the middle, a sweep carries the replacement on after it restarts. If the owner
moves the mark by hand, or clears it, the replacement is abandoned and nothing is closed. The
Fleet Manager runs in its own folder, not a repository, and the Gateway refuses to start one on a
Director too old to understand that.

**Step 6 - the Fleet Manager page.**
The desktop opens on it. The conversation is the Fleet Manager's own session history with each
filed record drawn in place as a card. Every open record is a card with its buttons, however old
it is; only the answered history is capped. A right-hand panel shows what is waiting on the owner,
what is under way, and what is answered today, and refreshes itself. Answering a card is ONE call
to the Gateway: the answer is saved and the "the owner answered" event is queued in the same
transaction, then typed into the Fleet Manager when it is next idle, and sent again until it is
acknowledged, so it survives a restart of either side. The page decides nothing for itself - every
sentence, label, button word, tone and enabled-or-not comes folded from the Gateway, as rule 7
requires.

**Step 7 - "Take me through them", the walkthrough.**
One waiting item at a time, in the same order the page uses. Each item shows the Wingman's reading
of that session verbatim - or the Gateway's sentence for why there is none - the Fleet Manager's
one line of advice and its pick, the session's own last lines, and the answer buttons the Wingman
offered. The owner answers, snoozes, skips, opens or closes, and the round remembers what is
settled. Closing a session is offered only on positive evidence that it is safe: its Director
reports no uncommitted files, and the account holds a repository report from that same Director,
covering that session's folder, taken after the session's last activity, clean, on a named branch
contained in the default branch. Anything else, including "cannot tell", withholds the button and
says why. What the owner's answer records is what the Gateway actually sent: the answer route
stores the verdict, the option positions and their words in the same save as the answered mark,
and the walkthrough records that stored choice, not positions a client names. An answer to an
earlier stop can never close a later record.

**Step 8 - pinning and hand over.**
The running Fleet Manager is the first row of the session list on both surfaces, marked, with the
sessions it owns collapsed under it. The owner can hand a running session to the Fleet Manager or
take it back, from the session list, the Fleet Manager page or the command line
(`cc-devthrottle session hand-over <session> --to fleet-manager|owner`), and every place that
treats a session as the Fleet Manager's follows the owner it has now - the Wingman's routing, the
digest, the page's Under way, the walkthrough. The Fleet Manager may also hand sessions over
itself with its own session key, under the owner's rules: it takes a session that answers to the
owner, hands back one it owns, never takes one another running session owns, and only ever reads
its own account. An owner change is compare-and-set at the Director: of two hand overs sent at
once, exactly one is made. A change that cannot be written to the Director's crash journal is
undone and reported as not made, so the owner is durable across a Director restart.

**Step 9 - the Assistant is gone.**
The Fleet Manager page replaced it, so the Assistant view, its stylesheet, its icon, the phone's
Assistant page and menu entry, the shared Assistant hook and turn machine, the fleet brain client,
the hidden Assistant settings tab, and on the Gateway the whole fleet brain, its turn and keep-warm
routes, the two car-mode settings keys and their routes and fields, and the three Car Mode
configuration classes are all removed. The old addresses redirect to the Fleet Manager page (the
phone, which has no Fleet Manager page, lands on the session list). A migration in each provider
deletes the two stored settings rows.

## Review rounds passed, and the blocking findings fixed

- **Steps 5 and 6, independent inspection, then three rounds of fixes.** Six findings, four of them
  blocking: a move or restart took authority from a Fleet Manager mid-turn; the old one was closed
  without waiting for its Director to report it idle; answering a card was two calls that could half
  happen; old open records were dropped out of the card list; and the page decided for itself what a
  state meant. Round two: a replacement closed whatever was marked rather than the session it
  recorded, the mark move was not one transaction, and events were typed into a Fleet Manager being
  replaced. Round three: the replacement now compares the current mark with the session it recorded
  before anything else, and a session started to take over is told exactly once, whenever it is
  marked.
- **Step 7, one round of fixes.** The walkthrough recorded option positions a client named rather
  than what the answer route stored, and an answer given to a different stop could close a record.
- **Step 8, two rounds of fixes.** Round one: the Fleet Manager could not hand over at all, and an
  owner change was not durable - the Director's crash journal, which is what a restarted Director
  reads, carried no owner. Round two: the hand over overwrote the owner without checking, so it is
  now compare-and-set and the Gateway refuses a hand over to any Director that does not say it can
  do that; and a record filed about one stop could be closed by an answer to another, so a record
  now stores the stop it is about and closes only for that stop.
- **A Fleet Manager waiting to take over was never judged.** Its first turn ends before the mark
  moves to it, and turn ends were kept only for the marked session, so its takeover event waited for
  a turn end that would never come. Turn ends are now kept for every session recorded as waiting to
  take over.

## What is proven

Everything below was run on this branch on a Mac, after a rebase onto the current `main`
(44d5b661).

- **Build**: the whole solution, 0 errors and 0 warnings.
- **Gateway unit tests**: 6,100 passed, 7 failed, 8 skipped, 6,115 total. A clean worktree of
  `origin/main` was built and run for comparison: 5,763 passed, 7 failed, 8 skipped, 5,778 total.
  The seven failures are the SAME seven names on both, and all seven are this computer rather than
  this branch - two file-rename cases, two rules cases that need symbolic-link behaviour, and three
  process-liveness cases that need `cmd.exe` and Windows process security. **No new failure.** The
  branch adds 337 tests.
- **Command line suite**: 3,356 passed, 0 failed.
- **Web**: `npm run typecheck` clean across all four packages; shared client 1,296 passed in 112
  files, desktop 417 passed in 49 files, phone 86 passed in 14 files. 0 failed.
- **Migrations**: the five migrations this branch adds are renumbered so that both providers sort
  them after every migration now on `main`, in the same order in each. Each Designer model is the
  model at that point, both snapshots equal the current model, and a chain test walks from step 4's
  delivery migration proving that the difference between each pair of Designer models is exactly
  what that migration's `Up` performs - no more, no less. The ordering and apply-from-empty tests
  pass in both providers.
- Each fix on this branch was watched failing with the fix reverted, with the reported symptom, and
  then restored.

## What is NOT proven

- **Nothing was run against a live Gateway, a live Director, or a real browser.** Every proof here
  is a test process. The booted-host tests start the Gateway in-process; they are not the deployed
  service.
- **The PostgreSQL migrations were never applied to a real database by this branch.** Docker is not
  running on this computer, so the PostgreSQL apply-from-empty proofs are gated off and the
  PostgreSQL boot smoke test is skipped. Their PostgreSQL correctness rests on the model comparison,
  not on an applied database.
- **Screenshots and fixtures use test data**, not a real account's sessions.
- The parked suites (`Gateway.Tests`, `Core.Tests`) were not run here, and no release gate was run.
- The phone has no walkthrough and no hand-back action; restoring an interrupted session creates a
  new session and does not re-apply its owner (the owner is now in the crash journal for that to
  read); the phone's redirect routes have no test of their own.


thefrederiksen:
**Verdict: No significant problems.**

I read the branch in my own worktree (`origin/fm/step59`, a47454097) and reviewed the five steps against the risks this mission's owner cares about: a mark moved or a session closed while it still had work; an event or a card answer lost, doubled, or delivered to the wrong session; the Cockpit ruling for itself; a hand over taking a session another session owns; Assistant-removal leftovers; and migrations that would not apply in order. I found nothing that is wrong, unsafe, loses work, or breaks for the owner.

What I checked, and what I found:

**Migrations (both providers).** The five new migrations (110200 through 110600 in the SQLite provider, 110200-prefixed twins at 110209/…/110609 in PostgreSQL) sort after every migration now on `origin/main` in both providers, in the same order, with identical names. No migration already on main was renamed, so no applied migration can be re-applied after deploy. The `RemoveAssistantSettings` twins delete exactly the two rows by key (`car_mode_model`, `car_mode_end_phrase`); the PostgreSQL one correctly schema-qualifies `"gateway"."tenant_settings"` and quotes the `"Key"` column, which matches the PostgreSQL snapshot. The edits to existing Designer files (`AddDevReports`, `AddTurnVerdictTraceRowAndClock`) are content-only, design-time-only fixes (the runtime never executes `BuildTargetModel`), and the chain test in `FleetManagerLaterStepsMigrationChainTests` proves each Designer differs from the previous by exactly its own `Up`, for both providers.

**Restart, retirement and the mark move (step 5).** `FleetManagerPlacementService.RetireAfterTurnAsync`/`AdvanceReplacementAsync` closes the old Fleet Manager only when its Director reports it Idle on two consecutive looks with an unchanged delivery generation, inside the per-account placement gate and the shared `FleetManagerDeliveryGate`; `Working`, `WaitingForInput` and `WaitingForPerm` are never closed and the status says so. The mark is compared against the recorded old one before anything else, so an owner's mark change or clear abandons the replacement and closes nothing; the promotion (mark, history, replacement removal and the one `marked` event) is one transaction in `FleetManagerPromotionStore`, and a failure leaves the replacement recorded for the next look. Interrupted starts are recovered by the sweep (`FindInterruptedStart`) by remembering the started sessions as waiting - nothing closed, nothing marked. Both services share one `FleetManagerDeliveryGate` instance (a field on `GatewayHost`), and the lock order (placement gate, then delivery gate) is consistent everywhere, so there is no deadlock path.

**Events and card answers (steps 4-6 surfaces).** An owner's card answer is one call: `FleetOutcomeStore.Answer` writes the answer and queues the `answered` event in the same transaction (`AnswerOutcomeAsync` passes `FleetManagerEventStore.AnsweredEvent` as `tellFleetManager`), so neither can exist without the other. Delivery is at-least-once by design (prompt typed before the delivery is saved) with `Owed` excluding events already delivered to that same session, so a new Fleet Manager receives everything still open but a living one is not woken forever. The `marked` event is stored once per session (`AddMarkedIn` dedup), nothing is typed into a Fleet Manager being replaced (the `ReplacementPending` check sits inside the delivery gate), and deliveries are held until the Wingman's reading of the Fleet Manager's latest turn says it asks the owner nothing.

**The walkthrough (step 7).** `AnsweredAsync` records the options the answer route stored (`taken.Words`, `taken.OptionIndexes`), not what a client named, and refuses with a specific conflict when the verdict answered is not the record's stop, was answered before the record was filed, or is no longer the session's newest stop - so an answer to an earlier stop cannot close a later record. Close is decided by `FleetManagerCloseRule` with positive evidence only (no uncommitted files, a repository report from the same Director covering the folder, taken after the session's last activity, clean, on a named branch contained in the default branch), and the close route asks the rule again before stopping - a hidden button is not a refusal.

**Hand over (step 8).** `FleetManagerHandOverService` refuses to take a session a live session owns, refuses the Fleet Manager itself and the waiting successor, and the owner change is compare-and-set: the Director honours `set-controller` only while the owner is still the one checked (`ChangesOwnerIfExpected`, refused for Directors that did not say so), and of two concurrent hand overs exactly one lands. `Session.SetController` undoes the change when the crash-journal write fails, and `PersistOwnerChange` refreshes the full roster when the session was not yet in the journal, so the owner is durable across a Director restart. The Fleet Manager's own session key is checked against the live marked Fleet Manager before the target session is even looked up, and then meets the owner's own rules.

**The client is dumb (rule 7).** The Cockpit's Fleet Manager page, walkthrough, Settings tab, session-list pinning and the phone's surfaces render folded answers: every sentence, label, tone, button word and enabled-or-not boolean (`ComposerUsable`, `QuickPromptsUsable`, `Pin`, `OwnerChange`, the close permission) comes from the Gateway folds (`FleetManagerPageFold`, `FleetManagerWalkthroughFold`, `FleetManagerPlacementFold`, `FleetManagerRosterFold`). The only client-side logic I found is layout (interleaving cards with history bubbles in `conversation.ts`), which is legitimate.

**Assistant removal (step 9).** Nothing still calls anything removed: no references remain to `AssistantView`, `useAssistant`, `assistantApi`, `brainApi`, `AssistantTab`, `FleetBrainEndpoint`, or any `CarMode` class outside tests, in either the TypeScript or the C# trees. The old addresses redirect (the Cockpit's `/assistant` to the Fleet Manager page, the phone's `/assistant` and `/car` to the session list), the OpenAPI schema was regenerated without the brain routes, and the two stored settings rows are deleted by the migration in each provider. `SessionKeyGuard` classifies every new route deliberately: placement/start/restart/move/page/walkthrough refuse a session key, hand-over lets only the POST through and the service refuses every session key but the live Fleet Manager's.

Two observations, explicitly **not** findings, both narrow and already accounted for:

1. During a replacement the owner's own prompt (typed into the conversation or a quick prompt) does not pass through the delivery gate, so a prompt sent in the seconds between the retirement loop's second Idle look and the Director carrying out the close would be cut short by the close. The two-look rule makes the window a few seconds wide, the bar says a replacement is under way, and the consequence is limited to a prompt typed into a session the owner just asked to restart - the same thing the owner can already do deliberately from any session's stop button. I judged this below the bar for a finding.

2. The gap stated in `FleetManagerEndpoints` predates this pull request (step 2 is on main): `PUT /gateway/fleet-manager` is reachable by any session key of the account, so a session can mark itself and pass the Fleet Manager's own route checks - and step 8's hand-over means a self-marked session would then pass `RefuseUnlessFleetManager` too. It is documented in the code as a known gap "until that route is narrowed", and this branch neither introduced nor widened it beyond that amplification; worth narrowing in a later step.

The branch's own "What is NOT proven" section is honest and I agree with its scope: nothing here ran against a live Gateway, Director or browser, and the PostgreSQL apply proofs are model-comparison based because Docker was not running on the author's machine. Those are gaps in evidence, not defects I found in the code.

PR #3098, opened 2026-09-18
A Linux Director installs the update it has downloaded, without a person restarting it (#3091)

### LINKED ISSUE #3091: A Linux Director installs the update it has downloaded, without a person restarting it
## Why

A Linux Director stages every update it finds and installs none of them. It is silent: no error in any log, and the only visible symptom is a staging directory that grows by about 105 MB per release.

On the test machine (Ubuntu 24.04, one Director, a build of `main` taken after #2979) three builds were downloaded and left unused in a single day:

| version | staged at |
|---|---|
| 2.5.0 | 2026-09-18 01:15 |
| 2.6.0 | 2026-09-18 09:16 |
| 2.7.0 | 2026-09-18 18:17 |

From `config/director/updater-state.json`:

```json
{
  "stagedVersion": "2.7.0",
  "stagedExecutable": ".../updates/2.7.0/cc-director-linux-x64",
  "installTarget": ".../app/cc-director",
  "applyAttempts": 0,
  "lastCheckOutcome": "Staged"
}
```

`applyAttempts: 0` is the whole finding. The install was never attempted - not attempted and failed.

## What is missing

#1033 moved the install out of the Director and into the launcher, for a good reason: a process cannot witness its own replacement. `DirectorUpdateOwner` holds the rule - a staged update plus no running sessions means install it - and the launcher is what outlives the swap and can put the old build back.

#2979 then made a Linux Director check, download, stage, and be installable. What neither did was put a running launcher on a Linux machine to do the installing.

Nothing starts a launcher on Linux:

- `LauncherAutostart` is the only autostart mechanism there is. It writes `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, and every method carries `[SupportedOSPlatform("windows")]`. There is no Linux counterpart - no systemd user unit, no `~/.config/autostart` entry.
- `LinuxDesktopEntry.BuildEntry` writes `Exec={QuoteExec(directorExe)}`, so the application-menu entry added in #2985 starts the Director. That is correct and matches Windows - but it means nothing on the machine ever starts the launcher.

The launcher binary is built, shipped and installed on Linux - `ComponentRegistry` carries `LinuxAsset: "cc-launcher-linux-x64"` and it lands in `.local/share/cc-director/launcher/cc-launcher` - and #2979 even fixed `DirectorSupervisor.Start` so that a Linux launcher could restart a Director. It is Linux-capable, installed, and never started.

So the gate everyone reaches for first - "it had no sessions, why did it not update itself?" - is never evaluated on Linux at all, because the process that owns that decision is not running.

## What this is not

It is not a broken install. The test machine happened to start its Director from a shell script, but the supported path has the same shape: the application-menu entry starts the Director directly too.

It is not a detection or a network problem. The same machine's Python tool bundle reached 2.7.0 on its own that afternoon, and the hourly check correctly reports `Release v2.7.0 is already staged`.

## The escape hatch, and why it did not fire

`UpdateInstaller.TryApplyStagedUpdateAtStartup` survives for exactly this case - its own comment says removing it "would permanently strand any install whose launcher is missing". But it only runs on a start, and this Director had been up since before 2.5.0 existed. `applyAttempts: 0` confirms it has never run on this machine, so the Linux path through it is still unproven.

## The rule this breaks

Every platform should behave the same way; only the implementation differs. A launcher runs from login on
every machine, stages its own update, and installs the Director's when the Director is idle. How it is
started at login is the only part that is allowed to be platform-specific - and that is the part Linux never
got:

| Platform | Mechanism | Implementation | Exists |
|---|---|---|---|
| Windows | HKCU Run key | `LauncherAutostart` | yes |
| macOS | launchd user agent | `LauncherLaunchdAutostart` | yes |
| Linux | - | - | **missing** |

`LauncherLaunchdAutostart` describes itself as "the macOS twin of `LauncherAutostart` (the Windows Run key),
living in the engine for the same reason: the installer, the uninstaller, and the launcher itself must agree
on one label, one property list path, and one command-line format." This asks for the third twin, written to
the same sentence.

We have fixed this exact defect once already, for macOS. `LauncherCore.RunUpdateLoopAsync` records it:

> THE WHOLE OF JOB 1 USED TO SIT BEHIND AN `IsWindows` GATE, and that is the defect this shape replaces: a
> Mac never looked for a launcher release, so there was never anything staged, so the Director had nothing to
> install, so NO LAUNCHER ON A MAC EVER UPDATED ITSELF.

Read with "Linux" for "Mac", that is this issue.

## What would close this

1. **The Linux autostart twin** - one unit or entry, one command-line format, shared by the installer, the
   uninstaller and the launcher. New file in `cc-director-setup-engine`, modelled on
   `LauncherLaunchdAutostart`, with tests beside `LauncherLaunchdAutostartTests`.
2. **Let Linux through the gate.** `LauncherCore.RegisterAutostartSafe` and the equivalent check in
   `LauncherTrayController` currently fall through to `// no mechanism on this platform`.
3. **Start it at install, remove it at uninstall.** `InstallFinalizer` and `Uninstaller`.
4. **Decide the launcher's shape on Linux.** It is a tray application with no headless mode -
   `LauncherAppOptions` parses only `--no-autostart` and `--managed` - and GNOME does not show legacy tray
   icons without an extension. Either add a headless mode (the two jobs need no window), or ship a real Linux
   tray. This is an open decision, not a detail.

**Done means a presence, not an absence:** on a freshly installed Linux machine, after a logout and login,
`ps` shows a running `cc-launcher`, and a deliberately staged older Director is replaced with nobody touching
the machine, with the new version answering afterwards. "No errors in the log" is not evidence - this bug
produced no errors across three consecutive releases.

## Not in scope

Rescuing Linux machines that are already stuck. They cannot receive this fix automatically, which is the bug
itself, so they are reinstalled.

## Correction to the original text of this issue

As first filed, this issue said `LinuxDesktopEntry` should start the launcher rather than the Director. That
was wrong and has been removed. On Windows the Start Menu shortcut targets the **Director**
(`InstallFinalizer.CreateDirectorShortcut` resolves `ComponentRegistry.Director`), and the launcher is started
separately by the Run key. Parity therefore says the Linux desktop entry should keep starting the Director
exactly as it does now, and the launcher gets its own autostart beside it.

thefrederiksen:
Closes #3091.

## Why

A Linux Director found every release, downloaded it, verified it and staged it - and installed none of them, silently and for ever. Three full builds were staged and unused on one machine in a single day, and `updater-state.json` said exactly why:

```json
{ "stagedVersion": "2.7.0", "applyAttempts": 0, "lastCheckOutcome": "Staged" }
```

`applyAttempts: 0` is the finding. Not a swap that failed - an install that was never attempted, with no error in any log.

The install is not the Director's job: since #1033 the launcher owns it, because a process cannot witness its own replacement. Windows starts the launcher from a Run key, macOS from a launchd agent, and on Linux `LauncherCore.RegisterAutostartSafe` fell out of the bottom of an "is this Windows or macOS" condition with `// no mechanism on this platform`. So the launcher - which ships for Linux and sits on disk - never ran, and the thing that installs updates was not there.

We have fixed this same defect once before, for macOS, when the launcher's update job sat behind a Windows-only gate. `LauncherCore` still records it: *"NO LAUNCHER ON A MAC EVER UPDATED ITSELF"*. Linux is that story one platform over.

| Platform | Mechanism | Before | After |
|---|---|---|---|
| Windows | Run key (`LauncherAutostart`) | yes | yes |
| macOS | launchd agent (`LauncherLaunchdAutostart`) | yes | yes |
| Linux | systemd user unit (`LauncherSystemdAutostart`) | **missing** | **added** |

## What changed

1. **`LauncherSystemdAutostart`** - the Linux twin, and the direct counterpart of the Gateway's existing `GatewaySystemdAutostart`. `WantedBy=default.target` to start at login; `Restart=on-failure` so a crash is resurrected but a clean exit stays exited (the systemd analogue of launchd's `KeepAlive SuccessfulExit=false`).
2. **Linux goes through the autostart gate**, with each branch naming its platform positively rather than leaning on a trailing `else` - the platform analyzer cannot reason through one, and it caught a real defect in review of my own first draft.
3. **Headless on Linux by intent, not as a fallback.** The launcher is a tray app on Windows and macOS; Linux has no dependable tray and neither of the launcher's two jobs needs a window. `Program` routes Linux to the headless core that already existed for degraded starts, and records `headless` rather than `degraded` so a healthy Linux launcher does not report a fault.
4. **The swap order moves to place-then-restart on Linux.** This is a consequence of (1), not a preference. `LauncherSwapOrder` documented Linux as stop-first *"where nothing supervises the launcher"* - a premise the new unit makes false. Stopping first would let systemd bring the **old** build back up in the gap before the swap lands, leaving a launcher that is alive and the wrong version: the exact race the enum already describes for macOS.
5. **The uninstaller removes the unit, and says so when it could not.** A unit left enabled keeps `Restart=on-failure`, so systemd could bring the launcher back after the uninstall's stop had already been certified.

## Proof, and what it does not cover

Run on Ubuntu 24.04 with the .NET 10.0.301 SDK.

- `CcDirector.Launcher` builds clean (0 warnings, 0 errors), after the platform analyzer rejected the first draft of change 2.
- **Setup engine suite:** 578 passed / 22 failed. **Baseline `origin/main` at the same commit, zero changes: 565 passed / 22 failed.** The failing sets are byte-identical (`diff` empty) - 22 pre-existing Linux failures, none introduced. The +13 are the new tests, all passing.
- **Launcher suite:** 24 failed / 173 passed, and the baseline failing set is again byte-identical - 24 pre-existing, none introduced.

**Not covered, and it matters:**

- `scripts/test-local.ps1`, the repository's local gate, **was not run**: there is no PowerShell on this machine. The two affected suites were run directly instead, against a measured baseline. A reviewer on Windows should run the real gate.
- **No launcher was started on Linux, and no unit was installed.** These tests pin the unit's text; they cannot prove systemd does anything with it. The claim that a launcher comes up at login and that an idle Director then installs a staged update with nobody present needs a run on a real Linux machine, and that run is what should gate the merge.
- `ToolReconcilerTests` crashes the test host on Linux on a `/tmp` path, on baseline as well as here, so it is excluded from the comparison above. Unrelated and pre-existing.
- Windows and macOS were not re-tested. Their code paths are read, not run. Change 4 alters the default swap order for **macOS as well** in form (it was already place-then-restart there, now expressed as "not Windows"), so a Mac reviewer should confirm that reads correctly.


thefrederiksen:
REVIEW VERDICT: SENT BACK — one blocking defect. Reviewed independently against the reported symptom (issue #3091), not against the report, on a fresh worktree cut from this branch head, on Windows.

## The blocking defect: nothing ever starts the launcher on Linux, and the launcher is the only writer of the unit

The change gives the launcher the ability to register a systemd user unit for itself — but `LauncherCore.RegisterAutostartSafe` runs only when the launcher is already running, and nothing on Linux starts the launcher. It is a chicken-and-egg with no egg:

- The ONLY caller of `LauncherSystemdAutostart.EnsureRegistered` is `LauncherCore.RegisterAutostartSafe` (src/CcDirector.Launcher/LauncherCore.cs:315). The unit is written by the launcher, and the unit is the only thing that starts the launcher at login.
- The command-line installer does not start it and does not write the unit: `tools/cc-director-setup-cli/Commands.cs` still takes the Windows and macOS branches and falls into the Linux `else` that says, unchanged by this change: "The launcher is not started on Linux yet. The Director does not need it: a downloaded update installs the next time the Director starts." That sentence is what issue #3091 exists to refute — the escape hatch runs only on a Director start, and the machine in the issue had been up across three releases.
- The setup wizard does not either: `tools/cc-director-setup-avalonia/Services/EngineInstallRunner.cs` InstallLauncherAsync returns false for anything that is not macOS, and Linux finalize writes only the application-menu entry, which starts the Director.

Issue #3091, "What would close this", item 3: "**Start it at install, remove it at uninstall.** InstallFinalizer and Uninstaller." This pull request implements the uninstall half only. The issue's own definition of done — on a freshly installed Linux machine, after a logout and login, `ps` shows a running `cc-launcher` — fails with this change: a fresh install places the launcher, starts nothing, writes no unit, and login starts nothing, so the launcher never runs, the unit is never written, and `applyAttempts` stays at zero exactly as before. The "Not in scope" note says stuck machines are reinstalled — but a reinstall on this code produces a stuck machine again.

The pull request's "Not covered" framing understates this. It is not that the unit's behaviour needs a run on a real Linux machine to be proven; the reachable chain itself has a hole. The fix as merged would change nothing for any machine that does not hand-start `cc-launcher` at least once.

**What closes it:** register the unit and start the launcher at install on Linux, in both install surfaces (`tools/cc-director-setup-cli/Commands.cs` and `tools/cc-director-setup-avalonia/Services/EngineInstallRunner.cs`), mirroring what macOS already does — `LauncherMacInstaller` calls `LauncherLaunchdAutostart.EnsureRegistered` at install precisely so the agent exists before the launcher first starts (LauncherCore's own comment records this: "on macOS from the FIRST run, because the installer now registers the agent before the launcher starts"). The launcher-side self-registration this change adds then remains as the keep-fresh mechanism. While there, the stale Linux message in Commands.cs ("a downloaded update installs the next time the Director starts") should be corrected or removed — this branch's own documentation is the thing that proves it false.

## What I verified independently, and it holds

1. **The root cause is right.** Confirmed from the code, not from the report: since issue #1033 a running Director never installs its own staged update (`UpdateInstaller.TryApplyStagedUpdateAtStartup` is called from startup only, and its own comment says "A RUNNING Director never calls this any more (issue #1033)"); the unattended install route is the launcher's periodic pass (`LauncherCore.RunUpdateLoopAsync`, job two, through `DirectorUpdateOwner.RunOnceAsync`). On Linux, `RegisterAutostartSafe` fell out of the bottom of the "Windows or macOS" condition with no mechanism, so no launcher ever ran, so `applyAttempts: 0`. The cause attribution is correct.

2. **The swap-order change is sound, and Windows and macOS are not broken.**
   - The race argument is real: the stop path (`LauncherUpdateOwner.StopInstalledLauncher`) first asks the launcher to quit and then KILLS whatever does not comply — and a killed process is exactly the unclean exit that `Restart=on-failure` resurrects. The non-complying process is precisely the old build being replaced. A graceful stop would be no better: it ends in a manual start that systemd does not supervise, and a duplicate at the next login. Place-then-restart is correct, and replacing the file under a running process is safe on Unix because the process keeps its inode.
   - Windows: the default is unchanged (still `StopThenPlaceThenStart`), pinned by the new `LauncherSwapOrderByPlatformTests` Windows case, and the whole Windows gate below ran green through both affected suites.
   - macOS: unchanged in behaviour — same `PlaceThenRestart` default, same `LauncherLaunchdAutostart.Kickstart` restart, same `RegisterAutostartSafe` branch, same degraded-fallback. The headless label change ("degraded") affects the Linux branch only.
   - The branch is six commits behind main; I checked for overlap and conflicts with all six newer main commits and found none.

3. **The test changes strengthen, not weaken.** The shared test factory stays hermetic (every process delegate is injected); it now states the swap order it was written around and makes `RestartLauncher` throw if ever reached under stop-then-place-then-start. Before this change those tests inherited the host's platform default, which on macOS and Linux would have driven the REAL launchctl/systemctl from a unit test — a latent bug, and the new default would have tripped it on Linux hosts. The supervised order keeps its own dedicated tests that supply their own `RestartLauncher` (LauncherUpdateOwnerTests, and LauncherSelfUpdateTests), and the platform default is pinned separately. Nothing was loosened to make this change pass.

4. **The headless routing loses nothing Linux needs.** The Linux branch sits after the single-instance mutex and the `--apply-update` handling, so those guards are intact. `RunHeadless` keeps Gateway registration, the persistent command stream, the autostart registration, the self-update loop, SIGTERM/SIGINT handling, and a graceful core stop. The `userInterfaceState` value is recorded in the registration file but nothing branches on it, so the new "headless" string breaks no reader. Two cosmetic notes, not blocking: the Linux path returns before Main's final `finally`, so the "CC Launcher exited" line and the log stop are skipped (RunHeadless's own finally does the substantive cleanup); and `After=network.target` is a no-op inside a `--user` manager, which has no network target — harmless either way.

## The local gate (run on Windows, the thing the author could not run)

`.\scripts\test-local.ps1` — GREEN. All eight suites, zero failures, 2,236 tests:

| Suite | Result |
|---|---|
| CcDirector.Core.UnitTests | 666 passed |
| CcDirector.Avalonia.Tests | 550 passed |
| CcDirector.Engine.Tests | 63 passed |
| CcDirector.HostedAgent.Tests | 88 passed |
| CcDirector.Launcher.Tests | 197 passed |
| CcDirector.Terminal.Avalonia.Tests | 25 passed |
| cc-director-setup.Tests | 25 passed |
| cc-director-setup-engine.Tests | 622 passed (the 13 new tests included) |

The gate flags a coverage gap into the parked `Gateway.UnitTests` suite (it maps `src/CcDirector.Launcher` to it), so I ran that suite directly as well: 6,334 passed, 9 failed — all nine are the PostgreSQL-backed stats schema tests, which fail identically on any tree run without the throwaway database that only the `-Parked` gate provisions. They are environmental and unrelated to this change. At merge time, run the full `-Parked` gate as the instructions require (it needs Docker); do not merge on this comment alone.

On the author's Linux baseline comparison (byte-identical failing sets against a zero-change worktree): I cannot reproduce a Linux run on this Windows machine, so I did not re-check it — but the fully green Windows run of the same two suites makes it moot for this verdict.

## Summary

The mechanism, the routing, the swap order, the test changes, and the Windows and macOS paths are all correct and green under the real gate. But the change is unreachable where the defect lives: a fresh Linux install still starts no launcher, still writes no unit, and still leaves `applyAttempts` at zero — the issue's definition of done fails on the machine this was written for. Sent back for the install half of issue #3091 item 3: register the unit and start the launcher at install, in both install surfaces, and fix the stale Linux message that claims the Director installs a downloaded update itself.

PR #3100, opened 2026-09-18
One tools folder per machine: the resolver, in the installer and every reader

thefrederiksen:
The tools home is the machine root: `<root>/bin` for the launchers, `<root>/pyenv` and `<root>/python` for what runs them. A Director's own folder is never an install root.

## Why

Every Director points the `CC_DIRECTOR_ROOT` setting at its own folder, `<root>/instances/<name>`, and hands it to every session it starts. The installer layout and every reader that looked for an installed tool asked `CcStorage.Root()`, so an install, a tools repair, or a plain lookup run from inside a Director or one of its sessions took that Director's folder for the whole machine.

One machine ended up with seven complete copies of the tools, each with its own interpreter, each ageing at its own pace, and whichever copy reached the search path first answered for all of them. On 17 and 18 September a two-month-old copy told every agent to set a variable for a part of the product that no longer exists; it cost about forty minutes and was filed four separate times.

## The rule

One resolver, `CcStorage.MachineRoot()`, climbs out of a Director's folder and answers with the machine root above it. It climbs repeatedly, because the same leak produced a nested `instances/default/instances/default`, and one climb out of that lands on another Director's folder.

A root that is **not** a Director's folder - a throwaway root a test rig pins - comes back exactly as it went in and keeps its own tools. That separation is what makes a rig safe to run at all, and every test below asserts both halves.

The shape test now has one definition; `InstanceContext.LooksLikeAnInstanceHome` calls it, so the storage paths and the launcher's own refusal cannot drift apart.

## What changes, all in one pull request

Either the installer or the readers alone would leave a Director looking where nothing is installed, so they land together.

- The installer layout (`InstallLayout.Default`), and with it the tool reconciler, the tool updater and the setup command line - they all take that default.
- `setup_ops.py` in cc-devthrottle, which is run from inside a session more than anywhere else.
- The session path builder, the Tools page, the worktree pool, the browser harness installer, and the macOS terminal helper.

**Three found by sweeping beyond the list I was given:**

- What the About boxes report as the install root, and the installed-component manifest they read. A named Director would have reported every component as missing and every Python tool's version as unknown - it was reading a file only the machine root has.
- The browser harness's own environment. Its shim is written into the machine's `bin`, so an environment inside one Director's folder leaves every other Director on the machine pointing at a folder it does not own.

**Two comments corrected on purpose.** `InstanceContext` said shared binaries "stay at the shared root by design" in the present tense while the code had stopped doing it, and `InstallLayout.PyenvDir` called the environment shared without saying one per what. The Tools page carried a third, arguing at length that the tools belong under a Director's home; that argument is the leak, and it is rewritten rather than left to persuade the next reader.

Out of scope and untouched: the path repair, the deletion sweep, and the registry ship flags.

## Proof

**Each fix reverted once and watched failing, with the rig-root controls staying green.** Eleven of them: the tool directory, the bundled interpreter, the repeated climb, the installer layout, the session path, the Tools page, the worktree pool, the terminal helper, the harness environment, the About manifest, and the cc-devthrottle setup root. Three examples of what the red said:

- repeated climb: `Expected: "root\cc-director"` / `Actual: "root\cc-director\instances\default"`
- installer layout: `Expected: ...-machine-d152f64` / `Actual: ...d152f64\instances\slot-5`
- session path: `src/CcDirector.Core/Sessions/SessionManager.cs: var ownToolBin = Path.Combine(Instances.InstanceContext.InstanceHome, "bin");`

**The readers are watched, not just the path helper.** Each test puts the real thing on disk under the machine root, pins the setting at a Director's folder the way a running Director does, and asks the reader its own question. For the three that cannot be instantiated in a unit test - the session PATH, the Tools page inside a window class, and the macOS helper that only extracts on a Mac - a guard names each one and the call that proves it, so a reader that stops asking fails by name rather than by an absence.

**Two existing tests in the setup engine had to change**, and both are worth reading:

- `InstallRootResolutionTests.Default_UsesTheSameRootAsTheDirectorItself` asserted the exact claim this change overturns. It compared the installer's root with `CcStorage.Root()`, which held on a developer's shell and quietly licensed the leak. It now compares against the machine root, with a second test that the two answers still agree on an ordinary root - so nothing about where a normal install lands has moved.
- `GatewayAccountEnrollRunnerTests` resolves the hosted address from a process-wide environment variable but sat outside the collection that serialises it, so it raced the classes that point the override at a stub host: the cancellation test read `DEVTHROTTLE_HOSTED_GATEWAY_URL is set to ...` where it expected `Sign-in was cancelled`. Nothing about the code under test was wrong either time. Adding a collection beside it changed the scheduling enough to surface this; joining the collection it always belonged in fixes it, and three consecutive suite runs are green.

**The gate.** `.\scripts\test-local.ps1` on the rebased tree: eight suites, all `outcome=Completed`, zero failures.

## What is not proven here

- **The parked suites did not run.** `Core.Tests`, `Gateway.Tests` and `Gateway.UnitTests` are parked, and `Core.Tests` is where this change's new tests live. They were each run directly and pass - 26 in `Core.Tests`, 9 in the engine suite, 9 in the Python suite - and every revert above was watched against them. A full `-Parked` run was started and was stopped by the machine running low on memory, and Docker is not running here, which `-Parked` needs for the database-backed proofs. So the full parked run is outstanding, not green and not red.
- **macOS and Linux on a real machine.** The rule is written for all three and the tests are portable - the path assertions use relative segments so they mean the same thing on every platform - but nobody has watched the macOS terminal helper or a Linux install run.


thefrederiksen:
## Review finding answered: the absence guard now sees the shape the leak was written in

**The finding.** `ToolPathsNeverComeFromADirectorFolderTests` carried the claim *"nothing in the product composes a path to an installed tool out of a Director's own folder"*, but its scan recognised one idiom: a line naming `InstanceHome`. The Reviewer proved the gap with a probe - `Path.Combine(CcStorage.Root(), "bin")`, which is the original leak's own shape and the entire reason this phase exists - and watched it pass GREEN while the `InstanceHome` spelling was caught. A check whose pass condition is an absence, certifying more than its instrument can see.

**Accepted, and widened rather than narrowed.** Narrowing the comment was the honest option and would have been a real answer, but it would have left the `Root()` idiom watched only by the seven named readers - one rename away from unwatched, in the one place this mission exists to hold. Widening was available without the scan crying wolf: a sweep of every production C# statement naming `Root()`, `Base()`, `SharedRoot` or `InstanceHome` beside a quoted `bin`, `pyenv` or `python` finds **nothing** on this tree, so the wider instrument starts from the same clean reading the narrow one gave. That is what makes widening the better answer here rather than merely the more ambitious one.

**What changed** (commit `65ec1c33c`, one test file):

- **Four producers of a Director's folder instead of one**, each named in the failure message so the reader is told which way in was taken. `Root()` and `Base()` are the pair that matters: neither reads like a Director's folder, and both are one whenever the `CC_DIRECTOR_ROOT` setting points at one - which is inside every Director and every session it starts. `\bRoot\(\)` does not match inside `MachineRoot()`, because there is no word boundary in the middle of a word.
- **A statement at a time, not a line at a time**, so a wrapped `Path.Combine` cannot hide in the gap between two lines. The Reviewer named this as a second hole in the same family.
- **Comments dropped first**, so a line that writes the forbidden shape out in order to forbid it is not itself reported as a breach.
- **The Reviewer's probe is now a test.** All five shapes are held against the scan by hand, so the class comment's claim is answerable without running a sweep.
- **A companion test holds six innocent lines against it** - `Path.Combine(MachineRoot(), "bin")`, `Path.Combine(Base(), "config")`, `Path.Combine(SharedRoot, "instances", slug)`, and two comments that spell the forbidden shape out. A future widening that starts crying wolf fails here rather than in someone's pull request.
- **The sweep asserts it read more than a hundred files.** An absence found by an instrument that never looked is the failure this class is about.
- **The class comment now says what the scan sees and, plainly, what it does not**: nothing outside C# (the Python `bin()` resolves from LOCALAPPDATA and ignores the setting, so it is machine-root by construction rather than by this guard), no producer missing from the list, no composition split across two statements, no folder name that is not a literal in the same statement.

**Watched failing, twice, each with a real build.** The probe test is not decoration:

- Producers put back to `InstanceHome` alone: `Assert.Equal() Failure: Values differ / Expected: 5 / Actual: 1`. One of the five shapes seen - the Reviewer's probe result to the character.
- Scan put back to a line at a time: `Expected: 5 / Actual: 4`. The wrapped `Path.Combine` is the one that disappears.

Both reverts were restored, the project rebuilt (not `--no-build`), and the tree verified clean against `HEAD` before the gate was run. Both reds are recorded in the test's own comment, so the next reader does not have to take this on trust.

**The gate.** `.\scripts\test-local.ps1` in the worktree, foreground: **eight suites, all `outcome=Completed`, zero failures** (Core.UnitTests 666, Avalonia 550, Engine 63, HostedAgent 88, Launcher 197, Terminal 25, setup 25, setup-engine 614). No suite was killed. The four tests in this class were also run directly - `Failed: 0, Passed: 4` - because `CcDirector.Core.Tests`, where they live, is parked and the default gate does not reach it. The three parked suites remain outstanding for the release gate, unchanged by this commit; Docker is hung on this machine, so no `-Parked` run was attempted.

**What this answer does not cover.** The widening is in the scan only - no product code changed, so nothing about where the tools resolve to has moved. The three blind spots listed in the class comment are real and are now written down rather than papered over; the honest summary is that the guard covers the C# idioms we know how to produce a Director's folder with, and the seven named readers cover the rest.

PR #3102, opened 2026-09-18
feat(tools): the product installs cc-ship and cc-dev-reports

thefrederiksen:
Phase 4 of the mission "one tools folder per machine". The owner's words: "CC Ship and CC Dev Reports
should be installed by a new version of DevThrottle. They should absolutely be included. but not CC
Playwright."

## What changed

`cc-ship` and `cc-dev-reports` are now `"ship": true` in `tools/registry.json`, and everything that
counts the shipped set has been brought into line with them.

**cc-ship could not be shipped as it stood, so it was packaged.** A shipped Python tool must build a
wheel with a console script: the bundle builder runs `uv build --wheel` on each one and reads its
`[project.scripts]`, and the installer writes the bin shim from that. `cc-ship` had no build backend,
no packaging declaration and no console script, and its modules imported each other by bare top-level
names, which only works when its `src` folder is itself on the path. It is now packaged exactly the
way its sibling `cc-dev-reports` already was:

- `src/__init__.py` added, and every import inside `src` made relative, so the same code loads under
  the installed package name `cc_ship` and under `src` when run from a checkout.
- `pyproject.toml` gains the setuptools build backend, `packages = ["cc_ship"]` with
  `package-dir = {"cc_ship" = "src"}`, the console script `cc-ship = "cc_ship.cli:main"`, and the
  dependency on `cc-storage` that `runstore.py` had been satisfying by editing `sys.path`.
- `runstore.py` adds the checkout's `tools` folder to the path only when `tools/cc_storage` is
  actually there, instead of unconditionally.
- `cc-ship --version` added. It is not decoration: `--version` is one of the two health checks the
  Director's Tools page runs against every shipped tool, so a tool that cannot answer it shows as
  broken there.
- `main.py`, the five test files and the three probes follow the same package, so running from a
  checkout is unchanged.

`cc-playwright` is untouched and still does not ship.

**Everything that counts the shipped set:**

- `src/CcDirector.Core/Tools/tools-manifest.json` - both tools added, so the Tools page and the
  in-app health check know them. `cc-ship` gets a `--help` smoke; `cc-dev-reports` declares honestly
  that it has no zero-auth command and is checked on presence and version only.
- `tools/cc-director-setup-engine.Tests/PythonToolsInstallerTests.cs` - the live install test now
  asserts both tools install, shim, and answer `--version` WITH THEIR OWN NAME in the output, and
  that `cc-playwright` did not leak into the bundle.
- `docs/public/tools/01-overview.md` - ten becomes twelve, both rows in the table, a section each.
- `docs/cli-reference.md` - "Not in the installer yet" was true and is now false for both.
- `docs/CC_TOOLS.md` - both added to the fleet table.
- Both tools' READMEs - the hand-made launcher is no longer the way in, and each says why keeping one
  beside the installed tool is two sources for one name.

The bundle builder, the installer, the release workflow, `setup_ops.py` and the legacy-shim purge all
read the registry or the bundle manifest, so they needed no edit - and that was checked by running
them, not assumed.

## Proof

**A test watched failing.** `tools-manifest.json` was edited first, with the registry left alone, and
`ShippedToolsManifestGuardTests.ShipTruePythonTools_MatchCoreHealthManifest` went red:

```
Assert.Equal() Failure: Collections differ
        v (pos 0)
Expected: ["cc-devthrottle", "cc-gmail", "cc-html", "cc-image", "cc-outlook", ...]
Actual:   ["cc-dev-reports", "cc-devthrottle", "cc-gmail", "cc-html", "cc-image", ...]
        ^ (pos 0)
Failed!  - Failed: 1, Passed: 3
```

Expected is the registry's shipped set, Actual is the manifest. Flipping the two registry flags turns
it green: `Passed! - Failed: 0, Passed: 4`.

**A fresh install into a throwaway root, and both tools run.** `scripts/build-python-bundle.ps1` was
run for real: it selected twelve shipped tools and wrote a bundle whose manifest carries `cc-ship`
and `cc-dev-reports`, each with its console script. That bundle was then installed the way the
product installs it - bundled CPython extracted, a venv made from it, `pip install --no-index
--find-links wheelhouse` - into a rig root under the temporary directory. Never the real root.

```
=== cc-ship --version ===         cc-ship 0.1.0            exit=0
=== cc-dev-reports --version ===  cc-dev-reports 0.1.0     exit=0
cc-ship --help exit=0 ; contains the smoke string the manifest declares: True
cc-playwright.exe present in the rig: False
```

`cc-ship status` was also run from that installed copy against this repository and gave its real
answer - `code: no-run`, naming this branch - so the installed tool does actual work, not just print a
version.

The same install runs as a test: `PythonToolsInstallerTests` with `CC_PYBUNDLE_DIR` pointing at the
bundle. `Passed! - Failed: 0, Passed: 1, Duration: 3 m 38 s`.

**The gate.** `.\scripts	est-local.ps1`: eight suites, all `outcome=Completed`, 2,223 tests, zero
failures. Run twice - once after the registry and manifest work, once after the skill change.

**The parked suites.** The gate's coverage-gap line names `CcDirector.Core.Tests`,
`CcDirector.Gateway.Tests` and `CcDirector.Gateway.UnitTests`. `-Parked` could not be run: it needs
Docker, and Docker on this machine did not answer `docker info` within three minutes. The tests in
those suites that this change can reach were run by name instead, and all pass -
`ShippedToolsManifestGuardTests`, `FleetToolsShipGuardTests` and `ToolCatalogServiceTests` (23 tests),
and the Gateway's skill tests `BuiltInSkillsHaveNoDeadDoor`, `SkillStoreTests` and the seeder (61
tests). A full `CcDirector.Core.Tests` run was started and abandoned after fifty minutes at about two
seconds of processor time per ten minutes; it is a parked suite and it was contending with another
session's Gateway run on this machine.

**cc-ship's own suite.** This paragraph first said `170 passed, 1 failed` and named the one failure.
The review of this pull request could not reproduce that exactly, so the suite was measured properly:
three full runs of all 171 tests on this branch's copy and three on `origin/main`'s copy, extracted to
a separate folder, interleaved on the same loaded Windows machine on 18 September 2026.

| Run | This branch | origin/main's copy |
|---|---|---|
| 1 | 169 passed, 2 failed | 169 passed, 2 failed |
| 2 | 170 passed, 1 failed | 169 passed, 2 failed |
| 3 | 168 passed, 3 failed | 168 passed, 3 failed |

`170 passed, 1 failed` is one sample of that, not the suite's settled result, and the two copies behave
the same. Two separate things are going on.

**One deterministic failure, and the stated cause was wrong.**
`test_run_NonAsciiRepositoryPath_BriefsStillWritten` failed in all six runs, on both copies. It is not a
file lock held by git: the test renames the repository folder while that folder is the process's own
current directory, and Windows refuses that outright. Reduced to three lines - `os.chdir(d)` then
`os.rename(d, ...)` - it gives the same `PermissionError: [WinError 32]`, while renaming a folder that is
not the current directory to the same non-ASCII name succeeds. So the test fails before it reaches the
non-ASCII path it is about. Pre-existing, deterministic, nothing to do with packaging.

**A whole family of load-sensitive tests, not one flaky test.** Forty-four of the 171 tests call
`cc-ship wait --seconds 5`, and `engine.wait` treats that as a real wall-clock deadline. On a busy
machine the five seconds run out before the scripted run reaches its end state and the test fails with
`assert 'working' == 'merged'` (or whichever state it expected). Six runs turned up **six different**
members of that family, none in one run and two in another, on both copies:
`test_run_CommitDuringReview_WholeRoundRunsAgainOnTheNewHead` (twice),
`test_run_WebsiteApiChange_RatedHighAndParked`,
`test_run_InvalidReviewFromReapedSessionsThreeTimes_FailsAfterTwo`,
`test_risk_AuthenticationCodeNotListedByRepository_StillHigh`,
`test_run_VercelPreview_WaitsForItThenRemovesTheCookieFile` and `test_run_FailedCheck_RefusesMerge` -
a seventh, `test_run_InvalidVerifyFromReapedVerifier_FreshVerifierWithFreshCookie`, was seen by the
review. Re-running one on its own does not settle it: alone, eight times, four on each copy, it passed
four and failed four. The margin is thin by construction - the test gives the engine five seconds while
its own call needs eight to thirty-five seconds of real git work in the same measurements.

Neither belongs to this change. This pull request's edits to `engine.py` and to the five test files are
import rewrites only, with no timing and no logic touched, which is why the two copies measure the same.
Neither is fixed here either, and no test was skipped, quietened or retried to tidy the number. What is
recorded instead is a new **Its own tests** section in `tools/cc-ship/README.md`, where the next person
to run the suite will find both facts and the failure signature that tells them apart.

**cc-dev-reports' suite:** `28 passed`. **The shipped-tool contract guard**, which now walks both tools'
sources: `52 passed`.

## Not proven

- macOS and Linux. `scripts/build-python-bundle.sh` reads the registry the same way and was read, not
  run.
- The Tools page rendering the two new rows in a running Director.
- The full `CcDirector.Core.Tests`, `CcDirector.Gateway.Tests` and `CcDirector.Gateway.UnitTests`
  suites, for the reason above. Only the tests in them that this change can reach were run.
- The skill change reaches agents only when the Gateway is deployed. This pull request does not
  deploy it.


thefrederiksen:
## Review finding 1, and the answer

**The finding.** The review of this phase could not reproduce the result this pull request stated for
cc-ship's own suite. The body said `170 passed, 1 failed` and named the one failure as
`test_run_NonAsciiRepositoryPath_BriefsStillWritten`. In one of the review's own runs a different test
failed instead - `test_run_InvalidVerifyFromReapedVerifier_FreshVerifierWithFreshCookie` - and it passed
in three other runs on either copy. So the suite has at least one load-sensitive test and the stated
result is not a stable description of it. The finding asked the seat that built the work to accept it
or decline it with the reason.

**Accepted, and it turns out to be wider than one test.** The suite was measured rather than sampled:
three full runs of all 171 tests on this branch's copy and three on `origin/main`'s copy, extracted to a
separate folder with `tools/cc_storage` beside it so it resolves the way a checkout does. The runs were
interleaved on the same loaded Windows machine so that drifting load could not favour one copy.

| Run | This branch | origin/main's copy |
|---|---|---|
| 1 | 169 passed, 2 failed | 169 passed, 2 failed |
| 2 | 170 passed, 1 failed | 169 passed, 2 failed |
| 3 | 168 passed, 3 failed | 168 passed, 3 failed |

Two distinct things are in there, and neither is a single flaky test.

**1. One deterministic failure, whose stated cause was wrong.**
`test_run_NonAsciiRepositoryPath_BriefsStillWritten` failed in six runs out of six, on both copies. This
pull request called it a Windows file lock on a directory rename, which reads as incidental. It is not:
the test renames the repository folder while that folder is the process's own current directory, and
Windows refuses that outright. Reduced to three lines - `os.chdir(d)` then `os.rename(d, ...)` - it
reproduces the same `PermissionError: [WinError 32]`, while renaming a folder that is *not* the current
directory to the same non-ASCII name succeeds. The test therefore fails before it reaches the non-ASCII
path it exists to check. Deterministic, pre-existing, present identically on `origin/main`.

**2. A family of forty-four tests on one five-second wall clock.** Forty-four of the 171 tests call
`cc-ship wait --seconds 5`, and `engine.wait` treats that as a real deadline. When the machine is busy
the five seconds expire before the scripted run reaches its end state, and the test fails with
`assert 'working' == 'merged'` - or whichever state it expected. Six runs turned up **six different**
members of that family, none in one run and two in another, on both copies:

- `test_run_CommitDuringReview_WholeRoundRunsAgainOnTheNewHead` (twice)
- `test_run_WebsiteApiChange_RatedHighAndParked`
- `test_run_InvalidReviewFromReapedSessionsThreeTimes_FailsAfterTwo`
- `test_risk_AuthenticationCodeNotListedByRepository_StillHigh`
- `test_run_VercelPreview_WaitsForItThenRemovesTheCookieFile`
- `test_run_FailedCheck_RefusesMerge`

The review's own sighting is a seventh. Re-running one alone does not settle it, because the cause is
the machine rather than its neighbours: one of them run on its own eight times, four on each copy,
passed four and failed four. The margin is thin by construction - the test allows the engine five
seconds while the test's own call needs eight to thirty-five seconds of real git work in the same
measurements.

**Not caused by this change, and that is measured rather than argued.** The edits here to `engine.py`
and to the five test files are import rewrites only - no timing, no logic - and the two copies produce
the same distribution of failures.

**What changed as a result.** The pull request body no longer states `170 passed, 1 failed` as the
suite's result; it carries the table above and names both causes. `tools/cc-ship/README.md` gains an
**Its own tests** section recording the same two facts, because the next person to run this suite reads
the tool's README and not this thread - including the discriminator, that a failure with `'working'` on
the left is the clock and not the thing under test.

**What deliberately did not change.** No test was skipped, marked expected, retry-wrapped or quietened,
and neither defect was fixed here. They belong to cc-ship's own suite, they are on `origin/main` today,
and repairing them inside a packaging change would mix two jobs and remove the very evidence that they
are pre-existing. They are worth their own work item: the first is a two-line fix, the second is a
design question about giving those tests a budget that is not a wall clock.

**The gate.** `.\scripts\test-local.ps1` in this worktree: eight suites, every one `outcome=Completed`,
2,223 tests, zero failures - no failure outside the declared baseline list. `-Parked` was not run:
Docker is hung on this machine, the same declared gap as before, and this commit changes one markdown
file.

PR #3112, opened 2026-09-19
The per-session terminal log is off by default

thefrederiksen:
## What this changes

The per-session terminal log is **off by default**. A new `SessionLogConfig` reads
`"session_logs": { "enabled": true }` from config.json, with `CC_DIRECTOR_SESSION_LOGS`
overriding in both directions, and `SessionLogManager.Start()` returns early when it is off -
subscribing to nothing and opening no file, so the four streams are absent rather than empty.

## Why

It ran for every session on every install, with no setting to turn it off.

The expensive stream is `raw.jsonl`: every chunk of bytes the terminal emitted, base64-encoded.
That is not a record of what the agent said, it is a record of what the screen **painted**, so every
spinner frame, progress bar and status-line repaint was captured, inflated by a third by base64, and
wrapped in JSON with a timestamp per chunk. One Codex session measured 1.1 GB. One machine measured
35 GB across 3,422 sessions, 99.6% of them belonging to sessions that no longer existed.

**Nothing in the product ever read a byte of it.** The only readers anywhere are the unit tests and
the `raw-view-harness` developer tool. The streams were raw material for work that was never
finished: `agent-view.jsonl` is described in its own writer as a "slot reserved for a follow-up
slice", and the bytes were captured, in `ProcessHost`'s words, "so we could". The terminal evidence
the product actually reads is the wingman's bounded per-turn `ScreenTail`, which the Gateway holds.

This is the same defect and the same remedy as `SessionRecordingConfig` - the terminal screen
recorder, written up as "an internal engineering corpus, collected from every install, invisible in
the product, with no age limit" and turned off by default. That one was fixed and this one was
missed. The new config record, the setting shape, the environment override and the tests all follow
its pattern deliberately.

## Proof

- `SessionLogPolicyTests` holds the default, the switch, the override in both directions, and a
  nonsense override throwing rather than guessing.
- The two that matter watch the **manager** - the thing the Director actually calls - not the config
  record: one asserts nothing lands on disk for a new session, and its **twin** asserts the identical
  arrangement DOES write with the setting on. The absence is only evidence because the instrument is
  shown capable of reading a write.
- Revert proof: with the gate removed and rebuilt, `With_logging_off_a_new_session_writes_nothing_to_disk`
  fails and its twin still passes; restored and rebuilt, both pass. Committed before mutating, and no
  `--no-build` on the restore run.
- Local gate green: 2,202 tests, every suite `outcome=Completed`.

## Coverage this does NOT have

The default gate flagged a coverage gap for `CcDirector.Core.Tests` (where the new tests live),
`CcDirector.Gateway.Tests` and `CcDirector.Gateway.UnitTests`, all parked. The new tests and the
existing `SessionLogWriterTests` were run directly and pass (11 tests). The full parked `Core.Tests`
run was still going at the time of writing; the release gate runs `-Parked -Configuration Release` on
merged main before the tag.

## What this does not do

It does not touch existing `session-logs` directories, here or on any install. It changes what is
collected from now on, not what is already on disk. Clearing this machine's 35 GB was done separately
by hand.


thefrederiksen:
## Gate, and what it does not cover

**Ran, green:**
- Default local gate, twice (before and after the review fix): 2,202 tests, every suite `outcome=Completed`.
- **Parked `CcDirector.Core.Tests` in full** - the suite that holds the new tests and the code this
  touches: 4,490 passed, 8 skipped, 0 failed, 26 minutes.
- The session-log tests directly, including the revert proof: remove the gate and rebuild and
  `With_logging_off_a_new_session_writes_nothing_to_disk` fails while its twin still passes; restore
  and rebuild and both pass.

**Ran, with unrelated failures:**
- Parked `CcDirector.Gateway.UnitTests`: 6,379 passed, 2 skipped, **8 failed**. Every failure is in
  `HostedSchemaRefusesAnUnownedRowTests` and every one is
  `Npgsql.PostgresException : 28P01: password authentication failed for user "postgres"` - a
  connection failure before any test logic, in the stats schema's tenant-ownership proofs. Those are
  the PostgreSQL-backed tests that only run correctly through `-Parked`, which builds its own
  throwaway database. Nothing in this change touches the Gateway, a connection string, or that
  schema.

**Not run, and why:**
- `-Parked` in full, and `CcDirector.Gateway.Tests`, need Docker. Docker Desktop's process is up on
  this machine but the engine does not answer - `docker ps` times out at 90 seconds. Rather than let
  those proofs report as skipped, which reads identically to a pass, this says plainly that they did
  not run.

The release gate (`-Parked -Configuration Release` on merged main at the commit about to be tagged)
still has to run before any tag, and that needs the same Docker.

## Review

Reviewed by a different agent family (Copilot), after Codex, Gemini and Grok were each unavailable
for account reasons. It confirmed independently that no production code reads any of the four
streams, and raised one finding: `SessionLogManager`'s class comment named the `TurnSummaryCache`,
the wingman and a future agent-view pipeline as consumers pushing records through this manager.

Verified and true: `WriteTurnSummary` and `WriteAgentViewWidget` have **no callers anywhere** in the
product - the only calls are tests invoking the writer directly - so `turns.jsonl` and
`agent-view.jsonl` were never written at all. Fixed in d4ceeed64.

PR #3116, opened 2026-09-19
One tools folder per machine: the master owns the path, repaired at every Director start

thefrederiksen:
Phase 2 of the "one tools folder per machine" mission. Phase 1 put the resolver in place; this turns
the path repair the right way round and makes it run by itself.

## What was wrong

`FleetToolPathRepair` had the world inverted. It called the machine root's tools folder the "legacy
flat bin" that the move to per-Director folders had superseded, called a Director's own folder ours,
and deliberately kept every other Director's copy on the path because removing it "would be sabotage
dressed as hygiene". Its only caller was the Fix button on the Tools page, so nothing ran it at
Director start. The installer's path step added its own folder and removed nothing.

Nothing ever made a Director's folder an install root. Every Director points the `CC_DIRECTOR_ROOT`
setting at its own folder and hands it to every session, so an installer or a tools repair run from
inside one took that folder for the whole machine and filled it with a complete copy of the tools.
One computer ended up with seven copies, and whichever one happened to be first on the search path
answered.

## What this does

**The master owns the path.** The master is the machine root's `bin`: one installed copy per machine,
which every Director uses. A tools folder inside a Director's own folder of the same root is a copy
the master replaces and comes off the path - the default Director's, a named Director's, the nested
`instances\default\instances\default` the leak produced, and one whose folder has already been
deleted while its entry sits there waiting for something to recreate it.

**It will not reach outside that.** A DevThrottle tools folder belonging to a different machine root -
a rig serving its own throwaway root, a second install elsewhere on the disk - stays. A rig keeping
its own tools is what makes a rig safe to run at all. The name test is narrow on purpose: only `bin`,
`pyenv` and `python` inside a Director's folder qualify, so a bystander folder of the user's from
inside one is left alone.

**It runs at every Director start**, before the session manager is constructed - a session inherits
the Director's running path, so repairing it later would leave every session of the run on whatever
the Director was launched with. Both paths are put right: the saved one and this process's own. The
Fix button keeps working.

**A Director serving a throwaway root never writes the saved path.** The saved path is permanent
machine state and belongs to the install at the machine's default root. A rig, a wizard harness or an
unpacked bundle that wrote itself there would outlive its own directory by months - there is an entry
on the machine that prompted this work pointing at a harness root deleted in July. A rig still repairs
its own running path, and so does the button, which now obeys the same guard the start does.

**The installer's path step uses the same rule.** Appending the master while a Director's copy sits in
front of it puts the files in the right place and changes nothing about which copy answers a command.

Nothing on disk is touched. A copy that is off the path but still there can no longer answer a command
- that is the intended state between this phase and the deletion in the next one.

## Proof

**Tests on the rewrite rule** (`FleetToolPathRepairTests`, 38 tests) with folders that exist and
folders that have vanished, each of the four copies, a rig's own root, another install's master, a
bystander folder, the raw-versus-expanded path handling, and all of them at once ending with exactly
one entry. Plus `ThePathRuleRunsWhereItMustTests`, which names the two callers no unit test can reach
and fails when either stops asking for the rule.

**A real Director start** - `scripts\one-tool-path-rig-proof.ps1`, committed beside the code. It builds
a throwaway root shaped like the machine that prompted this (master, default Director's copy, named
Director's copy, the nested copy, a copy already deleted from disk, and a bystander), starts a slot 5
Director on it through a scheduled task with a crafted process path carrying all six, reads what the
Director reports about its own path, and stops it with its named signal. Green: the running path ended
with exactly one DevThrottle tools entry and it was the master, all four copies named as removed, the
bystander untouched, and the real user's saved path byte-for-byte unchanged.

**Each fix reverted once and watched fail:**

| Fix reverted | What went red, and what it said |
|---|---|
| The rule turned back the way it was (a Director's own copy kept) | Six tests. `Rewrite_EveryCopyAtOnce_LeavesExactlyOneDevThrottleEntry`: expected `C:\cc-director\bin;C:\windows`, actual still carried `C:\cc-director\instances...` - the copies were still on the path. The 36 controls stayed green, including the rig-keeps-its-own-tools case. |
| The Director's start-time call removed | Both guard tests: "App.axaml.cs no longer calls the start-time tool path repair at all." Then the rig proof itself, rebuilt and re-run: the Director started and logged 282 lines of normal startup with no `Tool path at start` line at all, and nothing was repaired. |
| The installer's path step put back to a plain append | `Every_caller_of_the_path_rule_still_asks_for_it`: "tools/cc-director-setup-engine/InstallFinalizer.cs: expected to find `FleetToolPathRepair.RewriteForInstall(current, layout.BinDir)`". |

Each was restored with a full rebuild afterwards, never a `--no-build` run, and re-run green.

**The local gate**, in this worktree, on this branch rebased onto `6edf77f44`: eight suites, 2,202
tests, every one `outcome=Completed`, zero failures.

**The parked `CcDirector.Core.Tests`**, run separately because the gate does not include it and it is
where the tests above live: 4,500 passed, 8 skipped, 1 failed - and that one is dealt with below
rather than waved through.

## The one failure, and what is actually known about it

`ContentTurnRuleTests.A_byte_inside_the_window_pushes_the_check_out_rather_than_being_dropped` failed
once in a full parallel run of `Core.Tests` on this branch: expected 129 bytes in one shadow row, got
40, meaning the settle check fired part-way through the first burst. It is outside the nine known
failures in `BASELINE.md`, so it does not get called pre-existing on the strength of an argument.

What was measured:

- The same test class alone on this branch: **25 of 25 pass.**
- A full `Core.Tests` run on a clean `origin/main` worktree cut for the purpose: **0 failed, 4,483
  passed, 21m29s.** It did not reproduce.
- The failing assertion is a settle deadline with a 40-millisecond margin, in turn detection, last
  touched in August. This change touches the search path and Director startup; the test constructs a
  detector directly and never starts the application, so the start-time repair cannot run in its
  process. The only `CcStorage` member it uses is `TurnDetectionShadow()`, which is unchanged - the
  only `CcStorage` change here is a new `DefaultRoot()` method that nothing else calls.

What is NOT known: a second full run on this branch, which is the measurement that would separate "a
flake that happened to land here" from "something this branch causes", was **stopped by the harness
because the machine was critically low on memory**, and the instruction on that stop is not to restart
it unasked. So the position is one failing run and one clean isolated run on this branch against one
clean full run on main - suggestive, not conclusive. Whoever merges should re-run
`Core.Tests` on a quiet machine and, if it fails again, treat it as this branch's problem until shown
otherwise.

## What is NOT proven

- **The saved-path write itself.** The saved path is one per-user registry value on this machine and
  the rig must never write it, so nothing here exercises the registry write. What it decides to write
  is the same `Rewrite` the tests cover; the two lines that read and write the registry are covered by
  nothing, exactly as before this change.
- **The Fix button's new guard, end to end.** It is the same `OwnsTheSavedPath` rule the start uses,
  with four tests on that rule, but no rig run presses the button. The rig's green run was against a
  tree whose `App.axaml.cs`, `RepairAtDirectorStart`, `Rewrite` and `CcStorage` are byte-identical to
  this one (verified by diff); the only delta since is inside `PutFirstOnPath`, which the rig never
  calls.
- **`-Parked`.** Docker is hung on this machine, so the two PostgreSQL-backed suites
  (`CcDirector.Gateway.Tests`, `CcDirector.Gateway.UnitTests`) did not run and stay outstanding for
  the release gate.
- **macOS and Linux.** The rule is written for all three and covered by the suites; nobody has watched
  it run on a Mac or a Linux box. On those the saved path is the shell profile's and is left alone, as
  before.
- **That the start-time call is reached at run time in source order.** The order test reads source
  text, which is not execution order. The rig proof is the end-to-end answer, and it shows only that
  the repair happened before the Director opened any session, because no session existed.


thefrederiksen:
Answers to the phase 2 review's two findings, plus the evidence for the two jobs that carried no evidence yet.

**Finding 1 - the CC_DIRECTOR_ROOT race in the setup-engine tests.** Accepted and fixed in `52051f4de`: `SelfUpdateIsolationTests` joined the collection that serializes every class that sets the process-wide `CC_DIRECTOR_ROOT`, so it can no longer run beside `InstallLayoutMachineRootTests`. Said plainly: I did not watch the race fail, and a fix for a race you cannot watch fail is a fix you reason about rather than observe. What I can report is that the mechanism is the one the review prescribed, and the default gate in this worktree is green twice (once as the default run, once inside the parked release run described below), with no cross-class root failure in either.

**Finding 2 - the rig proof's stop path was a force-kill while claiming a named signal.** Accepted and fixed in `acda782af`. The mechanism the previous half-finished edit proposed was verified against the product code first: the Director writes `Lifecycle signals listening for directorId=<id>` (App.axaml.cs, `StartLifecycleSignals`) only after both named signals are listening, through the same file log the proof already reads. The instance registration is written later, on a background task, which is why no run ever found it in time. The script now reads the identifier from that log line, and a new check fails the whole proof unless the Director was stopped by its named signal. Proven end to end: a fresh slot 5 build of this branch's head, a fresh throwaway rig root, launched through the scheduled task - the Director reported its identifier, the signal was opened and set, and the Director exited on its own: "stopped by its named signal". All twelve checks passed, including the new one. The force-kill branch still exists as the last resort and says so when it fires; it did not fire.

**The guards move, with the counts.** The five moved classes - `MachineRootTests`, `ToolReadersUseTheMachineRootTests`, `ToolPathsNeverComeFromADirectorFolderTests`, `FleetToolPathRepairTests`, `ThePathRuleRunsWhereItMustTests` - all appear in a default gate run now. `CcDirector.Core.UnitTests` was 640 tests in the review's own pre-move green run and is 706 in mine (the moved classes contribute 66 tests once theories expand to cases). Whole default gate: 2,268 tests, every suite completed, zero failures.

**The parked release gate, run in full.** `.\scripts\test-local.ps1 -Parked -Configuration Release` with Docker answering (29.6.1): the default eight suites green; `CcDirector.Core.Tests` green (4,442 passed, 8 skipped, 0 failed); `CcDirector.Gateway.UnitTests` green (6,387 passed, 2 skipped, 0 failed - none of the nine named pre-existing baseline failures fired); `CcDirector.Gateway.Tests` red with four failures, none of them on the baseline list. Those four were then run alone (the baseline's own rule: run a database-backed failure alone before calling it anything) and they fail alone, identically. They were then run on a clean worktree of this branch's base commit `70bdb6603` - before any of this branch's work - and the same four fail identically there, so they are pre-existing on main and not this branch's damage. Two are migration pin tests that assert their migration is the last in the set, broken when main gained a newer migration on 18 September; one is a collation proof whose expected natural-key list no longer matches the schema; one asserts a session owned by another session is offered no change of owner, broken by the take-over feature that landed on 18 September. Also said plainly: in the full parked run the `CcDirector.Gateway.Tests` suite stopped without printing a summary line (four failures logged before it stopped); the isolated re-run completed normally with the same four failures. Nothing was skipped silently - the gate reported the failure loudly.

So the parked release gate is red by the letter of the bar (failures outside the baseline list), and the red is main's, not this pull request's. Whoever merges should expect those four to be red until main fixes them; they are a main defect with its own owner to find, not a wave-through.

Issue #3096, opened 2026-09-18
The user can tell a session to TAKE ownership of another session - and a session may never be put under a third session


thefrederiksen:
A session may already RELEASE a session it owns to the user (#3086, merged). The other direction is
still refused to every session, and it should not be: the user must be able to tell a session to
TAKE a session over.

## The owner's ruling, 18 September 2026

> "The user should be allowed to move sessions around. If I command a session to take over another
> session, that should be allowed if it comes with user approval. But it shouldn't be possible to
> tell a session to go in underneath another session. A session can be directed by the user to take
> ownership. That's what we should implement."

So there are two changes of owner a session may make, and one it may never make:

- **TAKE, to itself** - allowed, on the user's direction. New.
- **RELEASE, to the user** - allowed, on its own. Already built (#3086).
- **PUT A SESSION UNDER A THIRD SESSION, or put ITSELF under another session** - never, however it is
  asked, and whoever asks.

**The one sentence that covers all three: the only owner a session may ever name is ITSELF.** It is
the rule `spawn` already enforces for a session it starts - a session may only name itself as the
owner of what it spawns - applied to sessions that are already running. Pull toward me, yes. Push
toward somebody else, never.

**Why the asymmetry is not arbitrary.** Releasing moves work toward the person, which is where
everything lands by default, so it is safe unasked. Taking moves work AWAY from the person: a session
that is taken stops going red for him and reports to the session that took it. That is a real loss of
attention, which is why it needs his direction - and why the session that took it must be the one
that answers for it. A three-party move has nobody to answer for it: the session that made the move
is not the one the work now reports to, and the session that receives it never agreed to hold it.

## What exists today, on main

- `POST /gateway/fleet-manager/hand-over` takes `{ "session", "to" }` and `to` is one of exactly two
  values, `fleet-manager` or `owner` (`SessionOwnerChangeDto.Directions`). There is no way to say
  "to me", and no way to name a session id as the destination - so the forbidden three-party move is
  currently unexpressible, and must STAY unexpressible.
- A session key is allowed through for exactly two things: the account's marked Fleet Manager doing
  what the owner asked, and any session releasing a session it owns.
- The owner's own phone or browser may hand a session to the Fleet Manager, and hand back a session
  THE FLEET MANAGER owns - but it is refused a session any other session owns
  ("... is owned by session X, not by the Fleet Manager. Only the Fleet Manager's sessions are handed
  back here").

[...]

Issue #3051, opened 2026-09-18
The test signal is not usable: two runs of one commit disagreed, main has not been green since 16:11, and the database proofs never run in continuous integration


thefrederiksen:
## Why this exists

Over one evening (17 September 2026) I ran the full test gate twice on identical code while finishing an
unrelated piece of work. The two runs disagreed with each other, the failures they did report were mostly not
real, and the ones that were real had been red on main for hours without anyone acting. The signal is not
currently usable: green means little and red means "go and look it up" rather than "something broke".

This is the state-of-the-signal report. Two existing issues, #2915 and #2878, are single instances of one
symptom below; this one is the whole picture and the measurements behind it.

### 1. The gate gives different answers for the same code

`.\scripts\test-local.ps1 -Parked` run twice on the same commit, nothing else changed:

| | Run 1 | Run 2 |
|---|---|---|
| Failures | 2 | 4 |
| `Gateway.Tests` | `PostgresProviderProofTests.Collation_ExplicitC_OnExactlyTheDeclaredNaturalKeys_OnRealPostgres` | same |
| `Gateway.UnitTests` | `TurnsVerbUnresolvedTranscriptTests...(agent: Grok)` | same, plus `AuthMiddlewareTests.Bearer_with_a_valid_device_key_is_accepted` |
| `Core.Tests` | none | `ContentTurnRuleTests.A_byte_inside_the_window_pushes_the_check_out_rather_than_being_dropped` |

Both of the run-2-only failures pass when their class is run alone: 27 of 27 and 25 of 25.

The `AuthMiddleware` one shows the mechanism. It does not fail in the code under test at all - it dies in the
test's own setup building a `DeviceRegistry`:

```
System.InvalidOperationException : The Gateway SQLite database at
'...\cc-authmw-4667959ae0e8432a83a7c007b92581e9.json.gateway.db' could not be opened or migrated:
Cannot access a disposed object. Object name: 'SQLitePCL.sqlite3'.
```

A process-global SQLite handle was disposed by a different test running in parallel. The test never reached
the thing it exists to check.

### 2. Continuous integration on main has not been green since 16:11 on 17 September

Of the last 60 runs of `ci.yml` on `main`: **19 success, 9 failure, 31 cancelled.** Of the most recent 12:
**zero successes** - 4 failures and 7 cancelled.

The three tests failing in run
[35270394665](https://github.com/thefrederiksen/devthrottle/actions/runs/35270394665):

[...]

PR #3125, opened 2026-09-19
The four proof tests the parked Gateway suite was red with, caught up to the schema they prove

thefrederiksen:
## What this is

The release gate for v2.7.1 (the full parked set, Release configuration) ran `CcDirector.Gateway.Tests` to completion for the first time since 17 September and found **four stale test expectations — all already on main, none caused by the session-log change** the release carries. The suite is parked in the default gate and never runs in continuous integration (issue #3051), so nothing else had looked at them.

All four are tests pinned to a moment that the product correctly moved past, without the test being updated:

1. **`PostgresProviderProofTests` collation census** — the nine `fleet_manager_events`, `fleet_manager_owned_sessions` and `fleet_outcomes` columns that the 17 September Fleet Manager migrations declare with collation C were missing from the census. Issue #3051 had already documented this test as permanently red on main.
2. **`TurnVerdictAnswerChoicePostgresTests`** — pinned its migration as second from the end of the migration set; `AddWingmanNarrationCallTrace` (#3105, 18 September) moved it to third.
3. **`FleetOutcomeStopIdentityPostgresTests`** — pinned its migration as the last in the set, in two places; the same migration moved it to second from the end.
4. **`FleetManagerRoutesHostTests`** — expected a session owned by another live session to offer no change of owner; the hand-over feature (#3096 / #3101) deliberately added the `Hand back to me` undo to exactly that row, and this host test was missed when the fold's own tests were updated.

**No product code is changed.** Each fix updates the expectation to the behaviour the product already has, with a comment naming the change that moved it — the same convention the census file itself uses for every earlier catch-up.

## Why this is a pull request now

The release gate is red on main and a red gate cannot be tagged around: this unblocks the v2.7.1 release, whose only outstanding step was one clean full parked gate.

## Checks

- All 51 tests of the four repaired classes pass against a throwaway PostgreSQL (`pg-stats-proof-rig.ps1`, instance `repair1`).
- The default gate is green on this branch, every suite `outcome=Completed`.
- The full parked `CcDirector.Gateway.Tests` suite runs as the release gate on the merged commit, which is the definitive run for it — that is the plan immediately after this merges.
- Why the parked suites did not run on the branch: the definitive full parked run happens on the merged commit as the release gate; running it twice would cost two hours for no additional signal over the four-class run against the real database.


thefrederiksen:
## Gate, and what it covers

**Ran, green:**
- All 51 tests of the four repaired classes against a real throwaway PostgreSQL (built with `scripts/pg-stats-proof-rig.ps1`, instance `repair1`): `PostgresProviderProofTests`, `TurnVerdictAnswerChoicePostgresTests`, `FleetOutcomeStopIdentityPostgresTests`, `FleetManagerRoutesHostTests` - 0 failed, 0 skipped.
- The collation census verified a second way: the live migrated database carries exactly 51 columns with collation C, and the census list now has exactly 51 entries, matching one-for-one.
- Default local gate on this branch: every suite `outcome=Completed`.

**Not run on this branch, and why:**
- The full parked `CcDirector.Gateway.Tests` suite. It serializes machine-wide and takes over an hour; the definitive run is the release gate itself (`-Parked -Configuration Release`), which runs immediately after this merges on the exact commit about to be tagged - the same suite, the same configuration, and the answer that matters for the release. Running it twice buys no signal over the four-class run against the real database above.

## Where the four failures came from

The v2.7.1 release gate ran `CcDirector.Gateway.Tests` to completion for the first time since 17 September and found these four red - all stale expectations on main, all predating the session-log change the release carries, all invisible to continuous integration because that suite never runs there (issue #3051 documents one of them as permanently red). Each fix updates the expectation to behaviour the product already has, with a comment naming the change that moved it. No product code is changed.

## Review

Reviewed by a different agent family (Copilot), as a tracked session reading the branch. Verdict: approve. Its evidence, verified from the source rather than from the brief: the collation census now matches the `UseCollation("C")` declarations in `GatewayDbContext` one-for-one and the migration files carry the same columns with `collation: "C"`; the migration-order pins assert the real sequence ending at `20260918181205_AddWingmanNarrationCallTrace`; the owner-change expectation matches the fold's `TakeBack()` rule for a session any other live session owns (issue #3096); and no proof was deleted or weakened - the assertions are more precise, not looser.

thefrederiksen:
## Second review, by a separate seat, after the release shipped

Approve. A takeover seat re-reviewed this change and the v2.7.1 close-out around it, adversarially and independently, reading only `origin/main` and the run artifacts on disk rather than the earlier reports. Nothing was weakened, every updated expectation matches behaviour the product already has, and the release was cut on the commit the gate actually ran.

### The four expectations, each judged against the product

**The collation census (`PostgresProviderProofTests`).** Complete and in order. `GatewayDbContext` on `origin/main` declares `UseCollation("C")` 51 times; the census list now holds exactly 51 entries, and they map one-to-one onto those declarations with nothing left over on either side. The order matches the query it is compared against (`ORDER BY c.relname, a.attname`) at every one of the nine new entries. The nine are truthful about the schema, not only about the model: `20260917110009_AddFleetManagerEvents` creates `Kind`, `SessionId`, `VerdictId` and `DeliveredTo` with `collation: "C"`, `20260917110409_AddFleetManagerEventOutcomeAnswer` adds `OutcomeId` the same way, `20260917110109_AddFleetManagerEventDelivery` creates the `fleet_manager_owned_sessions` pair, and `20260917090009_AddFleetManagerOutcomes` creates `fleet_outcomes.Kind` and `Status`. The census is the side that had gone stale; catching it up is what it is for.

**The two migration-position pins.** They match the real set. There are 47 Postgres migrations on `origin/main` and `20260918181205_AddWingmanNarrationCallTrace` is the last of them, so second-from-the-end and third-from-the-end are the true positions for `AddFleetOutcomeStopIdentity` and `AddTurnVerdictAnswerChoice`. The assertion that actually carries each proof, that the migration under test follows `MigrationBefore`, is untouched.

**The change of owner (`FleetManagerRoutesHostTests`).** This is the one edit that flips an expectation from "nothing is offered" to "something is offered", so it was checked hardest. It is correct. `FleetManagerRosterFold.OwnerChangeFor` returns `TakeBack()` for any session a live session owns, and `TakeBack()` is `To = owner`, `Label = "Hand back to me"` - the two values the test now asserts. That branch is deliberate and documented in the fold as the owner's undo for a taking (issue 3096), on the reasoning that a taking feature whose undo is missing is a trap. In the fixture `_nextFleetManagerId` is pushed as a live session and `_ownedByNextId` is controlled by it, so the row genuinely reaches that branch. The old expectation was the stale one, and no safety property was given up, because the fold never offers a hand to the Fleet Manager for such a row.

### The release gate, verified from its own artifacts

The gate run was found on disk and read directly rather than taken from any report: `cc-test-local-70efaaf7`. It ran in Release configuration out of the worktree `devthrottle-session-log-off`, and the Gateway suite lock names its holder as this session. All eleven suites report green, including the three parked ones, and `CcDirector.Gateway.Tests` reports `Failed: 0, Passed: 2670, Skipped: 8, Total: 2678` over one hour six minutes. It finished at 15:37:17 local and the tag was pushed about two minutes later, so the gate preceded the tag rather than the other way round.

Two things were checked that a green summary alone would not settle.

**The gate ran on the merged commit, not the branch head.** Those two trees are not the same: the branch was cut before pull request 3123 landed, so its tree lacks that change. The gate result files contain all fourteen tests pull request 3123 introduced (`TerminalLineWrapTests` in `Core.Tests`, `TerminalWrappedUrlTests` in `Terminal.Avalonia.Tests`) while also passing the four repaired classes, and only the merged tree satisfies both of those at once.

**The repaired tests were executed, not skipped.** All six `PostgresProviderProofTests`, the one `TurnVerdictAnswerChoicePostgresTests`, the one `FleetOutcomeStopIdentityPostgresTests` and all forty-three `FleetManagerRoutesHostTests` record outcome `Passed`. The eight not executed are the pre-existing environment-gated `DoorbellEndToEndProof` and `GatewayDatabaseLivePostgresProofTests`, unrelated to this change. A skipped database proof would have been indistinguishable from a pass in the summary line, so this was read per test.

A second run that finished a minute earlier, `cc-test-local-6889f06c`, holds a `CcDirector.Gateway.Tests` result file with zero tests in it. That file belongs to an unrelated Debug run out of `devthrottle-usage-limit-signatures` which queued behind this gate's lock and was aborted. It is not this gate, and it was not mistaken for it.

### The release itself

Tag `v2.7.1` is an annotated tag on `729a6afbe`. `Directory.Build.props` reads `2.7.1` and `docs/public/release-notes/v2.7.1.md` exists, both from `6edf77f44`, which is an ancestor of the tagged commit. The release is published rather than draft, with the full asset set and a manifest reading version `2.7.1`. Correctly cut.

### Findings, none of them blocking

**1. The position pins will go red again at the next migration.** `FleetOutcomeStopIdentityPostgresTests` now asserts that `AddWingmanNarrationCallTrace` is the last migration and that `index == all.Count - 2`; `TurnVerdictAnswerChoicePostgresTests` asserts `all.Count == index + 3`. Neither adds anything to what these tests prove, which is that the migration applies from its predecessor and preserves a row filed before it - and that rests on `all[index - 1] == MigrationBefore`, which is sound. The tail assertions only record that nothing has been added since, and they are the reason this suite was red in the first place. This change caught the values up but kept the shape, so the next migration re-reds both, in a parked suite that surfaces at a release gate, which is the most expensive moment to find it. Worth a small follow-up: compare the fully migrated database against the set's own last entry rather than a written-out name, and drop the count arithmetic.

**2. A count in the census comment contradicts the list beneath it.** The comment still reads "27 declarations, 27 columns - verified one-to-one against the model on 2026-09-02" above a list of 51. This predates the change, since it read 27 when the list held 42, but this was the moment to correct it, in a test whose whole purpose is that the count is consciously acknowledged.

**3. Two provenance sentences in the new comments are imprecise.** `fleet_outcomes.Kind` and `Status` are described as coming from "the same 17 September migrations as the two groups above", meaning the eleven o'clock ones; they come from `20260917090009_AddFleetManagerOutcomes`. And the `fleet_manager_events` group cites `20260917110109` among its sources, which contributes the `fleet_manager_owned_sessions` pair rather than any `fleet_manager_events` column. Same day and same mission, so the conclusion holds, but a reader tracing a column would be sent to the wrong file.

**4. The release notes cover one of the three product changes in the tag.** Between the version bump at `6edf77f44` and the tagged commit sit pull request 3102, which makes the product install `cc-ship` and `cc-dev-reports`, and pull request 3123, the terminal wrapped-url change. `docs/public/release-notes/v2.7.1.md` is written solely about the session terminal log. Nothing to do about a pushed tag; noted for the next one.

**5. Continuous integration on the tagged commit was cancelled, not passed.** The run on `729a6afbe` was superseded four minutes later by `ec6793e96`, so the aggregate check on that commit reads as a failure. The run on main's head is green. The local gate is the authority here by design, so this is not a defect, but "continuous integration is green" is true of main's head rather than of the tagged commit, and the two are worth keeping apart.

**6. Issue 3051 is still open.** The part of it that documented this census as permanently red is now fixed.

Issue #3120, opened 2026-09-19
Reclaim the Disk: say what fills a disk, recommend what is safe to remove


thefrederiksen:
## The mission

Build a DevThrottle feature that tells the owner what is filling a disk and recommends what is safe to remove, and that removes only what it can prove is disposable, in a way that can be undone.

## Why

On 18 September 2026 the C: drive of the owner's main machine reached zero bytes free of 931 GB and broke running commands in the middle of a session. A hand cleanup recovered 130 GB, and nothing on the machine prunes anything, so it all comes back. A machine that runs a fleet of sessions fills its disk faster than a person notices, and a full disk stops the fleet.

## The mission document

`docs/missions/reclaim-the-disk-2026-09-18/mission.md` on the branch `mission/reclaim-the-disk`, landing on main with phase 1. It holds the decisions in the owner's words, the design, the six phases, the check, and the refusal list. It is the whole mandate; this issue only points at it.

## Measured on the day

- 800 GB used on C:. An ordinary scan sees 703 GB; about 97 GB is invisible without an administrator.
- `C:\Windows\Installer` holds 58.8 GB, of which 211 files and 27.9 GB are orphans that no installed product or patch points at.
- About 31 GB of package caches, 31.9 GB in Temp, 47.1 GB of DevThrottle's own data.

## Done when

Merged to main, with the QA report described in section 3 of the mission document. No removal is ever run on the owner's machine by the mission itself; removal is proven on fixture trees.

thefrederiksen:
Delivery Lead, 18 September 2026.

The record is on main: pull request #3121 merged the mission document, the Architect's handover and the two read-only measurement scripts behind the numbers. The record exists before the code does, which is the point of landing it first.

Phase 1 - scan and report - is underway in its own worktree cut from origin/main, on branch `reclaim/phase1-scan-and-report`. It builds three things and nothing else: the platform-neutral engine library `src/CcDirector.Reclaim` (scanner, saved index, report), the `scan` and `report` commands of `tools/cc-cleanup-storage`, and `src/CcDirector.Reclaim.Tests` added to the default list in `scripts/test-local.ps1`.

Phase 1 deliberately contains no removal code of any kind - no delete, no move, no holding folder, not even unreachable. That is what makes it safe to merge quickly. Removal arrives in phase 3, with a Tech Lead watching and every refusal proven to fail when the refusal is taken out.

[...]

PR #3135, opened 2026-09-19
Reclaim the Disk phase 2: rules, the classifier and the recommend command

thefrederiksen:
Phase 2 of the Reclaim the Disk mission. Part of #3120. Phase 1 merged in #3122.

Mandate: `docs/missions/reclaim-the-disk-2026-09-18/mandate-phase-2.md`.
The judgements taken while building it: `phase-2-decisions.md`.
Proof: `phase-2-proof.md`, with both real-machine runs committed in `evidence/`.

## The one idea

**A rule may recommend removing something only when it holds one of exactly three proofs, and it can
hold no other.** A record the system keeps says nothing needs it; the tool that made the data has its
own command to clear it; or we made it, by an exact name, and it is old and closed.

**Anything no rule matched is reported as unclassified and is never offered for removal** - on the
owner's machine that is 596 gigabytes, and the size of that number is not a problem to be solved by
loosening a rule. This is an allow-list: the tool enumerates what to remove, never what to skip. A
path pattern from somebody else's product is not a proof, and nothing was imported from one.

## Fail closed, decided in one place

Every rule reports what it counted. `RuleFold` is the ONLY place that decides whether those counts
mean the rule worked, and it is folded there rather than inside each rule for the same reason the
Gateway folds a session's display state: a rule that ruled on itself could forget to, every new rule
would have to remember, and the forgetting is invisible.

A rule that compares two lists and finds one of them empty reports **broken**, never "nothing to
remove". A comparison against a record set that failed to load finds nothing every single time and
reads exactly like a clean disk - and for the installer rule specifically, it would mean every cached
package looks like an orphan and the tool would confidently recommend deleting the whole Windows
package cache, leaving the machine unable to repair or remove anything it has installed.

A rule that counts nothing at all is broken too, so a rule written later cannot fail open by simply
not counting.

## Six Windows rules

| Rule | Proof | Found here |
|---|---|---|
| Orphaned Windows installer packages | a system record | 204 files, 26.6 gigabytes |
| The NuGet package caches | its own command | 10.9 gigabytes |
| The uv package cache | its own command | 5.5 gigabytes |
| The npm package cache | its own command | 4.5 gigabytes |
| The pip package cache | its own command | 4.4 gigabytes |
| DevThrottle test scratch folders | we made it | 4,464 matched, all inside the age gate, 0 offered |

**Gradle is deliberately absent although the mission's table names it.** It ships no command that
clears its cache; its own guidance is to delete the folder, which is exactly the reasoning this tool
refuses. It has no proof of any of the three kinds, so no rule is written for it. Its cache is still
reported by the scan; it is simply never offered. Reasoning in `phase-2-decisions.md`.

## The mission's own check, run read-only on the owner's machine

    cc-cleanup-storage scan "C:\"              exit 0, 1409 seconds
    cc-cleanup-storage recommend "C:\" --json  exit 0,  116 seconds

Both committed in `evidence/`. Read by exit code and parsed output, never by searching printed text:

- exit **0**
- the installer rule with all three controls greater than nought: `records-read` **576**,
  `records-found-on-disk` **576**, `candidates-examined` **787**
- the unseen-gap line: **71.9 gigabytes** the volume counts as used that the scan did not see

**It lands on the design's hand measurement.** The Architect measured 576 packages Windows still
points at and 211 orphans on 18 September with an independent throwaway script. The rule found 576,
and 204 offered plus 7 held back by the thirty-day age gate - exactly 211.

**The best result in the run is a rule that offered nothing.** The scratch rule matched 4,464 folders
and every one was inside the seven-day age gate, so it offered nought. Its controls say that, instead
of printing "nothing to remove" - which is the same answer it would give on a machine where the name
list had been emptied by mistake.

## Two defects found by running it rather than reading it

- **Rules were applying machine-wide to a question about one folder.** A recommendation for a five
  kilobyte scratch folder reported 51.9 gigabytes reclaimable and nought unclassified. Every rule now
  declares where it looks, and `recommend` runs the rules whose folder is inside the folder asked
  about - **naming the ones it left out, with where they look**, because a tool that quietly ran fewer
  rules than it has reports less and reads like a cleaner disk.
- **The reach line was being picked out of the scan report's text by matching how it starts.** That is
  a check whose pass condition is finding something: reword a line in the scan report and the
  recommendations silently lose their reach line, with nothing failing anywhere. `ScanReport` now
  carries `ReachLines` as its own list, written once and read by both.

## No removal code of any kind

Not a delete, not a move, not a holding folder, not even unreachable, not even behind a flag that is
never passed. Nothing starts another process, so nothing runs anybody else's cleanup command - the
rules print them. Nothing raises itself to administrator. A sweep of the whole reclaim surface for
removal verbs returns five hits, every one of them prose. Phase 2 cannot remove anything because the
code to do it does not exist.

Nothing was removed from the owner's machine. The only things run against it are the two read-only
commands above.

## Proof

Local gate at this tip, in a worktree of its own: **nine suites, 2,464 tests, every suite
`outcome=Completed`, nothing failed and nothing skipped.** `CcDirector.Reclaim.Tests` is 191, up from
phase 1's 135.

**Four revert proofs**, each run by breaking the fix by hand, rebuilding, seeing the named tests red,
restoring, rebuilding, seeing them green. No run used a stale binary. Breaking the emptiness check in
`RuleFold` alone turns **five** tests red across two files. Full output in `phase-2-proof.md`.

**The COVERAGE GAP line names three parked suites and they were not run.** The selector cannot rule
anything out when a solution file changes and new paths appear, so that is a "cannot tell" verdict
rather than a "this touches it" one. Measured against `origin/main`, this branch **adds 25 files,
deletes none, and the only modified files outside the reclaim surface are `cc-director.sln`, which
registers the new project, and one description string in `tools/registry.json`**. No source file those
three suites compile or exercise is changed.

## What the proof does not cover

Written out in `phase-2-proof.md`. In short: nothing ran on macOS or Linux; the registry reader is the
one piece behind an interface and has no unit test, only the real-machine run that matched the
design's independent script; no cache was cleared and no folder removed, so nothing proves the printed
commands free the bytes the rules measured, only that the caches hold them; the scan took 1,409
seconds on a Debug build, which is the number phase 5 should design the background scan around; and an
elevated scan was not run, so the 71.9 gigabytes the scan could not see is neither measured nor judged.


thefrederiksen:
## Fix round one: the review is answered

The phase 2 review returned **not approved** - one blocking finding on the fold, four corrections to
the record. All five are accepted, none declined. `docs/missions/reclaim-the-disk-2026-09-18/review-phase-2-answers.md`
answers each one.

**The blocking finding was real and it was the one that mattered.** `RuleFold.BrokenReasonFor` refused
a rule that reported no controls at all, but accepted a rule whose every control was declared
may-be-empty. Such a rule can never report broken, because no count it reports can ever alarm - folded,
it answered `verdict: ok` and `items: 0`, which is "nothing to remove" said by a rule with nothing
behind its answer. That is the exact failure this whole phase exists to make impossible, arriving
through the one door the guard did not cover.

It now refuses that rule in the same breath as one declaring no controls at all. Two tests hold it,
the second so the fix cannot be satisfied by refusing everything. No rule that exists today changes
its answer - all six Windows rules already declare a must-not-be-empty control - so this is for the
rules that do not exist yet, which is what makes it worth blocking on: phase 5 turns rules into data
refreshed from the Gateway, and this fold is the single gate a refreshed rule passes through.

The four record corrections, each a place this phase's own proof claimed more than it had:

- The proof said both runs were committed in `evidence/` when only the recommend run was - and the
  file count, folder count and 245 refused folders in its table had been read off a console that was
  not kept. The scan run's evidence is now committed, and the elapsed time corrected from 1,409 to the
  scan's own measure of 1,406.2.
- Revert proof three turned **three** tests red, not two. Re-run unfiltered here to confirm rather
  than accept. The cause is the useful part: the original ran under a two-name filter, so the third
  test was invisible rather than absent. A revert proof runs against the whole suite.
- The test table said 8 where the file holds 9.
- A sentence in the decisions document asserted something about the temporary folder that nothing had
  established. The real reason for that decision was already in the same sentence; the unsupported
  clause is gone.

## The gate

`.\scripts\test-local.ps1` in its own worktree: **2,466 tests, all nine suites `outcome=Completed`**,
zero failed.

**On the coverage gap, and why `-Parked` was not run.** The gap names `Core.Tests`, `Gateway.Tests`
and `Gateway.UnitTests`. This branch touches no Core or Gateway source at all -
`git diff --name-only origin/main...HEAD` reaches only `src/CcDirector.Reclaim*`,
`tools/cc-cleanup-storage`, the mission documents and the solution file. The gap fires because
`CcDirector.Reclaim` references `CcDirector.Core` for its logger, not because anything under Core or
the Gateway changed, so those suites would prove nothing about this change. Stated here rather than
run, which is the documented choice.

**Nothing was removed from this machine.** The phase contains no removal code; that is phase 3. The
only things run against the real disk are a read-only scan and a read-only recommendation, and both
are committed as evidence.

Issue #3124, opened 2026-09-19
The daily report states zeros it never measured: sessions/work tiles zero-fill a silent feed, the work tile is labelled 'merged to main' but never looks at git, and the microphone section is too much


thefrederiksen:
## What the owner saw

The daily report email of 2026-09-19 opened with two stat tiles - **"0 agent sessions ran yesterday"** and **"0 pieces of work merged to main"** - on a day the owner had run sessions and merged work all day. It ended with a microphone section enumerating every device on the account: dictation counts, speech levels, narrowband and clipping shares. Owner ruling on the third: *"way too much emphasis on the microphones. Listing all the people's microphones is kind of creepy - don't do that."*

This issue is the mission record for **what the daily report should contain and what every box must be measured from**.

### 1. A silent feed was rendered as a confident zero

`MorningReportBuilder.SessionsRan` (src/CcDirector.Gateway/Reports/MorningReportBuilder.cs) emitted `0` whenever the tenant had *any* ledger history and none inside the window. But "no events yesterday" has two causes the gateway cannot tell apart after the fact: nothing ran, or nothing reached us. The owner's sessions run on Directors that push to the hosted gateway - the reported account's ledger simply had nothing recent in it (a tenant/feed mismatch that still needs verifying against live data; the report endpoint requires the service token and no agent session may read the admission surface).

The same hole existed for `WorkDelivered` (`Any()` probe) and `HostedAiSpendUsd`.

### 2. The work tile's label claimed a measurement no code ever made

`WorkDelivered` counts **workflow runs with `AcceptanceStatus == Accepted`** completed in the window. It has never looked at git. The email labelled it *"pieces of work merged to main"* - so a user who merges fifty PRs reads a confident zero. The label was false advertising for the number behind it.

### 3. The microphone section

A 30-day rolling per-device table - device names with platforms, samples, dB, shares, advice - in a daily email. The full detail already lives on the Cockpit's Transcription Health page, which is where it belongs.

## What has been implemented (this issue's first slice)

**Gateway (this repo):**
- **Freshness bars for all three stats.** A zero is now only a measurement when the feed behind it has been heard from recently: `SessionStatFreshness` (48h - a connected Director pings the session funnel continuously), `RunStatFreshness` (7d - runs complete sporadically), `SpendStatFreshness` (7d). Outside the horizon the stat is ABSENT - the email's honesty rule, extended from "no data at all" to "no data recent enough to certify a zero".
[...]

PR #3140, opened 2026-09-19
test(gateway): the four DirectorHub funnel tests named in #3134 land on main

thefrederiksen:
## What

PR #3134 (issue #3124) stated its proof as *"53 tests green ... 4 new DirectorHub funnel tests: accepted snapshot and delta feed the funnel, a stale rejected push feeds nothing, an observer that throws never tears the tunnel down."*

Those four never reached `main`. `DirectorHubTests.cs` on main mentions the issue zero times, and nothing in either test project names `SessionStateObservationSink`. They existed only as uncommitted edits in a shared working copy on the build machine.

That left the change the email family mission rests on - hosted sessions reaching the governance ledger, the reason eight consecutive daily reports read `sessionsRan=0` - with no executed proof anywhere, while a merged pull request said it had 53 tests. The end-to-end test that did land (`HostedDirectorTunnelGovernanceTests`) has never run either: the machine-wide gateway test-suite lock has been held continuously by other sessions since yesterday afternoon.

This restores the four, unchanged from what was written, onto main. No production code is touched.

## Proof

Run in a worktree cut from `main`:

- **29 `DirectorHubTests` pass, 0 fail**, the four named in the output: `AnAcceptedSnapshotFeedsTheSessionStateFunnelForEverySession`, `AnAcceptedDeltaFeedsTheFunnelForThatSession`, `ARejectedStaleSnapshotFeedsTheFunnelNothing`, `ASessionStateObservationThatThrowsNeverTakesTheTunnelDown`.
- **Whole unit project: 6,356 pass, 8 fail.** The same 8 fail on clean `main` without this change - the database-backed `HostedSchemaRefusesAnUnownedRow` cases plus one `FleetOutcomeStore` paging case, which need a local Postgres that is not running. Verified by stashing the change and re-running the same filter.
- **Negative control.** Switching off the two lines in `DirectorHub` that feed the funnel turns **three of the four RED**. The fourth, `ARejectedStaleSnapshotFeedsTheFunnelNothing`, keeps passing: it asserts an *absence*, so it also passes when nothing ever feeds the funnel. It is a guard only alongside the other three, and that is worth knowing when reading it. Fix restored, 29 of 29 green again.

## Scope

What this does NOT cover: the end-to-end hosted test still has not been executed locally. The lock passed straight from one session to another at 22:08 UTC. CI on this PR is the runner of record for it, and I will run it locally the moment the lock frees.


thefrederiksen:
Merging on a red build, deliberately. Run 35481758999 on this branch fails three tests: VoiceServingLoopIsolationTests.Voice_sweep_reaches_only_the_owning_tenants_director, VoiceSweepBudgetTests.Cached_audio_cannot_hide_a_terminal_failure_after_a_later_user_message, WingmanMenuGuardProofTests.Prompt_WithoutMenuGuard_MenuOnScreen_ForwardsAndNeverReadsTheScreen. Run 35478491468 on main (736d9afe2) fails the same three, plus HostedDirectorTunnelGovernanceTests - which passes here for the first time in CI - plus two others. So this branch's failures are a strict subset of main's: it takes one red test off main and adds none. The three are tracked in #3152 and none is touched by this change.

PR #3145, opened 2026-09-19
Reclaim the Disk phase 2: the review record

thefrederiksen:
The phase 2 review and the fix-round review, kept with the phase they judged.

Two documents. The first read the whole phase and returned **not approved**: the fold accepted a rule whose every control was declared may-be-empty, and such a rule can never report broken - folded, it answered `verdict: ok` and `items: 0`, which is "nothing to remove" said by a rule with nothing behind its answer. Four corrections to the record came with it.

The second read only the fix-round delta and returned **approved**, having re-run the finding's own reproduction, both revert proofs, the local gate and the counts itself rather than taking the answers document's word.

Both reviewers were a different agent family from the seat that built the phase. The code they judged is already merged as #3135; this is the record of how it was judged.


thefrederiksen:
Superseded: this was opened from a branch that still carried the phase 2 commits, which are already squash-merged into main as #3135, so it showed the whole phase again rather than the review record. Reopened cleanly as a two-file change off main.

Issue #3152, opened 2026-09-20
Three CcDirector.Gateway.Tests failures on the 2.8.1 release gate, none of them from 2.8.1 - and one reads a session name where an id belongs


thefrederiksen:
The full release gate was run for 2.8.1 on merged main at `643304d3f36b6744220b52256cdff9457caf068a`,
in Release configuration with the parked suites. **Twelve suites, 16,033 tests, three failures - all
three in `CcDirector.Gateway.Tests`.** Every other suite reported `outcome=Completed` with zero
failures, including `CcDirector.Core.Tests` (4,456 executed) and `CcDirector.Gateway.UnitTests`
(6,360 executed).

## They are not from 2.8.1, and that is checkable rather than asserted

`git diff --name-only v2.8.0..643304d3f -- src/CcDirector.Gateway/ src/CcDirector.Gateway.Contracts/
src/CcDirector.Gateway.Tests/ src/CcDirector.Gateway.UnitTests/` is **empty**. Not one Gateway file
changed between the published v2.8.0 and this commit, so the release cannot have introduced these.
They are present in v2.8.0, which is already out.

## The three

```
VoiceSweepBudgetTests.Cached_audio_cannot_hide_a_terminal_failure_after_a_later_user_message
  Assert.False() Failure   Expected: False   Actual: True

VoiceServingLoopIsolationTests.Voice_sweep_reaches_only_the_owning_tenants_director
  Assert.DoesNotContain() Failure: Filter matched in collection (pos 2)
  Collection contained DirectorCommand { SessionId = "voice-loop-sess-a", Verb = "set-resolved-role" }

HostedDirectorTunnelGovernanceTests.A_tunnel_push_reaches_the_ledger_and_the_morning_report
  Assert.Equal() Failure: Strings differ
  Expected: "s-gov-1"
  Actual:   "gov - the waiting session"
```

All three are assertion failures in well under a second of test time (0.09s, 0.01s, 3.1s), not
timeouts or hangs.

## A hypothesis worth checking first, because it would explain all three cheaply

**The third one's actual value is a session NAME where an id was expected** - `"gov - the waiting
session"` rather than `"s-gov-1"`. That is the shape of a test reading real fleet state instead of
its own fixture. The machine this ran on had 28 live sessions at the time. The second failure is also
an isolation assertion finding one MORE item than it expected, which fits the same explanation.

This repository already has form here: a Gateway turns test fails on this machine specifically
[...]

PR #3154, opened 2026-09-20
feat(gateway): the report recipient list carries the time zone an account chose (#3124)

thefrederiksen:
Part of #3124 (Email Improvements, phase one). Owner ruling, 19 September 2026: the daily report arrives at 7am "and that would be the time set by the user, whatever time zone they're in".

## What changes

`GET /gateway/reports/recipients` now returns `{ account, email, timeZone? }`.

- `timeZone` is the zone the account CHOSE, always as an IANA id (a Windows id is converted; one with no IANA mapping is dropped), because the sender runs on a JavaScript clock that only knows IANA.
- An account that never chose carries **no key at all** - not null, not the operator default. The sender keeps those accounts on 7am Eastern, which is what everybody had before. A default dressed up as a choice would silently move them.
- One address on two accounts is still mailed once; the account that chose a zone wins over the one that said nothing, whichever was minted first.
- The log line gains a count of accounts on their own zone. Still counts only, never addresses.

No new data is collected: `TenantSettingKeys.TimeZone` already exists per tenant. The website half (hourly cron, per-reader send) is devthrottle_internal, same mission. It is safe in either merge order: a sender that does not know the key ignores it, and a sender that knows it treats a missing key as Eastern.

## Proof

`ReportRecipientsEndpointTests`: 18 pass (14 existing, 4 new). The new guards were proved by mutation - with the endpoint changed to fill in `settings.TimeZone(...)` and the chosen-zone-first ordering removed, `AnAccountThatNeverChoseATimeZone_CarriesNoTimeZoneKeyAtAll` and `OneAddressOnTwoAccounts_TheChosenTimeZoneWins_WhicheverAccountCameFirst` both fail; restored, all 18 pass.

Not run: the full Gateway suite locally. Three tests on main are red today for reasons recorded in #3152, none of them touched here.

This does not deploy anything. The hosted Gateway deploys after phases one and two, on the owner's say.


thefrederiksen:
Merging on a red build, deliberately. Every failure in this run is one main already has and none is touched here: HostedDirectorTunnelGovernanceTests (fixed on main by #3140, which merged after this branch was cut), VoiceServingLoopIsolationTests and WingmanMenuGuardProofTests (both red on main run 35478491468, tracked in #3152), and on one of the two runs a TunnelExplicitRouteProofTests case - a different case of that class fails from run to run, on main too. The tests this change adds ran in this build and passed.

PR #3155, opened 2026-09-20
feat(gateway): the unmerged-branches row counts the branches it could not place (#3124)

thefrederiksen:
Part of #3124 (Email Improvements, phase one). Owner decision, 19 September 2026: branch counts report what they could not work out, instead of silently dropping it.

## What changes

`UnmergedBranchesAttentionDto` gains `undetermined`: how many of the repository's branches could not be placed either way - merge state never determined (`MergedIntoDefault` null), or unmerged with no tip date to age them by.

- They are still never NAMED in `branches`. Nobody established that they are unfinished work; the rule that only a definite `false` is an unmerged branch stands.
- They are no longer dropped. Dropping them made the email's count read as the whole truth when it was only the part that could be checked.
- A definite merged branch, a branch touched within the day, the default branch and the checked-out branch are known quantities and are not counted.
- The key is absent from the wire when zero, like every other section with nothing to say.
- An item can now carry `undetermined` with an EMPTY `branches` list. The website renderer handles that (devthrottle_internal#2146) and must be live before this deploys - it is, the site deploys on merge and the Gateway deploys later on the owner's say.

## Proof

`RepoHygieneFoldTests`: 19 pass (2 new, 2 existing tests changed from "no item" to "an item with no named branches and a count of 1" - that change IS the feature). With `Undetermined` forced to 0, three of them fail; restored, 19 pass. The 47 tests matching RepoHygieneFold and MorningReport in `CcDirector.Gateway.UnitTests` pass.

Not run: the full Gateway suites locally. Not established: why the Directors leave the merge state undetermined for so many branches in the first place. This makes the email admit the gap; it does not close it.


thefrederiksen:
Merging on a red build, deliberately. Every failure in this run is one main already has and none is touched here: HostedDirectorTunnelGovernanceTests (fixed on main by #3140, which merged after this branch was cut), VoiceServingLoopIsolationTests and WingmanMenuGuardProofTests (both red on main run 35478491468, tracked in #3152), and on one of the two runs a TunnelExplicitRouteProofTests case - a different case of that class fails from run to run, on main too. The tests this change adds ran in this build and passed.

PR #3157, opened 2026-09-20
THROWAWAY path-filter proof: an unrelated change leaves the PostgreSQL proofs job out

thefrederiksen:

This pull request exists only to produce a real workflow run whose head touches
.NET code that is NOT database-facing, so the run record shows the new
"PostgreSQL proofs" job NOT starting while the .NET job still does. It carries
the package 3 branch plus one scratch commit that adds three comment lines to an
Avalonia dialog.

The other half of the same proof - a migration change, where the job must start -
is in the sibling pull request.

Both are closed as soon as the run numbers are recorded.


thefrederiksen:
Closing: this head carries the whole package branch, so its diff against main contains database-facing files and the PostgreSQL proofs job correctly starts. It cannot show the other half of the path filter. A controlled pair against a scratch base replaces it.

PR #3158, opened 2026-09-20
THROWAWAY: one migration file changed - the PostgreSQL proofs job must run

thefrederiksen:

One half of a controlled pair. Both pull requests are opened against the same
scratch base, and each differs from it by exactly ONE file. This one changes a
PostgreSQL migration; the sibling changes an Avalonia dialog. Everything else
about the two heads is identical, so the only thing that can explain a
difference in which jobs ran is the path filter under test.

Expected here: "PostgreSQL proofs" RUNS.

Closed as soon as the run numbers are recorded.


thefrederiksen:

PR #3159, opened 2026-09-20
THROWAWAY: one unrelated .NET file changed - the PostgreSQL proofs job must not run

thefrederiksen:

The other half of a controlled pair. Both pull requests are opened against the
same scratch base, and each differs from it by exactly ONE file. This one
changes an Avalonia dialog - .NET code that touches no database; the sibling
changes a PostgreSQL migration. Everything else about the two heads is
identical, so the only thing that can explain a difference in which jobs ran is
the path filter under test.

Expected here: the .NET job runs and "PostgreSQL proofs" does NOT.

Closed as soon as the run numbers are recorded.


thefrederiksen:

PR #3160, opened 2026-09-20
THROWAWAY, EXPECTED RED: take the database away and the job must refuse the run

thefrederiksen:
THROWAWAY. Do not merge and do not review. THIS RUN IS EXPECTED TO BE RED.

It removes the three environment variables that point the proofs at the database
this job builds, so every proof skips itself - the exact state continuous
integration has been in every day since these tests were written. The run must
go RED on the step that reconciles the tests the assembly declares against the
results the run produced.

A check nobody has watched fire is a check nobody has tested. This is that.

Closed as soon as the run number is recorded.


thefrederiksen:

PR #3173, opened 2026-09-20
feat(gateway): the daily report names the Directors that are behind the newest release (#3124)

thefrederiksen:
Part of #3124 (Email Improvements, phase one). Owner decision, 19 September 2026: the daily report gets a row for machines running an out-of-date Director.

## What changes

A new attention item, `outdated-directors`: `{ type, newest, directors: [ { machine, version } ] }`. One item per account listing every Director that is behind - three machines behind is one thing to do, not three rows.

- **Same rule as the Cockpit.** "Behind" is `FleetMachinesFold.CompareToNewest`, and the newest release comes from the SAME `NewestReleaseWatch` instance the Machines view reads (it was constructed inline for that view; it is now one field shared by both). The email and that page cannot disagree.
- **Not knowing is never reported as an answer.** A newest release that has not been read yet means no row. A version that cannot be parsed is never called behind.
- **Who is named.** A Director that was not stopped and was heard from in the last 24 hours. Not "connected right now": the email goes out at seven in the morning, when a laptop is often still shut, and that would hide exactly the machines the row is for. A Director stopped or silent for over a day is not nagged about.
- Tenant-scoped by construction: it reads `Registry.ListDirectors(tenant)`.

## Proof

`MorningReportOutdatedDirectorsTests` (in `CcDirector.Gateway.UnitTests`): 9 pass, including a positive control on the unread-release case, the exact wire JSON, and one-account-never-names-another's-machine. With the stopped and heard-from filters removed, `A_stopped_Director_and_one_not_heard_from_in_a_day_are_not_nagged_about` fails; restored, 9 pass. `CcDirector.Gateway.Tests` still builds.

Not run locally: `CcDirector.Gateway.Tests` - another session held the machine-wide test lock on a full run. CI is the runner for that. Not proven by any test: the `GatewayHost` wiring itself (that the lambdas reach the live registry and watch) - it compiles and is three lines, and the first hosted report after deploy is the check.

Known limit: after a Gateway restart the newest release is unread until something asks for it, so the first report after a restart may carry no row. It is never wrong, only absent.

The website renderer for this row is in devthrottle_internal and must be live before this deploys. Nothing deploys here; the hosted Gateway deploys after phases one and two, on the owner's say.


thefrederiksen:
Merging on a red build, deliberately. The nine tests this change adds ran in this build and passed, and nothing in CcDirector.Gateway.UnitTests failed. The five failures are in code this change does not touch - a handover tunnel test, the turn pusher, the Voice isolation and Wingman menu guard tests already red on main (#3152), and a disk-cleanup test in CcDirector.Reclaim.Tests - and the set differs from run to run on this repository.

PR #3174, opened 2026-09-20
feat(gateway): the daily report lists sessions stopped on a usage limit (#3124)

thefrederiksen:
Part of #3124 (Email Improvements, phase one). Owner ruling, 19 September 2026: "usage limit in, died-overnight not yet" - a crash cannot be told from a clean finish, and an email that cries wolf stops being read.

## What changes

A new attention item, `usage-limit-stops`: `{ type, sessions: [ { session, stoppedUtc } ] }`. One item per account.

- **An observed fact, not an inference.** The session supervisor already classifies the live screen when a turn ends and writes a `supervisor-fault-detected` line to the recovery log carrying the matched signature. The row reads those lines for exactly the three usage-limit blocks the classifier knows - now exposed by one spelling as `TerminatingFaultClassifier.UsageLimitSignatures`, which the classifier's own list is built from. A bad API key, "out of credits", a rate limit and a model verdict are all non-recoverable too and are NOT called a usage limit.
- **Still stopped.** A session with an `active` transition in the governance ledger after the stop has resumed and is dropped; so is one the Gateway can see has exited. A session that stopped twice is listed once, at its latest stop.
- Stops older than 48 hours are retired. An account with the supervisor switched off writes no such lines and gets no row - absent, never "none stopped".
- Named by friendly name when the session is live, otherwise its id - the same rule as the waiting row.

Note for the reader: the supervisor already emails the owner at the moment of an escalation (capped per day). This row is the morning reminder that the session is STILL stopped, not the first notice.

## Proof

`MorningReportUsageLimitStopsTests`: 8 pass, including the seam test - every signature the row reads is fed through `Classify` and must come back as the same string the row matches, so the two cannot drift silently - tenant isolation, and the exact wire JSON. Mutation 1 (resumed sessions kept): `A_session_that_worked_again_after_the_stop_is_not_news` fails. Mutation 2 (every fault counts): `A_stop_for_any_other_reason_is_not_called_a_usage_limit` fails. Restored: 126 pass across the classifier, supervisor and new tests in `CcDirector.Gateway.UnitTests`.

Not run locally: `CcDirector.Gateway.Tests` (the machine-wide test lock was held). Not established: how often the supervisor's screen read actually catches a usage-limit block in production - the row can only be as complete as that detection.

This and #3173 both add a type to `MorningReportDtos.cs`; whichever merges second needs a trivial rebase. The website renderer lands first in devthrottle_internal. Nothing deploys here.


thefrederiksen:
Merging without waiting for the rebuild, deliberately. This branch was rebased onto main after #3173 (both add a type to MorningReportDtos.cs and a row to MorningReportBuilder.cs); the rebased code builds, including CcDirector.Gateway.Tests, and the 98 unit tests covering the morning report, the hygiene fold and the fault classifier pass locally. The pre-rebase build (run 35487811616) ran this change's eight tests and they passed; its failures were tunnel, screenshot, registry and disk-cleanup tests in code this change does not touch, plus the Voice and Wingman tests already red on main (#3152). A rebuild takes about 100 minutes and would be red for those same reasons.

PR #3185, opened 2026-09-20
feat(gateway): the daily report names a session working unusually long by the account's own measure (#3124)

thefrederiksen:
Part of #3124 (Email Improvements, phase one - the last piece). The rule is the owner's, approved 20 September 2026: a session is running too long when it is **still working after three hours AND at least ten times the account's own usual**. Both must hold, so a long build does not fire it and somebody whose sessions routinely run for hours is not nagged.

## What changes

A new attention item, `long-running-sessions`: `{ type, usualMinutes, sessions: [ { session, runningHours } ] }`. One item per account.

- **A working stretch** is the time from an `active` transition in the governance ledger to the same session's next transition. The account's usual is the MEDIAN of its finished stretches over 14 days - the median, so one 30-hour marathon does not move it.
- **Too little history says nothing.** Under 20 finished stretches the row is absent and the log says why. It never guesses a usual.
- **The ledger alone is not believed for "still working".** The state emitter closes an open WAIT when a session exits, but writes nothing when an active session simply vanishes (the Director closed, the lid shut), so that session reads `active` for ever. An open stretch is reported only when the Gateway can SEE the session live, right now, saying Working. No live store, or a session missing from it, means no claim.
- A session whose later transition is a wait or idle has no open stretch, whatever a stale roster says.

## Proof

`MorningReportLongRunningSessionsTests` (CcDirector.Gateway.UnitTests): 9 pass - the floor, the multiple, the minimum history with a positive control, the vanished-session trap, the median against a marathon, tenant isolation, exact wire JSON. Mutation 1 (a session not seen Working still counts): the vanished-session test fails. Mutation 2 (floor removed): the under-three-hours test fails. Restored: all 54 morning-report unit tests pass, and CcDirector.Gateway.Tests builds.

Not established: what the real distribution of working stretches looks like for any account - the production ledger was not read. The rule is the owner's choice, made knowing that. If it cries wolf or stays silent in practice, the three constants at the top of the method are the knobs.

The website renderer lands first in devthrottle_internal. Nothing deploys here.


thefrederiksen:

**Why not wait.** Build & Test (.NET) takes about 100 minutes and is red on main for tests this branch does not touch (tracked in #3152: `VoiceServingLoopIsolationTests`, `WingmanMenuGuardProofTests`, `VoiceSweepBudgetTests`, plus the flaky tunnel and screenshot tests). A green run here is not available to anybody today, so waiting buys no information about this change.

- `MorningReportLongRunningSessionsTests`: 9 of 9 pass.
- Every morning-report unit test: 54 of 54 pass.
- `CcDirector.Gateway.Tests` builds against the changed constructor and DTOs.
- Two mutations, each caught: removing the live "Working" requirement fails `A_session_the_ledger_calls_active_but_the_Gateway_cannot_see_working_is_not_claimed_to_be_running`; replacing the median with a mean fails `The_usual_is_the_MEDIAN_so_one_marathon_does_not_move_it`.

**What this does not prove.** The full Gateway.Tests suite was not run locally - another session held the machine-wide test lock. The change adds one DTO, one derived-type registration and one private method on the builder; it changes no existing row. If this run later shows a failure outside the #3152 list, that is mine to fix and I will.

The website already renders the new row (devthrottle_internal #2151, merged first), so the email never sees a type it does not know.

PR #3225, opened 2026-09-20
The daily report says whether the account did anything that day (#3124)

thefrederiksen:
## Why

Owner's rule, 20 September 2026: *"If you didn't work in DevThrottle yesterday, I don't think we should send a daily email. It's just annoying. And if you didn't make any changes, you already have an email from the day before."*

The sender cannot answer that on its own - the website holds accounts and email preferences and has never seen a session. This is the Gateway's half.

## What

`MorningReportDto.workedInWindow` - true when a turn was submitted **or** a session became active inside the reported day.

**Two signals, either of which means yes.** A submitted turn is a person driving an agent, and is the normal case. A session becoming active covers the day that produced no turn record, which is the hosted case when a Director is mid-reconnect. Reporting "you did nothing" for that day would silently stop the email of somebody who worked all day, so one signal alone is not safe.

**It is a boolean, not a count, and that is the point.** The same owner ruled on the same day that the daily is a briefing and not a scoreboard - *"telling me how much shit I did yesterday is not going to help me today"* - and the `stats` key came off this wire because of it. A count would be that key returning under a new name. A yes/no cannot be rendered as an achievement; it can only answer whether there is any reason to write to this person at all. The sender reads it to decide whether to send and never prints it. One test asserts `stats` is still absent from the serialized report.

**The window is the reader's own day.** A turn at 15:00Z on the 18th is inside a Sydney reader's 19th and outside a UTC reader's. A test pins exactly that, because getting it wrong means mailing people on the wrong day's evidence.

## What is deliberately NOT here

The owner also asked that the daily only go to people who have **actually installed**. That needs no field: a session cannot run on a machine that never installed, so `workedInWindow` already contains the install rule.

I wrote an `installed` flag on `/gateway/reports/recipients` and then removed it. It changed no outcome, and `RegisteredDeviceDto` does not carry a device type - the registry normalises a blank type to `workstation`, so the flag would have counted almost every enrolled device and been a weaker signal wearing a stronger name. Saying so here rather than leaving a field nothing reads.

## Proof

- 9 new tests in `MorningReportWorkedInWindowTests`; **63 morning-report unit tests pass**.
- Two mutations, both caught and both restored:
  - dropping the window bounds from the turn query: fails `Work_on_a_different_day_is_not_work_on_this_one` and `The_window_is_the_readers_own_day_not_UTCs`
  - accepting any session state rather than only `Active`: fails `A_session_merely_sitting_in_some_other_state_is_not_work`

## Deployment order matters

This must be deployed **before** the website starts reading the field. The website treats a missing `workedInWindow` as "we cannot tell" and skips the send, so shipping the website first would stop the daily report for everybody until this lands.

Full CI here takes about 100 minutes and main is red for Voice/Wingman tests this branch does not touch (#3152).


thefrederiksen:

**Why not wait.** Build & Test (.NET) takes about 100 minutes and is red on main for tests this branch does not touch (tracked in #3152: `VoiceServingLoopIsolationTests`, `WingmanMenuGuardProofTests`, `VoiceSweepBudgetTests`, plus the flaky tunnel and screenshot tests). A green run is not available to anybody today, so waiting buys no information about this change.

- `MorningReportWorkedInWindowTests`: 9 of 9 pass.
- Every morning-report unit test: 63 of 63 pass.
- Two mutations, each caught and restored:
  - dropping the window bounds from the turn query fails `Work_on_a_different_day_is_not_work_on_this_one` and `The_window_is_the_readers_own_day_not_UTCs`
  - accepting any session state rather than only `Active` fails `A_session_merely_sitting_in_some_other_state_is_not_work`

**What this does not prove.** The full `CcDirector.Gateway.Tests` suite was not run locally - it takes a machine-wide lock and another session was holding it. This change adds one property to a DTO and one private read on the builder; it changes no existing row and no existing endpoint. If this run later shows a failure outside the #3152 list, that is mine to fix and I will.

**Why it merges first.** The website half (devthrottle_internal #2175) treats a missing `workedInWindow` as "we could not tell" and skips the send, so it cannot land until this is deployed - otherwise the daily report stops for everybody.

PR #3250, opened 2026-09-20
fix(gateway): the usage-limit row names its sessions instead of printing their ids (#3124)

thefrederiksen:
The owner opened his own daily report today and found six raw identifiers under "6 sessions
stopped on a usage limit". That row tells him to open each one and tell it to continue - which
nobody can do from a string that no search will match.

It is the same defect fixed yesterday in the waiting rows (#3244), one row type over, and I
left it standing while repairing its neighbour. One line:

```csharp
Session = string.IsNullOrWhiteSpace(liveSession?.Name) ? sessionId : liveSession!.Name!,
```

The lookup was allowed to miss and the id stood in for the name, so a session the Gateway can
no longer see came out looking exactly like one it knows.

## What changed

- A stop the Gateway can see is listed with the session's **number, name and short repository
  name**, plus its id so the email can link straight into the Cockpit.
- A stop it cannot see is **counted** in `LostContactCount` and never listed. Absent from the
  wire rather than zero when there are none, like every other section.
- A **snoozed** session is dropped. Snoozing is the owner saying "not now"; telling him to go
  and restart it contradicts his own decision.

## Why the tests did not catch it

They seeded no live roster at all, so they asserted the identifier and called it a name. Every
test now seeds one - including the ones that assert silence, so a null is the filter under test
and not the new lookup quietly emptying the row for an unrelated reason.

## Proof

Control first, green at 76 morning-report tests. Then eight mutations, each restoring a red suite:

| Mutation | Tests failed |
|---|---|
| the identifier fallback put back | 2 |
| the lost ones dropped instead of counted | 2 |
| a closed session told to resume | 1 |
| a snoozed session nagged about again | 1 |
| the count zero-filled rather than absent | 4 |
| repository swapped for the disk path | 2 |
| the session number dropped | 2 |
| the link id dropped | 3 |

Full project: 6934 passed, 8 skipped, 0 failed.


thefrederiksen:
Merging with the .NET job red. The argument, so it can be checked rather than taken on trust.

**The two failures this branch caused are fixed, and they passed by name.** The first run failed
`MorningReportEndpointTests.The_JSON_matches_the_agreed_contract_and_omits_what_the_Gateway_does_not_know`
and `A_second_tenants_data_never_appears_in_the_first_tenants_report`. Both were asserting the behaviour
the owner ruled against - an unseeable waiting session still appearing, named by its identifier - and both
seeded no live roster, so each was asserting the shape of a row the report no longer produces. Both now
seed a connected Director. In the second run they appear as `Passed` by name, along with **99 passing
MorningReport tests and 14 passing UsageLimitStops tests**. That is presence, not the absence of a failure
line.

**The remaining ten are not in anything this branch touches.** The diff is three files: the morning-report
DTOs, the morning-report builder, and morning-report tests.

**And the failing set CHANGED between two runs of near-identical code**, which is the signature of
contention rather than breakage:

| | run 1 | run 2 |
|---|---|---|
| SkillEndpoints, SkillLibrary, VoiceServingLoop, VoiceSweepBudget, WingmanMenuGuard, Reclaim DiskCleanup | fail | fail |
| TenantSessionMap, TunnelExplicitRoute x2 | fail | pass |
| RepositoryRegistryConcurrency, DirectorCommandRouterTimeout, MicrophoneQualityFold, WorkflowAuthoring | pass | fail |

Six are stable red on this repository; seven others moved in or out between the two runs. None is a
morning-report test.

**What this evidence does not cover.** There is no completed CI run on `main` to compare against - every
main run today was cancelled by the next push - so this is not a measured "strict subset of main's
failures". It is the weaker claim that the failures lie outside the diff and that the set is unstable.

Locally, on the merge content: `CcDirector.Gateway.UnitTests` 6934 passed / 8 skipped / 0 failed; the
morning-report filter green in both projects (76 and 23); and every other consumer of the changed types -
`ReportCadence`, `HostedDirectorTunnelGovernance` - green. Eight mutations of the new rules were each
caught against a green control.

Issue #3275, opened 2026-09-21
Gateway polling re-reads unchanged rows from the database


thefrederiksen:
The Gateway re-read unchanged database rows on timers (roster fold, Director hello, conversation polling, history sweep, inbox counts), which made database traffic far larger than the data that actually changes.

Fixed by #3282: reads that do not change between writes are held in memory and discarded on write, and polling reads only what is new.

Measurement and follow-up are tracked internally.

PR #3282, opened 2026-09-21
perf(gateway): stop re-reading unchanged rows from the database on every poll

thefrederiksen:
Draft - not to be merged or deployed without the owner's go.

Refs #3275. The per-query root cause is written up on the internal cost issue (devthrottle_internal#2199).

## What was wrong

The Gateway re-read the same database rows on a timer instead of only what changed. The worst case was the roster fold: on every roster poll, every 5-second display sweep and every accepted Director push it re-read every session's latest Wingman verdict, including its JSON, although those rows change only when a verdict is written.

## What this changes

| Read | Before | After |
|---|---|---|
| Wingman verdict snapshot (`TurnVerdictStore.SnapshotLatest`) | Full rows incl. `AnswerJson`, every fold | Held per account until the next verdict write; only `SessionId`, `JudgedAtUtc`, `VerdictJson` selected. Each caller still gets fresh objects. |
| Turn watermarks (`DirectorHub.Hello`) | Every Hello - every 10 s per Director, since the re-push sends Hello | First Hello of a connection only. A repeat Hello answers `TurnWatermarksKnown = false`, which the Director already treats as "keep what you have". Reconnect / restart / deploy are new connections and still get the list. |
| Stored conversation (`SessionTurnStore.ReadCurrent`) | Whole generation incl. `PartsJson` on every Chat poll (2.5 s), narration, turn log, Wingman Now | Rows already read are held (raw columns, never the handed-out objects); only ordinals past them are read, checked against the head row every time. Generation switch or a shorter head rebuilds. Idle 10 min or over 50,000 held rows lets go. |
| History sweep roll-ups (`SessionHistorySummarizer.RefreshRollupsAsync`) | 30 days of full rows every 2 min per account | Staleness from a narrow read (the columns the grouping and hash use); full rows only for the groups rewritten. The saved hash is the one the History report computes. |
| Unread inbox counts (`FleetMessageStore.UnreadCountsByRecipient`) | Every fold | Held per account; discarded by send, reply, read, overdue, stuck and purge. A ring changes nothing counted and leaves it. |
| Push-subscription count, tenant census | Every notifier poll / every sweep cycle | Held; discarded by the store's own writes. |

**Freshness rule for every held answer:** any write through the owning store discards it at once (a version read before the query, so a write that lands mid-read cannot leave a stale answer held). A 30 s ceiling (60 s for the tenant census) covers the one writer a process cannot hear: a second Gateway process during a deploy.

## Proof

- New tests, each checked to go **red** with its fix disabled (negative control run for every change, and again with the write-invalidation removed):
  - `TurnVerdictStoreTests`: second snapshot reads nothing; every write (store, supersede, purge) is seen by the next snapshot; per-account isolation; ceiling; callers get their own objects; the snapshot query does not carry `AnswerJson`.
  - `DirectorHubTests.Hello_HandsBackTheWatermarks_OncePerConnection`: repeat Hello answers "not read" (not an empty list), a new connection gets the full list. The existing multi-Director test now reconnects on a new connection, which is the real scenario.
  - `SessionTurnStoreTests`: a read with nothing new issues one command (the head); the held prefix is deleted behind the store's back and the next read still returns it, proving only the tail is read; generation switch and retention-and-return rebuild; accounts isolated; idle eviction.
  - `SessionHistoryRollupRefreshTests` (new - the roll-up pass had no tests): prompt built from full records; saved hash equals the report's hash; second pass writes nothing; narrow and full reads group and hash identically.
  - `FleetInboxLineTests`, `PushSubscriptionStoreTests`, `TenantRegistryTests`: held, every write seen, ceiling.
- `GatewayReadCutsPostgresTests` (gated on `CC_GATEWAY_TEST_PG_CONNECTION`): the three changed query shapes on a real PostgreSQL 16. Run locally against a throwaway container: 10/10 passed with the neighbouring Postgres proofs.
- `CcDirector.Gateway.UnitTests` full suite: 7,173 passed, 1 failed - `GatewayStatsSqliteBaselineEquivalenceTests.Baseline_CreatesTheSameTablesAndIndexesAsTheHandRolledVersion5Code`, which passes in isolation twice and in the previous full run (not touched here).
- `CcDirector.Gateway.Tests` (longer than 10 min in full, run in batches): Wingman / verdict / roster 154 passed; history / conversation / hub 24 passed; push / tenancy / sweeps / tunnel 178 passed, 6 failed - all `AWorktreeIsNotARepositoryTunnelProofTests`, failing inside the test's own `RealPath` helper (`DirectoryNotFoundException: C:\`) before any Gateway code runs; a machine filesystem issue, not this change.

## Not in this PR

`tenant_settings` by key, `snoozes` by session and `session_keys` lookups are very high in call count but tiny in bytes; the Stats page reads (`concurrency_hour`, token hours, origin totals) run only when Stats is opened. Left for a follow-up if the after-numbers call for it.


thefrederiksen:
Second review and gate, before merge (owner's go given 2026-09-21).

**Review, looking for stale or missing data on a screen: no defect found.**
- Every writer of `turn_verdicts`, `fleet_messages`, `push_subscriptions` and `tenants` in the Gateway goes through its own store, and each store is constructed once in `GatewayHost`. Every committed write discards the held answer, and the version is read before the query, so a write that lands during a read cannot leave a stale answer held. The verdict `Store` commits in `SaveChanges` with no outer transaction, so the discard always comes after the commit.
- Held conversation: turn rows are insert-only (a push skips ordinals that are already there), the generation only moves forward, and `Append`, `ReadCurrent` and `PurgeOlderThan` all share one lock. A retention purge removes the head, and the next read sees no head and drops what it held.
- Hello once per connection: the Director's `TurnPusher` takes the Gateway's watermark from every `PushTurns` reply, including a lower one after a gap or a purge, so the repeat Hello was never the only thing that corrected it.
- History narrow read: it skips the fold's empty-to-null normalisation of `EndingKind` and `SummaryKind`, but `InputHash` formats an empty string and null the same way, so the hashes are identical and no roll-up is rewritten every pass.

**Local gate on 50d1dcd06**
- `test-local.ps1` default run: every suite completed and passed except 2 `LauncherDeclaredCapabilitiesTests`, which check the machine-wide restart signal that the live Launcher on this machine holds. This change touches no Launcher code.
- `CcDirector.Gateway.UnitTests` in full: 7,174 passed, 0 failed, 8 skipped.
- `GatewayReadCutsPostgresTests` on the run's own throwaway PostgreSQL: 3 of 3 passed, none skipped.
- `CcDirector.Gateway.Tests`, the Wingman, verdict, roster, history, push, tenancy and fleet slice: 80 passed.

**Continuous integration is red, but not because of this change.** The same families fail on main (`AWorktreeIsNotARepository*`, `RepositoryRegistryConcurrencyTests`, `VoiceServingLoopIsolationTests`, the factory switch-route tests).

**No review from a different agent family.** Codex and Grok were at their usage limits and Gemini's API key was rejected. This is a gap in the gate, recorded here.

